PR #1, opened 2024-11-30
Add echo server

jlowin:


imadNaseer:
Can you confirm if FastMCP 2.0 supports WebSockets?

PR #17, opened 2024-11-30
typing updates

zzstoatzz:
looks like main is hot so will rebase and update

---

this PR
- fixes typing in `server.py`
- adds precommits + CI job
- updates publish CI to use `uv`

lemme know if you want to avoid any of this for now, just getting oriented and yak shaving


pyproject.toml
```diff
@@ -24,15 +24,18 @@ build-backend = "setuptools.build_meta"
 [tool.setuptools_scm]
 write_to = "src/fastmcp/_version.py"
 
-[dependency-groups]
```
zzstoatzz, line 27: `dev` was not being recognized as a valid group

src/fastmcp/server.py
```diff
@@ -176,7 +180,7 @@ async def list_resource_templates(self) -> list[MCPResourceTemplate]:
             for template in templates
         ]
 
-    async def read_resource(self, uri: _BaseUrl) -> Union[str, bytes]:
+    async def read_resource(self, uri: AnyUrl | str) -> str | bytes:
```
zzstoatzz, line 183: we can have nice union syntax now 🙂

jlowin (approved):
🎉

PR #29, opened 2024-12-01
Surge example

zzstoatzz:
adds example using https://surgemsg.com/ to text myself from claude

```console
fastmcp install examples/text_me.py -f .env && fastmcp run examples/text_me.py
```

also updates tests to only run on changes to src, tests, and dep-related files

### DIFF
--- a/examples/text_me.py
+++ b/examples/text_me.py
@@ -1,0 +1,48 @@
+# /// script
+# dependencies = ["fastmcp"]
+
+"""
+FastMCP Echo Server
+"""
+
+import httpx
+from pydantic_settings import BaseSettings, SettingsConfigDict
+
+from fastmcp import FastMCP
+
+
+class SurgeSettings(BaseSettings):
+    model_config: SettingsConfigDict = SettingsConfigDict(
+        env_prefix="SURGE_", env_file=".env"
+    )
+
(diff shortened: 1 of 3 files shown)


examples/text_me.py
```diff
@@ -0,0 +1,59 @@
+# /// script
+# dependencies = ["fastmcp"]
+# ///
+
+"""
+FastMCP Echo Server
```
jlowin:
```suggestion
FastMCP Surge Server
```

examples/text_me.py
```diff
@@ -0,0 +1,59 @@
+# /// script
+# dependencies = ["fastmcp"]
```
jlowin, line 2: This *should* be unecessary as fastmcp always adds itself as a dependency but 👍 good practice. I did learn that if you have a script declaration like this it overrules all other dependencies though

jlowin (approved):
Thanks! Do you mind adding some brief documentation to the top of the example file - maybe what env vars to set or how to deploy with env

zzstoatzz:
good calls @jlowin - added in [cbb8dbe](https://github.com/jlowin/fastmcp/pull/29/commits/cbb8dbeccfa62479900d12f38cff6267cea61249)

PR #31, opened 2024-12-01
Handle complex inputs

jurasofish:
This updates tools to handle complex function arguments including
- Pydantic models
- Things like `list[str] | str`
- Things like `Annotated[str, Field(max_length=10)]`

A key part of this is that there's pre-processing to parse inputs from the client from JSON.
This is because Claude Desktop seems to send complex things - sub-models, lists of strings - as JSON strings. If you naively try to validate them with pydantic they'll fail because they're just strings.

The code here is conceptually very similar to FastAPI
- See [`analyze_param`](https://github.com/fastapi/fastapi/blob/8255edfecfe539f09c2655016ea3ef3e4fb50881/fastapi/dependencies/utils.py#L348)
- See [`request_params_to_args`](https://github.com/fastapi/fastapi/blob/8255edfecfe539f09c2655016ea3ef3e4fb50881/fastapi/dependencies/utils.py#L740)

FastAPI creates a [ModelField class](https://github.com/fastapi/fastapi/blob/8255edfecfe539f09c2655016ea3ef3e4fb50881/fastapi/_compat.py#L88), with a FieldInfo, for each argument. Then it validates them one by one [with a typeadapter](https://github.com/fastapi/fastapi/blob/8255edfecfe539f09c2655016ea3ef3e4fb50881/fastapi/_compat.py#L120). Here, however, I've basically taken all the FieldInfos and used them to construct a new pydantic model with all those fields and used that for validation and schema generation. FastAPI has far more faff to it as it's handling both Pydantic v1 and v2, and it has all kinds of logic for dependency (via Depends) and Path/Query/etc.

As part of this I've also
- Ensured context is not included in tool json schema as passed to client
- Added a test for tool json schema generation (well it's a unit test for the internal thing that FastMCP uses)

### DIFF
--- a/README.md
+++ b/README.md
@@ -212,6 +212,27 @@ async def fetch_weather(city: str) -> str:
         return response.text
 ```
 
+Complex input handling example:
+```python
+from pydantic import BaseModel, Field
+from typing import Annotated
+
+class ShrimpTank(BaseModel):
+    class Shrimp(BaseModel):
+        name: Annotated[str, Field(max_length=10)]
+
+    shrimp: list[Shrimp]
+
+@mcp.tool()
+def name_shrimp(
+    tank: ShrimpTank,
+    # You can use pydantic Field in function signatures for validation.
+    extra_names: Annotated[list[str], Field(max_length=10)],
+) -> list[str]:
+    """List all shrimp names in the tank"""
+    return [shrimp.name for shrimp in tank.shrimp] + extra_names
+```
+
 ### Prompts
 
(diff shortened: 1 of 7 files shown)


jurasofish:
The example wasn't actually working before - have fiddled handling of `x: None` and should be okay now. Will see if I can get this ready for review later today. Need to sit down and study how FastAPI solves this problem too, see if there's anything that can be pilfered

jurasofish:
This is good for review now

jlowin:
Works as advertised, very nice! 
image

Issue #36, opened 2024-12-02
installation issues on windows


plasmax:
Love this idea, trying to install/get it working on windows, so if anyone else has please let me know your steps.

Other mcp servers are working with this caude_desktop_config.json:

```json
{
  "mcpServers": {
    "filesystem": {
      "command": "node",
      "args": [
        "C:\\Users\\Max\\AppData\\Roaming\\npm\\node_modules\\@modelcontextprotocol\\server-filesystem\\dist\\index.js", 
        "E:\\Repositories\\sandbox"
      ]
    },
    "git": {
      "command": "python",
      "args": ["-m", "mcp_server_git", "--repository", "E:\\Repositories\\bloom\\"]
    },
    "mcpserver": {
      "command": "uv",
      "args": [
        "--directory",
        "E:\\Repositories\\sandbox\\mcpserver",
        "run",
        "mcpserver"
      ]
    },  
    "Demo": {
      "command": "uv",
      "args": [
        "run",
        "--with",
        "fastmcp",
        "fastmcp",
        "run",
        "E:\\Repositories\\sandbox\\demo.py"
      ]
    }
  }
}
```

I've tried (in windsurf cmd prompt):
```
E:\Repositories\sandbox>uv pip install fastmcp      
Audited 1 package in 13ms

E:\Repositories\sandbox>fastmcp install demo.py
'fastmcp' is not recognized as an internal or external command,
operable program or batch file.

E:\Repositories\sandbox>C:\Users\Max\AppData\Roaming\Python\Python312\Scripts\fastmcp.exe install demo.py
[12/02/24 09:17:43] 
INFO     Added server 'Demo' to Claude config            claude.py:108                    
INFO     Successfully installed Demo in Claude app       cli.py:402
```

So the installation went well, but when restarting Claude Desktop, it failed to recognise fastmcp. The cmd terminal launches in the background, but then closes, suggesting an error somewhere, but there's no traceback.

Not sure if we can run servers externally? I tried this and it just sits there, and Claude doesn't recognise it even after a restart:
```
E:\Repositories\sandbox>C:\Users\Max\AppData\Roaming\Python\Python312\Scripts\fastmcp.exe run demo.py    
```

then I tried this:
```
E:\Repositories\sandbox>C:\Users\Max\AppData\Roaming\Python\Python312\Scripts\fastmcp.exe run E:\Repositories\sandbox\demo.py
File not found: E:\Repositories\sandbox\E
```

[...]

PR #42, opened 2024-12-02
Changes to accomodate windows users.

justjoehere:
This fixes _parse_file_path for windows users.  Also fixes 1 test, and adds 3 more for this function.

I did have to bypass 1 file permissions test on windows because I couldn't get it to fail permissions properly.

### DIFF
--- a/Windows_Notes.md
+++ b/Windows_Notes.md
@@ -1,0 +1,15 @@
+# Getting your development environment set up properly
+```bash
+uv venv
+.venv\Scripts\activate
+uv pip install -e ".[dev]"
+```
+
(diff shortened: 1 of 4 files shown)


justjoehere:
This also fixes Issue #36

jlowin:
Thank you! I'm also adding windows tests to the test matrix in #44. Really appreciate it

jlowin:
@justjoehere I believe I fixed the pyreadline issue by excluding dev utilities that aren't compatible with windows (changed `uv sync --extra dev` to `uv sync --extra test` which is a smaller set of requirements), so hopefully the windows note is unecessary - if you can confirm, we can remove it.

PR #44, opened 2024-12-03
Set up multiple os tests

jlowin:


jlowin:
Merging, based on #45 I believe #42 will fix the failing windows tests

PR #49, opened 2024-12-03
add pgvector memory example

zzstoatzz:
a memory system for Claude.

- `remember` tool (save one or many memories)
- `get_profile` tool (render the aggregate profile)

I have yet to let this go over a lot of memories, but will be using it and come back to this

The goal is to have a self-regulating, steady state system that can take everything the chat interface wants to remember, jostle and phase unused/old memories out, and render a reasonably sized summary of history at any given time

---

using pydantic_ai for gpt wrapper, pgvector for vector db, openai for llm/embeddings models

---

### usage

```
» mkdir -p postgres-data
docker run -d --rm -e POSTGRES_PASSWORD=postgres \
    -p 54320:5432 \
    -v `pwd`/postgres-data:/var/lib/postgresql/data \
    pgvector/pgvector:pg17

» fastmcp install examples/memory.py -f .env

» open -a Claude
```


src/fastmcp/cli/cli.py
```diff
@@ -423,7 +423,11 @@ def install(
         # Load from .env file if specified
         if env_file:
             try:
-                env_dict.update(dotenv.dotenv_values(env_file))
+                env_dict |= {
+                    k: v
+                    for k, v in dotenv.dotenv_values(env_file).items()
+                    if v is not None
```
zzstoatzz, line 431: excluding `None`

examples/memory.py
```diff
@@ -0,0 +1,326 @@
...
+        input=text,
+        model=EMBEDDING_MODEL,
+    )
+    # Return raw list of floats for internal use
+    return embedding_response.data[0].embedding
+
+
+async def summarize_text(text: str, deps: Deps) -> str:
+    summary_agent = Agent(
+        "openai:gpt-4",
```
jlowin: 4o?

pyproject.toml
```diff
@@ -10,6 +10,9 @@ dependencies = [
     "pydantic>=2.5.3,<3.0.0",
     "typer>=0.9.0",
     "python-dotenv>=1.0.1",
+    "pydantic-ai-slim[openai]>=0.0.8",
```
jlowin: Want to make sure these dont leak into the parent pyproject

examples/memory.py
zzstoatzz: oops!

pyproject.toml
zzstoatzz: removed!

PR #50, opened 2024-12-03
Add mypy

jurasofish:
Have made a handful of tweaks to get things passing.
mypy runs via pre-commit, and in CI.
To get it going with pre-commit I've set it up to run a custom script that uses uv to ensure all dependencies are installed. This is tricky using raw pre-commit with e.g. https://github.com/pre-commit/mirrors-mypy/tags hence the custom script.

Haven't done `/tests`.

I figure it's best to run mypy under python 3.10 - in support of this I've changed `.python-version` to specify 3.10. Otherwise there's environment churn and/or poor caching when running mypy. imo this is best as I don't want to be using py3.12 while I'm developing and start using features that aren't available 🤷 but this shouldn't be critical - mypy has a `python_version` in `pyproject.toml` which should make it irrelevant which version mypy is actually run under. Happy to change back.


jlowin:
@jurasofish appreciate this effort, but after discussion with @dsp-ant, adopted pyright in #52 to meet the requirements of the Python SDK

jurasofish:
🙌

PR #51, opened 2024-12-03
fix: use stderr for logging

sd2k:
Without this, logs get sent to stdout which is captured by the
JSONRPC protocol (which also causes some errors with clients when
the invalid JSON is sent over the stdout pipe).


jlowin:
Thanks! This was sort of driving me crazy

PR #54, opened 2024-12-05
Relicense from Apache 2.0 to MIT

jlowin:
FastMCP's code will be merged into the main MCP codebase, which is MIT licensed. To facilitate that, I am relicensing the codebase as MIT. 

This will be part of a new 0.4 release for clarity.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,201 +1,21 @@
-                                 Apache License
-                           Version 2.0, January 2004
-                        http://www.apache.org/licenses/
+MIT License
 
-   TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
+Copyright (c) 2024 Jeremiah Lowin
 
-   1. Definitions.
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
 
-      "License" shall mean the terms and conditions for use, reproduction,
-      and distribution as defined by Sections 1 through 9 of this document.
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
 
-      "Licensor" shall mean the copyright owner or entity authorized by
-      the copyright owner that is granting the License.
-
-      "Legal Entity" shall mean the union of the acting entity and all
-      other entities that control, are controlled by, or are under common
-      control with that entity. For the purposes of this definition,
-      "control" means (i) the power, direct or indirect, to cause the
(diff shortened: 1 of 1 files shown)


sd2k:
## Test Failure Analysis

**Summary**: Test `test_callback_error_returns_html_page` times out on Windows due to SQLite locking when creating OAuthProxy without isolated storage.

**Root Cause**: The test creates an OAuthProxy instance without providing a `client_storage` parameter (line 1387 in `tests/server/auth/test_oauth_proxy.py`). When `client_storage` is `None`, OAuthProxy defaults to creating a DiskStore at `settings.home / "oauth-proxy"`. On Windows, when tests run in parallel, multiple tests trying to access the same SQLite database causes locking/timeout issues during the SQLite connection initialization.

The stack trace shows the timeout occurs at:
```python
File "diskcache/core.py", line 623, in _con
    con = self._local.con = sqlite3.connect(
```

**Suggested Solution**: Provide an isolated in-memory storage for this test to avoid SQLite file contention:

1. Modify `test_callback_error_returns_html_page` to use an in-memory key-value store
2. Add `client_storage` parameter to the OAuthProxy initialization

Example fix:
```python

async def test_callback_error_returns_html_page(self):
    """Test that OAuth callback errors return styled HTML instead of data: URLs."""
    from unittest.mock import Mock

    from starlette.requests import Request
    from starlette.responses import HTMLResponse

    from fastmcp.server.auth.oauth_proxy import OAuthProxy

    # Create a minimal OAuth proxy with isolated storage
    provider = OAuthProxy(
        upstream_authorization_endpoint="https://idp.example.com/authorize",
        upstream_token_endpoint="https://idp.example.com/token",
        upstream_client_id="test-client",
        upstream_client_secret="test-secret",
        token_verifier=JWTVerifier(
            jwks_uri="https://idp.example.com/.well-known/jwks.json",
            issuer="https://idp.example.com",
            audience="test-client",
        ),
        jwt_signing_key="test-signing-key",
        client_storage=MemoryStore(),  # Add this line
    )
    # ... rest of test
```

<details>
<summary>Detailed Analysis</summary>

The test was added/modified in this PR and exposed a Windows-specific issue that wasn't caught because:
1. Linux/macOS handle SQLite file locking differently than Windows
2. The test runs without `--numprocesses` on Windows (see workflow), so it's not parallelized but still conflicts with other tests from previous runs or shared state

Other tests in the file use the `oauth_proxy` fixture which also doesn't provide isolated storage, but this particular test might be more susceptible due to timing or ordering.

</details>

<details>
<summary>Related Files</summary>

- **tests/server/auth/test_oauth_proxy.py:1376** - Failing test that needs isolated storage
- **src/fastmcp/server/auth/oauth_proxy.py:821** - Where DiskStore is created by default
- **tests/server/auth/test_oauth_proxy.py:312** - `oauth_proxy` fixture that could also benefit from isolated storage

</details>

jlowin:
@sd2k https://github.com/modelcontextprotocol/python-sdk/pull/106

PR #99, opened 2025-04-07
Support mounting FastMCP instances as sub-MCPs

jlowin:
```python
main_app = FastMCP("MainApp")
sub_app = FastMCP("SubApp")

# Add a tool to the sub-app
@sub_app.tool()
def sub_tool() -> str:
    return "This is from the sub app"

# Mount the sub-app to the main app
main_app.mount("sub", sub_app)
```


aaazzam:
@DamyanBG this is addressed in [FastMCP 2.0 ](https://gofastmcp.com/servers/composition#mounting-live-linking)

DamyanBG:
@aaazzam This is nice implementation, but still not in the official SDK.

PR #123, opened 2025-04-12
Improve client return types

jlowin:
The Client class changes focus on API simplification and improved error handling:

1. **Unwrapped return values**: Methods now return direct results instead of wrapper objects
   - `list_resources()` returns `list[Resource]` instead of `ListResourcesResult`
   - `list_tools()` returns `list[Tool]` instead of `ListToolsResult` 
   - Similar pattern for other methods like `read_resource()`, `list_prompts()`, etc.

2. **Enhanced error handling**:
   - Added `ClientError` class as base exception type
   - `call_tool()` now raises `ClientError` with meaningful messages on failure

3. **Improved type safety**:
   - Added `@overload` annotations for `call_tool()` method
   - Better typing with `Literal` types for the raw result parameter

4. **New parameter options**:
   - Added `_return_raw_result` parameter to `call_tool()` for accessing underlying response data

5. **Commented out unused methods**:
   - Removed `subscribe_resource()` and `unsubscribe_resource()` functionality

These changes provide a more intuitive interface while maintaining compatibility with existing code.

### DIFF
--- a/src/fastmcp/client/client.py
+++ b/src/fastmcp/client/client.py
@@ -1,7 +1,7 @@
 import datetime
 from contextlib import AbstractAsyncContextManager
 from pathlib import Path
-from typing import Any
+from typing import Any, Literal, cast, overload
 
 import mcp.types
 from mcp import ClientSession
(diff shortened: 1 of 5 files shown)


regismesquita:
Hi @jlowin does it mean that this library doesn't support subscribing to resources? I am trying to implement that notification but I am having a hard time to understand the flow on the client.

adamkovari:
Hi @jlowin, I'm curious about the question of @regismesquita ,as well. Why wasn't implemented subscribe_resource() and unsubscribe_resource() functionality in fastmcp? Is there any update coming about these features?

Issue #174, opened 2025-04-15
Decorating class methods


strawgate:
I was thinking a bit more about the class method decoration problem and ran through a little POC on doing delayed registration. The rough idea is that the decorator gains a `delay_registration` flag (or we add specific decorators for class instance methods, or we add a heuristic for detecting wrapping of a class method) and when this is called, the decorator does not register the tool/resource/etc but instead marks the class method as needing registration.

Finally, the MCP server provides a `perform_delayed_registration` method which takes a prefix and the object, identifies any wrapped methods, and performs registration.

This allows the continued use of Decorators, which makes the experience between class methods and other setups similar, while enabling the wrapping of class methods and enabling the instantiation of several objects, with each registered under its own prefix.

Here's a sample implementation in FastMCP for playing around with:
https://github.com/jlowin/fastmcp/pull/175

The decorator:
```
        def decorator(fn: AnyFunction) -> AnyFunction:

             if delay_registration: # Mark for Delayed registration
                 fn.__perform_registration = lambda prefix, real_fn: self.add_tool(real_fn, name=f"{prefix}_{name or real_fn.__name__}", description=description, tags=tags)

             else: # Register the tool immediately
                 self.add_tool(fn, name=name, description=description, tags=tags)

             return fn
```

The method on the Server for registration:

```
    def perform_delayed_registration(self, prefix: str, object: object) -> None:
        """Perform delayed registration.

        This method is used to register tools/etc that were decorated with @tool(delay_registration=True).
        It allows them to be registered with a prefix that is determined at runtime.

        Args:
            prefix: The prefix to use for the registration
            object: The object containing methods marked with @tool(delay_registration=True)
        """
        
        methods_to_register = [
            getattr(object, method_name)
            for method_name in dir(object)
            if callable(getattr(object, method_name))
            if hasattr(getattr(object, method_name), "_FastMCP__perform_registration")
        ]
        for method_to_register in methods_to_register:
            method_to_register._FastMCP__perform_registration(prefix, method_to_register)
```

In he PR I provide a server that leverages this, which is what I took the screenshot of.

Here's a minimal viable implementation outside of FastMCP for playing around with. This sample is much more complex looking than the PR but is useful for confirming decorator behavior outside of FastMCP:
```
import inspect

# Global registry for tools
AVAILABLE_TOOLS = {}

print(f"Initial AVAILABLE_TOOLS: {AVAILABLE_TOOLS}")

def _register_item(key, item):
[...]

PR #175, opened 2025-04-15
[Idea] Delayed Registration of Instance Methods

strawgate:
Related to: https://github.com/jlowin/fastmcp/issues/174

Add `delay_registration` to the tool decorator (as an example) and a method on the MCP Server to perform delayed registration. 

See both objects get their own tools, which are callable, and return data from the data in the corresponding instance.
Screenshot 2025-04-15 at 5 55 22 PM

The `prefix` implementation is a bit sloppy, ideally prefix would be supported natively on the decorators but anyway

### DIFF
--- a/sample.py
+++ b/sample.py
@@ -1,0 +1,29 @@
+from src.fastmcp import FastMCP
+
+mcp = FastMCP()
+
+class Sample:
+    def __init__(self, name):
+        self.name = name
+
+    @mcp.tool(delay_registration=True)
+    def first_tool(self):
+        """First tool description."""
+        return f"Executed first tool {self.name}."
+    
+    @mcp.tool(delay_registration=True)
+    def second_tool(self):
+        """Second tool description."""
+        return f"Executed second tool {self.name}."
+    
+first_sample = Sample("First")
+second_sample = Sample("Second")
+
+mcp.perform_delayed_registration("first", first_sample)
+mcp.perform_delayed_registration("second", second_sample)
+
+def main():
+    mcp.run("sse")
+
+if __name__ == "__main__":
+    main()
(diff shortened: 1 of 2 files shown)


jlowin:
This feels to me like it has a lot of overloaded responsibility and extremely tight coupling -- the decorators are tied to a specific `mcp` instance, which needs to be invoked again later to call `mcp.perform_delayed_registration`. In the given sample, it seems preferable to simply call `mcp.tool()(first_sample.first)` explicitly which leaves no mystery about what's happening. It feels a bit outside the scope of the actual FastMCP responsibility. 

My suggestion instead is to create a class (below, `MCPRegisterable`) that has its own decorator (`mcp_tool`) which collects functions to be used as tools, and then have that class have a `register_tools()` method that applies those tools to any FastMCP instance. This will avoid the tight coupling and require no changes to FastMCP, while achieving the same goal with considerably greater flexibility

The pseudocode would be:

```python
from src.fastmcp import FastMCP
from some_module import MCPRegisterable, mcp_tool

mcp = FastMCP()

class Sample(MCPRegisterable):
    def __init__(self, name):
        self.name = name
        super().__init__()

    @mcp_tool()
    def first_tool(self):
        """First tool description."""
        return f"Executed first tool {self.name}."

    @mcp_tool()
    def second_tool(self):
        """Second tool description."""
        return f"Executed second tool {self.name}."

first_sample = Sample("First")
second_sample = Sample("Second")

first_sample.register_tools(mcp)
second_sample.register_tools(mcp)

def main():
    mcp.run("sse")

if __name__ == "__main__":
    main()
```

strawgate:
Thanks for taking a look and for the feedback. I'll take a look at making something like McpRegisterable and will follow-up unless you've already got an implementation in mind

PR #200, opened 2025-04-18
Fix missing parentheses in `@mcp.tool` decorator example

Minseok0917:
The example in the documentation shows the `@mcp.tool` decorator without parentheses, which leads to a runtime error:

`TypeError: The @tool decorator was used incorrectly. Did you forget to call it? Use @tool() instead of @tool`

This PR adds the missing parentheses (`@mcp.tool()`) in the usage example to avoid confusion and ensure the example works out of the box.

### DIFF
--- a/docs/getting-started/quickstart.mdx
+++ b/docs/getting-started/quickstart.mdx
@@ -32,7 +32,7 @@ from fastmcp import FastMCP
 
 mcp = FastMCP("My MCP Server")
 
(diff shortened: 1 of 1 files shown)


jlowin:
Thank you! That's a major oversight. I want to update the behavior to permit both but just haven't prioritized.

jlowin (approved):

PR #215, opened 2025-04-20
Example: Bulk Tool Caller

strawgate:
While I was working on reducing the size of the input schemas I realized that if I could just give the LLM a generic bulk tool caller on my server that I could get rid of almost all of my "bulk" operation tools. 

I thought the tool was neat and so would love to see if you're interested in having it as a Contrib item.

https://github.com/user-attachments/assets/12dfd0c8-da0f-4bbf-a87a-1d7c98743df6


jlowin (approved):

jlowin:
Cool idea

PR #220, opened 2025-04-21
fix issues with proxy mcp server

yihuang:
Fix two issues with proxy mcp server, let me know if you want to separate into two PRs:
1. tolerate METHOD_NOT_FOUND error, because the legacy server might not support all methods
2. Fix nested client context manager behavior which is important for proxy server use case.

  ```python
  with Client(...) as client:
    proxy_mcp = MCPServer.from_client(client)
    # operations will create nested context manager on client, and they should reuse the existing session.
    await proxy_mcp.list_tools()
  ```

### DIFF
--- a/src/fastmcp/server/proxy.py
+++ b/src/fastmcp/server/proxy.py
@@ -3,16 +3,16 @@ from urllib.parse import quote
 
 import mcp.types
 from mcp.server.lowlevel.helper_types import ReadResourceContents
+from mcp.shared.exceptions import McpError
 from mcp.types import (
+    METHOD_NOT_FOUND,
     BlobResourceContents,
     EmbeddedResource,
     GetPromptResult,
     ImageContent,
     TextContent,
     TextResourceContents,
-    METHOD_NOT_FOUND,
 )
-from mcp.shared.exceptions import McpError
 from pydantic.networks import AnyUrl
 
 from fastmcp.client import Client
--- a/src/fastmcp/server/proxy.py
+++ b/src/fastmcp/server/proxy.py
@@ -180,6 +180,8 @@ class FastMCPProxy(FastMCP):
             except McpError as e:
                 if e.error.code == METHOD_NOT_FOUND:
                     client_tools = []
+                else:
+                    raise e
             for tool in client_tools:
                 tool_proxy = await ProxyTool.from_client(self.client, tool)
                 tools[tool_proxy.name] = tool_proxy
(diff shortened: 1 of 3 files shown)


src/fastmcp/server/proxy.py
```diff
@@ -183,7 +190,12 @@ async def get_resources(self) -> dict[str, Resource]:
         resources = await super().get_resources()
 
         async with self.client:
-            for resource in await self.client.list_resources():
+            try:
+                client_resources = await self.client.list_resources()
+            except McpError as e:
+                if e.error.code == METHOD_NOT_FOUND:
+                    client_resources = []
```
yihuang, line 199:
```suggestion
                if e.error.code == METHOD_NOT_FOUND:
                    client_resources = []
                else:
                    raise e
```

src/fastmcp/server/proxy.py
```diff
@@ -193,7 +205,12 @@ async def get_resource_templates(self) -> dict[str, ResourceTemplate]:
         templates = await super().get_resource_templates()
 
         async with self.client:
-            for template in await self.client.list_resource_templates():
+            try:
+                client_templates = await self.client.list_resource_templates()
+            except McpError as e:
+                if e.error.code == METHOD_NOT_FOUND:
+                    client_templates = []
```
yihuang, line 216:
```suggestion
                if e.error.code == METHOD_NOT_FOUND:
                    client_templates = []
                else:
                    raise e
```

src/fastmcp/server/proxy.py
```diff
@@ -203,7 +220,12 @@ async def get_prompts(self) -> dict[str, Prompt]:
         prompts = await super().get_prompts()
 
         async with self.client:
-            for prompt in await self.client.list_prompts():
+            try:
+                client_prompts = await self.client.list_prompts()
+            except McpError as e:
+                if e.error.code == METHOD_NOT_FOUND:
+                    client_prompts = []
```
yihuang, line 233:
```suggestion
                if e.error.code == METHOD_NOT_FOUND:
                    client_prompts = []
                else:
                    raise e
```

tests/client/test_client.py
```diff
@@ -159,6 +164,38 @@ async def test_client_connection(fastmcp_server):
     # After connection
     assert not client.is_connected()
 
+async def test_client_nested_context_manager(fastmcp_server):
```
yihuang, line 168: this test case fail without the patch.

jlowin (approved):
Great, thank you! I have a PR to remove the mock in the test (seems to work ok) but can't push to your branch so I will add it in another PR

PR #222, opened 2025-04-21
fix fastmcp version ,python path valueError

lingfengchencn:
run fastmcp version  , when codes in other volume
```
ValueError: '/Volumes/Codes/../.venv/lib/python3.13' is not in the subpath of '/Users/<user home>'
```


jlowin (approved):
Thank you for catching the issue! Made a small change to make it even simpler, since we don't need 3 levels of parents for installed versions, and just knowing the location of the root is enough.

PR #237, opened 2025-04-24
small typing fixes

zzstoatzz:
not sure if `client` was meant to be exported top level but it seems `Client` should be from readme and the init

also was getting partially unknowns in `mount` because `FastMCP` is generic


jlowin (approved):
🏅

Made minor change to use PEP 649 annotations without strings

PR #260, opened 2025-04-25
Added feature : Load MCP server using config 

sandipan1:
The current Client supports loading server from ClientTransport , FastMCP server , script path, URL . It does not support config dictionary to add a server which is a standard way in many clients . 
I have added this feature for adding a single server through config dictionary. 

Example client.py 
```python
from fastmcp import Client
import asyncio
config = {
  "mcpServers": {
    "time": {
      "command": "uvx",
      "args": ["mcp-server-time", "--local-timezone=Asia/Kolkata"]
    }
  }
}
client = Client(config)
print(client.transport.args)
async def main():
    async with client:
        print(f"Client connected: {client.is_connected()}")

        print(f"Available tools: {[tool.name for tool in tools]}")
      
        if any(tool.name == "get_current_time" for tool in tools):
            result = await client.call_tool("get_current_time", {"timezone": "Asia/Kolkata"})
            print(f"Time result: {result[0].text}")

    print(f"Client connected: {client.is_connected()}")

if __name__ == "__main__":
    asyncio.run(main())
``` 

Output : 
```powershell
['mcp-server-time', '--local-timezone=Asia/Kolkata']
Client connected: True
Available tools: ['get_current_time', 'convert_time']
Time result: {
  "timezone": "Asia/Kolkata",
  "datetime": "2025-04-26T02:11:10+05:30",
  "is_dst": false
}
Client connected: False
```


jlowin:
Thanks! Great idea. I think this just needs some tests to ensure it works, nothing complex.

sandipan1:
Hey @jlowin  I have run a few example server tests on my local. Please let me know how to be more comprehensive about it

jlowin (approved):
Actually I'm realizing we dont have comprehensive transport tests right now so I will add those and include this. There is a small chance I will change it very slightly in testing but I don't anticipate that. Thank you for the contribution, this is really great functionality. I think I will also add a facility for starting from the CLI!

PR #276, opened 2025-04-29
create FastMCP Client from command string

huang-sh:
when input string is not path, consider it is a command

### DIFF
--- a/src/fastmcp/client/transports.py
+++ b/src/fastmcp/client/transports.py
@@ -431,14 +431,26 @@ def infer_transport(
     elif isinstance(transport, FastMCPServer):
         return FastMCPTransport(mcp=transport)
 
-    # the transport is a path to a script
-    elif isinstance(transport, Path | str) and Path(transport).exists():
-        if str(transport).endswith(".py"):
-            return PythonStdioTransport(script_path=transport)
-        elif str(transport).endswith(".js"):
-            return NodeStdioTransport(script_path=transport)
+    # the transport is command or path to a script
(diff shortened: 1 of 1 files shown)


jlowin:
I'm not crazy about blindly treating strings a shell commands, I think that could get into trouble pretty quickly. All the other inferences are fairly type bound: things that looks like paths are treated as paths; things that look like urls are treated as urls, things that look like servers are treated as in-memory servers. This would make a FastMCP client always effectively execute an arbitrary command which feels like a can of worms, especially because not all commands are stdio compatible -- let's think if there are ways to confine this. Otherwise I have no issue asking folks to import the correct transport for their use case.

huang-sh:
@jlowin Yes, I agree you. there are some complex servers, not simple single server file, maybe need this...although there are another way to call, but this one is fast and simple.
if you think strings is not good.  Just think input is shell command. and check if command  exist?

huang-sh:
Like .py or .js file,  just think they are stdio(I think they also maybe sse transport🤣....) and safe.

jlowin:
No, I don't see an advantage here - it is too magic. Inference works well if there is one clear option (e.g. a FastMCP instance must be in memory; a python path must be STDIO). Providing support for arbitrary commands and then running them as stdio creates a ton of really negative user experiences who provide commands that aren't starting stdio servers and end up with trapped error messages.

Without the clear hinting, and needing intense documentation to cover all the potential bad outcomes, it's difficult to argue that this is a better user experience than instantiating a StdioTransport itself.

PR #283, opened 2025-04-30
Ensure openapi tool responses are properly converted

### LINKED ISSUE #275: api endpoints returning lists of items cause error in openapi
### Description

Given there's an endpoint in openapi spec consisting of a list as root.
When the endpoint gets called and returns back the list
The response is not converted to TextContent and therefor throws an error

### Example Code

```Python

```

### Version Information

```Text
Looks like the issue is fixed by adding list here:
https://github.com/jlowin/fastmcp/blob/ee68d1429d4fc3cf10271e3b0a4abb1404db3a09/src/fastmcp/server/openapi.py#L594-L599
```

### Additional Context

_No response_

jlowin:
Closes #275, supersedes #277 

If an openapi tool returned a list, it was not properly converted to an MCP response. This ensures _convert_to_content is always called.

### DIFF
--- a/src/fastmcp/server/openapi.py
+++ b/src/fastmcp/server/openapi.py
@@ -10,12 +10,12 @@ from re import Pattern
 from typing import TYPE_CHECKING, Any, Literal
 
 import httpx
-from mcp.types import TextContent
+from mcp.types import EmbeddedResource, ImageContent, TextContent
 from pydantic.networks import AnyUrl
 
 from fastmcp.resources import Resource, ResourceTemplate
 from fastmcp.server.server import FastMCP
-from fastmcp.tools.tool import Tool
+from fastmcp.tools.tool import Tool, _convert_to_content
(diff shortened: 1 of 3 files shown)


src/fastmcp/server/openapi.py
```diff
@@ -239,9 +239,14 @@ async def _execute_request(self, *args, **kwargs):
...
-    async def run(self, arguments: dict[str, Any], context: Any = None) -> Any:
+    async def run(
+        self,
+        arguments: dict[str, Any],
+        context: Context[ServerSessionT, LifespanContextT] | None = None,
+    ) -> list[TextContent | ImageContent | EmbeddedResource]:
         """Run the tool with arguments and optional context."""
-        return await self._execute_request(**arguments, context=context)
+        response = await self._execute_request(**arguments, context=context)
+        return _convert_to_content(response)
```
jlowin, line 249: This is the primary change of this pR

PR #304, opened 2025-05-02
feat: Streamable HTTP client support

Akshit97:


jlowin:
Nice! Looks like this should be pretty close if not great once the upstream releases that client context

jlowin:
@Akshit97 I've pulled your PR's changes into #361, which also has the server-side changes

Akshit97:
@jlowin Would it be possible to merge this separately once the python-sdk is released? It would be helpful for me in my work organisation :)

jlowin:
Certainly

jlowin:
@Akshit97 going to merge and then bump pins to fix tests

PR #306, opened 2025-05-03
Add test coverage + tests on 4 examples

alainivars:
Refactor tests (remove duplicate). Add test coverage for simple_echo complex_inputs, desktop, echo.


jlowin:
Hi @alainivars -- lgtm though I don't see the test refactors in the PR. Did you mean to include? Or perhaps they were overwritten by the upstream 1.7.1 pin (I see the changes to add `annotations=None` which would have been superseded)

tests/test_examples.py
```diff
@@ -0,0 +1,113 @@
+"""Tests for example servers"""
+
+import pytest
+
+from mcp.shared.memory import (
+    create_connected_server_and_client_session as client_session,
```
jlowin: I'm generally not concerned with testing the example files in this way, as it requires essentially copying them and maintaining them twice (running the file via subprocess and ensuring it exits cleanly could suffice) but if you do, it needs to use a FastMCP `Client` rather than a low-level one, as the test needs to focus on FastMCP capabilities.

alainivars:
Hi @jlowin, the re-factorization was about to merge 2 similar test in one, the new test are include now.

tests/test_examples.py
alainivars: Yes you are right, I work on it.

alainivars: The tests for examples use now the fastmcp Client, and that even make the tests shorter. Thanks for the remark.

Issue #207, opened 2025-04-18
Allow alternative serialization to str when objects are returned


strawgate:
Right now there's a content formatter that gets called when tools return. Anything other than a string basically.

The content formatter just dumps it to JSON

It would be nice if this was customizable. At the moment I have to wrap all my tool return calls in a formatter call to return something other than compressed json

For example, I'd like to return yaml from tool calls as that tends to be readable for humans and for machines or alternatively maybe I'd like to return prettified json

Perhaps a configurable formatting callable that defaults to dumping json?

strawgate:
For now I'm doing:

```

        # We need to wrap the underlying call in a way that captures the response and then calls the formatter.
        def format_response(func: Callable) -> Callable:
            @functools.wraps(func)
            async def wrapper(*args, **kwargs):
                response = await func(*args, **kwargs)
                return self.response_formatter(response)

            return wrapper

        # Register the tools with the MCP server.
        mcp.add_tool(format_response(self.knowledge_base_client.create))
        mcp.add_tool(format_response(self.knowledge_base_client.get))
```

I would really like to switch to using the mixin and the decorators and will try to figure out if I can do this probably by nesting the decorators

strawgate:
@jlowin for many clients it would be nice to dump responses to pretty yaml or pretty json instead of flattened JSON. Especially if this could be configured as an env var (though that could be part of what the user implements)

For example in Roo Code, responses w/ objects come as a single line of json which is probably fine for responses that only an LLM needs to see, but if it's desirable for the response to be usable by both the user and the LLM then the MCP author needs to perform their own serialization and return strings instead of objects.

It would be really nice if there was some convenient way to customize the response serialization if an object is returned.

For example:

<img width="776" alt="Image" src="https://github.com/user-attachments/assets/364567ec-9691-4a9a-a23a-b58ce7bbdc4a" />

vs 

<img width="779" alt="Image" src="https://github.com/user-attachments/assets/370fb130-3a4f-48ee-84d9-a682ca25594b" />

I could perhaps offer a contrib module for this, perhaps a decorator which goes before the FastMCP decorator that changes the return type but I think a decorator that changes the return type breaks function docs/type hints in the IDE (in vscode at least)

In this way, we can continue to write methods and functions which take complex inputs and return complex outputs but when used as a tool serialize appropriately for the use-case without requiring custom wrapping or MCP-specific versions of each method/function

strawgate:
I'd be happy to offer a PR to either:
[...]

PR #308, opened 2025-05-03
Support custom Serializer for Tools

strawgate:
Example for #207

running the example prints:

```
Tool Result:
[TextContent(type='text', text='name: Test\nvalue: 123\nstatus: true\n', annotations=None)]
```

### DIFF
--- a/src/fastmcp/server/server.py
+++ b/src/fastmcp/server/server.py
@@ -205,6 +205,7 @@ class FastMCP(Generic[LifespanResultT]):
             | None
         ) = None,
         tags: set[str] | None = None,
+        serializer: Callable[[Any], str] | None = None,
         **settings: Any,
     ):
         self.tags: set[str] = tags or set()
@@ -214,6 +215,7 @@ class FastMCP(Generic[LifespanResultT]):
                 seconds=self.settings.cache_expiration_seconds
             )
         )
+        self._serializer = serializer
 
         self._mounted_servers: dict[str, MountedServer] = {}
 
@@ -226,7 +228,8 @@ class FastMCP(Generic[LifespanResultT]):
             lifespan=_lifespan_wrapper(self, lifespan),
         )
         self._tool_manager = ToolManager(
-            duplicate_behavior=self.settings.on_duplicate_tools
+            duplicate_behavior=self.settings.on_duplicate_tools,
+            serializer=self._serializer,
         )
         self._resource_manager = ResourceManager(
             duplicate_behavior=self.settings.on_duplicate_resources
--- a/src/fastmcp/server/server.py
+++ b/src/fastmcp/server/server.py
@@ -205,7 +205,7 @@ class FastMCP(Generic[LifespanResultT]):
             | None
         ) = None,
         tags: set[str] | None = None,
-        serializer: Callable[[Any], str] | None = None,
+        tool_serializer: Callable[[Any], str] | None = None,
         **settings: Any,
     ):
         self.tags: set[str] = tags or set()
@@ -215,7 +215,6 @@ class FastMCP(Generic[LifespanResultT]):
                 seconds=self.settings.cache_expiration_seconds
             )
         )
-        self._serializer = serializer
 
         self._mounted_servers: dict[str, MountedServer] = {}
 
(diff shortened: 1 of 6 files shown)


jlowin:
I think this is all possible (and considerably more customizable, flexible, and supported) by using Pydantic's existing functionality, so I'd prefer to lean on that. For example, this script permits the same outcome with no modification to FastMCP:

```python
import asyncio
import yaml
import pydantic

mcp = FastMCP()


# new base class with serializer
class PrettyPrintModel(pydantic.BaseModel):
    @pydantic.model_serializer
    def serialize(self) -> str:
        return yaml.dump(dict(self), width=100, sort_keys=False)


# pretty-printed custom model
class ReturnModel(PrettyPrintModel):
    x: int
    y: str
    
    
# model with custom serializer
class ReturnModel2(pydantic.BaseModel):
    x: int
    y: str

    @pydantic.model_serializer
    def serialize(self) -> str:
        return f"A return model with {self.x} and {self.y}"




@mcp.tool()
async def demo_pretty(x: int, y: str) -> ReturnModel:
    """Demonstrate a pretty-printed model."""
    return ReturnModel(x=x, y=y)

@mcp.tool()
async def demo_custom(x: int, y: str) -> ReturnModel2:
    """Demonstrate a custom serializer."""
    return ReturnModel2(x=x, y=y)


async def main():
        result1 = await client.call_tool("demo_pretty", dict(x=1, y="hello"))
        print(f"Pretty result: {result1}")
        result2 = await client.call_tool("demo_custom", dict(x=1, y="hello"))
        print(f"Custom result: {result2}")


if __name__ == "__main__":
    asyncio.run(main())
```

strawgate:
That's certainly true but it comes with some challenges -- I believe that in pydantic, model serializers returning strings is a pretty blunt tool and doing this would break many other use cases for model dump as generally you expect a model to serialize to a dictionary (which then can be separately serialized to a string).

My overall goal is to be able to build a server that I wrap with FastMCP and so customizing models in a way that I think breaks other pydantic functionality is less than ideal. 

I totally understand if this is not functionality you're interested in making configurable but I think the overall argument is that if the code is going to call a serializer with specific settings like indent, etc that serializer should be customizable.

jlowin:
Ah, ok ok I understand better now. I am a little nervous about the "blunt instrument" of making serialization globally configurable but I agree with you now it is the best solution, and preserves the "scalpel" of Pydantic serializers for users who want it.

src/fastmcp/server/server.py
```diff
@@ -205,6 +205,7 @@ def __init__(
             | None
         ) = None,
         tags: set[str] | None = None,
+        serializer: Callable[[Any], str] | None = None,
```
jlowin: Minor nit, but could we call this `tool_serializer` because (at least at this time) I think it only applies to tools as resources/prompts have much more fixed representations. I could imagine a different resource_serializer being specified for e.g. converting resources bytes to some other format in the future (not rushing to implement that!)

I think the arg is great staying as `serializer` on the tool / tool manager classes because obviously already scoped to tools, so this is only a comment for the server kwarg itself

src/fastmcp/server/server.py
```diff
@@ -214,6 +215,7 @@ def __init__(
                 seconds=self.settings.cache_expiration_seconds
             )
         )
+        self._serializer = serializer
```
jlowin: I think this could be removed safely?

src/fastmcp/tools/tool.py
```diff
@@ -142,6 +147,7 @@ def __eq__(self, other: object) -> bool:
 def _convert_to_content(
     result: Any,
     _process_as_single_item: bool = False,
+    _serializer: Callable[[Any], str] | None = None,
```
jlowin: This could be `serializer` (no leading underscore) and appear before `_process_as_single_item` in the kwargs order -- the reason `_process_as_single_item` has the underscore is because its only meant to be called in recursive calls to `_convert_to_content` and is essentially a "private" implementation detail of the function itself.

jlowin (commented):
Thanks @strawgate! Left some extremely minor comments on variable names

strawgate:
Perhaps limiting it to text_content_object_serializer or moving the serializer into a module var that can be monkey patched?

strawgate:
addressed feedback

jlowin:
LGTM, thanks @strawgate! Made small push to get typing to pass and should be good to go

PR #324, opened 2025-05-05
Update README.md

kaddynator:


jlowin:
In the readme I'm trying to minimize boilerplate as much as possible to illustrate each component; the linked documentation is completely runnable and should apply the (proper) standard of every code block being standalone-runnable. Here I'm not sure these changes are even helpful - why import httpx and create a BaseModel just to show how to decorate a function?

PR #361, opened 2025-05-07
Streamable HTTP support

jlowin:
This implementation is based on changes introduced to the low-level server in https://github.com/modelcontextprotocol/python-sdk/pull/641, and requires the new objects introduced there. However we need to do more work to abstract out auth and routes to avoid polluting the `FastMCP` object further, as it is at risk of becoming as configuration-heavy as the low-level server. Since Streamable HTTP is so popular, it's ok to merge this and risk some breaking changes in auth (which can be made backwards compatible via deprecation) in order to get to a more sane configuration setup. 

Opening as draft until 641 is merged. Supersedes #304

### DIFF
--- a/src/fastmcp/client/transports.py
+++ b/src/fastmcp/client/transports.py
@@ -18,6 +18,7 @@ from mcp.client.session import (
 )
 from mcp.client.sse import sse_client
 from mcp.client.stdio import stdio_client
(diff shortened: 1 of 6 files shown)


Akshit97:
Nice work @jlowin, I was thinking of implementing this today :)

PR #379, opened 2025-05-09
Fix tool result serialization when the tool returns a list

### LINKED ISSUE #378: Tool serializer is not passed during recursive convert content calls
### Description

If you send a list of objects the custom tool serializer is never called

### Example Code

```Python

```

### Version Information

```Text
na
```

### Additional Context

_No response_

strawgate:
Fixes #378


jlowin (commented):
Good catch!

PR #385, opened 2025-05-09
Fixing various text typos

didier-durand:
Hi,

As title says, fixing various typos found when reading the docs.

For details, see commit diff.

BTW, fastmcp is great! Thks for that

Didier


didier-durand:
My pleasure!

PR #387, opened 2025-05-09
Adding a few tests to Image type

didier-durand:
Hi again,

Added some tests to Image type to improve code coverage. They pass ok when run.


jlowin (approved):
Thank you!

PR #388, opened 2025-05-09
Adding tests for TimedCache

didier-durand:
Hi, 

Added tests for TimedCache class

Best,

Didier


jlowin (approved):
Thank you! Cleaned up some small typing issues and LGTM

PR #413, opened 2025-05-12
logging stack trace for easier debugging

jbkoh:
When an exception happens in a tool, most of the details are dropped including stack traces, which makes debugging pretty challenging. The PR is to add a logging for exceptions at the low level to capture what's happening properly.

I'd be happy to update it to generalize the solution if the maintainer can give more guidance.

### DIFF
--- a/src/fastmcp/tools/tool.py
+++ b/src/fastmcp/tools/tool.py
@@ -143,6 +143,7 @@ class Tool(BaseModel):
 
             return _convert_to_content(result, serializer=self.serializer)
         except Exception as e:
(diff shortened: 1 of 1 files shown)


jlowin:
@jbkoh typically I'd log the traceback here but I see you declined the copilot suggestion -- ok with it either way and can always enhance. Thanks!

jbkoh:
Thanks! I ignored the copilot suggestion as `logger.exception` automatically captures current exceptions afaik. Thanks for accepting my PR!

PR #417, opened 2025-05-12
Test Cases for Two Bugs in FastMCP

DanielAvdar:
This PR adds test cases for two bugs in the FastMCP library:

## 1. Resource Template Bug with Special Characters in Parameter Names

The current implementation of resource templates has issues with special characters in parameter names. The tests demonstrate:
- Parameter names with hyphens cause matching issues
- URL encoding/decoding of parameter values needs to be handled consistently

## 2. SSE Transport Bug in Nested Starlette Applications

The current SSE server transport has issues when used in deeply nested Starlette applications:
- The root_path is not properly considered when constructing event stream URLs
- Nested apps in Starlette's Mount components break SSE connections

## Tests Added

1. `tests/resources/test_resource_template_special_chars.py` - Tests demonstrating the resource template bug
2. `tests/resources/test_resource_template_fix.py` - Tests verifying the resource template fix
3. `tests/client/test_buggy_sse_nesting.py` - Tests reproducing the SSE nesting bug
4. `tests/client/test_deep_sse_nesting.py` - Tests for deeply nested SSE paths

## Custom Test Targets

Added custom test targets to the Makefile for easy testing:
- `test-resource-templates`
- `test-resource-template-fix`
- `test-sse-bug`

These tests fail on the current codebase, demonstrating the bugs. Fixes will be provided in a separate PR.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,31 @@
+# Makefile for common project tasks
+
+.PHONY: install deps test lint format docs test-resource-templates test-sse-bug test-resource-template-fix
+
+install:
+	@echo "Installing all dependencies via uv..."
+	uv sync --all-extras --all-groups --frozen
+	uv run pre-commit install
+
+test: 
+	@echo "Running tests..."
+	uv run pytest tests
+
+test-resource-templates:
+	@echo "Running resource template tests with special characters in parameters..."
+	uv run pytest tests/resources/test_resource_template_special_chars.py -v
+
+test-resource-template-fix:
+	@echo "Running resource template fix tests..."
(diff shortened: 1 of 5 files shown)


jlowin:
This is duplicative of #421 ?

DanielAvdar:
> This is duplicative of #421 ?

Yes, this is indeed a duplicate of the next PR that includes the fix. However, this PR focuses solely on the tests to ensure the issue can be reproduced independently.

Let me know if you need further clarification!

jlowin:
Closing alongside #421 for similar reasons. PRs must be self-contained.

Issue #439, opened 2025-05-13
Emit "tool list changed notifications" when tools are added / removed


jlowin:
### Enhancement Description

When a server is running, adding or removing tools should emit change notifications, as described in #433 and #437

### Proposed Implementation

```Python

```

davenpi:
I’ve been digging into this and I think the missing piece lives in the MCP SDK. FastMCP now declares the `toolsChanged` capability, and the low-level `ServerSession` has a `send_tool_list_changed()` method—but there’s no public API on the `Server` to access active sessions or emit this notification when the tool list changes.

I can open an issue on the SDK side proposing an API addition (e.g., `send_tools_list_changed_notification()` on `Server`) unless you think there’s a better approach. Happy to coordinate!

PR #450, opened 2025-05-14
Declare toolsChanged capability for stdio server.

davenpi:
### Enable `toolsChanged` capability for stdio server

This PR updates the stdio version of the `FastMCP` server to declare the `tools/list_changed` capability, as defined in the MCP spec.  Related to #439. 

### Why this matters
- The [MCP spec](https://modelcontextprotocol.io/specification/2025-03-26/server/tools#list-changed-notification) states: *"When the list of available tools changes, servers that declared the `listChanged` capability SHOULD send a notification."*
- Declaring `listChanged = True` signals to clients that this server may notify them when the tool list changes.
- It does not require immediate notification support, nor does it change any current behavior.

### What this doesn’t do
- The server still does not send `tools/list_changed` notifications when tools are added or removed.
- Emitting those would require access to an active session, which is not currently available.
- That could be added in a follow-up PR.

Totally fine to close if you’d rather bundle this with notification support later. I mainly wanted to surface the capability gap now.

### DIFF
--- a/src/fastmcp/server/server.py
+++ b/src/fastmcp/server/server.py
@@ -19,7 +19,7 @@ import pydantic
 import uvicorn
 from mcp.server.auth.provider import OAuthAuthorizationServerProvider
 from mcp.server.lowlevel.helper_types import ReadResourceContents
-from mcp.server.lowlevel.server import LifespanResultT
+from mcp.server.lowlevel.server import LifespanResultT, NotificationOptions
 from mcp.server.lowlevel.server import Server as MCPServer
 from mcp.server.stdio import stdio_server
 from mcp.types import (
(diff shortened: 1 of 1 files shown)


jlowin:
It's frustrating/odd that the low-level SDK has this passed at a moment when the server isn't available always to us (e.g. in the STDIO server you obviously have access as in this PR but in the SSE/SHTTP ones its buried in implementation-specific code). This may be a moment where it's useful for us to actually modify the low-level server as if it took init kwargs e.g. overwriting this https://github.com/modelcontextprotocol/python-sdk/blob/main/src/mcp/server/lowlevel/server.py#L146. 

Let's get this merged for now because its a good enhancement and then look at more holistic approaches.

davenpi:
Thanks for merging. I haven’t figured out a clean path for emitting notifications yet, but happy to keep thinking about it or follow along if something broader comes together, especially around HTTP transport.

PR #475, opened 2025-05-16
feat: Add middleware option in server.run

Maxi91f:
I'm using a middleware on a project, and I could use a quicker way to add one from run instead of creating a new app using `mcp.http_app`.

LMK if there's something I forgot to do here.


zzstoatzz (approved):
lgtm! thanks @Maxi91f

PR #477, opened 2025-05-16
doc(asgi): Change custom route example to PlainTextResponse

mcw0933:
## What

Updates the doc examples of custom routes to use `PlainTextResponse` instead of `JSONResponse`

## Why

`streamable_http` expects the client to connect with a header for `text/event-stream` - so if you actually use the old example, you'll fail when trying to return JSON from Starlette/FastAPI with this error message:

```
{
  "jsonrpc": "2.0",
  "id": "server-error",
  "error": {
    "code": -32600,
    "message": "Not Acceptable: Client must accept text/event-stream"
  }
}
```

which makes it seem like a client issue, but it's actually just the response body not being plain text.


jlowin (approved):
Good catch, thank you!

docs/deployment/running-server.mdx
```diff
@@ -260,13 +260,13 @@ You can also add custom web routes to your FastMCP server, which will be exposed
...
-from starlette.responses import JSONResponse
+from starlette.responses import PlainTextResponse
 
 mcp = FastMCP("MyServer")
 
 @mcp.custom_route("/health", methods=["GET"])
-async def health_check(request: Request) -> JSONResponse:
-    return JSONResponse({"status": "healthy"})
+async def health_check(request: Request) -> PlainTextResponse:
+    return PlainTextResponse("OK")return JSONResponse({"status": "healthy"})
```
zzstoatzz:
```suggestion
async def health_check(request: Request) -> PlainTextResponse:
    return PlainTextResponse("OK")
```

mcw0933:
No prob!  Love your documentation!   ❤️   We're out here rolling our MCP server now and FastMCP has been a godsend 😁 

Any chance you will be at PyCon this weekend?

jlowin:
Unfortunately not this one but maybe we’ll meet another time!

PR #478, opened 2025-05-16
Introduce MCP client oauth flow

jlowin:
This is a first cut at adding oauth support to fastmcp clients, with the goal of enabling e.g. `fastmcp run https://mcp.paypal.com/sse` from the CLI (or Claude Desktop / Cursor / local client of choice) to interact with any remote + authenticated server. 

This needs careful review for spec compliance.

Quick demo script:

```python
from rich import print
from fastmcp import Client
from fastmcp.client.transports import SSETransport

async with Client("https://mcp.paypal.com/sse") as client:
    await client.ping()
    print(tools)
```

### DIFF
--- a/src/fastmcp/client/auth/httpx_client.py
+++ b/src/fastmcp/client/auth/httpx_client.py
@@ -1,0 +1,424 @@
+from __future__ import annotations
+
+import asyncio
+import socket
+import time
+from collections.abc import AsyncIterator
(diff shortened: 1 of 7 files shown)


jlowin:
MCP # 734 would allow us to avoid the ugly patch

juan-abia:
https://github.com/modelcontextprotocol/python-sdk/pull/752 was just merged. It allows customisation to avoid the ugly path

PR #479, opened 2025-05-16
Doc(quickstart): Fix import statements

mai-nakagawa:
- Removing the unused import statement
- Adding the missing import statement


zzstoatzz (approved):
thanks @mai-nakagawa !

Issue #480, opened 2025-05-16
make `session_manager` as a property of the streamable HTTP app


softyoungha:
### Enhancement Description

Hi
we're creating a streamable HTTP app and mounting it onto our main fastapi. 
However, when registering a sub-application with our fastapi, the lifespan o sub-app does not execute.

FastMCP includes an internal session_manager, which is supposed to run within the lifespan of sub-app. 
But since the lifespan does not propagated when registered as a sub-app, it throws an error stating that `session_manager.run()` needs to be executed.
> fastapi lifespan events - [link](https://fastapi.tiangolo.com/advanced/events/#sub-applications)

When using `from fastmcp import FastMCP`, session_manager is defined within a function, making it inaccessible. 
https://github.com/jlowin/fastmcp/blob/a31b4233583ebda511d5a791561e036dd4fcfe3a/src/fastmcp/server/http.py#L251-L292

But in the MCP repository, the session_manager is defined as a property, so it can be accessed directly.
 
- [link](https://github.com/modelcontextprotocol/python-sdk/blob/6353dd192c41b891ef3bf1dfc093db46f6e2175a/src/mcp/server/fastmcp/server.py#L180-L213) 

Therefore, to solve this issue, I am using the import `from mcp` instead of `fastmcp`. 

```python
import contextlib
from typing import Optional, AsyncIterator
from mcp.server.fastmcp.server import FastMCP   # not fastmcp
from fastapi import FastAPI

mcp_app = FastMCP()

@contextlib.asynccontextmanager
async def mcp_lifespan(app: Starlette) -> AsyncIterator[None]:
    async with mcp_app.session_manager.run():
        yield
        

# fastapi app
app = FastAPI(
    ...,
    lifespan=mcp_lifespan
)

app.mount("/", mcp_app.streamable_http_app(), name="mcp")
```

### Proposed Implementation

```Python
It seems like the session_manager could be defined as a property in FastMCP as well. 
Would it be okay to contribute towards this?
```

jlowin:
We can consider this but the situation is explicitly documented [here](https://gofastmcp.com/deployment/asgi#starlette-integration) and shows how to access the lifespan context without writing a new async context manager, which is necessary if we only expose the session manager as a attribute

jlowin:
See #483 for an alternative that I think avoids these downsides - considering this

softyoungha:
[...]

PR #481, opened 2025-05-16
make StreamableHTTPSessionManager as property of FastMCP

softyoungha:
- this fixes https://github.com/jlowin/fastmcp/issues/480

Here's the concept for the code. 
If it's in the right direction, I'll add the test code.

### DIFF
--- a/src/fastmcp/server/http.py
+++ b/src/fastmcp/server/http.py
@@ -284,18 +284,19 @@ def create_streamable_http_app(
     server_middleware: list[Middleware] = []
 
     # Create session manager using the provided event store
-    session_manager = StreamableHTTPSessionManager(
-        app=server._mcp_server,
-        event_store=event_store,
-        json_response=json_response,
-        stateless=stateless_http,
-    )
+    if server._session_manager is None:
+        server._session_manager = StreamableHTTPSessionManager(
+            app=server._mcp_server,
+            event_store=event_store,
+            json_response=json_response,
+            stateless=stateless_http,
+        )
 
     # Create the ASGI handler
     async def handle_streamable_http(
         scope: Scope, receive: Receive, send: Send
     ) -> None:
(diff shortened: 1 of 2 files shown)


src/fastmcp/server/server.py
```diff
@@ -179,6 +181,25 @@ def name(self) -> str:
...
+            RuntimeError: If called before create_streamable_http_app() has been called.
+        """
+        if self._session_manager is None:
+            raise RuntimeError(
+                "Session manager can only be accessed after"
+                "calling create_streamable_http_app()."
+                "The session manager is created lazily"
+                "to avoid unnecessary initialization."
+            )
+        return self._session_manager
```
softyoungha, line 202: - [link](https://github.com/modelcontextprotocol/python-sdk/blob/6353dd192c41b891ef3bf1dfc093db46f6e2175a/src/mcp/server/fastmcp/server.py#L196-L213) 

The session_manager is defined exactly the same way as the property in MCP. 
Since it was copied as is, the doc message might need some modifications.

src/fastmcp/server/http.py
```diff
@@ -284,18 +284,19 @@ def create_streamable_http_app(
...
-        json_response=json_response,
-        stateless=stateless_http,
-    )
+    if server._session_manager is None:
+        server._session_manager = StreamableHTTPSessionManager(
+            app=server._mcp_server,
+            event_store=event_store,
+            json_response=json_response,
+            stateless=stateless_http,
+        )
```
softyoungha, line 293: I'm hesitant about setting the private `_session_manager` outside of the class,
but it seems impossible without refactoring since it's currently separated into a different function

jlowin:
I have some hesitation about this approach but agree with making this more discoverable - I'd prefer not to go down this route but have a (hopefully simpler!) alternative in #483

Issue #474, opened 2025-05-16
fix flaky test


zzstoatzz:
### Description

https://github.com/jlowin/fastmcp/actions/runs/15057691794/job/42326885053

### Example Code

```python
FAILED tests/client/test_client.py::TestTimeout::test_timeout - mcp.shared.exceptions.McpError: Timed out while waiting for response to ClientRequest. Waited 0.01 seconds.
================== 1 failed, 882 passed, 3 skipped in 18.20s ==================
```

### Version Information

```Text
main
```

zzstoatzz:
should be closed by #486

PR #486, opened 2025-05-16
Fix flaky timeout test by increasing timeout (#474)

davenpi:
This PR addresses #474 by increasing the client timeout in `TestTimeout::test_timeout` from 0.01s to 0.05s.

**Background:**  
With the original 0.01s timeout, the test would occasionally fail due to event loop scheduling and system timing, as confirmed by running:

```sh
pytest tests/client/test_client.py::TestTimeout::test_timeout --count=5000
```

Results:
- timeout=0.01s: 1 failure in 5,000 runs
- timeout=0.05s: 0 failures in 5,000 runs

**Rationale:**  
A slightly longer timeout gives the event loop enough time to enforce it reliably, making the test more robust.

### DIFF
--- a/tests/client/test_client.py
+++ b/tests/client/test_client.py
@@ -512,11 +512,11 @@ class TestErrorHandling:
 class TestTimeout:
     async def test_timeout(self, fastmcp_server: FastMCP):
         async with Client(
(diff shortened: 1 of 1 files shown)


jlowin (approved):
Thanks!

PR #509, opened 2025-05-20
Store the initialize result on the client

### LINKED ISSUE #500: Feature request: expose server capabilities
### Enhancement Description

hello, currently the MCP Clients initialize themselves, but do not store the capabilities of the server anywhere. It would be useful to be able to grab those.

Consider the implementation of `StreamableHttpTransport`:

```python
    @contextlib.asynccontextmanager
    async def connect_session() -> AsyncIterator[ClientSession]:
        async with streamablehttp_client(...) as transport:
            async with ClientSession(...) as session:
                await session.initialize() # <-- the capabilities are discarded here
                yield session
```

jlowin:
Closes #500

When creating a client session, we now store the initialize result so users can access it as `client.initialize_result` within the `async with Client()` context manager. Outside the client, it will error.

### DIFF
--- a/src/fastmcp/client/client.py
+++ b/src/fastmcp/client/client.py
@@ -1,5 +1,5 @@
 import datetime
-from contextlib import AsyncExitStack
+from contextlib import AsyncExitStack, asynccontextmanager
 from pathlib import Path
 from typing import Any, cast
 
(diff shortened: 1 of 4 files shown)


davenpi:
Nice! I was thinking along similar lines, but didn’t consider the context management improvements. One side effect is that the `initialize` request now gets request ID 1, which matches the spec’s connection lifecycle—I noticed you updated the tests to reflect this.

jlowin:
Ah I was trying to figure out why that changed and whether it was deterministic! Is that an issue? I'm surprised by that

davenpi:
I was also a little surprised—I’m not totally sure how context injection works yet. My understanding is that client requests deterministically increment the counter, so now that `initialize` is always the first request, it takes ID 1. I’d be interested to see if there’s a scenario where this could cause problems.

jlowin:
The thing I don't understand is we were making the initialize request before so why didn't it increment? All we did is move the request location. But it's ok

PR #514, opened 2025-05-20
Skip timeout tests on windows

jlowin:
@zzstoatzz what do you think? These tests are flaky. I do wonder if it's a timing thing, like if we increased the timeouts to ~1 second it would have enough time to properly respond, but I don't want to add such overhead to the test suite. Maybe only on windows we use higher timeouts? Happy to merge this now regardless.


zzstoatzz (approved):
lgtm

PR #515, opened 2025-05-20
Added PyProject Reading to Dev mode to ignore multiple `--with` arguments.

ZiadAmerr:


jlowin:
@ZiadAmerr we already have similar functionality in `fastmcp dev` via the `--with-editable` kwarg, which has the additional benefit of ensuring that the inspector always reflects in-flight changes to your code:

```
fastmcp dev --with-editable path/to/server/repo
```

If there's another use case you have that isn't met by `--with-editable` I'd rather use `uv`'s -r flag rather than adding our own dependency parsing logic.

ZiadAmerr:
Hey @jlowin. Would you mind taking a look at this PR?

I have noticed [this part of the documentation](https://gofastmcp.com/deployment/cli#options:~:text=This%20command%20runs%20your%20server%20in%20an%20isolated%20environment.%20All%20dependencies%20must%20be%20explicitly%20specified%20using%20the%20%2D%2Dwith%20and/or%20%2D%2Dwith%2Deditable%20options.) and thought of improving it a bit.

ZiadAmerr:
I don't think it works as expected, or am I running it incorrectly?

The MCP server seems to be working normally with run but not with dev.

image


It also fails even with `--with-editable` argument is specified.
image

jlowin:
No that's definitely unexpected 
image


Please open an issue if you can create an MRE

PR #538, opened 2025-05-21
Fix all_routes_as_tools route map getting overwritten

SebastianRubina:
- `route_maps` in `openapi.py` were something like this if using all_routes_as_tools in `server.py`:
```
[
    RouteMap(
        methods=['GET'],
        pattern=r'.*\{.*\}.*',
        route_type=RouteType.RESOURCE_TEMPLATE
    ),
    RouteMap(
        methods=['GET'],
        pattern=r'.*',
        route_type=RouteType.RESOURCE
    ),
    RouteMap(
        methods=['POST', 'PUT', 'PATCH', 'DELETE', 'OPTIONS', 'HEAD'],
        pattern=r'.*',
        route_type=RouteType.TOOL
    ),
    RouteMap(
        methods='*',
        pattern=r'.*',
        route_type=RouteType.TOOL
    ),
]
```
- Although we had the `methods='*'`    to set as TOOL, the others were still taking precedence, so `GET` endpoints were still being assigned as `RESOURCE`.
- Changed this to ONLY use the `methods='*'` RouteMap if all_routes_as_tools.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -55,7 +55,6 @@ dmypy.json
 .python-version
 .envrc
 .direnv/
-lib
 
 # Logs and databases
 *.log
--- a/.gitignore
+++ b/.gitignore
@@ -22,6 +22,7 @@ venv/
 env/
 ENV/
 .env
+*venv
 
 # System files
 .DS_Store
@@ -54,6 +55,7 @@ dmypy.json
 .python-version
 .envrc
 .direnv/
+lib
 
 # Logs and databases
 *.log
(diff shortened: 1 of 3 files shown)


jlowin:
I'm a little confused -- route_maps are always added before the default routes https://github.com/jlowin/fastmcp/blob/4437594e1f1b0dfc441461d5f2942ac86c650152/src/fastmcp/server/openapi.py#L605 and we have an entire test class that affirms that GET routes are properly handled (https://github.com/jlowin/fastmcp/blob/4437594e1f1b0dfc441461d5f2942ac86c650152/tests/server/test_openapi.py#L2031-L2154). 

Can you create an MRE that shows the specific concern as an issue?

Additionally, `all_routes_as_tools` is meant to be a user-facing shortcut on the main FastMCP class's constructor classmethods, where this type of sugar is appropriate. I would prefer to avoid it on the FastMCPOpenAI class, which should only have logic for explicit RouteMaps.

PR #563, opened 2025-05-22
Add client method to send cancellation notifications

davenpi:
### Summary

- Implements `Client.cancel()` to emit `notifications/cancelled` as defined in the [spec](https://modelcontextprotocol.io/specification/2025-03-26/basic/utilities/cancellation#cancellation)

- Allows clients to request cancellation of in-progress requests with an optional reason.

- No tests added, matching current notification method coverage.

Happy to close if there's another way to do this. Didn't see the option anywhere else.


jlowin (approved):
LGTM - to my knowledge this doesnt have server side support yet but no reason to leave it out on the client!

PR #570, opened 2025-05-23
FastAPI integration: add support to forward headers from MCP Server to API endpoints

asaintsever:
This PR allows to define a list of headers we would like the MCP Server to forward to the API endpoints.

By default the feature is disabled. It can be enabled via the settings (and env var `FASTMCP_SERVER_PASSTHROUGH_HEADERS`).

For e.g., to enable it using an `.env` file:

```
FASTMCP_SERVER_PASSTHROUGH_HEADERS=True
FASTMCP_SERVER_PASSTHROUGH_HEADERS_LIST='["Authorization", "x-account-id"]'
```

### DIFF
--- a/src/fastmcp/server/server.py
+++ b/src/fastmcp/server/server.py
@@ -1171,12 +1171,16 @@ class FastMCP(Generic[LifespanResultT]):
             **settings,
         )
 
+    _cached_server_settings: fastmcp.settings.ServerSettings | None = None
+
     @staticmethod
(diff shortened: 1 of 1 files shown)


jlowin:
Thanks @asaintsever ! I'm going to close this in favor of #575, which is more holistic. It adds the ability to pass client kwargs to the underlying httpx client, so if you prefer that approach (as in this PR) you can do it!

Issue #604, opened 2025-05-26
MCP Servers have 1 second to come online


strawgate:
### Description

The context manager on Client is responsible for bring MCP servers online and [currently gives MCP servers 1s to respond before failing startup](https://github.com/jlowin/fastmcp/blob/c27385da78595fc4a47bef689f259317a3739904/src/fastmcp/client/client.py#L171-L173):

```
    @asynccontextmanager
    async def _context_manager(self):
        with catch(get_catch_handlers()):
            async with self.transport.connect_session(
                **self._session_kwargs
            ) as session:
                self._session = session
                # Initialize the session
                try:
                    with anyio.fail_after(1):
                        self._initialize_result = await self._session.initialize()
                    yield
                except TimeoutError:
                    raise RuntimeError("Failed to initialize server session")
                finally:
                    self._exit_stack = None
                    self._session = None
                    self._initialize_result = None

```

### Example Code

```Python

```

### Version Information

```Text
2.5.1
```

jlowin:
Someone raised a similar issue - I thought the initialize call was a rapid request response exchange of capabilities but I think that’s not correct!

strawgate:
So I often just point uvx at whole repos and so it has to fetch the repo, get dependencies,  build the package, etc and that's what it is failing during

It becomes much more annoying when I'm on coffee shop wifi

Issue #546, opened 2025-05-22
session initialization fails for longer spawning processes


sakalys:
### Description

https://github.com/jlowin/fastmcp/pull/523/files#diff-1bcb7e49a0d8ca93f1e30298eff7ff7c0b3ce68ed57cee274f77253d98651ac6R163

This part simply prevents sessions from initializing if they take longer to boot. Reproducible with github's npx 
MCP server.

This looks like a breaking change also.

As a fix for that the timeout should be configurable and the default value should be something in the tens.

### Example Code

```Python

```

### Version Information

```Text
fastmcp version: 2.4.0
```

jlowin:
That’sa shame, my understanding was that initialization is a rapid request response handshake

jlowin:
@sakalys can you provide an MRE or confirm that it is the initialization call itself that's timing out, and not that the server is failing to start for another reason? As far as I can tell there is no way to put long-running logic in the initialize call.

sakalys:
Simply plug in `docker run --rm -i --cpus=0.1 --init node:24-slim npx -y @modelcontextprotocol/server-github` as a server. (Note the `--cpus=`)

JasonGao180:
+1
same issue with npx mcp-remote

my workaround:
1. rollback to 2.3.5
2. write my own config parser

strawgate:
oops yeah i have the same issuehttps://github.com/jlowin/fastmcp/issues/604

jfouret:
Same issue here, although a default init timeout of 1 would make sense since most mcp server shall be ready fast.
An easy workaround is to setup the env before and in the config just run the env.

When it's just installing the env, a workaround is easy, however there should be case, in complex mcp tools, where the main script needs to setup things before to spawn the server (database, dll models...).

PR #607, opened 2025-05-27
add init_timeout for mcp client

jfouret:
relative to #604  and #546


jfouret:
@jlowin ready to merge, I think ?

jlowin:
Thanks @jfouret -- I can't push to your branch so I've added #609 which I'll merge immediately after this one to add global setting dependnecy and relax the default to disable the timeout.

Issue #612, opened 2025-05-27
Support throwing `McpError` (errors with codes)


andrewgazelka:
### Enhancement Description

Problem:
- Right now there is no way to return an error with a custom code even though it is part of the JSON RPC spec.

Alternatives:
- returning a non-error JSON

### Use Case

Being able to have errors that can be handled programatically. i.e., an error of a certain code might indicate retrying / etc if there is a proxy between the agent and the MCP server.

### Proposed Implementation

Right now McpError is defined in `mcp` as 

```Python

class McpError(Exception):
    """
    Exception type raised when an error arrives over an MCP connection.
    """

    error: ErrorData

    def __init__(self, error: ErrorData):
        """Initialize McpError."""
        super().__init__(error.message)
        self.error = error

class ErrorData(BaseModel):
    """Error information for JSON-RPC error responses."""

    code: int
    """The error type that occurred."""

    message: str
    """
    A short description of the error. The message SHOULD be limited to a concise single
    sentence.
    """

    data: Any | None = None
    """
    Additional information about the error. The value of this member is defined by the
    sender (e.g. detailed error information, nested errors etc.).
    """

    model_config = ConfigDict(extra="allow")

```
I propose being able to throw `McpErrors` and it being returned back to MCP without getting converted to a `ToolError` and losing its code.

zzstoatzz:
closing per https://github.com/jlowin/fastmcp/pull/614#issuecomment-2917887231

lmk if you have a different assessment @andrewgazelka !

PR #614, opened 2025-05-28
allow raising `McpError`

zzstoatzz:
closes https://github.com/jlowin/fastmcp/issues/612

see https://modelcontextprotocol.io/docs/concepts/transports#responses

### DIFF
--- a/src/fastmcp/server/server.py
+++ b/src/fastmcp/server/server.py
@@ -18,6 +18,7 @@ from typing import TYPE_CHECKING, Any, Generic, Literal
 import anyio
 import httpx
 import uvicorn
+from mcp import McpError
 from mcp.server.auth.provider import OAuthAuthorizationServerProvider
 from mcp.server.lowlevel.helper_types import ReadResourceContents
 from mcp.server.lowlevel.server import LifespanResultT, NotificationOptions
@@ -25,9 +26,12 @@ from mcp.server.lowlevel.server import Server as MCPServer
 from mcp.server.stdio import stdio_server
 from mcp.types import (
     AnyFunction,
+    CallToolRequest,
+    CallToolResult,
     EmbeddedResource,
     GetPromptResult,
     ImageContent,
+    ServerResult,
     TextContent,
     ToolAnnotations,
 )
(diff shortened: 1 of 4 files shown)


andrewgazelka:
note the client side still has some issues with respect to this. I do not know if there is a way to catch an `McpError` and look at its code but I haven't really looked into a lot.

jlowin:
@zzstoatzz following #550 you may want to test that this error makes it all the way out to the client (see `test_client.py` in https://github.com/jlowin/fastmcp/pull/550/files#diff-f91d7d1bfecb41d80a4cc73133b82560b7284a2af2d6630c49b983ca9366ea13) -- I think you've put the error trap in the right place but there are different higher-level places where errors are trapped so I just want to make sure we ensure MCPErrors survive a full client request/response round trip

src/fastmcp/server/server.py
```diff
@@ -1304,6 +1310,30 @@ def from_client(cls, client: Client, **settings: Any) -> FastMCPProxy:
...
+            # Let McpError bubble up to _handle_request for JSON-RPC error response
+            raise
+        except Exception as e:
+            # Keep backward compatibility for other exceptions
+            return ServerResult(
+                CallToolResult(
+                    content=[TextContent(type="text", text=str(e))],
+                    isError=True,
+                )
+            )
```
zzstoatzz, line 1335: this maybe should happen upstream

zzstoatzz:
@jlowin how does [this](https://github.com/jlowin/fastmcp/pull/614/commits/daf1184e2e96027a7bdaf329c13f0b5d0caab823) look?

zzstoatzz:
hrm ok thanks to chatting with @jlowin - i think the original issue and subsequently I misunderstood that the request/response linked in the issue is the wire format, not the user facing format and [this](https://modelcontextprotocol.io/specification/2025-03-26/server/tools#calling-tools) is the protocol for tools which seems consistent with [this](https://github.com/modelcontextprotocol/python-sdk/blob/main/src/mcp/server/lowlevel/server.py#L415-L427) impl upstream

PR #619, opened 2025-05-28
 fix for issue number #601 Client's Header is getting replaced by mcp_header in OpenAPI implementation when API is deployed remote

mansura-habiba:


mansura-habiba:
#601

jlowin:
This PR seems to address a very specific and idiosyncratic case - please see comments on the initial report

PR #623, opened 2025-05-29
Improve type inference from client transport

jlowin:
```python
from pydantic import AnyUrl
from fastmcp import Client
from fastmcp.client.transports import StdioTransport


client = Client(transport=AnyUrl("http://localhost:8000"))
client.transport # inferred: SSETransport | StreamableHttpTransport


client = Client(transport=StdioTransport(command="python", args=[""]))
client.transport # inferred: StdioTransport
```

### DIFF
--- a/src/fastmcp/client/client.py
+++ b/src/fastmcp/client/client.py
@@ -125,8 +125,6 @@ class Client(Generic[ClientTransportT]):
         instance = super().__new__(cls)
         return instance
 
-    transport: ClientTransportT
-
     def __init__(
         self,
         transport: ClientTransportT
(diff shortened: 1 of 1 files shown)


Oaklight:
```
Call id_FED1 error: Error executing math_server-add: Client.__new__() missing 1 required positional argument: 'transport'
```

This error started to pop up after upgrading to fastmcp >= 2.5.2
My codebase was developed since fastmcp < 2.2, and everything worked fine until recently, when I upgraded the fastmcp to the latest.
After trying every release, I identified that the breaking changes happened in 2.5.2

@jlowin

PR #625, opened 2025-05-29
Support for uppercase letters at the log level

ksawaray:
Previously, fastmcp commands using uppercase log levels would fail to execute.

This update adds support for uppercase letters in log level specifications.

e.g.
fastmcp run --transport streamable-http --host 0.0.0.0 -p 8000 --log-level DEBUG examples/echo.py


zzstoatzz (approved):
ty @ksawaray !

Issue #630, opened 2025-05-29
Some methods of FastMCP does not need to be async


DamyanBG:
### Enhancement Description

I checked the codebase and some methods of FastMCP does not need to be async.

### Proposed Implementation

```Python

```

jlowin:
FastMCP subclasses implement these as async

PR #631, opened 2025-05-29
remove not needed async

DamyanBG:
#630


jlowin:
FastMCP subclasses implement these as async

DamyanBG:
@jlowin, get out from the vide coding and take a look... Even in subclasses you can remove the "await" when you are calling these methods.

jlowin:
Excuse me? Please do not attempt to participate in this community again. 

The proxy server subclass makes async client calls to remote servers in order to implement those methods.

PR #684, opened 2025-06-03
Fix : fix path error to CLI Documentation

yrangana:
CLI Documentation path was wrongly pointed under servers/development, fixed it to patterns


jlowin:
Thank you!

PR #708, opened 2025-06-04
Fix pr labeler permissions

jlowin:
We are sometimes seeing permissions error in the PR labeler like

```
Error: HttpError: You do not have permission to create labels on this repository.: {"resource":"Repository","field":"label","code":"unauthorized"} Error: You do not have permission to create labels on this repository.: {"resource":"Repository",
```

Based on https://github.com/actions/labeler/issues/870 and https://github.com/orgs/community/discussions/156181 it seems like the solution is to add this additional `issues: write` permission

cc @zzstoatzz

### DIFF
--- a/.github/workflows/labeler.yml
+++ b/.github/workflows/labeler.yml
@@ -7,6 +7,7 @@ jobs:
     permissions:
       contents: read
       pull-requests: write
+      issues: write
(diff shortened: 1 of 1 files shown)


zzstoatzz:
oh nice find!

PR #723, opened 2025-06-05
Fix links in README.md

alainivars:
Just a quick fix 2 links provided in the README.md was wrong.


jlowin:
Thanks, overlooked those!

PR #734, opened 2025-06-06
Docs : fix client to mcp_client in Gemini example

yrangana:
fixed `async with client:` to `async with mcp_client:`


jlowin (approved):
Thanks for catching!

PR #752, opened 2025-06-08
chore: fix prompt docstring

danb27:
@jlowin Thanks for all the amazing work. 

I found a small issue in the documentation, likely caused by a bulk find+replace operation in this recent [commit](https://github.com/jlowin/fastmcp/commit/099d3402082b1649d3aa4698e17cd446b8918bae).


jlowin:
Good catch @danb27, we just added that syntax and must have overlooked it!

PR #803, opened 2025-06-11
Ensure http args are passed through

### LINKED ISSUE #799: `json_response` and `stateless_http` parameters ignored in 2.8.0
### Description

If you look at the code, the `json_response` and `stateless_http` function arguments are never read:

https://github.com/jlowin/fastmcp/blob/57be94e93e0ef1ff4547817ae141f4080bd5367d/src/fastmcp/server/server.py#L1293-L1332

Instead, the code reads `self._deprecated_settings`:

https://github.com/jlowin/fastmcp/blob/57be94e93e0ef1ff4547817ae141f4080bd5367d/src/fastmcp/server/server.py#L1319-L1320

I also tested it by running the app, and I did see it use stateful connection on 2.8.0, despite setting `stateless_http=True`.

### Example Code

```Python

```

### Version Information

```Text
FastMCP 2.8.0
```

### Additional Context

_No response_

jlowin:
Closes #799

### DIFF
--- a/src/fastmcp/server/server.py
+++ b/src/fastmcp/server/server.py
@@ -1332,8 +1332,12 @@ class FastMCP(Generic[LifespanResultT]):
                 or self._deprecated_settings.streamable_http_path,
                 event_store=None,
                 auth=self.auth,
-                json_response=self._deprecated_settings.json_response,
-                stateless_http=self._deprecated_settings.stateless_http,
+                json_response=(
+                    json_response or self._deprecated_settings.json_response
+                ),
+                stateless_http=(
+                    stateless_http or self._deprecated_settings.stateless_http
+                ),
                 debug=self._deprecated_settings.debug,
                 middleware=middleware,
             )


src/fastmcp/server/server.py
```diff
@@ -1332,8 +1332,12 @@ def http_app(
                 or self._deprecated_settings.streamable_http_path,
                 event_store=None,
                 auth=self.auth,
-                json_response=self._deprecated_settings.json_response,
-                stateless_http=self._deprecated_settings.stateless_http,
+                json_response=(
+                    json_response or self._deprecated_settings.json_response
```
kkom, line 1336: i think if you pass `json_response=False` the value will be ignored and `self._deprecated_settings.json_response` will be used instead

`json_response if json_response is not None else self._deprecated_settings.json_response` would be safer!

jlowin: 🤦 of course

PR #833, opened 2025-06-14
Add audio support

### LINKED ISSUE #832: Audio as tool output
### Enhancement Description

I see the `Image` is implemented, but not `Audio`.

jlowin:
- Requires mcp >= 1.9.4
- Adds `fastmcp.utilities.types.Audio`, which is analogous to `fastmcp.utilities.types.Image`
- Deprecates top-level Image import, which we've needed to do for a while

Closes #832


jlowin:
I can't replicate the test error and suspect it is because of CI latency failing to start the server before the 0.01 second timeout

jlowin:
It's unclear what the issue is. I'm going to merge the feature and explore the fix separately.

PR #838, opened 2025-06-15
fix: report_progress missing passing related_request_id causes notifications not working in streaming-http

alexsee:
This pull request fixes an issue related to `report_progress` that misses passing the `related_request_id` which causes notifications not working in `stateless_http=True` mode.

This issue was also reported for the FastMCP implementation in the Python MCP SDK: https://github.com/modelcontextprotocol/python-sdk/issues/953

I also added a test for `stateless_http=True` mode introducing a parameterized pytest (not sure if you prefer having this in a separate file or keep it there).

### DIFF
--- a/tests/client/test_streamable_http.py
+++ b/tests/client/test_streamable_http.py
@@ -99,8 +99,12 @@ def run_nested_server(host: str, port: int) -> None:
 
 
 @pytest.fixture()
-async def streamable_http_server(stateless_http: bool = False) -> AsyncGenerator[str, None]:
-    with run_server_in_process(run_server, stateless_http=stateless_http, transport="streamable-http") as url:
+async def streamable_http_server(
+    stateless_http: bool = False,
+) -> AsyncGenerator[str, None]:
+    with run_server_in_process(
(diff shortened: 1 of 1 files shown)


jlowin:
Thank you!

jlowin:
I wonder why the low-level doesn't fill that in by default 🤔 But appreciate the fix!

alexsee:
> I wonder why the low-level doesn't fill that in by default

Thats a very good question and if the server is run in stateful mode, the requests are handled differently through the `GET_stream` (SSE) that is established. In that case, the `related_request_id` is not used.

For other logging methods, the `related_request_id` is actually set and it was added as part of the streaming-http implementation:
- https://github.com/modelcontextprotocol/python-sdk/blob/f3cd20c9200de001fb58f83a130ab83c6a6ed5fd/src/mcp/server/fastmcp/server.py#L976
- https://github.com/modelcontextprotocol/python-sdk/pull/553

For the context class in FastMCP (in this repo), these are also missing as I see.

What I'm not entirely sure is if there are any side effects that I may have introduced due to this change.

PR #842, opened 2025-06-16
Add version parameter to FastMCP constructor

mkyutani:
## Summary

This PR adds a `version` parameter to the `FastMCP` constructor, allowing users to specify their own server version in the `serverInfo` field of JSON-RPC responses.

## Problem

Currently, FastMCP always uses the MCP library version in the `serverInfo.version` field, which doesn't allow users to specify their own package version. This is inconsistent with the MCP Python SDK's `Server` class, which accepts a `version` parameter.

## Solution

- Added `version: str | None = None` parameter to `FastMCP.__init__()`
- Pass the version parameter to the underlying `MCPServer` initialization
- Maintains full backward compatibility - when no version is specified, it defaults to the MCP library version

## Changes Made

- **src/fastmcp/server/server.py**: Added version parameter to FastMCP constructor and MCPServer initialization
- **tests/test_server.py**: Added comprehensive test that verifies both custom version setting and backward compatibility

## Testing

- All existing tests pass
- New test verifies that custom version appears in serverInfo
- Backward compatibility confirmed when no version is specified

## Example Usage

```python
# With custom version
app = FastMCP("MyApp", version="2.1.0")

# Without version (backward compatible)
app = FastMCP("MyApp")  # Uses MCP library version
```

This change enables FastMCP users to properly version their MCP servers while maintaining full backward compatibility.

### DIFF
--- a/tests/client/test_client.py
+++ b/tests/client/test_client.py
@@ -341,7 +341,9 @@ async def test_server_info_custom_version():
         assert result.serverInfo.name == "DefaultVersionServer"
         # Should fall back to MCP library version
         assert result.serverInfo.version is not None
-        assert result.serverInfo.version != "1.2.3"  # Should be different from custom version
+        assert (
+            result.serverInfo.version != "1.2.3"
+        )  # Should be different from custom version
 
 
 async def test_client_nested_context_manager(fastmcp_server):


jlowin:
Thanks! I'm adding a minor change because I think name/instructions are common positional args, so adding version as the second arg might break some people's servers. I can't push to your branch but will open a separate PR

PR #843, opened 2025-06-16
Add File utility for wrapping binary data

gorocode:
Not entirely sure if this fits within the intended scope of the repo, but I ran into a need to embed arbitrary binary files in the same way `Audio` and `Image` resources are handled.

This PR introduces a `File` helper class, following the same structure and conventions as `Audio` and `Image`, to make it easier to return binary content as an `EmbeddedResource`. It supports both file paths and raw byte data, infers MIME types, and handles base64 encoding internally.

### What’s Included

- `File` class to wrap binary content (from path or raw data) and encode it as base64
- MIME type inference from extension or explicit format
- Optional `annotations` support
- `.to_resource_content()` method to generate an `EmbeddedResource`, automatically choosing between `TextResourceContents` or `BlobResourceContents` based on the data
- Tests for key usage scenarios (path, data, annotations)
- Example included to show typical usage

### Notes

- No breaking changes
- Mirrors the existing `Audio` and `Image` helpers
- Happy to adjust if the approach doesn’t quite fit the intended design


jlowin:
That's great, thanks!

Issue #875, opened 2025-06-19
Context.session get session_id


triangle959:
### Enhancement Description

If multiple tools are used to process data, but each tool needs to pass a large amount of data as parameters for the tool, it can be a very troublesome thing. 
Can we use **session** to manage and ensure that each tool can process the same batch of data after accessing the same session? Of course, my idea is to use session ID as the key and store the data in Redis or other places

### Proposed Implementation

```Python

```

PR #881, opened 2025-06-19
Add session_id property to Context for data sharing

jlowin:
## Summary

Adds `session_id` property to FastMCP's `Context` class to enable session-based data sharing between tools. This addresses the Redis-based data sharing architecture requested in #875.

```python
@server.tool
def store_data(data: dict, ctx: Context) -> str:
    if session_id := ctx.session_id:
        redis_client.set(f"session:{session_id}:data", json.dumps(data))
        return f"Data stored for session {session_id}"
    return "No session available (stdio/memory transport)"
```

## Key Changes

- **New property**: `Context.session_id` returns MCP session ID from HTTP headers
- **Transport aware**: Returns `None` for stdio/memory transports (per MCP spec)
- **HTTP only**: Works with SSE and StreamableHTTP transports that use session IDs
- **Thread safe**: Uses existing `get_http_headers()` utility

## Session ID Behavior

| Transport | Session ID | Use Case |
|-----------|------------|----------|
| HTTP (SSE/StreamableHTTP) | ✅ UUID | Web APIs, data sharing |
| Stdio | ❌ None | CLI tools |
| In-memory | ❌ None | Testing |

## Test Coverage

- ✅ Session ID extraction from HTTP headers
- ✅ Graceful fallback for non-HTTP transports  
- ✅ Missing header handling
- ✅ Integration test with real server

This enables the exact data sharing workflow described in #875 and addresses persistent community requests as mentioned in https://github.com/modelcontextprotocol/python-sdk/issues/986, https://github.com/modelcontextprotocol/python-sdk/issues/942, and https://github.com/modelcontextprotocol/python-sdk/issues/485.

### DIFF
--- a/src/fastmcp/server/context.py
+++ b/src/fastmcp/server/context.py
@@ -181,15 +181,15 @@ class Context:
     @property
     def session_id(self) -> str | None:
         """Get the MCP session ID for HTTP transports.
-        
+
         Returns the session ID that can be used as a key for session-based
         data storage (e.g., Redis) to share data between tool calls within
         the same client session.
-        
+
         Returns:
             The session ID for HTTP transports (SSE, StreamableHTTP), or None
             for stdio and in-memory transports which don't use session IDs.
-            
+
         Example:
             ```python
             @server.tool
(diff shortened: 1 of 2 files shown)


qqaatw:
Hi @jlowin, thank you for adding the session concept to the library. 

Do we have the ability to set the `session_id` as a response for the client with some api? From the code changes in this PR, it looks read-only.

I suppose that session_id is set by the server in most practices.

PR #885, opened 2025-06-20
Enable and disable component endpoints

gorocode:
🔄 Changes Summary

> `server.py` – Added enable and disable route controllers
> `http.py` – Introduced set_up_component_management_routes() to register new endpoints
> `tools_manager.py`, `resource_manager.py`, `prompt_manager.py` – Added enable and disable methods
> `Docs` – Extended usage documentation:
>
> >  docs/servers/tools.mdx
> >  docs/servers/resources.mdx
> >  docs/servers/prompts.mdx  
>  
> `Tests` – Added test_component_management_routes.py for full coverage (auth, non-auth, mount scenarios)

This PR exposes new HTTP endpoints for enabling and disabling tools, resources, and prompts in FastMCP.

Previously, these actions were only possible from within the application code. With this update, they can now be triggered externally via authenticated (or unauthenticated) API calls — depending on the server’s configuration.

### 🛠️ New API Endpoints
- `POST /tools/{tool_name}/enable`
- `POST /tools/{tool_name}/disable`
- `POST /resources/{uri:path}/enable`
- `POST /resources/{uri:path}/disable`
- `POST /prompts/{prompt_name}/enable`
- `POST /prompts/{prompt_name}/disable`

>  Templates are also supported under `/resources` path
>  Mounted servers are supported too

These routes are separated from the existing/custom path routes to avoid mixing them with standard MCP protocol behavior. This separation keeps the codebase modular and organized, making it easy to migrate these endpoints under the MCP protocol in the future if such functionality is added to the standard.

### 🔐 Optional Authentication
These endpoints reuse the existing authentication logic, so protection via scopes is available when the server is configured with authentication.

### 🧪 Tests
Added comprehensive test coverage in `tests/server/test_component_management_routes.py`. It covers non-auth and auth posibilities and mounted servers too.

### 📘 Docs
Usage examples and endpoint details have been added to:
- `docs/servers/tools.mdx`
- `docs/servers/resources.mdx`
- `docs/servers/prompts.mdx`


src/fastmcp/prompts/prompt_manager.py
```diff
@@ -111,6 +111,58 @@ async def list_prompts(self) -> list[Prompt]:
         prompts_dict = await self._load_prompts(via_server=True)
         return list(prompts_dict.values())
 
+    async def enable_prompt(self, key: str) -> Prompt:
```
jlowin, line 114: Note: I would prefer not to add these methods, instead users should do e.g. 

```
mcp = FastMCP()

@mcp.tool
def my_tool():
    pass

tool = await mcp.get_tool('my_tool')
tool.disable()
```

jlowin:
I appreciate the thought here, however I think this needs a little consideration. Even on an authenticated server, this gives any authenticated user the power to modify the server behavior, which is almost certainly not going to be desirable. Also these changes are *global* so if a user enables/disables a tool, it's that way for everyone. This is too much power to expose by default.

src/fastmcp/prompts/prompt_manager.py
gorocode: When I use `get_tool`, it works as expected for tools from the main server — I can disable them and the change sticks. But for tools coming from mounted servers, it seems like `get_tool` returns a copy. You can still use the tool from this copy, but calling `disable()` doesn’t affect the actual tool.

I found this in documentation:

> `import_server`: For a one-time copy of components with prefixing (static composition).
> `mount`: For creating a live link where the main server delegates requests to the subserver (dynamic composition).
> Dynamic Updates: Changes to the mounted server are immediately reflected when accessed through the parent.  

This seems to suggest that updates should propagate, so I’m wondering if I might be missing something in how I’m using it.

```python
mcp = FastMCP()
mount = FastMCP()

mcp.mount(mount, prefix="mon")

@mount.tool
def my_tool():
    pass
    
tool = await mcp.get_tool('mon_my_tool')
tool.disable()
```

gorocode:
Thanks so much for the thoughtful feedback — that totally makes sense, and I really appreciate the guidance. I'm still fairly new to contributing to open source, so apologies if this PR was a bit overreaching.

I'll go ahead and refactor this into a module with opt-in behavior and a more robust and configurable permission system, so it's clearly separated from core functionality and only used when explicitly enabled.

Thanks for sharing your work — I’m learning a lot from this!

jlowin:
Hey @gorocode not overreaching at all! Just opening more surface area than we want right now :) 

Since I'm not sure of how to think about this becoming first-class and won't be able to fully accept until we know how it aligns with e.g. other exposing admin routes, my recommendation is to package it as a `contrib` module for now, which will relax the acceptance criteria a bit

PR #892, opened 2025-06-20
Fix JWT issuer validation to support string values per RFC 7519

### LINKED ISSUE #890: FastMCP JWT issuer claim enforcement doesn't match JWT Specification
### Description

The FastMCP issue strictly enforces the issuer to a valid URL String format when the specification references a String OR a URL: https://www.rfc-editor.org/rfc/rfc7519.html#section-4.1.1. The prevents developers from using the BearerAuthProvider to validate JWT Tokens where the issuer uses a string instead of a URL String. Example: "my-iss". Please reference the JWT Specification: https://www.rfc-editor.org/rfc/rfc7519.html#section-4.1.1

### Example Code

```Python

```

### Version Information

```Text
https://www.rfc-editor.org/rfc/rfc7519.html#section-4.1.1
```

### Additional Context

_No response_

jlowin:
## Summary
Fixes JWT issuer validation to allow string values in addition to URLs, per RFC 7519 specification. Resolves #890 where `BearerAuthProvider` rejected valid tokens with non-URL issuer claims like `"my-service"`.

## Solution
Works around the underlying MCP SDK's URL validation requirement by using a default URL for the parent class while preserving the original string for JWT claim validation. This maintains RFC 7519 compliance without breaking existing functionality.

## Test Plan
- ✅ String issuers work correctly
- ✅ URL issuers continue to work as before  
- ✅ All existing authentication tests pass

### DIFF
--- a/src/fastmcp/server/auth/providers/bearer.py
+++ b/src/fastmcp/server/auth/providers/bearer.py
@@ -17,7 +17,7 @@ from mcp.shared.auth import (
     OAuthClientInformationFull,
     OAuthToken,
 )
-from pydantic import AnyHttpUrl, SecretStr
+from pydantic import AnyHttpUrl, SecretStr, ValidationError
 
 from fastmcp.server.auth.auth import (
     ClientRegistrationOptions,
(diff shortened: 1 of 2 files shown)


sstene1:
How long before this is available? No pressure :)

PR #977, opened 2025-06-27
Add OpenAPI extensions support to HTTPRoute

maddymanu:
**Context**

We have a very complicated `openapi.json` file, that is used for various purposes. 
Including our documentation. 

We wanna add support for things like 
- `x-cortex-mcp-enabled: False`
- `x-cortex-mcp-context: This is an llm description..`
etc.

The OpenAPI spec (and Pydantic models) already support this:
https://swagger.io/docs/specification/v3_0/openapi-extensions/

In this PR, I add support for said `extensions` and tests :)


jlowin:
Thanks!

PR #985, opened 2025-06-29
Update link to OpenAI MCP example

mossbanay:
I noticed that this repository has been deleted or renamed and doesn't seem to redirect. I'm not sure if this cupcake MCP server is the same as the original repo but still looks like a perfectly good example of an MCP server implementing `search` & `fetch` tools.


jlowin:
Thank you! It looks like it has indeed moved and there are a few -- let me merge this and then revisit how to best direct.

PR #1007, opened 2025-07-01
FEAT: SPIKE - OAuth proxy

jessesanford:
This PR adds an experimental capability that allows for fastmcp to be used with upstream oauth authentication providers that do not support DCR similar to the design of fastapi_mcp's `setup_proxies=true` flag: https://github.com/tadata-org/fastapi_mcp/blob/main/examples/09_auth_example_auth0.py#L126

See more info about that design decision here: https://fastapi-mcp.tadata.com/advanced/auth#setup-proxies%3Dtrue

I have not tested this thoroughly, nor have I done a in-depth security audit of these changes, but in essence, it's just manipulating some of the 3LO request/responses and the token minting/client registration is all done with the upstream auth server.


jlowin:
Also... that FastAPI-MCP documentation looks suspiciously familiar. Beside the styling, looks like some content is taken directly from our docs?? Sort of shameless.

src/fastmcp/server/auth/providers/transparent_proxy.py
```diff
@@ -0,0 +1,511 @@
...
+        # Map relations for rotation / revocation
+        self._access_to_refresh: dict[str, str] = {}
+        self._refresh_to_access: dict[str, str] = {}
+
+        # Validate signature and expiration using upstream JWKS, but do NOT enforce
+        # issuer matching because the tokens were issued by the upstream provider
+        # (e.g., "https://developer.api.autodesk.com"), not by this proxy.
+        self._bearer_validator = BearerAuthProvider(
+            jwks_uri=self._upstream_jwks_uri,
+            issuer=None,
```
dacamposol, line 116: Since we have the `issuer_url` from the upstream Authorization Server as parameter for the initialization of the proxy, why not adding it here?

src/fastmcp/server/auth/providers/transparent_proxy.py
```diff
@@ -0,0 +1,511 @@
...
+        # Merge the supplied client metadata (redirect URIs, scopes, etc.) with
+        # the fixed credentials.
+        enriched = OAuthClientInformationFull(  # type: ignore[call-arg]
+            **client_info.model_dump(
+                exclude={"client_id", "client_secret", "grant_types", "token_endpoint_auth_method"}
+            ),
+            client_id=upstream_id,
+            client_secret=upstream_secret,
+            grant_types=client_info.grant_types or ["authorization_code", "refresh_token"],
+            token_endpoint_auth_method="none",
```
dacamposol, line 188: 🔴 Clients with `token_endpoint_auth_method` set to `none` are assumed to **do not have a client secret**. This includes applications using authorization code flow with PKCE or implicit flow, typical in mobile apps that can't securely store secrets.

If the MCP Client relies solely on this field to determine the authentication method against the client, it risks failing.

A better approach would be for the MCP Server to explicitly set the correct `token_endpoint_auth_method` when provisioning the proxy. This shifts responsibility to the developer or integrator to specify whether the client should authenticate with `client_secret_basic`, `client_secret_post`, or `client_secret_jwt`, depending on the actual security model of the OAuth 2.0 client application.

jlowin:
Going to close this as stale - some of the functionality has been incorporated elsewhere; the rest will emerge shortly. Thank you for contributing!

PR #1017, opened 2025-07-02
Docs: add /sse/ URL note for chatgpt

thap2331:


jlowin (changes requested):
Thanks but this is not necessarily correct - /sse/ is the default URL for SSE transport only, but users could change it (and it is the wrong url for streamable http, which is what this example doc uses). I think it would be better to say something like "the full URL to your MCP server" and mention including the path which defaults to /sse/ for SSE transport and /mcp/ for MCP transport

thap2331:
> Thanks but this is not necessarily correct - /sse/ is the default URL for SSE transport only, but users could change it (and it is the wrong url for streamable http, which is what this example doc uses). I think it would be better to say something like "the full URL to your MCP server" and mention including the path which defaults to /sse/ for SSE transport and /mcp/ for MCP transport

Thanks for catching that! I realized the error a bit later and was planning to update it—sorry for the inconvenience. Also, I just updated the doc.

jlowin (approved):
Thanks!

PR #1022, opened 2025-07-03
Proxy support advanced MCP features

hopeful0:
### Summary

This PR introduces a `ProxyClient` class, enabling the `FastMCPProxy` to support advanced MCP features. These features include:

- Roots
- Sampling
- Elicitation
- Logging
- Progress

This change allows the proxy to seamlessly forward these advanced requests from the remote client to the proxy, creating a more feature-complete proxy. Corresponding integration tests have been added by adapting existing tests for these features from the client-side test suite, simply redirecting them to target the new proxy service.

### Motivation

The proxy is a compelling feature within FastMCP, offering a powerful mechanism for integrating services from multiple upstream MCP sources. However, the current implementation limits the integration of tools that rely on advanced MCP features, such as sampling or elicitation. These capabilities were effectively hidden because the simple client within the proxy did not handle them.

This PR addresses this limitation by introducing a custom proxy client (`ProxyClient`) designed specifically to forward these advanced feature requests. By doing so, it exposes the full capabilities of the upstream server to any client connected to the proxy. This enhancement is achieved without altering the existing proxy logic for core features, ensuring backward compatibility while making the proxy a more powerful and transparent component in a FastMCP architecture.

---

### Checklist

-   [x] Implemented `ProxyClient` to enable advanced MCP features in the proxy.
-   [x] Added handlers to forward `roots`, `sampling`, `elicitation`, `log`, and `progress` requests.
-   [x] Added integration tests for the new proxy capabilities in `tests/server/proxy_client/`.
-   [x] Modified the default client in `mount`, `as_proxy` and `test_proxy.py`.
-   [x] All existing and new tests pass.
-   [x] Code is documented with comments.


jlowin (changes requested):
This is a fantastic feature, thank you!

Some minor comments:
- your proxy server tests are a little overcomplete. Not that too many tests is a bad thing, but these go beyond testing the ProxyClient's ability to forward interactions. For example, we need to test that the elicitation handler passes inputs to the ultimate client, and passes responses back to the ultimate server. All the tests for e.g. response types and all possible parameterizations are unecessary and only increase the surface area of change friction for those features. I'd suggest a single tests/server/proxy/test_proxy_client.py file with just a handful of tests ensuring that all arguments are passed up and down the proxy chain. As long as you are confident arguments are faithfully passed on, then testing the actual arguments should be left to the rest of the codebase.
- This is a very powerful feature and it may be worth adding brief documentation to explain how proxy servers handle this type of forwarded interaction. At a minimum, explain the difference between a `ProxyClient` and a `Client` and explain that users should prefer ProxyClients (and FastMCP will automatically create them) unless they have specific reason not to.

hopeful0:
Thank you for the positive feedback and for the detailed comments! I'm glad you find this feature useful.

Here are my thoughts on your suggestions:

1.  **Proxy Server Tests:** I agree that the initial test suite was a bit overcomplete. My thought process was to be thorough in ensuring that the proxy could correctly forward complex and special types. For example, in elicitations, the `action` parameter can take on special values like `accept` which require careful handling during forwarding. My initial comprehensive testing was designed to discover such cases. Having identified these, I will now refactor `tests/server/proxy/test_proxy_client.py` as you suggested, retaining only the essential tests that confirm arguments are passed faithfully through the proxy chain.

2.  **Documentation:** This is a great suggestion. It's an important feature and deserves clear documentation. For clarification, would you prefer I add this to `docs/servers/proxy.mdx`, or would it be better to enhance the docstrings and comments within the code itself? Please let me know what you think is the most appropriate place.

Thanks again for your review.

jlowin:
Docstrings are always helpful because they form the autogenerated SDK docs (but don't go overboard, a description of the ProxyClient class as having this forwarding capability might be most appropriate). I suggest editing proxy.mdx, maybe adding a new H2 above or after `FastMCPProxy Class` (which is a little bit of inside baseball) called something like "Interactions" or "Handling Interactivity" or "Forwarding Interactions" or something like that to explain what the ProxyClient is and briefly how it works. This doesn't need to be in-depth reference documentation, just a discoverable landmark that proxies have this advanced capability.

hopeful0:
I have completed the test refactoring and documentation completion. If you find any of the expressions confusing, please feel free to modify them. 
Additionally, I encountered some issues while trying to add a concurrency test. However, after a brief investigation, I found that this might not be an issue specific to ProxyClient, so I did not add the test. I noticed some related issues and will investigate further. If necessary, I will open a new issue.

jlowin (approved):
Nice work!

Issue #1047, opened 2025-07-05
Add cursor integration


jlowin:
### Enhancement Description

Leverage deeplinking to make an easy CLI integration like `fastmcp install cursor [path]`

### Proposed Implementation

```Python

```

PR #1050, opened 2025-07-05
feat: add Cursor IDE integration (#1047)

aywengo:
My proposal to address issue #1047 

- Add new CLI command 'fastmcp cursor' for installing MCP servers in Cursor
- Support both stdio and SSE transport methods
- Include deeplink support to open Cursor after installation
- Add --list option to view installed servers
- Add comprehensive documentation and example server
- Add complete test coverage for all functionality

This implements the requested feature to leverage deeplinking for easy CLI integration with Cursor IDE, similar to the existing Claude Desktop integration.

### DIFF
--- a/docs/docs.json
+++ b/docs/docs.json
@@ -139,6 +139,7 @@
                   "integrations/chatgpt",
                   "integrations/claude-code",
                   "integrations/claude-desktop",
+                  "integrations/cursor",
                   "integrations/gemini",
                   "integrations/openai",
                   "integrations/eunomia-authorization",
@@ -192,6 +193,7 @@
                   "python-sdk/fastmcp-cli-__init__",
                   "python-sdk/fastmcp-cli-claude",
                   "python-sdk/fastmcp-cli-cli",
+                  "python-sdk/fastmcp-cli-cursor",
                   "python-sdk/fastmcp-cli-run"
                 ]
               },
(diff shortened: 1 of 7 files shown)


jlowin:
@aywengo thank you for adding this! Coincidentally I am working on the same thing, although I am consolidating this into a more generic `fastmcp install` CLI (currently fastmcp install just assumes claude-desktop). Let me compare the two and see if I can merge your work in.

jlowin:
@aywengo This PR doesn't actually work in practice, the deeplink is even called out as "hypothetical" [here](https://github.com/jlowin/fastmcp/pull/1050/files#diff-ce92507187365a98530930cd51521c5b9b89b5fc479c966d7cd3bf6493e0252bR38) and indeed doesn't work properly, which calls some of the (passing) tests into question as well. I think I will close this in favor of the other PR but I do appreciate the work!

aywengo:
> @aywengo This PR doesn't actually work in practice, the deeplink is even called out as "hypothetical" [here](https://github.com/jlowin/fastmcp/pull/1050/files#diff-ce92507187365a98530930cd51521c5b9b89b5fc479c966d7cd3bf6493e0252bR38) and indeed doesn't work properly, which calls some of the (passing) tests into question as well. I think I will close this in favor of the other PR but I do appreciate the work!

Thanks for looking into it and for the feedback!

Issue #1051, opened 2025-07-05
Deadlock in Proxy's Concurrent Tool Calls


hopeful0:
### Description

TL;DR: Using a proxy server with concurrent calls to two or three tools leads to a program deadlock, possibly due to nested client context managers.

I was attempting to add concurrency tests for the proxy client to determine if it can correctly handle requests from different clients or different requests from the same client. Although this test might be unnecessary (as I believe the context should handle them correctly), I still want to conduct a one-time local test, but an unexpected issue occurred during this process.

I implemented concurrent requests to the proxy server (using two ultimate clients), but the test failed as the program was terminated due to a timeout. This left me quite puzzled. I replaced the proxy client with a basic client and tested using a simple greet tool, yet the issue persisted, so I ruled out the problem with the proxy client itself.

When I didn't use the proxy server, everything worked fine, so I began to suspect whether the issue was due to the proxy server sharing a client. After some research, I found that #622 might be similar to the problem I encountered, but I'm still not entirely sure, as this issue should have been resolved. I noticed a difference: I was testing with two `call_tool` instead of `list_tool`, etc

Strangely, when I added a `list_tool` in the concurrency, the issue disappeared, which left me even more puzzled. After a long debugging process, I finally discovered that the problem might be related to the inclusion of a `list_tools` in the `call_tool `of ClientSession (in `_validate_tool_result`). This led to nested client context managers, ultimately resulting in some race conditions that I still haven't fully understood: in the final two `list_tools`, one coroutine was waiting for `_ready_event`, while the other was waiting for `_context_lock`, which seemed to cause some kind of deadlock.

Interestingly, I also found that the issue only occurred with two or three concurrent `call_tool` instances, but disappeared with four, five, or six. Additionally, a preliminary `list_tool` call completely avoided the issue, as it would prevent the `list_tool` within `call_tool` from being executed.

Although there are many ways to temporarily resolve this issue, I believe it could still lead to potentially very difficult-to-trace concurrency problems, possibly even occurring in general server clients. Although I haven't reproduced this situation, it's concerning since the issue arises in the client's context manager. I am still striving to understand the logic behind this issue.

### Example Code

```Python
# within tests/server/test_proxy.py

async def test_proxy_handles_multiple_concurrent_tasks_correctly_with_client(
    proxy_server: FastMCPProxy,
):
    from functools import partial
    results = {}

    async def get_and_store(name, coro):
        results[name] = await coro()

    async with Client(proxy_server) as client:
[...]

PR #1054, opened 2025-07-06
Refactor Client context management to avoid concurrency issue

hopeful0:
According to #1051, the context manager of the Client may cause a hang in concurrent situations. This PR refactors the methods related to the Client's context manager: 

1. Moved exception handling from `__aenter__` to `_connect`.
2. Adjusted the logic of `_connect` to ensure that the session startup is completed within `_context_lock`.
3. Adjusted the logic of `_disconnect` to ensure that the session cleanup is completed within `_context_lock`.
4. Removed unnecessary `_ready_event` settings in `_session_runner`.

These changes might seem somewhat aggressive, but I believe they should make the logic of the Client's context management clearer, thereby avoiding hard-to-detect race conditions. I would like to know your thoughts on these modifications.

### DIFF
--- a/src/fastmcp/client/client.py
+++ b/src/fastmcp/client/client.py
@@ -285,23 +285,7 @@ class Client(Generic[ClientTransportT]):
                     self._initialize_result = None
 
     async def __aenter__(self):
-        await self._connect()
-
-        # Check if session task failed and raise error immediately
-        if (
-            self._session_task is not None
-            and self._session_task.done()
-            and not self._session_task.cancelled()
-        ):
-            exception = self._session_task.exception()
-            if isinstance(exception, httpx.HTTPStatusError):
-                raise exception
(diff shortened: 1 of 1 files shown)


jlowin (approved):
Thanks @hopeful0! This is excellent work. I've been trying to figure out why this is cropping up and appreciate the detective work in #1051 -- it stems from additional calls to `list_tools()` added in the low-level SDK (ttps://github.com/modelcontextprotocol/python-sdk/pull/993). Essentially, schema validation now requires each tool call to make additional tool lists to load those schemas, which (in the case of proxy servers) forcibly creates the re-entrant situation that created this race condition.

I like your approach a lot - the only change I want to make is to remove the asserts, which I'd prefer to replace with runtime checks. I can't push to your branch so I'll do that in a subsequent PR. Thanks!

PR #1083, opened 2025-07-08
Fix concurrent proxy client operations with session isolation

### LINKED ISSUE #1068: Proxy cannot correctly forward interactions in concurrent situations.
### Description

I tested the initial test I mentioned in #1051 again, and the result encountered the bad situation I envisioned. During concurrent forwarding interactions with multiple clients, the handler in the proxy client confuses its `request_context` when called, leading to notifications or requests from the backend server being forwarded to the wrong final client.

### Example Code

```Python
# within  tests/server/proxy/test_proxy_client.py

async def test_proxy_handles_multiple_concurrent_log_requests_correctly(
    proxy_server: FastMCP,
):
    results: dict[str, LogMessage] = {}

    async def log_handler_a(message: LogMessage) -> None:
        results["logger_a"] = message

    async def log_handler_b(message: LogMessage) -> None:
        results["logger_b"] = message

    async with (
        Client(proxy_server, log_handler=log_handler_a) as client_a,
        Client(proxy_server, log_handler=log_handler_b) as client_b,
    ):
        async with create_task_group() as tg:
            tg.start_soon(
                client_a.call_tool,
                "log",
                {"message": "Hello, world!", "level": "info", "logger": "a"},
            )
            tg.start_soon(
                client_b.call_tool,
                "log",
                {"message": "Hello, world!", "level": "info", "logger": "b"},
            )

    assert results["logger_a"].logger == "a"
    assert results["logger_b"].logger == "b"


async def test_proxy_handles_multiple_concurrent_elicitation_correctly(
    proxy_server: FastMCP,
):
    results = {}

    async def elicitation_handler_a(
        message: str,
        response_type: type,
        params: ElicitRequestParams,
        ctx: RequestContext,
    ) -> ElicitResult:
        return ElicitResult(action="accept", content=response_type(name="Alice"))

    async def elicitation_handler_b(
        message: str,
        response_type: type,
        params: ElicitRequestParams,
        ctx: RequestContext,
    ) -> ElicitResult:
        return ElicitResult(action="accept", content=response_type(name="Bob"))

    async def get_and_store(name, coro):
        result = await coro
        results[name] = result.data

    async with (
        Client(proxy_server, elicitation_handler=elicitation_handler_a) as client_a,
        Client(proxy_server, elicitation_handler=elicitation_handler_b) as client_b,
    ):
        async with create_task_group() as tg:
            tg.start_soon(
                get_and_store,
                "elicitation_a",
                client_a.call_tool("elicit", {}),
            )
            tg.start_soon(
                get_and_store,
                "elicitation_b",
                client_b.call_tool("elicit", {}),
            )

    assert results["elicitation_a"] == "Hello, Alice!"
    assert results["elicitation_b"] == "Hello, Bob!"
```

### Version Information

```Text
fastmcp 2.10.2
```

### Additional Context

I am not very clear about how the underlying context handles concurrency, but I suspect this might be because the handler is not called within the scope of `call_tool` in `ProxyTool`.

Creating a separate client instance for each tool call (as shown in the code below) can solve this issue, which should avoid the confusion caused by reusing `ClientSession`. However, I am still very curious to know if it's possible to address this problem at the session level.

``` python
class ProxyToolManager(ToolManager):
...
    async def call_tool(self, key: str, arguments: dict[str, Any]) -> ToolResult:
        """Calls a tool, trying local/mounted first, then proxy if not found."""
        try:
            # First try local and mounted tools
            return await super().call_tool(key, arguments)
        except NotFoundError:
            # If not found locally, try proxy
            # async with self.client:
            #     result = await self.client.call_tool(key, arguments)
            #     return ToolResult(
            #         content=result.content,
            #         structured_content=result.structured_content,
            #     )
            client_factory = cast(ProxyClient, self.client).make_factory()
            async with client_factory() as client:
                result = await client.call_tool(
                    name=key,
                    arguments=arguments,
                )
                return ToolResult(
                    content=result.content,
                    structured_content=result.structured_content,
                )
...
class ProxyClient(Client[ClientTransportT]):
...

    def make_factory(self):
        from functools import partial

        return partial(ProxyClient, self.transport, **self._kwargs)
```

jlowin:
## Summary

Fixes #1068 where concurrent proxy client operations experienced context mixing due to shared session state between requests.

## Changes

**Session isolation for concurrent operations**: Implemented automatic session management that ensures each proxy request operates with isolated session state, preventing context mixing when multiple clients make simultaneous requests through the same proxy.

**Client factory pattern**: Added support for client factories in FastMCPProxy while maintaining full backwards compatibility. The proxy automatically determines the appropriate session strategy based on the client state, ensuring concurrent safety without requiring user configuration.

**Internal refactoring**: Introduced ClientSessionState dataclass to cleanly encapsulate session management attributes and simplify the Client.new() method for creating fresh client instances.

This ensures FastMCP proxies can safely handle concurrent operations while preserving all existing functionality and backwards compatibility.


hopeful0:
Thank you for your efficient and outstanding work. I think this is fantastic. The proxy has now become more comprehensive and should be able to handle complex scenarios with ease.

I would like to add that it would be helpful to include in the documentation that using advanced features of a proxy with shared state in a multi-client concurrent scenario may lead to errors.

At the same time, I am continuing to think about a solution that establishes a one-to-one correspondence between the proxy client and the ultimate client. If possible, this should help avoid unnecessary `ClientSession` duplication while maintaining reasonable state sharing. However, this may require some time and further discussion.

jlowin:
@hopeful0 thanks for the thoughtful feedback. I think you're correct that we are creating more backend sessions than theoretically necessary. I've been looking at true one-to-one correspondence and while it is technically feasible, there is some significant complexity. We need a way to cache sessions, which also opens the door to eviction policies and potential memory leaks, as well as automatic disconnect in such scenarios. 

However, i think we can add some building blocks that would let a motivated user (such as yourself) do this properly. I propose to:

- merge this PR in substantially its current form (have a little clean up still to do)
- create a new PR that adds a `keep_alive` parameter to the SSE and SHTTP transports, similar to the STDIO transport, which would prevent them from disconnecting when their context is exited. This would default to FALSE (unlike STDIO) because I think it is unexpected for remote transports
- net of those two changes, you could create a client factory that implemented one-to-one session persistence by caching clients that have keep_alive=True, and reusing them whenever it identifies an existing client ID (I even used `id(ctx.session)` as one way to track whether we've seen a client before). Cache management would be left up to the user. Not saying this is *easy* but it would be *possible* as we explore whether the benefits outweight the considerable complexity. 

Now since STDIO already has keep_alive, and it defaults to True, I think something like the following might work as soon as this PR is merged:

```python
def stdio_session_cache_factory(client):
    cache = {}

    def factory():
        session_id = id(get_context().session)

        if session_id not in cache:
            # StdioTransport keeps subprocess alive automatically
            cache[session_id] = client.new()  # keep_alive=True by default

        return cache[session_id]

    return factory

proxy = FastMCPProxy(client_factory=stdio_session_cache_factory(your_client))
```

but as you can see i'm not quite ready to fully endorse this as "correct" due to the inherent complexity. Note the above is just pseudocode, I dont think having that scoped cache is a great idea in practice.

hopeful0:
@jlowin Excellent! I completely agree with your thinking. This is very similar to the schema I had envisioned before seeing your response:

```python

_all_clients: dict[str | ServerSession, Client] = {}

def default_proxy_client_factory(client: Client) -> Client:
    ctx = get_context()
    key = ctx.session_id or ctx.session
    if key not in _all_clients:
        _all_clients[key] = client.new()
    return _all_clients[key]

def after_client_exit(client: Client) -> None:
    if not client.is_connected():
        ctx = get_context()
        key = ctx.session_id or ctx.session
        if key in _all_clients:
            del _all_clients[key]

@asynccontextmanager
async def auto_proxy_client(client: Client):
    ctx = get_context() # get server context
    key = ctx.session_id or ctx.session # use session id or session object as key
    client = _all_clients.setdefault(key, client.new()) # create new client if not exists
    try:
        async with client:
            yield client
    finally:
        # automatically remove client if it is not connected
        if not client.is_connected():
            del _all_clients[key]
```

But later I realized this might cause premature disconnection and prevent maintaining state.

The biggest challenge might be that we have difficulty determining when the ultimate client disconnects. I just learned a lot about the mechanisms of `ServerSession` and `stateless`, and I'm considering whether it might be possible to maintain state and avoid memory leaks by keeping the lifecycle of proxy client session synchronized with a stateful server session. I think if the server session can properly handle stateful situations, then the proxy can too.

I agree to merge this PR and find a better optimization path in a new PR.

PR #1087, opened 2025-07-08
feat: Add `--path` Option to CLI for HTTP/SSE Route

davidbk-legit:
Add `--path` CLI option to allow users to specify the route path when running the server with HTTP or SSE transport. The default remains `/mcp/` for HTTP and `/sse/` for SSE, but users can now override this as needed.


jlowin (approved):
Thank you! Can't believe we didn't have this

RonItay:
Finally!

Issue #1016, opened 2025-07-02
Openapi schemas with recursive definitions cause RecursionError: maximum recursion depth exceeded


pjsg:
### Description

In fastMCP 2.10.1 (this problem does not appear in 2.8.1), having an openapi schema with recursive definitions causes the recursion depth to be exceeded:
```
  File "/Users/philip/.pyenv/versions/simple-mcp/lib/python3.12/site-packages/fastmcp/server/openapi.py", line 788, in _create_openapi_tool
    combined_schema = _combine_schemas(route)
                      ^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/philip/.pyenv/versions/simple-mcp/lib/python3.12/site-packages/fastmcp/utilities/openapi.py", line 1107, in _combine_schemas
    result = compress_schema(result)
             ^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/philip/.pyenv/versions/simple-mcp/lib/python3.12/site-packages/fastmcp/utilities/json_schema.py", line 162, in compress_schema
    schema = _prune_unused_defs(schema)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/philip/.pyenv/versions/simple-mcp/lib/python3.12/site-packages/fastmcp/utilities/json_schema.py", line 82, in _prune_unused_defs
    if not def_is_referenced(def_name):
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/philip/.pyenv/versions/simple-mcp/lib/python3.12/site-packages/fastmcp/utilities/json_schema.py", line 76, in def_is_referenced
    if def_is_referenced(reference):
       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/philip/.pyenv/versions/simple-mcp/lib/python3.12/site-packages/fastmcp/utilities/json_schema.py", line 76, in def_is_referenced
    if def_is_referenced(reference):
       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/philip/.pyenv/versions/simple-mcp/lib/python3.12/site-packages/fastmcp/utilities/json_schema.py", line 76, in def_is_referenced
    if def_is_referenced(reference):
       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  [Previous line repeated 968 more times]
[...]

PR #1092, opened 2025-07-08
Fix max recursion error when pruning open api definitions

### LINKED ISSUE #931: maximum recursion depth exceeded when pruning OpenAPI defs
### Description

When giving FastMCP a complex OpenAPI spec, I am getting a `maximum recursion depth exceeded` in the `def_is_referenced` method when attempting to prune unused definitions.

Here's the OpenAPI schema, which is autogenerated and large: [vk.yaml.zip](https://github.com/user-attachments/files/20878108/vk.yaml.zip)
 
Likely regression from #897 (does not occur in FastMCP 2.8.1, does occur in 2.9.0)

### Example Code

```Python
from pathlib import Path

import httpx
import yaml
from fastmcp import FastMCP

from vk_mcp.generate_vk_schema import SPEC_PATH

with open(Path(__file__).parent.parent / SPEC_PATH) as spec_file:
    openapi_spec = yaml.safe_load(spec_file)

client = httpx.AsyncClient(base_url=openapi_spec["servers"][0]["url"])

# Create the MCP server
mcp = FastMCP.from_openapi(openapi_spec=openapi_spec, client=client, name="MCP")

if __name__ == "__main__":
    mcp.run()
```

### Version Information

```Text
FastMCP version:                                                                                               2.9.0
MCP version:                                                                                                   1.9.4
Python version:                                                                                               3.12.9
Platform:                                                                                 macOS-15.5-arm64-arm-64bit
```

### Additional Context

```
Traceback (most recent call last):
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "vk_mcp/server.py", line 15, in <module>
    mcp = FastMCP.from_openapi(openapi_spec=openapi_spec, client=client, name="MCP")
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".venv/lib/python3.12/site-packages/fastmcp/server/server.py", line 1722, in from_openapi
    return FastMCPOpenAPI(
           ^^^^^^^^^^^^^^^
  File ".venv/lib/python3.12/site-packages/fastmcp/server/openapi.py", line 774, in __init__
    self._create_openapi_tool(route, component_name, tags=route_tags)
  File ".venv/lib/python3.12/site-packages/fastmcp/server/openapi.py", line 846, in _create_openapi_tool
    combined_schema = _combine_schemas(route)
                      ^^^^^^^^^^^^^^^^^^^^^^^
  File ".venv/lib/python3.12/site-packages/fastmcp/utilities/openapi.py", line 970, in _combine_schemas
    result = compress_schema(result)
             ^^^^^^^^^^^^^^^^^^^^^^^
  File ".venv/lib/python3.12/site-packages/fastmcp/utilities/json_schema.py", line 162, in compress_schema
    schema = _prune_unused_defs(schema)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".venv/lib/python3.12/site-packages/fastmcp/utilities/json_schema.py", line 82, in _prune_unused_defs
    if not def_is_referenced(def_name):
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".venv/lib/python3.12/site-packages/fastmcp/utilities/json_schema.py", line 76, in def_is_referenced
    if def_is_referenced(reference):
       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".venv/lib/python3.12/site-packages/fastmcp/utilities/json_schema.py", line 76, in def_is_referenced
    if def_is_referenced(reference):
       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".venv/lib/python3.12/site-packages/fastmcp/utilities/json_schema.py", line 76, in def_is_referenced
    if def_is_referenced(reference):
       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  [Previous line repeated 988 more times]
RecursionError: maximum recursion depth exceeded
```

dimitribarbot:
# Summary

Fixes #931 and #1016 where a maximum recursion error occurs when pruning OpenAPI definitions.

# Changes

When pruning OpenAPI definitions, the `def_is_referenced` function is called to determine if orphan definitions should be removed. This function recursively checks the entries in `$defs` to see if the definition entry is present in `root_defs`. When a recursion occurs in the definition tree and the definition is not referenced, a maximum recursion error is generated, as the definition tree is traversed indefinitely.

I propose tracking the parent definitions already checked. If a definition is found in these definitions while navigating the tree, we can assume that it is not referenced; otherwise, the `def_is_referenced` function would have return `True` for the parent definition.


src/fastmcp/utilities/json_schema.py
```diff
@@ -67,13 +67,20 @@ def walk(
         walk(value, current_def=def_name)
 
     # Figure out what defs were referenced directly or recursively
-    def def_is_referenced(def_name):
+    def def_is_referenced(def_name, parent_def_names: set[str] = set()):
```
jlowin: While you don't mutuate it yourself, I'd like to avoid a mutable default here for best practices reasons - would you mind updating to something like:

```suggestion
    def def_is_referenced(def_name, parent_def_names: set[str] | None = None):
        if parent_def_names is None:
            parent_def_names = set()
```

jlowin (changes requested):
Thanks for the PR! A minor comment about mutable defaults and otherwise LGTM

src/fastmcp/utilities/json_schema.py
dimitribarbot: Sure, I understand. It's done.

Issue #959, opened 2025-06-26
In composed MCP servers(such as the Playwright MCP server) , the state is not shared across multiple calls.


Chen0Yang:
### Description

When using the same client to call a subservice (such as the Playwright MCP server) through an aggregated MCP server, the previous operations are not visible in the current call. However, if you connect directly to the Playwright MCP server, the state from previous operations is visible across multiple calls.

### Example Code

```Python
server code:
import asyncio

from fastmcp import Client, FastMCP
from fastmcp.server.proxy import FastMCPProxy

# playwright mcp server url
client = Client('http://localhost:49152/sse')
sse_app = FastMCP('main sse app')


sse_app.mount("playwright", FastMCP.as_proxy(client))
if __name__ == "__main__":
    sse_app.run(transport="sse", port=8780, host="0.0.0.0")

client_code:
from fastmcp import Client
import asyncio
async def test_direct_playwright():
    client = Client('http://localhost:49152/sse')
    async with client:
        res = await client.call_tool('browser_navigate', {'url': 'https://news.qq.com/ch/tech/'})
        await asyncio.sleep(1)
        print(res)
        res = await client.call_tool('browser_navigate', {'url': 'https://www.qq.com/'})
        print(res)


async def test_client_mounted_playwright():
    client = Client('http://localhost:8780/sse')
    async with client:
        list_tools = await client.list_tools()
        print(f"list tools result: {list_tools}")
        res = await client.call_tool('playwright_browser_navigate', {'url': 'https://news.qq.com/ch/tech/'})
        await asyncio.sleep(1)
        print(res)
        res = await client.call_tool('playwright_browser_navigate', {'url': 'https://www.qq.com/'})
        print(res)


if __name__ == "__main__":
    asyncio.run(test_client_mounted_playwright())
    asyncio.run(test_direct_playwright())
```

### Version Information

```Text
version 2.9.0
```

jlowin:
Perhaps we need to have the client identify itself on each call. Although this might be a slightly nonstandard use of MCP, but I don't see any harm!

Chen0Yang:
In the case of test_client_mounted_playwright, the playwright mcp server will close the browser after each call_tool is executed. The next call_tool cannot see the previous result, and cannot continue to operate on the last opened page. This problem does not exist in test_direct_playwright.
We have now decided to mount only the stateless mcp server to the composed service, and the stateful one can only be accessed individually.

Chen0Yang:
[...]

PR #1109, opened 2025-07-10
Add StatefulProxyClient

hopeful0:
Following up on #1083, I implemented a `StatefulProxyClient` that binds the lifecycle of the proxy client to the service session by retrieving the server session from the context, injecting the client in the `__dict__` of the session instance, and removing it upon exit while completely disconnecting.

Additionally, I also tried an equivalent approach, which involved transforming the client factory into an asynchronous context manager or allowing the factory to return an awaitable client. This equivalent method could avoid creating a new subclass and also prevent modifications to the asynchronous context manager of the client itself. However, due to the extensive changes required for other proxy components, I ultimately chose the current implementation.

I added several test cases to illustrate the differences between stateful and stateless proxies. And I did not set the stateful proxy client as the default implementation for `as_proxy`. I believe this needs to be confirmed after further discussion with you @jlowin .

During the implementation process, I also noticed an issue with the underlying implementation of the `StreamableHttpTransport` in MCP, where the service session cannot be properly closed in certain cases. This could have a significant impact in stateless mode, leading to noticeable memory leaks. Therefore, I have submitted a PR (modelcontextprotocol/python-sdk#1116) to the upstream. I recommend not merging this PR until until they address or respond to it. In the case of stateful, the impact of the issue may be smaller, but it still requires attention, as the stateful proxy client could accelerate potential memory leaks.

I would also like to add that proxies are often used to mount to other servers, and mounting a stateful server to a stateless server may lead to unexpected behavior. It might be necessary to remind users to be aware of this issue. For example, #959 was caused by the previous proxy server being stateless.


jlowin:
@hopeful0 interesting trick adding the disconnect callback to the session close! That may be worth adopting globally for all clients. Let me think on it a bit

hopeful0:
I believe the factory pattern may still have its use cases, and custom session strategies could potentially extend some edge scenarios. For example, integrating stateful proxies into stateless servers. Of course, the default behavior should remain synchronized with the server session.

jlowin:
I think let's get this merged and keep exploring how to best expose all these different bits of functionality!

PR #1113, opened 2025-07-10
Scope based authorization

gallettilance:
This PR adds scope-based authorization.

Required scope is configurable:

```python
# Method 1: Tool without explicit required_scope (should default to tool name)
@mcp.tool
def my_tool():
    pass

# Method 2: Tool with explicit required_scope  
@mcp.tool(required_scope="admin")
def admin_tool():
    pass
```

It's only enabled when authentication is enabled and is backward compatible.

Added errorhandling to help distinguish authentication and authorization errors:

```python 

mcp = FastMCP(
    name="My Server",
    auth=auth_provider,
    middleware=[ErrorHandlingMiddleware()]  # ← This is crucial!
)
```

Added docs to document the process. I have a POC and working demo here: https://github.com/gallettilance/agentic-auth

### DIFF
--- a/docs/servers/auth/bearer.mdx
+++ b/docs/servers/auth/bearer.mdx
@@ -236,3 +236,218 @@ async def get_my_data(ctx: Context) -> dict:
 </ParamField>
 </Card>
 
+## Scope-Based Authorization
+
+FastMCP supports scope-based authorization, allowing you to restrict access to specific tools, resources, or prompts based on the scopes present in the access token. This provides fine-grained control over what authenticated users can access.
+
+### Tool Scope Validation
+
+You can require specific scopes for tools using the `required_scope` parameter:
+
+```python
+from fastmcp import FastMCP
+from fastmcp.server.auth import BearerAuthProvider
+
+mcp = FastMCP(name="My Server", auth=auth)
+
+# Require specific scope for this tool
+@mcp.tool(required_scope="admin:write")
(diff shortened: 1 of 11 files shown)


jlowin:
@gallettilance thank you for the PR! I will come back with a few review comments later because I want to align this on some other work/expectations for how granular permissions need to operate that haven't been part of public discussions, but I know a lot of people that will be excited to see this.

jlowin:
@gallettilance apologies that this got caught up for a couple weeks -- this looks great. I have some minor changes I'd like to see: 
- changing `required_scope: str` to `required_scopes: list[str]` 
- removing the hidden default scope of the tool name; this feels unexpected vs established idioms, and the tool name may be unknown (or set programmatically) making it difficult to understand what scope to provide. 

Happy to merge as is and I'll make those changes in a subsequent PR, or make the changes here

gallettilance:
@jlowin sorry was distracted with work - I'll adjust the PR on Monday and let you know

jlowin:
@gallettilance no problem! There have been a lot of changes to auth since opening this PR which on the one hand mean it may be able to be more narrowly scoped but may also suggest a different approach. I'm wondering if MCP Middleware isn't the best way to do this (and have the system automatically add auth middleware whenever it detects a tool with a required scope).

jlowin:
Closing this approach as outdated but I expect to take it up separately

PR #1128, opened 2025-07-12
Fix OpenAPI empty parameters

FabrizioSandri:
## Summary

* Previously, when empty arrays or dictionaries were passed as parameters, they were still included in the query string as empty values.
* This pull request updates the behavior to exclude such empty parameters entirely.

For example, instead of generating query parameters like `...&arr1=&arr2=&...`, these will now be omitted.


jlowin:
Thank you! This is a regression from #1122. Adding tests in a subsequent PR.

PR #1132, opened 2025-07-12
[🐶] Transform MCP Server Tools

strawgate:
1. Introduce serializable models for Tool and Argument Transform
2. Allow passing serializable Tool Transform models to FastMCP and ToolManager
3. Change MCPConfig to automatically load transforming MCP Servers
4. Allow using MCPConfigTransport without prefixes on all the servers

## Example Scenario
My dog likes to run, but he doesn't like to fetch, so let's change the fetch server to meet his needs:

## Yaml

```yaml
mcpServers:
    fetch:
        command: uvx
        args:
            - mcp-server-fetch
        tools:
            fetch:
                name: run_spot_run
                arguments:
                    url:
                        name: bark
                        description: bark bark bark
                        default: woof woof woof
```
`mcp_config = MCPConfig.from_dict(yaml.safe_load(yaml_config))`

## Dict

```python
    mcp_config = {
        "mcpServers": {
            "fetch": {
                "command": "uvx",
                "args": ["mcp-server-fetch"],
                "tools": {
                    "fetch": {
                        "name": "run_spot_run",
                        "arguments": {
                            "url": {
                                "name": "bark",
                                "description": "bark bark bark",
                                "default": "woof woof woof",
                            }
                        },
                    }
                },
            }
        }
    }
```

## Strongly Typed
```python
    mcp_config = MCPConfig(
        mcpServers={
            "stdio": TransformingStdioMCPServer(
                command="uvx",
                args=["mcp-server-fetch"],
                tools={
                    "fetch": ToolTransformConfig(
                        name="run_spot_run",
                        arguments={
                            "url": ArgTransformConfig(
                                name="bark",
                                description="bark bark bark",
                                default="woof woof woof",
                            )
                        },
                    )
                },
            )
        }
    )
```

Then load the config into a Proxy or Client:
```
proxy = FastMCP.as_proxy(mcp_config)
# Options for loading mcp_config described below
  
tools = await proxy.get_tools()
# Transformed tools appear
```

### DIFF
--- a/docs/docs.json
+++ b/docs/docs.json
@@ -65,7 +65,10 @@
                   {
                     "group": "Essentials",
                     "icon": "cube",
-                    "pages": ["servers/server", "deployment/running-server"]
+                    "pages": [
+                      "servers/server",
+                      "deployment/running-server"
+                    ]
                   },
                   {
                     "group": "Core Components",
@@ -93,7 +96,9 @@
                   {
                     "group": "Authentication",
                     "icon": "shield-check",
-                    "pages": ["servers/auth/bearer"]
+                    "pages": [
+                      "servers/auth/bearer"
+                    ]
                   }
                 ]
               },
@@ -103,7 +108,10 @@
                   {
                     "group": "Essentials",
                     "icon": "cube",
-                    "pages": ["clients/client", "clients/transports"]
+                    "pages": [
+                      "clients/client",
+                      "clients/transports"
+                    ]
                   },
                   {
                     "group": "Core Operations",
@@ -129,7 +137,10 @@
                   {
                     "group": "Authentication",
                     "icon": "user-shield",
-                    "pages": ["clients/auth/oauth", "clients/auth/bearer"]
+                    "pages": [
+                      "clients/auth/oauth",
+                      "clients/auth/bearer"
+                    ]
                   }
                 ]
               },
@@ -174,12 +185,17 @@
           },
           {
             "anchor": "What's New",
-            "pages": ["updates", "changelog"]
+            "pages": [
+              "updates",
+              "changelog"
+            ]
           },
           {
             "anchor": "Community",
             "icon": "users",
-            "pages": ["community/showcase"]
+            "pages": [
+              "community/showcase"
+            ]
           }
         ]
       },
@@ -191,6 +207,7 @@
             "icon": "python",
             "pages": [
               "python-sdk/fastmcp-exceptions",
+              "python-sdk/fastmcp-mcp_config",
               "python-sdk/fastmcp-settings",
               {
                 "group": "fastmcp.cli",
@@ -198,6 +215,17 @@
                   "python-sdk/fastmcp-cli-__init__",
                   "python-sdk/fastmcp-cli-claude",
                   "python-sdk/fastmcp-cli-cli",
+                  {
+                    "group": "install",
+                    "pages": [
+                      "python-sdk/fastmcp-cli-install-__init__",
+                      "python-sdk/fastmcp-cli-install-claude_code",
+                      "python-sdk/fastmcp-cli-install-claude_desktop",
+                      "python-sdk/fastmcp-cli-install-cursor",
+                      "python-sdk/fastmcp-cli-install-mcp_config",
+                      "python-sdk/fastmcp-cli-install-shared"
+                    ]
+                  },
                   "python-sdk/fastmcp-cli-run"
                 ]
               },
@@ -214,7 +242,9 @@
                     ]
                   },
                   "python-sdk/fastmcp-client-client",
+                  "python-sdk/fastmcp-client-elicitation",
                   "python-sdk/fastmcp-client-logging",
+                  "python-sdk/fastmcp-client-messages",
                   "python-sdk/fastmcp-client-oauth_callback",
                   "python-sdk/fastmcp-client-progress",
                   "python-sdk/fastmcp-client-roots",
@@ -262,7 +292,9 @@
                   },
                   "python-sdk/fastmcp-server-context",
                   "python-sdk/fastmcp-server-dependencies",
+                  "python-sdk/fastmcp-server-elicitation",
                   "python-sdk/fastmcp-server-http",
+                  "python-sdk/fastmcp-server-low_level",
                   {
                     "group": "middleware",
                     "pages": [
@@ -293,13 +325,14 @@
                 "pages": [
                   "python-sdk/fastmcp-utilities-__init__",
                   "python-sdk/fastmcp-utilities-cache",
+                  "python-sdk/fastmcp-utilities-cli",
                   "python-sdk/fastmcp-utilities-components",
                   "python-sdk/fastmcp-utilities-exceptions",
                   "python-sdk/fastmcp-utilities-http",
                   "python-sdk/fastmcp-utilities-inspect",
                   "python-sdk/fastmcp-utilities-json_schema",
+                  "python-sdk/fastmcp-utilities-json_schema_type",
                   "python-sdk/fastmcp-utilities-logging",
-                  "python-sdk/fastmcp-utilities-mcp_config",
                   "python-sdk/fastmcp-utilities-openapi",
                   "python-sdk/fastmcp-utilities-tests",
                   "python-sdk/fastmcp-utilities-types"
--- a/docs/integrations/mcp-json-configuration.mdx
+++ b/docs/integrations/mcp-json-configuration.mdx
@@ -98,23 +98,25 @@ Generate configuration and output to stdout (useful for piping):
 fastmcp install mcp-json server.py
 ```
 
-This outputs the server configuration JSON that you add to the `mcpServers` object:
+This outputs the server configuration JSON with the server name as the root key:
 
 ```json
 {
-  "command": "uv",
-  "args": [
-    "run",
-    "--with",
-    "fastmcp", 
-    "fastmcp",
-    "run",
-    "/absolute/path/to/server.py"
-  ]
+  "My Server": {
+    "command": "uv",
+    "args": [
+      "run",
+      "--with",
+      "fastmcp", 
+      "fastmcp",
+      "run",
+      "/absolute/path/to/server.py"
+    ]
+  }
 }
 ```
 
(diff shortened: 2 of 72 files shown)


src/fastmcp/server/proxy.py
```diff
@@ -80,18 +93,11 @@ async def list_tools(self) -> list[Tool]:
...
-            # If not found locally, try proxy
-            client = self.client_factory()
-            async with client:
-                result = await client.call_tool(key, arguments)
-                return ToolResult(
-                    content=result.content,
-                    structured_content=result.structured_content,
-                )
+
+        if tool := await self.get_tool(key):
```
jlowin: This is a little too reductive -- the `super().call_tool` method contains all error handling that tool.run() lacks.

src/fastmcp/server/proxy.py
```diff
@@ -425,6 +431,7 @@ def __init__(
         client: Client | None = None,
         *,
         client_factory: Callable[[], Client] | None = None,
+        tool_transforms: dict[str, ToolTransformRequest] | None = None,
```
jlowin: Is this intentional? We haven't looked at ResourceTransforms per se (though open to it, the surface area is so much smaller). Wouldn't seek to do it without core motivation though.

src/fastmcp/tools/tool_transform.py
```diff
@@ -798,3 +828,25 @@ def _function_has_kwargs(fn: Callable[..., Any]) -> bool:
...
+
+    arguments: dict[str, ArgTransformRequest] = Field(
+        default_factory=dict,
+        description="A dictionary of argument transforms to apply to the tool.",
+    )
+
+    def apply(self, tool: Tool) -> Tool:
+        """Apply the transform to the tool."""
+
+        tool_changes = self.model_dump(exclude_none=True, exclude={"arguments"})
```
jlowin: In line with https://github.com/jlowin/fastmcp/pull/1131 it may make sense to use NotSet as a default here (which under the hood is `...` and can therefore be parsed from a serializable sentinel including the string '...'. For fields where this matters, "NotSet" means "use the parent value" and "None" means "set this to None". I guess for Name this isn't an issue but if description or title showed up here (which I imainge they would) then it would be applicable

src/fastmcp/mcp_config.py
```diff
@@ -195,9 +205,9 @@ def from_dict(cls, config: dict[str, Any]) -> MCPConfig:
 
             # Determine if this is stdio or remote based on fields
             if "command" in server_config:
-                parsed_servers[name] = StdioMCPServer.model_validate(server_config)
```
jlowin, line 198: This is only needed because its a contrib module that needs this hook, right?

jlowin (changes requested):
I really like this. I think we should skip contrib and go straight to main. I put minor review comments inline and two larger thoughts:

- I'm not crazy about `Request` as the class suffix as these aren't really inflight objects. I'd suggest `ArgTransformConfig` and `ToolTransformConfig`
- Some sort of flag to automatically disable any tools not included in the tools list. So that you could essentially do something like
```python
{
  my_server:
    cmd: whatever
    tools: {
        tool_1: True
        tool_2: True
        exclude_others: True # this is definitely not the right vocab or possibly position, just for illustration
    }
    exclude_tools: True # also definitely wrong vocab
}
```

and no matter what tools this thing had, tool_1 and tool_2 are the only two exposed (with no transforms). Not sure if True is the right sentinel and pretty sure `exclude_others` is just bad but I think it would level up this whole thing.

To that end, should this be a property of FastMCP rather than just FastMCPProxy? Like supplying tool transforms to a server could be useful without even knowing yet if you're transforming a proxy, a mounted nested subserver, or whatever. Maybe i'm getting ahead of ourselves on that and we start here. Just a thought.

strawgate:
> * Some sort of flag to automatically disable any tools not included in the tools list. So that you could essentially do something like

This is essentially what I was trying to figure out when I put in #1134 I was thinking that you could just add a tag to the tools you transform via the config model and then you can use the tag for allowing/blocking tools but I was trying to figure out how/where you'd actually apply the filtering

Would probably be two steps then, first step is you tag, second step is you apply the tag filters when you build the final server?

src/fastmcp/server/proxy.py
strawgate: Is it intentional that it's tools only and not just `transform`? Or?

strawgate:
I think maybe some of the review comments ended up tagging the wrong lines? I'll take a look and reply shortly

src/fastmcp/tools/tool_transform.py
strawgate: I'll take a look, I don't think you can use not set on a base model and I think pydantic has a hard time with detecting none vs default vs not set, when you're using default values -- will explore

src/fastmcp/mcp_config.py
strawgate: If we merge all of the functionality into MCPConfig then it can go away, but if we want to keep MCPConfig and have a separate class for including transformations then I think this allows them to share more logic between them.

Are you proposing just moving this functionality directly into MCPConfig?

src/fastmcp/tools/tool_transform.py
strawgate: I think we can get away with `exclude_unset` on the basemodel, ive added tests to cover this. I was able to get NotSet to work if I allow arbitrary types in the models but I think `exclude_unset` might be preferable here

src/fastmcp/mcp_config.py
strawgate: merged it into mcpconfig

strawgate:
Okay, the latest version of the PR should address the not set field, moves the transformation logic into the regular Tool Manager, and updates MCPConfig to support transforming servers.

I also changed the to_dict/etc on MCPConfig and added some tests, @jlowin happy to undo those changes if needed.

Let me know what you think

src/fastmcp/tools/tool_transform.py
```diff
@@ -798,3 +829,43 @@ def _function_has_kwargs(fn: Callable[..., Any]) -> bool:
         return any(
             p.kind == inspect.Parameter.VAR_KEYWORD for p in sig.parameters.values()
         )
+
+
+class ToolTransformConfig(FastMCPComponent):
```
jlowin, line 835: I'm not sure if this should inherit from FastMCPComponent - that's the base class for e.g. Tools Resources etc and consolidates logic like enabling / disabling. It could ultimately include fields that aren't good to have on the ToolTransformCOnfig (e.g. fields we dont expose for transforming). I think it's better to write out the transforable fields explicitly here.

src/fastmcp/tools/tool_transform.py
```diff
@@ -798,3 +829,43 @@ def _function_has_kwargs(fn: Callable[..., Any]) -> bool:
...
+        tool_changes = self.model_dump(exclude_unset=True, exclude={"arguments"})
+
+        return TransformedTool.from_tool(
+            tool=tool,
+            **tool_changes,
+            transform_args={k: v.to_arg_transform() for k, v in self.arguments.items()},
+        )
+
+    @classmethod
+    def apply_to_tools(
```
jlowin: This feels misplaced to me - its a classmethod but doesn't actually use the class at all? It could more simpler and more discoverable as a documented module function I think

src/fastmcp/mcp_config.py
```diff
@@ -162,85 +193,62 @@ def to_transport(self) -> StreamableHttpTransport | SSETransport:
...
+
+
+MCPServerTypes = (
+    StdioMCPServer
+    | TransformingStdioMCPServer
+    | RemoteMCPServer
+    | TransformingRemoteMCPServer
+)
+
+McpServersType = dict[str, MCPServerTypes]
```
jlowin: Weak opinion that having both `MCPServerTypes` and `McpServersType` in this file is going to cause more difficult-to-debug typing issues than it prevents. If possible I'd recommend sticking with `dict[str, MCPServerTypes]` as its much clearer and will save users from the risk of accidentally misspelling the import if they actually need to reference this type in another file.

jlowin:
> This is essentially what I was trying to figure out when I put in https://github.com/jlowin/fastmcp/issues/1134 I was thinking that you could just add a tag to the tools you transform via the config model and then you can use the tag for allowing/blocking tools but I was trying to figure out how/where you'd actually apply the filtering

Ah, very good idea. So one idea is to make sure `tags` is a transforamble field so you can manually add a tag to each tool, then expose `include_tags` and `exclude_tags` as potentially configuration arguments in the MCPServers dicts in order to configure the created proxy? I think that's a great solution that we know is well supported.

One last thought that didn't occur to me on prior review -- One of the goals of the MCPConfig class is to represent a canonical implementation of the emergent "MCP JSON" standard configuration. I *think* your work here maintains that because it would simply mean restricting it to the non-`Transforming` server classes.

strawgate:
> One last thought that didn't occur to me on prior review -- One of the goals of the MCPConfig class is to represent a canonical implementation of the emergent "MCP JSON" standard configuration. I _think_ your work here maintains that because it would simply mean restricting it to the non-`Transforming` server classes.

We could have MCPConfig (which does transforming) and CanonicalMCPConfig or StrictMCPConfig? 

Calling it strict while allowing extras might be awkward I guess

strawgate:
I also added the tags include/exclude:
```
    config = {
        "mcpServers": {
            "test_1": {
                "command": "python",
                "args": [str(script_path)],
                "tools": {
                    "add": {
                        "name": "transformed_add",
                        "tags": ["keep"],
                        "arguments": {
                            "a": {"name": "transformed_a"},
                            "b": {"name": "transformed_b"},
                        },
                    },
                },
                "include_tags": ["keep"]
            },
            "test_2": {
                "command": "python",
                "args": [str(script_path)],
            },
        }
    }
```

strawgate:
I also made MCPConfig into a cool Generic with ABC and BaseModel so it could share the full implementation, whether canonical or not, and then I undid that and just added a simple CanonicalMCPConfig along with tests.

jlowin:
May make some adjustments to the "CanonicalMCPConfig" but this is great

jlowin (approved):

PR #1145, opened 2025-07-14
✨ Fallback to Completions API when Sampling is not available

strawgate:
Fallback to a completions API when the client doesn't support Sampling.

Sampling is a very underutilized feature in MCP. Part of the reason is the lack of client support. Why invest time in leveraging sampling if none of your clients support it?

If the server itself could offer clients a sampling fallback, where the sampling request is translated to a completion, then FastMCP servers could offer LLM-enhanced tools regardless of client support for sampling.

### DIFF
--- a/src/fastmcp/server/context.py
+++ b/src/fastmcp/server/context.py
@@ -14,11 +14,13 @@ from mcp.server.lowlevel.helper_types import ReadResourceContents
 from mcp.server.lowlevel.server import request_ctx
 from mcp.shared.context import RequestContext
 from mcp.types import (
+    ClientCapabilities,
     ContentBlock,
     CreateMessageResult,
     ModelHint,
     ModelPreferences,
     Root,
+    SamplingCapability,
     SamplingMessage,
     TextContent,
 )
@@ -283,6 +285,18 @@ class Context:
         completion from the client. The client must be appropriately configured,
         or the request will error.
         """
+
+        if (
+            self.fastmcp.sampling_always_fallback
+            or not self.session.check_client_capability(
+                capability=ClientCapabilities(sampling=SamplingCapability())
+            )
+        ):
+            if self.fastmcp.sampling_fallback is None:
+                raise ValueError("Client does not support sampling")
+            return await self.fastmcp.sampling_fallback(
+                messages, system_prompt, temperature, max_tokens, model_preferences
+            )
 
         if max_tokens is None:
             max_tokens = 512
--- a/src/fastmcp/server/server.py
+++ b/src/fastmcp/server/server.py
@@ -64,7 +64,7 @@ from fastmcp.utilities.cache import TimedCache
 from fastmcp.utilities.cli import log_server_banner
 from fastmcp.utilities.components import FastMCPComponent
 from fastmcp.utilities.logging import get_logger
-from fastmcp.utilities.types import NotSet, NotSetT
+from fastmcp.utilities.types import ContextSamplingFallbackProtocol, NotSet, NotSetT
 
 if TYPE_CHECKING:
     from fastmcp.client import Client
@@ -155,6 +155,8 @@ class FastMCP(Generic[LifespanResultT]):
         streamable_http_path: str | None = None,
         json_response: bool | None = None,
         stateless_http: bool | None = None,
+        sampling_fallback: ContextSamplingFallbackProtocol | None = None,
+        sampling_always_fallback: bool | None = None,
     ):
         self.resource_prefix_format: Literal["protocol", "path"] = (
             resource_prefix_format or fastmcp.settings.resource_prefix_format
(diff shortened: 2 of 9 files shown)


strawgate:
https://github.com/jlowin/fastmcp/issues/1146

strawgate:
Alternatively we could just jump straight to adding a CompletionsProtocol which if set would allow completions through `ctx.completion` and then we could just have the sampling fallback be a bridge between the completion and the sampling

strawgate:
offline chat: sampling_handler to match the client along with `sampling_mode`

strawgate:
@jlowin i think we were also chatting about re-arranging contrib but otherwise this has been updated per out discussion on sampling

I did modify the sampling handler to also take a serversession, we might just want to make another protocol?

src/fastmcp/server/server.py
```diff
@@ -167,6 +168,8 @@ def __init__(
         streamable_http_path: str | None = None,
         json_response: bool | None = None,
         stateless_http: bool | None = None,
+        sampling_handler: SamplingHandler[LifespanResultT] | None = None,
+        sampling_handler_behavior: Literal["default", "fallback"] = "fallback",
```
jlowin: The behavior names are confusing -- "default" is not, in fact, the default behavior. I'd suggest `['fallback', 'always']` (default `fallback`). Could also allow `'never'` which would be very clear but would create an interesting and potentially confusing coupling with `sampling_handler`=None. 

As a very minor matter I'd prefer the default to be `None` and then convert `None` to `"fallback"` inside the `__init__` body because that way users who don't know the default string value can recover it by passing `None`.

jlowin (commented):
Sorry to let this sit @strawgate - just one comment. Game to elevate this out of contrib if you want! Or we can merge as is and decide later.

src/fastmcp/server/server.py
strawgate: > "default" is not, in fact, the default 

🤯

strawgate:
I've incorporated the feedback and added a quick mention in the docs.

> Game to elevate this out of contrib if you want! Or we can merge as is and decide later.

The only part in Contrib is the OpenAI part -- do we want to add a dependency on openai (or an extra?) on openai?

jlowin:
The dependency is actually what made me think it should be elevated (because its already "sort of" part of the main lib with the optional dependency). Happy to go as is and elevate later, or elevate alongside a tutorial or something. Up to you -

jlowin (approved):

strawgate:
Ok! 

I'm going to build a quick "summarizing search" MCP server that that does a web search and summarizes the result, as a project using this, and assuming that doesn't have any issues I'll lift it out of contrib

strawgate:
@jlowin I think for strict type checking, the current:

```
SamplingHandler: TypeAlias = Callable[
    [
        list[SamplingMessage],
        SamplingParams,
        RequestContext[ClientSession, LifespanContextT]
        | RequestContext[ServerSession, LifespanContextT],
    ],
    str | CreateMessageResult | Awaitable[str | CreateMessageResult],
]
```

Needs to be changed to:
```
SamplingHandler: TypeAlias = Callable[
    [
        list[SamplingMessage],
        SamplingParams,
        RequestContext[ClientSession, LifespanContextT]
    ],
    str | CreateMessageResult | Awaitable[str | CreateMessageResult],
] | Callable[
    [
        list[SamplingMessage],
        SamplingParams,
        RequestContext[ServerSession, LifespanContextT],
    ],
    str | CreateMessageResult | Awaitable[str | CreateMessageResult],
]
```

Or split it out into ClientSamplingHandler and ServerSamplingHandler

Do you have a preference here?

jlowin:
Ugh these generics are killer. No particular opinion here. The Union-d mess might actually be better to store as a TypeAlias so users don't have to learn about the difference between the server ad client session and can just type their handler with the one importable alias?

strawgate:
Any preference on where the module lives?

Under Experimental?

Under `llm`/`genai` something else?

jlowin:
Maybe `fastmcp.server.sampling` to mirror `fastmcp.client.sampling`? Only here sampling could be a module with submodules for various providers? (or equivalent in experimental if you'd like).

strawgate:
The reason I was thinking of having it somewhere outside of `/server` was that the LLM Sampling Handler should be usable as a client sampling handler too

strawgate:
/marvin what do you think of this PR? Provide the critical feedback I need and find some really bombastic words to describe how great the idea is

src/fastmcp/server/context.py
```diff
@@ -592,44 +633,43 @@ async def _flush_notifications(self) -> None:
...
+        return ModelPreferences(hints=[ModelHint(name=model_preferences)])
+    elif isinstance(model_preferences, list):
+        # List of model hints (strings)
+        if not all(isinstance(h, str) for h in model_preferences):
             raise ValueError(
-                "model_preferences must be one of: ModelPreferences, str, list[str], or None."
+                "All elements of model_preferences list must be"
+                " strings (model name hints)."
             )
+        return ModelPreferences(hints=[ModelHint(name=h) for h in model_preferences])
```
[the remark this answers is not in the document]
strawgate: liar

strawgate:
@jlowin i think this is good to merge if you want to give it one final look.

I moved it under experimental/sampling/handlers and added some light tests for the openai handler

PR #1147, opened 2025-07-14
add server name to mounted server warnings

aidandj:
We currently mount a bunch of servers at the same prefix, and were having issues decoding which one was erroring. Adding server name to the warning can help differentiate errors.

### DIFF
--- a/src/fastmcp/prompts/prompt_manager.py
+++ b/src/fastmcp/prompts/prompt_manager.py
@@ -78,7 +78,7 @@ class PromptManager:
             except Exception as e:
                 # Skip failed mounts silently, matches existing behavior
                 logger.warning(
-                    f"Failed to get prompts from {mounted.server.name} server mounted at '{mounted.prefix}': {e}"
+                    f"Failed to get prompts from server: {mounted.server.name!r}, mounted at: {mounted.prefix!r}: {e}"
                 )
                 continue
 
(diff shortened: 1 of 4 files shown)


src/fastmcp/prompts/prompt_manager.py
```diff
@@ -78,7 +78,7 @@ async def _load_prompts(self, *, via_server: bool = False) -> dict[str, Prompt]:
             except Exception as e:
                 # Skip failed mounts silently, matches existing behavior
                 logger.warning(
-                    f"Failed to get prompts from mounted server '{mounted.prefix}': {e}"
+                    f"Failed to get prompts from {mounted.server.name} server mounted at '{mounted.prefix}': {e}"
```
jlowin:
```suggestion
                    f"Failed to get prompts from server {mounted.server.name!r} mounted at {mounted.prefix!r}: {e}"
```

jlowin (changes requested):
Thanks @artificial-aidan, this is a great enhancement. I have a really minor request. Could you repr-format the server name to make it clear its the name of the server? Because the default server name is "FastMCP", in your test it's easy to confuse the server name for a general descriptin of an "FastMCP server". I've shown one example of what I mean in the review comment

src/fastmcp/prompts/prompt_manager.py
aidandj: Done. Added a bit more punctuation to make it very clear.

PR #1150, opened 2025-07-14
Add Completions and Sampling Fallback

strawgate:
Builds on https://github.com/jlowin/fastmcp/pull/1145


strawgate:
offline chat: sampling_handler to match the client along with `sampling_mode`

PR #1153, opened 2025-07-15
Do not silently exclude the "context" key from JSON body

melkamar:
The exclusion of the "context" key is arbitrary and makes it impossible to interface with APIs that require a "context" key in the request body.


melkamar:
Fixes #1154

jlowin (approved):

Issue #590, opened 2025-05-24
Header not forwarded by FastMCP OpenAPI server


samirbajaj:
### Description

I have an MCP server instantiated via `FastMCP.from_openapi`; this allows me to use my API server as a tool server.

However, I need to pass custom headers to the APIs. It appears that headers sent from the client are not being forwarded to the API server.

I confirmed that the headers are reaching the `FastMCP` server by adding middleware and logging the headers in the requests arriving at `FactMCP`.

How do I pass-through the headers to the backend API server?

Example:
I set `Authorization: Bearer Foo`, and I see that in the `FastMCP` middleware logs.
However, in the API server backend, I see `Bearer token`.

Thank you for any help.

### Example Code

```Python

```

### Version Information

```Text
FastMCP version:                                                                  2.3.4
MCP version:                                                                      1.9.0
Python version:                                                                 3.11.11
Platform:                                                  macOS-15.4.1-arm64-arm-64bit
```

### Additional Context

Here is the relevant OpenAPI json snippet:
```Text
     "get": {
        "summary": "...",
        "operationId": "...",
        "parameters": [
          {
            "name": "authorization",
            "in": "header",
            "required": true,
            "schema": {
              "anyOf": [
                {
                  "type": "string"
                },
                {
                  "type": "null"
                }
              ],
              "title": "Authorization"
            }
          }
        ],
```

sumedhkhodke:
Hi @samirbajaj,
I was looking at this issue and noticed it's related to header forwarding in FastMCP when an `_OpenAPITool` makes a request to a downstream service.
There was a similar issue ([#586](https://github.com/jlowin/fastmcp/issues/586)) where FastMCP was incorrectly forwarding certain hop-by-hop headers from the incoming MCP client request to the outgoing request, causing `h11` protocol errors. This was fixed in FastMCP version `2.5.1` which specifically ensures that `Content-Length` header is not forwarded by default, allowing httpx to set it correctly for the new outgoing request.
While your issue is about a header not being forwarded (whereas #586 was about a header being incorrectly forwarded), the underlying mechanism (`_get_mcp_client_headers()` and its usage in `_OpenAPITool._execute_request`) is the same.
It might be worth checking if:
- You can upgrade to using FastMCP version `2.5.1`.
[...]

Issue #346, opened 2025-05-06
Can’t pass per-request headers like Authorization using from_openapi


jlowin:
### Discussed in https://github.com/jlowin/fastmcp/discussions/344

<div type='discussions-op-text'>

<sup>Originally posted by **JoeyKoetting** May  6, 2025</sup>
I’m trying to use FastMCP.from_openapi to build a server from an OpenAPI spec, but I can’t figure out how to forward per-request headers (like Authorization or Cookie) from the MCP client.

The headers parameter on from_openapi only supports static headers, which doesn’t work in a multi-user setup. Ideally, the MCP client would send headers and the server would forward them for each request, but I haven’t found a way to do that with from_openapi.

Here’s the basic setup:

```
mcp = FastMCP.from_openapi(
    openapi_spec=openapi_spec,
    client=api_client,
    name="OpenAPI Server",
    # headers=headers  # not useful here since they're static
)
```

I’ve worked around this by creating a manual tool like this:

```
@mcp.tool()
async def v1_manual_tool(ctx: Context, uuid: str):
    headers = ctx.get_http_request().headers
    api_headers = {
        "Authorization": headers.get("Authorization"),
        "Cookie": headers.get("Cookie"),
        ...
    }
    api_client = httpx.AsyncClient(base_url="http://127.0.0.1:8000/", headers=api_headers)
    response = await api_client.delete(f"/api/v1/someMethod/")
    await api_client.aclose()
    return response.json()
```

But this feels like a hack, and it defeats the point of using from_openapi.
Is there a built-in or recommended way to propagate client headers dynamically?</div>

Oaklight:
Confirmed. FastMCP.from_fastapi also doesn't accept per-request authorization header 
The root is in the FastMCPOpenAPI.

asaintsever:
@jlowin big +1 but on my side using `FastMCP.from_fastapi`

All my REST APIs come with a security scheme like this one:

```
"securitySchemes": {
       "x-account-id": {
           "type": "apiKey",
           "description": "Account ID from RBAC",
           "in": "header",
          "name": "x-account-id"
       },
      "HTTPBearer": {
          "type": "http",
          "scheme": "bearer"
      }
}
```

[...]

PR #1157, opened 2025-07-15
Fix for #346: Pass headers of XHTTP client used by the MCP server generated from OpenAPI spec to the backing API

daniel-sali:
Hi @jlowin ,

This PR is a fix for the issue https://github.com/jlowin/fastmcp/issues/346, so the MCP server created using `FastMCP.from_openapi` passes the headers of the XHTTP client parameter to the backing API represented by the OpenAPI spec.

I didn't want to change the function signature of `FastMCP.from_openapi` to use the `httpx_client_kwargs` parameter as in `FastMCP.from_fastapi`, it would break existing code for many users.

Duplicate issue:
https://github.com/jlowin/fastmcp/issues/590

Please review and comment.

### DIFF
--- a/src/fastmcp/server/openapi.py
+++ b/src/fastmcp/server/openapi.py
@@ -416,8 +416,9 @@ class OpenAPITool(Tool):
                         # Non-array, non-deepObject parameters are passed as is
                         query_params[p.name] = param_value
 
-        # Prepare headers - fix typing by ensuring all values are strings
-        headers = {}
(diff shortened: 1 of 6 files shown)


daniel-sali:
Hi @jlowin,

Win10 tests now also run successfully, please review the PR and merge if acceptable.

jlowin:
The referenced issues are already closed by other PRs and as far as I can tell, the tests in this PR pass against the main branch without modification. I believe the HTTPX client always includes its default headers.

PR #1158, opened 2025-07-15
Fix `fastmcp install` sub-command docs

richardkmichael:
Fix the docs for  the install sub-command:
 - `--name` to `--server-name`
 - `--env-var` to `--env`

Also, the docs gave me the impression `mcp-json` would emit the entire settings object, but it does not:

```
$ uv run fastmcp install mcp-json main.py -n "Enum MCP"
{
  "command": "uv",
  "args": [
    "run",
    "--with",
    "fastmcp",
    "fastmcp",
    "run",
    "/source/fastmcp/experiments/main.py"
  ]
}
```

This makes the `--server-name` argument useless for `mcp-json` (its existence is why I expected to receive the entire settings object).  Am I supposed to use `--server-spec <FILE>` to ask it to install to an existing file?  That option is marked as "required", but it is not ☝️, and:

```
$ uv run fastmcp install mcp-json main.py -n "Enum MCP" --server-spec foo.json
╭─ Error ──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ Unused Tokens: ['main.py'].                                                                                                  │
╰──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
```

A bug?  I can't quite intuit what is intended here.  Perhaps I've misunderstood the `mcp-json` command.

I glanced at the sub-command code, and see the shared aspect to the config object assembly.  I notice `install_mcp_config` is only used by the install sub-command, so perhaps it could emit with the wrapper if no `--server-spec` (and that argument was optional; as now, effectively).

Emitting a full config would be helpful, because I want to do this: `claude mcp-config foo.json` for quick testing, and *not install* the server to any Claude configuration.

Admittedly, this is trivial:

`$ uv run fastmcp install mcp-json main.py -n "Enum MCP" | jq --arg n "enum" '{mcpServers: {($n): .}}'`

But, one more thing to do , and I'm used to FastMCP doing so much for me.  😂  It's amazing, thank you!  🔥

### DIFF
--- a/docs/patterns/cli.mdx
+++ b/docs/patterns/cli.mdx
@@ -182,10 +182,10 @@ The `install` command supports the same `file.py:object` notation as the `run` c
 
 | Option | Flag | Description |
 | ------ | ---- | ----------- |
-| Server Name | `--name`, `-n` | Custom name for the server (defaults to server's name attribute or file name) |
+| Server Name | `--server-name`, `-n` | Custom name for the server (defaults to server's name attribute or file name) |
 | Editable Package | `--with-editable`, `-e` | Directory containing pyproject.toml to install in editable mode |
 | Additional Packages | `--with` | Additional packages to install (can be used multiple times) |
-| Environment Variables | `--env-var`, `-v` | Environment variables in KEY=VALUE format (can be used multiple times) |
+| Environment Variables | `--env`, `-v` | Environment variables in KEY=VALUE format (can be used multiple times) |
 | Environment File | `--env-file`, `-f` | Load environment variables from a .env file |
 
 **Examples**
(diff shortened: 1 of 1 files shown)


jlowin:
Ah, no, I think this is a classic "thought we were done, wrote the docs, then updated the logic and forgot to update docs" situation. I'll have another look at the logic - I believe it was intended to include the server name but NOT the `mcpServer` key, I'll confirm.

jlowin:
Thank you for highlighting this. The implementation was bad and the docs were directionally correct. I've implemented a full fix in #1185, which will supersede this PR.

richardkmichael:
I used this today, thank you!  Happy to help where I can, this is an excellent project.

Issue #1118, opened 2025-07-11
Adding custom arguments to Context


mukulmurthy:
### Enhancement Description

I was looking through the [Middleware](https://gofastmcp.com/servers/middleware) and [Context](https://gofastmcp.com/servers/context) docs and couldn't find an easy way to pass information through chained middleware or to the tool call. We can use a ContextVar as a workaround but were wondering if it made sense to build something into the framework. 

### Use Case

We plan to include an auth middleware that examines the Authorization header, verifies the token, and stores user information in the Context so the tool being called has access to it. 

### Proposed Implementation

Include a `data: Dict[String, Any]` or `Dict[Any, Any]` inside Context so that middleware and/or tool implementations can pass data around.

If this makes sense, I'm happy to put up a PR that adds this.

jlowin:
That's a good idea -- I have one hesitation that at this time I think we freely open multiple contexts on each call. For example `_mcp_list_resources()` opens a context and then calls `_list_resources()` which opens its own context, and any mounted subservers will have their own `_list_resources()` methods called as well. One of the reasons for this is so that the `fastmcp` attribute of the Context is always set to the "current" server in the event of multiple nested subservers. Certainly, this isn't an insurmountable problem (as nested contexts could simply inherit the data attribute of any parent context) but I just want to make sure there aren't any other side effects.

That said, I think this is a great idea and my only suggestion would be to call the attribute `state`, following convention from FastAPI. I would also suggest it is `dict[str, Any]` for simplicity. Happy to react to a PR if you want to open one including the implicit parent `state` load

mukulmurthy:
Thanks! I'll take a stab at implementing this in the next few days.

mukulmurthy:
Just put up a PR at https://github.com/jlowin/fastmcp/pull/1160! Haven't added documentation yet, but can do so if the API and semantics make sense.

tybalex:
Hi @mukulmurthy , would your PR allow tools to directly get access to request object? right now it is accessible via `get_http_request` but it is buggy and does not return request object correctly.

jlowin:
@tybalex can you open an issue with an MRE if `get_http_request` isn't working? Fixing it would be preferable than a workaround

tybalex:
cool I will open an issue instead.

mukulmurthy:
Completed in https://github.com/jlowin/fastmcp/pull/1160.

PR #1160, opened 2025-07-16
Add `state` dict to Context (#1118)

mukulmurthy:
Add an internal dict `state` inside `Context` that can be interacted with via the `get_state_value` and `set_state_value` APIs. This is intended for middleware or other stages of a request to pass information back and forth, since there is not an easy way to do that right now without using ContextVars. 

When a context is created, it inherits state from its parents. What that means is that:
* State set on a child context never affects the parent context.
* State set on a parent context after the child context is initialized is not propagated to the child context.

Proposed in https://github.com/jlowin/fastmcp/issues/1118. Tested with new tests in test_context.py.

### DIFF
--- a/src/fastmcp/server/context.py
+++ b/src/fastmcp/server/context.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import asyncio
+import copy
 import warnings
 from collections.abc import Generator
 from contextlib import contextmanager
@@ -83,8 +84,19 @@ class Context:
         request_id = ctx.request_id
         client_id = ctx.client_id
 
+        # Manage state across the request
+        ctx.state["user_id"] = "12345"
+        ctx.set_state_value("key", "value")
+        value = ctx.get_state_value("key")
+
         return str(x)
     ```
+
+    State Management:
+    Context objects maintain a state dictionary that can be used to store and share
+    data across middleware and tool calls within a request. When a new context
+    is created (nested contexts), it inherits a copy of its parent's state, ensuring
+    that modifications in child contexts don't affect parent contexts.
 
     The context parameter name can be anything as long as it's annotated with Context.
     The context is optional - tools that don't need it can omit the parameter.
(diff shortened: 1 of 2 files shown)


src/fastmcp/server/context.py
```diff
@@ -455,6 +473,14 @@ def get_http_request(self) -> Request:
 
         return fastmcp.server.dependencies.get_http_request()
 
+    def set_state_value(self, key: str, value: Any) -> None:
+        """Set a value in the context state."""
+        self._state[key] = value
+
+    def get_state_value(self, key: str) -> Any:
+        """Get a value from the context state. Returns None if the key is not found."""
+        return self._state.get(key)
```
mukulmurthy: Chose to expose helpers for this rather than the state dictionary directly, but I don't feel strongly about this. I can add documentation to https://gofastmcp.com/servers/context for this change once the API looks good.

src/fastmcp/server/context.py
```diff
@@ -83,9 +84,20 @@ def my_tool(x: int, ctx: Context) -> str:
...
+        value = ctx.get_state_value("key")
+
         return str(x)
     ```
 
+    State Management:
+    Context objects maintain a state dictionary that can be used to store and share
+    data across middleware and tool calls within a request. When a new context
+    is created (nested contexts), it inherits a copy of its parent's state, ensuring
+    that modifications in child contexts don't affect parent contexts.
```
mukulmurthy, line 98: Discussed the semantics of this a little in the PR description. If you think child state _should_ affect the parent, we could do it that way.

src/fastmcp/server/context.py
jlowin: I like the helper methods, my suggestion is to be slightly more parsimonious `get_state` and `set_state`

mukulmurthy:
Thank you for fixing up and merging @jlowin!

PR #1175, opened 2025-07-18
Fix icon size on Firefox

richardkmichael:
Unfortunately, with `size="14"` Mintlify emits unit-less non-zero lengths in violation of the CSS specification.. for the last 20 years or so.  🤯 Firefox strips such erroneous rules, which leaves the icons taking up a lot of space.

I'm not certain about ignoring the Node package files.  Obviously that wouldn't be possible if Node packages are ever used.

What's the docs workflow?  Something different with npm or a local-only gitignore?

A less invasive alternative:

`cd docs && npm i mint && npx mint dev`

```
# .gitignore
docs/node_modules
docs/package.json
docs/package-lock.json
```

I can adjust the README instructions, and/or drop that commit.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -63,3 +63,7 @@ dmypy.json
 
 # Claude worktree management
 .claude-wt/worktrees
+
+# Installing Mintlify for documentation will generate Node package files
+package.json
+package-lock.json


richardkmichael:
This is unrelated to OpenAPI -- the bot has added an incorrect label.  I don't seem able to remove it.

jlowin:
These are autogenerated by a different package. I've given @zzstoatzz a heads up to take a look.

PR #1178, opened 2025-07-18
Fix nesting when making OpenAPI arrays and objects optional

melkamar:
Prior to this change, the mangling of optional arrays and objects misses "related" properties. For example, given this OAS:

```json
  {
    "name": "someProperty",
    "required": false,
    "schema": {
      "type": "array",
      "items": {
        "type": "string"
      }
    }
  }
```

The resulting MCP tool schema would be:

```json
  {
    "someProperty": {
      "items": {
        "type": "string"
      },
      "anyOf": [
        {
          "type": "array"
        },
        {
          "type": "null"
        }
      ]
    }
  }
```

This is invalid (and e.g. OpenAI fails on it). Instead, the "items" property should be a part of the now-nested "array" type:

```json
  {
    "someProperty": {
      "anyOf": [
        {
          "type": "array",
          "items": {
            "type": "string"
          },
        },
        {
          "type": "null"
        }
      ]
    }
  }
```

This commit looks at all the properties defined in the JSON schema to belong to either array or object type and nests them.

### DIFF
--- a/src/fastmcp/utilities/openapi.py
+++ b/src/fastmcp/utilities/openapi.py
@@ -1114,10 +1114,56 @@ def _make_optional_parameter_nullable(schema: dict[str, Any]) -> dict[str, Any]:
     # Create a new schema that allows null in addition to the original type
     if "type" in schema:
         original_type = schema["type"]
+
         if isinstance(original_type, str):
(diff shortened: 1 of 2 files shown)


jlowin:
Thanks!

Issue #1195, opened 2025-07-20
FastMCP.from_openapi() takes too much time to create instance


itaru2622:
### Description

as described in https://github.com/jlowin/fastmcp/discussions/1191 , 
It takes too long time to create FastMCP instance from openAPI spec.
as the result, production mcp clients, such as VC copilot chat UI, often give up waiting FastMCP ready to serve when transport is stdio.

when replacing deepcopy implementation from builtin copy.deepcopy  into json.dumps+loads, it goes faster and 24% improved. i.e total time decreased to 3/4.

Ideally, we would get it faster, but unfortunately that may be difficult in a realistic time frame.

### Example Code

```Python
# the same as docs/integrations/openapi.mdx

from fastmcp import FastMCP
import httpx

client = httpx.AsyncClient()
# get spec from real production server, such as https://github.com/github/rest-api-description/tree/main/descriptions/*/*.json

spec = httpx.get("https://api.production.server.com/openapi.json").json()
mcp = FastMCP.from_openapi(spec, client=client)
#
# mcp client often gives up waiting FastMCP ready to serve.
#
print("mcp instance created.") # use logger and output message into file other than stdout/stderr, if transport==stdio.
mcp.run(transport='stdio')
```

### Version Information

```Text
FastMCP version:                                       2.10.6
MCP version:                                           1.12.0
Python version:                                        3.12.11

when I tested with python 3.13.5, the above investigation results are almost the same( just noise level differences)
```

jlowin:
I agree we can make improvements here but I just want to call out that OpenAPI specs of this size will be awful MCP servers. A good server should have on the order of 10 well defined tools. More than 100 would be extremely unusual and lead to bad performance. So while we should strive to make this better, the second we're talking about minutes to parse a file (the GitHub OpenAPI spec itself is 10 MB) we're not talking about effective MCPs anymore.

itaru2622:
@jlowin  I almost agree with you.
over 1000 tools in one MCP server may cause troubles, by AI miss understanding etc.
but to reduce num of MCP tools for existing REST server, it needs some tailoring works according to usecases  with try and errors.

I like features of this OSS, therefore it would be more nice to have performance.

PR #1196, opened 2025-07-20
getting faster for FastMCP.from_openapi()

itaru2622:
as described in https://github.com/jlowin/fastmcp/discussions/1191 and https://github.com/jlowin/fastmcp/issues/1195,

deepcopy implementation replaced from builtin copy.deepcopy into json.dumps+loads.


jlowin:
No, I don't like shadowing our copy operation with this imitation module. Why not just replace deepcopys with shallow copy's explicitly?

itaru2622:
@jlowin thank you for response.

when deepcopy replaced into shallow copy (copy.deepcopy -> copy.copy), then
updating copied object gives impacts into source object. 
in utilities/json_schema.py, it is going to prune something from instance, so I guess it is not safe to replace deepcopy into shallow copy.
but if you don't mind that, I will change the code as you mentioned.

regarding altcopy.copy,  I can delete it.

alternatives for deepcopy is:
backporting  copy.deepcopy from python main branch in locally  i.e. just https://github.com/python/cpython/blob/main/Lib/copy.py as newer  altcopy.py
in this case,  it gets almost the same performance as json.dumps+loads on my tiny investigation program.
but the volume of the code is longer than current PR code.

which one you like,  in below three choices:?
a)  replace deepcopy to shallow copy (copy.deepcopy -> copy.copy)
b)  backports copy(.deepcopy) from  the above cpython/Lib/copy.py  and memo its permanent link in backported source file.
c)  keep current deepcopy(json version) but delete copy function from altcopy.py

if you prefer others, let me know.

jlowin:
@itaru2622 I really appreciate how you've pushed this. In conjunction with the new parser (#1209), I have prepared some optimizations inspired by your analysis that I'll PR shortly. On my machine, parsing the entire GitHub API goes from 3+minutes to 1.5 seconds. Will be curious for your timing as well.

itaru2622:
@jlowin thank you for your work. 

I'm happy on hearing your good news.

PR #1205, opened 2025-07-21
Fix documentation regarding `@tool()` decorator usage

amureki:
We have plenty of doc examples that are failing because of the `TypeError: The @tool decorator was used incorrectly. Did you forget to call it? Use @tool() instead of @tool`

This PR:
- fixes all the places the decorator was missing parentheses
- adds missing newlines in the touched docs
- removes unnecessary whitespaces in the docs

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -19,7 +19,7 @@ When testing or investigating FastMCP servers, **always prefer the in-memory tra
 # Create your FastMCP server
 mcp = FastMCP("TestServer")
 
-@mcp.tool
+@mcp.tool()
 def greet(name: str) -> str:
     return f"Hello, {name}!"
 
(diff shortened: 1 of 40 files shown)


jlowin:
What version of FastMCP are you using? calling `mcp.tool` without parentheses was introduced way back in 2.7

amureki:
> What version of FastMCP are you using? calling `mcp.tool` without parentheses was introduced way back in 2.7

Aha, I figured, the import was coming from https://github.com/modelcontextprotocol/python-sdk

So they are vendoring the old version, which I oversaw...

jlowin:
Got it! Yup, if you import `from fastmcp import FastMCP` you should be able to use all the new features in 2.x

amureki:
> Got it! Yup, if you import `from fastmcp import FastMCP` you should be able to use all the new features in 2.x

Awesome, thank you!

Sorry for bothering! This makes sense 👍

Issue #1207, opened 2025-07-21
References is not pruned when using `oneOf`, `anyOf`, `allOf`


algirdasci:
### Description

When component has reference with `oneOf`, `anyOf` or `allOf` then it is treated as a root reference and not pruned from `$def` list even when no references to it exists.

Let's take schema:
```json
{
  "openapi": "3.0.0",
  "info": {
    "title": "Some API",
    "version": "1.0.0"
  },
  "paths": {
    "/test": {
      "get": {
        "summary": "tool for testing",
        "operationId": "123456",
        "responses": {
          "200": {
            "$ref": "#/components/schemas/Schema"
          }
        }
      }
    }
  },
  "components": {
    "schemas": {
      "SchemaWithOneOf": {
        "description": "Schema With One Of",
        "properties": {
          "boost": {
            "description": "Optional reference to SomeRelatedSchema (can be null)",
            "allOf": [
              {
                "$ref": "#/components/schemas/SomeRelatedSchema"
              }
            ],
            "nullable": true
          }
        },
        "type": "object"
      },
      "SomeRelatedSchema": {
        "properties": {
          "id": {
            "type": "integer"
          }
        },
        "type": "object"
      },
      "Schema": {
        "description": "Schema",
        "properties": {
          "some_field": {
            "type": "string",
            "example": "field_value"
          }
        },
        "type": "object"
      }
    }
  }
}
```

After this schema is loaded to FastMCP, `SomeRelatedSchema` will be added to `$defs`, even though it is referenced only in another schema, but not in the operation:

```json
[
    {
        "name": "toolForTesting",
        "title": null,
        "description": "tool for testing",
        "tags": [],
        "enabled": true,
        "parameters": {
            "type": "object",
            "properties": {},
            "required": [],
            "$defs": {
                "SomeRelatedSchema": {
                    "type": "object",
                    "properties": {
                        "id": {
                            "type": "integer"
                        }
                    }
                }
            }
        },
        "output_schema": null,
        "annotations": {
            "title": null,
            "readOnlyHint": null,
            "destructiveHint": null,
            "idempotentHint": null,
            "openWorldHint": null
        },
        "serializer": null
    }
]
``` 

[...]

PR #1208, opened 2025-07-21
fix: preserve def reference when referenced in allOf / oneOf / anyOf

algirdasci:
Suggested bugfix for #1207


jlowin:
Thank you for reporting and fixing!

Issue #1193, opened 2025-07-20
Claude Desktop Compatibility Issue: Null Parameters for Enum value


vl-kp:
## 🚨 **Problem Description**

Claude Desktop fails to properly handle JSON schemas containing `$ref` references, resulting in **null parameter values** instead of valid enum/model data. This causes parameter validation errors and breaks tool functionality.

<img width="713" height="560" alt="Image" src="https://github.com/user-attachments/assets/30d37a6a-ff6b-4c0d-8dab-d2aeb0732b6d" />

## 🔍 **Root Cause**

When FastMCP generates JSON schemas for complex Python types, it creates `$defs` sections with `$ref` references:

```json
{
  "$defs": {
    "TestColor": {
      "enum": ["red", "green", "blue"],
      "type": "string"
    }
  },
  "properties": {
    "color": {"$ref": "#/$defs/TestColor"}
  }
}
```

**Claude Desktop cannot process `$ref` references** and sends `null` values instead of the expected enum/model data.

## 🔬 **Steps to Reproduce**

```python
# test_claude_desktop.py
from enum import Enum
from typing import Dict, Any
from fastmcp import FastMCP

mcp = FastMCP("Test Server")

class TestColor(str, Enum):
    RED = "red"
    GREEN = "green"
    BLUE = "blue"

@mcp.tool
async def test_enum(color: TestColor) -> Dict[str, Any]:
    return {
        "status": "success",
        "received_color": color.value
    }

if __name__ == "__main__":
    mcp.run()
```
Try to use the tool in Claude Desktop

### **Expected Request:**
```json
{
  "color": "red"
}
```

### **Actual Request from Claude Desktop:**
```json
{
  "color": null
}
```

### **Resulting Error:**
```
Error executing tool: Cannot convert undefined or null to object
ValidationError: Input should be 'red', 'green' or 'blue'
```

### Proposed Implementation

```Python
## ✅ **Solution: Schema Dereferencing**

We've implemented automatic schema dereferencing that resolves `$ref` references in properties while preserving the original `$defs` section for compatibility.

### **Before (Problematic):**

{
  "$defs": {
    "TestColor": {"enum": ["red", "green", "blue"], "type": "string"}
  },
  "properties": {
    "color": {"$ref": "#/$defs/TestColor"}  // ❌ Claude Desktop can't handle this
  }
}


### **After (Compatible):**

{
  "$defs": {
[...]

PR #1215, opened 2025-07-22
RFC: Expose initialization requests to middleware, and provide session storage

richardkmichael:
This is a proof-of-concept I hacked up with Claude.  I'm not proposing it for merge as-is.

I haven't thought carefully about the consequences of these modifications.  This might be touching on stateful/less aspects of MCP; not sure and perhaps not any more so than progress tokens or other remembered data on either side of the connection.   If there is interest, I'll take a closer look.

I was intrigued by a middleware based solution for #1193 and #1159, but this requires modifying FastMCP.

Middleware documentation mentions all requests are exposed to middleware, but this is not true.  The initialization part of the lifecycle is handled in the session establishment and not available to middleware.

This hack adds:

- A new session class which exposes itself to middleware for `on_initialize`
- Adds the session to the MiddlewareContext so there is cross-request storage for data only available in specific requests

It allows a middleware [like this](https://github.com/richardkmichael/fastmcp-experiments/blob/development/dereference_for_claude_middleware.py#L223-L242):

```python
class DereferenceForClaudeMiddleware(Middleware):
    async def on_initialize(self, context: MiddlewareContext, call_next):
        """Detect Claude clients and store on session object."""
        client_info = getattr(context.message, "clientInfo", None)
        client_name = (
            getattr(client_info, "name", "unknown") if client_info else "unknown"
        )

        is_claude_client = self._is_claude_client(client_name)

        # HACK: Store client type directly on session object
        if hasattr(context, "session") and context.session:
            context.session.is_claude_client = is_claude_client

        if is_claude_client:
            logger.info(f"Claude ({client_name}) detected - will dereference schemas")
        else:
            logger.info(
                f"Non-Claude ({client_name}) detected - schemas will remain unchanged"
            )

    async def on_list_tools(self, context: MiddlewareContext, call_next):
        """Dereference tool schemas for Claude clients."""
        tools = await call_next(context)

        # HACK: Read client type from session object (stored during initialization)
        is_claude_client = False
        if context.fastmcp_context and context.fastmcp_context.session:
            is_claude_client = getattr(
                context.fastmcp_context.session, "is_claude_client", False
            )

        if is_claude_client:
            for tool in tools:
                original_schema = deepcopy(tool.parameters)
                dereferenced_schema = self.dereference_json_schema(tool.parameters)

                if original_schema != dereferenced_schema:
                    tool.parameters = dereferenced_schema

        return tools
```

I think exposing initialization to middleware would also be useful for:
- logging
- analytics (i.e. server usage by specific MCP clients)
- future proofing for other [misbehaved] clients

### DIFF
--- a/src/fastmcp/server/low_level.py
+++ b/src/fastmcp/server/low_level.py
@@ -1,5 +1,8 @@
-from typing import Any
+from contextlib import AsyncExitStack
+from typing import TYPE_CHECKING, Any
 
+import anyio
+import mcp.types as types
 from mcp.server.lowlevel.server import (
     LifespanResultT,
     NotificationOptions,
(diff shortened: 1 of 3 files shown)


jlowin:
@richardkmichael thanks! I went down a rabbit hole of supporting *every* JSONRPC call at first but it required way too much rewriting of the low-level internals to work. I really like that you cut through that by overwriting just the one method and ensuring it's in place in the overwritten .run() method. This is pretty compelling! I want to think on it a little just to make sure we aren't inadvertently taking on a broader maintenance burden than intended (though I strongly suspect your solution limits that as much as possible)

richardkmichael:
Thank you for the quick feedback!

I've just noticed #1160 -- but have not read it closely.  Perhaps its state dictionary can be used for storage.  Using the session was obviously a hack, but I needed that because there was no context at during initialization, IIRC. I certainly wouldn't want data stored in two places, so it would be worth looking at streamlining that aspect.

jlowin:
My suggestion is lets look at #1160 for state and keep this PR focused on initializae middleware; I think 1160 may have stalled so I'll look at getting it over the line today.

(update: merged 1160)

hopeful0:
Rewriting the `run` method of low-level server and extending `ServerSession` in Fastmcp is very useful. This gives us the opportunity to create a service session-level context to store some state that needs to be shared throughout the session, such as the session ID in #1242 and some stuff about proxy.

jlowin:
@hopeful0 I agree. I have hesitated on rewriting low-level SDK at all because of the increased maintenance burden it places on FastMCP but I think we're at a place where the surface area is small enough and the benefit is large enough that this is worthwhile.

rishid:
I have been toying with this myself and this is a good solution. I want to be able to get client info and collect metrics on this information. Even though documentation says `on_message - Called for ALL MCP messages (both requests and notifications)` it took me a while to figure out that `initialize` is not one of those message.

dacamposol:
I love this proposal!

As @rishid mentioned, this would allow us to collect metrics on some information across the entire MCP session; plus there are some cases where the MCP serves as a shim to other upstream services, where being able to communicate directly on server communication whether it's available to be used or not to the MCP hosts would be helpful.

jlowin:
I've added the middleware/initialize aspects of this PR to #1546 and will close this as I think it's become stale. Appreciate the head start though @richardkmichael!

richardkmichael:
Thank you!  I was going ask what to do here, but I've been busy.  Thanks for running with it! 👍

PR #1218, opened 2025-07-22
cache _load_tools() and has_tool() , both used by list_tools() and call_tool()

geobio:
Because each call to call_tool results in 5 calls to list_tools which can be slow when many MCP servers. In my case, I have 30+ MCP servers running. and each call to list_tools takes ~100ms x5 makes each request take a minimum of 500ms for call_tool().

To not change the functioanlity of fastmcp, by default, caching is disabled. User should do `Client(list_tools_cache_ttl=300)` to enable list_tools caching for 5 minutes.

### DIFF
--- a/src/fastmcp/client/client.py
+++ b/src/fastmcp/client/client.py
@@ -302,6 +302,16 @@ class Client(Generic[ClientTransportT]):
         for key in expired_keys:
             del _GLOBAL_LIST_TOOLS_CACHE[key]
 
+    async def _list_tools_for_schema_validation(self) -> None:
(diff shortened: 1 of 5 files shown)


jlowin:
Hi @geobio - there is already a tested cache implementation (configurable with `cache_expiration_seconds` on the FastMCP server) but we disabled it because of unclear behavior with middleware e.g. ensuring that middleware routes are called even when low-level tooling doesnt need to be exposed. I think we could more safely add this back in at this time, which will be a different implementation than the approach taken in this PR.

jlowin:
I also want to look into why the low-level SDK isn't caching like it's supposed to

PR #1221, opened 2025-07-22
fix: _replace_ref_with_defs; ensure ref_path is string, applying #1164 for new parser.

### LINKED ISSUE #1220: new experimental OpenAPI Parser fails when '$ref' is used for property name of schema, like github enterprise cloud openAPI spec
### Description

in new experimental OpenAPI parser, #1163 is raised again, because of the same reason. 
so it needs to apply  almost the same patch as #1164

https://github.com/jlowin/fastmcp/blob/d6bd800d88256918ebd7af45074c866d77c6cc2a/src/fastmcp/experimental/utilities/openapi/schemas.py#L94-L105

### Example Code

```Python
#code is the same as #1163
```

### Version Information

```Text
FastMCP version:                         2.10.7.dev20+d6bd800
MCP version:                                           1.12.0
Python version:                                       3.12.11

the above is latest main branch installed by

pip install git+https://github.com/jlowin/fastmcp@main
```

### Additional Context

using 
FASTMCP_EXPERIMENTAL_ENABLE_NEW_OPENAPI_PARSER=true

itaru2622:
fix #1220
related #1164

### DIFF
--- a/src/fastmcp/experimental/utilities/openapi/schemas.py
+++ b/src/fastmcp/experimental/utilities/openapi/schemas.py
@@ -93,15 +93,16 @@ def _replace_ref_with_defs(
     """
     schema = info.copy()
     if ref_path := schema.get("$ref"):
-        if ref_path.startswith("#/components/schemas/"):
-            schema_name = ref_path.split("/")[-1]
-            schema["$ref"] = f"#/$defs/{schema_name}"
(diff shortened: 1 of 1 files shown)


jlowin:
Thanks @itaru2622 - I think this is a duplicate of #1217 which has one additional check

PR #1222, opened 2025-07-22
Skip repeated type conversion and validation in proxy client elicitation handler

### LINKED ISSUE #1167: Default values in elicitation schemas error when attempting to Proxy
### Description

My setup is the folliowing: Client -> Proxy -> Server.  When the Server initiates elicitation, and the schema includes a default value, the proxy elicitation handler throws an error. The same schema works without using the proxy.

The issue arises because `json_schema_to_type` (incorrectly?) marks the field as nullable when the original type did not specify that, and the elicitation handler in the proxy tries to serialize the proxy type, it is invalid elicitation schema.

I have provided a minimal code which reproduces that while it's possible to get an elicitation schema for the initial type, the inverse type creates a different schema and throws an error

The relevant lines of code are here:
https://github.com/jlowin/fastmcp/blob/5fc7aa04aa82da6c06e757b4916df977a3f88f0f/src/fastmcp/utilities/json_schema_type.py#L551-L570

The other thing which I checked is that the ElicitRequestParams do have the json schema from the original server here:
https://github.com/jlowin/fastmcp/blob/5fc7aa04aa82da6c06e757b4916df977a3f88f0f/src/fastmcp/server/proxy.py#L545-L552

### Example Code

```Python
from pydantic import BaseModel, Field, TypeAdapter
from fastmcp.utilities.json_schema_type import json_schema_to_type
from fastmcp.server.elicitation import get_elicitation_schema
import json


def create_type() -> BaseModel:

    class TestModel(BaseModel):
        content: str = Field(description="Your reply content")
        acknowledge: bool = Field(
            default=False,
            description="Send immediately or save as draft",
        )
    return TestModel


initial = create_type()
initial_schema = json.loads(initial.schema_json())
# schema["type"] = "object"
inverse = json_schema_to_type(initial_schema)

print("Original:", get_elicitation_schema(initial))
print("Inverse Schema:", TypeAdapter(inverse).json_schema())
print("Inverse:", get_elicitation_schema(inverse)) # Errors out because acknowledge becomes nullable
```

### Version Information

```Text
FastMCP version:                                                                 2.10.5
MCP version:                                                                     1.10.1
Python version:                                                                  3.12.4
Platform:                                                    macOS-13.5-arm64-arm-64bit
```

### Additional Context

The output of the example code below is here:
```
Original: {'properties': {'content': {'description': 'Your reply content', 'title': 'Content', 'type': 'string'}, 'acknowledge': {'default': False, 'description': 'Send immediately or save as draft', 'title': 'Acknowledge', 'type': 'boolean'}}, 'required': ['content'], 'title': 'TestModel', 'type': 'object'}
Inverse Schema: {'properties': {'content': {'title': 'Content', 'type': 'string'}, 'acknowledge': {'anyOf': [{'type': 'boolean'}, {'type': 'null'}], 'default': False, 'title': 'Acknowledge'}}, 'required': ['content'], 'title': 'TestModel', 'type': 'object'}
```

chughtapan:
fixes #1167

Instead of going through the context, the proxy elicitation handler here assumes that the original server did schema validation properly and bypasses it by going through the session directly.

### DIFF
--- a/src/fastmcp/server/proxy.py
+++ b/src/fastmcp/server/proxy.py
@@ -566,11 +566,14 @@ class ProxyClient(Client[ClientTransportT]):
         A handler that forwards the elicitation request from the remote server to the proxy's connected clients and relays the response back to the remote server.
         """
         ctx = get_context()
-        result = await ctx.elicit(message, response_type)
-        if result.action == "accept":
-            return result.data
-        else:
-            return ElicitResult(action=result.action)
+        # Bypass schema conversion by calling session.elicit directly with the original schema
+        # This avoids the double conversion bug where fields with defaults become nullable
+        result = await ctx.session.elicit(
+            message=message,
+            requestedSchema=params.requestedSchema,
+            related_request_id=ctx.request_id,
+        )
+        return ElicitResult(action=result.action, content=result.content)
 
     @classmethod
     async def default_log_handler(cls, message: LogMessage) -> None:
--- a/src/fastmcp/server/proxy.py
+++ b/src/fastmcp/server/proxy.py
@@ -566,8 +566,6 @@ class ProxyClient(Client[ClientTransportT]):
         A handler that forwards the elicitation request from the remote server to the proxy's connected clients and relays the response back to the remote server.
         """
         ctx = get_context()
-        # Bypass schema conversion by calling session.elicit directly with the original schema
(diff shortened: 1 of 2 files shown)


jlowin:
Thank you for catching this! I think this has an important side effect of not "unwrapping" simple elicitation responses before forwarding to the remote server

Issue #1162, opened 2025-07-16
In-Memory Transport to support Auth


dacamposol:
### Enhancement Description

Some MCP Servers may have business logic attached to the `id_token` used to retrieve a certain resource.

In order to make this easily testable, the easiest would be to support `auth` for In-Memory transport.

### Use Case

```py
key_pair = RSAKeyPair.generate()

@pytest.fixture
def mcp_server():
    auth = BearerAuthProvider(
        public_key=key_pair.public_key,
        issuer="https://dev.example.com",
        audience="test-server",
    )

    server = FastMCP("TestServer", auth=auth)

    server.tool()(tools.MyFunctionOne) # The function is the actual tool, which relies on get_access_token() for some business logic
    server.tool()(tools.MyFunctionTwo)
    server.tool()(tools.MyFunctionThree)

    return server


async def test_MyFunctionOne(mcp_server):
    async with Client(
        mcp_server,
        auth=key_pair.create_token(
            subject="test@dev.example.com",
            issuer="https://dev.example.com",
            audience="test-server",
        ),
    ) as client:
        result = await client.call_tool(
            "MyFunctionOne",
            {
                "firstParameter": "foo",
                "secondParameter": "bar",
            },
        )
        assert isinstance(result, ExpectedEntity)
```

### Proposed Implementation

Probably, I would disable the possibility to provide a `jwks_uri` for the In-Memory Transport; as it wouldn't make much sense.

dacamposol:
Currently, if the previous code snippet is attempted, the following error appears from `pytest`:

```shell
========================================= short test summary info =========================================
FAILED src/platform/tests/tools/test_entity.py::test_MyFunctionOne - ValueError: This transport does not support auth
============================================ 1 failed in 1.59s ============================================
```

DeoLeung:
the pr has been closed, any update on this one?

dacamposol:
> the pr has been closed, any update on this one?

I don’t think it’s necessary anymore with the new `TokenVerifier` auth.

Do you have issues when passing an `JWTVerifier` using the `RSAKeyPair` to the test class?

DeoLeung:
in memory transport doesn't have `_set_auth` , it will complain it.

i use my own implementation token auth
```python
class CollectorAuthProvider(AuthProvider):
  authenticate_client: AuthenticateClient

  def __init__(self, authenticate_client: AuthenticateClient):
    self.authenticate_client = authenticate_client
    self.resource_server_url = None

[...]

PR #1223, opened 2025-07-22
feat (#1162): Introduce Bearer Auth for In-Memory Transport

dacamposol:
## Context

Developers may implement some business logic regarding the received JWT Token which is being provided as part of the Bearer Authentication, like claims verification or similar.

Since this logic may be introduced, I consider the easiest way to test it would by allowing bearer auth directly in the in-memory transport, as stated in #1162, plus it doesn't really affect the business logic on those not implementing it.

### DIFF
--- a/docs/patterns/testing.mdx
+++ b/docs/patterns/testing.mdx
@@ -37,6 +37,51 @@ async def test_tool_functionality(mcp_server):
 
 This pattern creates a direct connection between the client and server, allowing you to test your server's functionality efficiently.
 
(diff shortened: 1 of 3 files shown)


jlowin:
I dont think this is going to end up being a satisfactory solution. The purpose of the bearer auth utility is to provide an httpx-compatible Bearer header for remote transports. However the local server has no way (or expectation) of receiving this information, so implementing it in the client without a server-side protocol is misleading at best. This can't even be used for testing. There are proposals to add certain universal elements (such as headers) to all server transports, including local, but I do not want the cart to get ahead of the horse on that one.

PR #1226, opened 2025-07-22
Permit.io integration docs

orweis:
This PR adds comprehensive documentation for integrating Permit.io authorization middleware with FastMCP servers, enabling fine-grained policy-based access control for MCP tools, resources, and prompts.


jlowin:
Thanks @orweis - please add the document to docs.json so it shows up in the left hand nav; the integrations section is sorted alphabetically!

Issue #956, opened 2025-06-26
Session id is None in context


chinmoy12c:
### Description

Continuing the trail here:

https://github.com/jlowin/fastmcp/issues/933#issuecomment-3006732438

I'm currently not able to get session id in context inside my tool

### Example Code

```Python
@mcp.tool()
async def my_highly_useful_tool(
    ctx: Context, very_important_arg: int
) -> set[str] | str:
    

    print(f"Current session: {ctx.session_id}") # Session ID is None here
    pass
```

### Version Information

```Text
### Version Information

Fastmcp Version: 2.9.0
```

### Additional Context

I Have added relevant screenshots in the previous thread.

Also I'm using `http` transport.

jlowin:
Are you supply the MCP-SESSION-ID header from your client?

jlowin:
please provide a complete MRE

iutx:
I used the same multiple clients during the test, cursor, cherry-studio and MCP Inspector. 
Using [mcp-go](github.com/mark3labs/mcp-go) sse mode, can get the session id from context.

```go
return func(ctx context.Context, req mcp.CallToolRequest) (cr *mcp.CallToolResult, err error) {
		start := time.Now()
		sessionID := server.ClientSessionFromContext(ctx) // Can get
```

```python
  async def on_request(self, context: MiddlewareContext, call_next):
        print(f"{context.fastmcp_context.session_id}")
        return await call_next(context) -> None
```

chinmoy12c:
> please provide a complete MRE

```
def main() -> None:

    mcp: FastMCP = FastMCP(name="mymcpserver")

    @mcp.tool()
    async def ping(ctx: Context) -> str:
        """A simple ping tool to check if the server is running."""
        print(f"Current session: {ctx.session_id}")
        return "pong"

    # Run server
    mcp.run(transport="http", host=HOST, port=PORT, path="/mcp")


if __name__ == "__main__":
    main()

```

Using this code I've tried connecting this server to co-pilot and claude desktop both.
The session id is None.

chinmoy12c:
@jlowin any additional details required to triage this issue?

rishabhgupta592:
Facing the same issue, @chinmoy12c were you able to find any resolution on this one?

LaptopKing:
[...]

PR #1242, opened 2025-07-23
Refactor `get_http_request` and `context.session_id`

hopeful0:
According to the discussion by @gumgum27 in #956, in SHTTP transports, we can try to obtain the HTTP request from `request_ctx`. I think this aligns more with the original design intent of the `get_http_request` method, while also solving the issue of the session ID being `None`.

The second commit refactors the `session_id` of context; when the session ID cannot be found, it generates a UUID string as a substitute. This allows the concept of the session ID to extend from stateful SHTTP to SSE, stdio, in-memory, and other transports, which might be useful. This change caused the old tests to fail, so I rewrote the tests. 

Claude's review pointed out a thread safety issue with the generation process, but this implementation should be coroutine-safe, so I kept this commit.

### DIFF
--- a/src/fastmcp/server/dependencies.py
+++ b/src/fastmcp/server/dependencies.py
@@ -37,9 +37,14 @@ def get_context() -> Context:
 
 
 def get_http_request() -> Request:
-    from fastmcp.server.http import _current_http_request
+    from mcp.server.lowlevel.server import request_ctx
 
-    request = _current_http_request.get()
+    request = None
+    try:
+        request = request_ctx.get().request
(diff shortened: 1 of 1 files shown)


jlowin:
@hopeful0 thanks very much for this thoughtful implementation

qqaatw:
Thank you @hopeful0 and @jlowin

vmazi:
Commenting for those who come after

Use

`session_id = ctx.request_context.request.headers.get("mcp-session-id") `

Until this PR is released

PR #1245, opened 2025-07-23
Fix stateful proxy client mixing in multi-proxies sessions

hopeful0:
The stateful proxy client cached in the server session gets mixed in a server session that mounts multiple proxies, so this PR moves the cache inside the stateful proxy client and ensures it supports concurrent server sessions. Additionally, a related regression test was added.

### DIFF
--- a/src/fastmcp/server/proxy.py
+++ b/src/fastmcp/server/proxy.py
@@ -7,6 +7,7 @@ from typing import TYPE_CHECKING, Any, cast
 from urllib.parse import quote
 
 import mcp.types
+from mcp import ServerSession
 from mcp.client.session import ClientSession
 from mcp.shared.context import LifespanContextT, RequestContext
 from mcp.shared.exceptions import McpError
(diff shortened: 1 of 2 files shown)


src/fastmcp/server/proxy.py
```diff
@@ -566,12 +567,11 @@ async def default_elicitation_handler(
         A handler that forwards the elicitation request from the remote server to the proxy's connected clients and relays the response back to the remote server.
         """
         ctx = get_context()
-        result = await ctx.session.elicit(
```
jlowin: I modified this to use the underlying session's `elicit` instead of `ctx.elicit` to avoid any user-friendly wrapping/rewrapping which could inadvertently modify the response. Was thinking we should try to pass the elicitation through with as little handholding as possible.

hopeful0: Sorry, this is because I accidentally reverted this change when organizing the commit. I agree that this should use the session's `elicit`. I have updated the commit.

strawgate:
I've been running into a lot of issues with proxy servers that are easily explained if this was the issue, excited to see if this resolves it

Issue #1189, opened 2025-07-20
Hidden parameters in Transformed Tools keep their $Defs


strawgate:
### Description

now we have almost the opposite problem https://github.com/jlowin/fastmcp/issues/1065

if i have an argument which uses a model which results in the usage of a $def / ref, when the argument is hidden the $def / ref is still retained in the schema

### Example Code

```Python

```

### Version Information

```Text
latest
```

jlowin:
Closed in #1257

PR #1257, opened 2025-07-24
Fix: prune hidden parameter defs

muhammadkhalid-03:
I fixed issue #1189 by running the compress_schema method right after we create the "$defs" schema property from parent_defs. I also added a test case to ensure that the hidden parameter's $defs entry is removed in the output schema


jlowin (approved):
Thank you!

PR #1264, opened 2025-07-25
Implemented a client-side feature use call_tool to **stream server output in chunks**, i.e., yielding chunks.

1810912213qqcom:
I recently implemented this feature using a somewhat hacky approach while making as few changes as possible to the original SDK. Later, I tried integrating the feature as an extension of client.call_tool directly into the client class. I’ll now present both approaches.
Let's take a look at the results now. I test the function with postman.

This is the result response in chunks.
7c45b28f02ef70f61e598c7b3f32869d

And here is the mcp server tool code.

22ebdc36dc4a0fd72e873f1438075092
Of cource you can only yield one chunk , which is the tool finally return.

64ef50f1b3e6db6b462edee34d9f3700
0621b876d90e84debd6a40b9d03a31a5

What I did was provide a custom implementation of the ProgressHandler used inside client.call_tool. I had first tried to yield chunks directly from the handler, but that’s impossible when call_tool itself is awaited. So I extended call_tool so that the custom progress states can also be yielded out.

I didn’t change the original @server.tool at all, so you can still use it exactly as before. Whenever you want to return additional states, just send your custom protocol—say, a JSON chunk—via ctx.report_progress().

Now let’s look at the code, starting with the interface layer.
```python
@app.post("/v1/mcp/tinker")
async def call_mcp_tools(request: Request):
    body = await request.json()
    return StreamingResponse(
        custom_mcp(url=body["url"], funname=body["name"], args=body["args"]),
        media_type="text/event-stream",
        headers={"Cache-Control": "no-cache", "Connection": "keep-alive"},
    )


async def custom_mcp(url: str, funname: str, args: dict):

    try:
        try:
            async with Client(url) as client:

                async for chunk in client.call_tool_stream_chunks(
                    name=funname, arguments=args
                ):
                    yield f"data: {json.dumps({"chunk":chunk},ensure_ascii=False)}\n\n"

        except Exception as e:
            print("task failed", e)

    except Exception as e:
        logger.error(f"Error: {e}")
        async for sendchunk in stream_completions_create(
            content=str(e), type="mcp_tool_error"
        ):
            yield f"data: {json.dumps(sendchunk,ensure_ascii=False)}\n\n"
```
Next, we created an enhanced version of call_tool as a custom method inside the Client class.
```python

    async def call_tool_stream_chunks(
        self,
        name: str,
        arguments: dict[str, Any] | None = None,
        timeout: datetime.timedelta | float | int | None = None,
        progress_handler: ProgressHandler | None = None,
        raise_on_error: bool = True,
    ):

        mcp_progress_queue = asyncio.Queue()

        async def state_update(progress: float, total: float = None, message: str = None):

            try:
                if message: 
                    logger.info(f"catch {message}")
                    await mcp_progress_queue.put(message)

            except ValueError as e:
                logger.error(f"no message {str(e)}")


        async def fetch_mcp_progress():
            try:
                 
                result = await self.call_tool_mcp(
                    name=name,
                    arguments=arguments or {},
                    timeout=timeout,
                    progress_handler=state_update,
                )
                data = None
                if result.isError and raise_on_error:
                    msg = cast(mcp.types.TextContent, result.content[0]).text
                    raise ToolError(msg)
                elif result.structuredContent:
                    try:
                        if name not in self.session._tool_output_schemas:
                            await self.session.list_tools()
                        if name in self.session._tool_output_schemas:
                            output_schema = self.session._tool_output_schemas.get(name)
                            if output_schema:
                                if output_schema.get("x-fastmcp-wrap-result"):
                                    output_schema = output_schema.get("properties", {}).get(
                                        "result"
                                    )
                                    structured_content = result.structuredContent.get("result")
                                else:
                                    structured_content = result.structuredContent
                                output_type = json_schema_to_type(output_schema)
                                type_adapter = get_cached_typeadapter(output_type)
                                data = type_adapter.validate_python(structured_content)
                            else:
                                data = result.structuredContent
                    except Exception as e:
                        logger.error(f"Error parsing structured content: {e}")

                tempresult= CallToolResult(
                    content=result.content,
                    structured_content=result.structuredContent,
                    data=data,
                    is_error=result.isError,
                )
                logger.info(f"result:{tempresult} {type(tempresult)}")
                await mcp_progress_queue.put(tempresult)
                logger.info(f"run success, result:{tempresult}")
                await mcp_progress_queue.put(None)
            except Exception as e:
                logger.error(f"task failed: {str(e)}")
        mcp_task = asyncio.create_task(fetch_mcp_progress())
        mcp_done = False
        while not mcp_done:
            try:
                mcp_data = await mcp_progress_queue.get()
                if mcp_data is None:
                    mcp_done = True
                    break
                else:
                    if type(mcp_data) == str:
                        yield mcp_data
                    elif type(mcp_data) == CallToolResult:

                        data = {
                            "content": vars(mcp_data.content[0]),
                            "structured_content": mcp_data.structured_content,
                            "data": mcp_data.data,
                            "is_error": mcp_data.is_error
                        }
                        yield data
                    await asyncio.sleep(0.0001)

            except asyncio.QueueEmpty:
                await asyncio.sleep(0.01)  
        await asyncio.gather(mcp_task)
```

### DIFF
--- a/src/fastmcp/client/client.py
+++ b/src/fastmcp/client/client.py
@@ -901,22 +901,20 @@ class Client(Generic[ClientTransportT]):
         progress_handler: ProgressHandler | None = None,
         raise_on_error: bool = True,
     ):
-
         mcp_progress_queue = asyncio.Queue()
 
-        async def custom_progress_handler(progress: float, total: float | None, message: str | None) -> None:
-
+        async def custom_progress_handler(
+            progress: float, total: float | None, message: str | None
+        ) -> None:
             try:
-                if message: 
+                if message:
                     await mcp_progress_queue.put(message)
 
             except ValueError as e:
                 logger.error(f"no message {str(e)}")
 
-
         async def fetch_mcp_progress():
             try:
-                 
                 result = await self.call_tool_mcp(
                     name=name,
                     arguments=arguments or {},
@@ -935,10 +933,12 @@ class Client(Generic[ClientTransportT]):
                             output_schema = self.session._tool_output_schemas.get(name)
                             if output_schema:
                                 if output_schema.get("x-fastmcp-wrap-result"):
-                                    output_schema = output_schema.get("properties", {}).get(
+                                    output_schema = output_schema.get(
+                                        "properties", {}
+                                    ).get("result")
+                                    structured_content = result.structuredContent.get(
                                         "result"
                                     )
-                                    structured_content = result.structuredContent.get("result")
                                 else:
                                     structured_content = result.structuredContent
                                 output_type = json_schema_to_type(output_schema)
@@ -949,7 +949,7 @@ class Client(Generic[ClientTransportT]):
                     except Exception as e:
                         logger.error(f"Error parsing structured content: {e}")
 
-                tempresult= CallToolResult(
+                tempresult = CallToolResult(
                     content=result.content,
                     structured_content=result.structuredContent,
                     data=data,
@@ -959,6 +959,7 @@ class Client(Generic[ClientTransportT]):
                 await mcp_progress_queue.put(None)
             except Exception as e:
                 logger.error(f"task failed: {str(e)}")
+
         mcp_task = asyncio.create_task(fetch_mcp_progress())
         mcp_done = False
         while not mcp_done:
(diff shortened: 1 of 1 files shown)


1810912213qqcom:
The main reason I implemented this method is that some MCP servers' tools take a *very* long time to run. While the client and server can use `ctx.report_progress` to transmit messages, those messages can’t be yielded from the client’s `ProgressHandler`.  

This new approach lets me stream both the intermediate progress messages *and* the final `tool_call` result together—especially useful when the front-end and back-end are wired up via `openai.StreamingResponse`. My own skill set is limited, so I couldn’t alter the underlying protocol; this was the compromise that worked.Although the implementation is far from elegant, it has—after considerable effort—managed to achieve the desired behavior.

jlowin:
This is a neat approach; however, I do not think it has a place in FastMCP unless we also implement a canonical server version (which is something being debated in the upstream mcp protocol). Otherwise this is a client method that is tightly coupled to an undocumented and unimplemented server-side pattern, and anyone trying to use it without that pattern will get unexpected results. In this form, I think this is best used as a custom client implementation in your own library, rather than in the global framework.

PR #1265, opened 2025-07-25
Update mcp>=1.12.2 pyproject.toml

x0rb0t:
When making uv sync it installs old dependency by default which makes server incompatible with Copilot Studio Connectors, described here: https://community.powerplatform.com/forums/thread/details/?threadid=7ec056e9-f950-f011-877a-7c1e5247028a  and resolved here: https://github.com/modelcontextprotocol/python-sdk/pull/1178

So this change will ensure that latest version is installed.


jlowin:
Thanks for raising this @x0rb0t — we only raise dependency lower bounds when FastMCP requires functionality from a newer version. This bug is in mcp, not FastMCP, so the truth is we remain compatible with mcp>=1.10, even if newer versions of mcp behave better. We don’t pin to avoid upstream bugs except in extreme circumstances—if we did that for all dependencies, `pyproject.toml` would imply FastMCP only works with the latest of everything, which isn’t the case.

However, if you’re trying to upgrade your own environment, you can run the following to get the latest compatible version of all dependencies:

```
uv sync -U
```

That’ll get you `mcp 1.12.2` without changing our declared pin. The next FastMCP release will likely bump the lockfile as well.

x0rb0t:
Got you. Right, I have already fixed this issue locally. Hope it will save time for whoever finds this pull request at least 😀

Issue #1256, opened 2025-07-24
Introduce uv transport


strawgate:
### Enhancement Description

I think we should introduce a uv studio transport which would enable #53 and simplify the CLI #1249 

### Proposed Implementation

```Python

```

jlowin:
Close in #1270

Issue #1249, opened 2025-07-23
FastMCP cli call tool


strawgate:
### Enhancement Description

How would we feel about enabling simple CLI tool invocations?

Or documenting a "nice" way to do it via a third-party tool?

### Proposed Implementation

```Python

```

strawgate:
or minimally `--arguments-json '{stringified json goes here}'`

jlowin:
I like this idea! So the way I'm kind of thinking of it is its analogous to `fastmcp run` except instead of loading a server and running it indefinitely on a transport, it loads a server and exceutes a single interaction with it. 

`fastmcp call config.json --name add --args {'x': 1, 'y': 2}` or something

so it takes all the same args as `run` so it can run against local/remote/config servers, but then it also takes instructions for the interaction -- list tools, call tool, list resources, read resource, etc. 

It's basically a single-turn CLI MCP client

Is `call` the right subcommand if we include things other than tools (which I think will be the first enhancement request 😆 )?

jlowin:
`fastmcp call`
`fastmcp interact`
`fastmcp use`
`fastmcp do`

strawgate:
In my agents cli I was putting this all under a `cli call` flag but `cli` as a command is a bit awkward.

I think everything will be a little weird because these are all action verbs

`list tools`
`call tool`
`read resource`
`list resources`
`get prompt`
`list prompts`


`call list tools`
`call tool`
`call list resources`
`call read resource`
`call list prompts`
`call get prompt`

strawgate:
Maybe use is good because you'll have the server.py or the config.json

`fastmcp use config.json list tools`
`fastmcp use config.json read resource --name=...`

This is what I was trying to hash out on the call when we kept getting into the weeds 😂 because if you can call a tool from the CLI then you can call an agent ..

strawgate:
Cyclopts doesn't do a good job with kwargs (all values are strings) unless you do prefixed args which is a bit annoying so my guess is it'll be json

strawgate:
Cyclopts also doesnt support any sort of command hierarchy where parent commands take arguments and handle them like Click does.

I'm thinking we use the dataclass/pydantic support to build a server object that we can re-use across the commands.

I'm also thinking that we don't need to support taking transport/host/port/etc for this?

The following code works:

```
app.command((call_app := cyclopts.App(name="call")))

call_app.command((list_app := cyclopts.App(name="list")))



class UvStdioTransport(StdioTransport):
[...]

Issue #53, opened 2024-12-05
Support running modules


paulrobello:
Nearly all my projects are laid out as a module ie. python -m 'module_name'
It would be awesome if I could use fastmcp dev -m 'module_name'

iljabauer:
For the run command which uses uv it can be build my allowing the -m flag:

https://github.com/astral-sh/uv/pull/7754

Should I create a PR? Do we need support for the non uv style run_command?

jlowin:
I actually think this would be quite useful -- today, both `fastmcp run` and `fastmcp dev` (which generates a `fastmcp run` command under the hood) parse the file you provide, import the MCP server, and call its `run()` method. This is why we can change transports and stuff on the fly. However, this *ignores* any `if __name__ == "__main__"` blocks and any specific configuration therein. There is a persistent request to respect those blocks (and the typical response is to just use `python <file>` instead of `fastmcp run <file>`. So I think passing a -m flag into fastmcp run could be quite effective but may require a little more work than just passing the flag to `uv` because I do think we will want to avoid our "import server object" logic, if that makes sense.

PakitoSec:
Hi. Do you have any alternative for launching module project via fastmcp cli ? Would be great

PR #1270, opened 2025-07-25
Add UV Transport

strawgate:
Related to:
- https://github.com/jlowin/fastmcp/issues/1256
- https://github.com/jlowin/fastmcp/issues/53
- https://github.com/jlowin/fastmcp/issues/1249

If we have a UV transport we can switch the CLI to just initializing the Transport instead of doing all of the heavy lifting directly in the CLI


jlowin (commented):
Looks great but can't figure out why tests are failing! They are running for me locally

PR #1287, opened 2025-07-29
feat: Allow Resource Metadata URL as field in OAuthProvider

### LINKED ISSUE #1276: Allow to provide Resource Metadata URL to RequireAuthMiddleware
### Enhancement Description

The latest MCP Protocol modification doesn't require the MCP Server to implement the Authorization Server by itself, and it defines the process to configure an external Identity Provider within the whole flow.

The MCP Client will first issue a request, which will receive an `HTTP 401 Unauthorized` with the `WWW-Authenticate` header, where it's expecting the `resource_metadata` field as defined in [RFC 9728](https://datatracker.ietf.org/doc/html/rfc9728#name-www-authenticate-response).

Currently, when `.auth` is provided to the FastMCP Server, it is being wrapped by the `RequireAuthMiddleware`:

```py
    if auth:
        auth_middleware, auth_routes, required_scopes = (
            setup_auth_middleware_and_routes(auth)
        )

        server_routes.extend(auth_routes)
        server_middleware.extend(auth_middleware)

        server_routes.append(
            Mount(
                streamable_http_path,
                app=RequireAuthMiddleware(handle_streamable_http, required_scopes),
            )
        )
```

but since the last change, this middleware allows to provide a `resource_metadata_url` which will be forwarded in the `WWW-Authenticate` header.

By just allowing to provide this endpoint, some MCP Clients like GitHub Copilot would automatically catch the information about the external Identity Provider to issue an access token to forward.

### Use Case

_No response_

### Proposed Implementation

```Python
1. Consider changing the `auth` field in the FastMCP server from an `OAuthProvider` to a more general settings object. 
2. Use this settings object to construct a `TokenVerifier` instance, and pass the `resource_metadata_url` from the settings to the `RequireAuthMiddleware`.
```

dacamposol:
## Context

Currently, the `WWW-Authenticate` header isn't returning the `resource_metadata` field, so MCP Hosts like GitHub Copilot don't know where to look for the AS to be used according the the latest protocol update.

Resolves #1276

### DIFF
--- a/src/fastmcp/server/auth/auth.py
+++ b/src/fastmcp/server/auth/auth.py
@@ -21,6 +21,7 @@ class OAuthProvider(
         client_registration_options: ClientRegistrationOptions | None = None,
         revocation_options: RevocationOptions | None = None,
         required_scopes: list[str] | None = None,
+        resource_server_url: AnyHttpUrl | str | None = None,
     ):
         """
         Initialize the OAuth provider.
@@ -43,6 +44,9 @@ class OAuthProvider(
         self.client_registration_options = client_registration_options
         self.revocation_options = revocation_options
         self.required_scopes = required_scopes
+        self.resource_server_url = (
+            AnyHttpUrl(resource_server_url) if resource_server_url else None
+        )
 
     async def verify_token(self, token: str) -> AccessToken | None:
         """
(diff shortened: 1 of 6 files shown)


jlowin:
Thank you!

PR #1303, opened 2025-07-30
Remove unused TimedCache 

strawgate:


Nilzkool:
Hi @strawgate, I noticed you've been keeping docs up-to-date, always good to see.

I built a small GitHub app for keeping high-level docs (SDK guides, API references, examples, tutorials, etc.) up-to-date whenever you commit changes.

No pressure, just thought it might be useful. It's free for OSS: https://github.com/marketplace/deepdocsai

jlowin:
Now that the low-level SDK makes *multiple* calls to `list_tools` for every `call_tool` call (🙄) I actually think this might have new utility... although it's true we're not using it ATM. Originally was added for proxy servers and other dynamic tool generation situations to avoid potential latency when listing / loading tools.

strawgate:
It might be nice to find an alternative implementation or to at least make sure we're using monotonic timers

jlowin (approved):

PR #1306, opened 2025-07-30
docs: update get_state and set_state references

Maxi91f:
Only updates docs.

Fixes a small inconsistency in the docs where, I think, `get_state_value` and `set_state_value` shouldn't have `_value`.


Maxi91f:
Sidenote: I almost went crazy man 😅 I was using the documentation as a guide trying to use `set_state` until I realized this wasn't released yet 😅 .

Is there a release planned soon?

jlowin:
Good catch! Thought I spotted all of these. We are trying to prep a large auth refactor for 2.11 but it's taking longer than expected... though so many people want to use the context state I might release early

PR #1307, opened 2025-07-31
(RFC) contrib: better validation of external resources thorough MCP

chughtapan:
This PR introduces a simplified implementation of a pattern I'm exploring for providing a scalable and safe approach to exposing external data through MCP.

The core challenge is that many tools need to operate with external resources—web APIs, database URLs, files, etc. Currently, tools accessing these resources must either be open-world (allowing access to any URL/file) or implement custom guardrails in each client/server-tool combination.

This implementation offers a straightforward solution using middleware to validate resource access, with hints and annotations defining expected behavior. A key design challenge was determining how to define the universe of valid endpoints/URLs that a server should access. My approach uses MCP resources (and resource templates) to declare valid endpoints, which the middleware then queries for validation.

I'm happy to hear thoughts, feedback, and suggestions around this space in general!

### DIFF
--- a/src/fastmcp/contrib/external_resources/README.md
+++ b/src/fastmcp/contrib/external_resources/README.md
@@ -1,0 +1,91 @@
+# External Resources for FastMCP
+
+A safe, controlled, and scalable approach to exposing external data to LLMs using MCP as a gateway protocol.
+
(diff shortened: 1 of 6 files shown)


takauji99:
@chughtapan This looks promising and here is my take.

This RFC adds important security features like fine-grained access control and policy enforcement, giving teams better ways to manage and restrict usage. It does increase complexity for developers though, so clear documentation and examples will be essential. With the right DX support, this could provide a strong path toward a secure and standardized MCP setup.

Have you tried this out as a concept or prototype in an SMB or enterprise setting with real data?

PR #1326, opened 2025-08-01
Structured client-side logging

cjermain:
This PR adds support for structured client-side logging, allowing easier processing of the structured logs in downstream systems for security monitoring and other applications. It adds the `extra` argument from the Python-standard `logging` library (see [the Python docs on `debug` for their argument definitions](https://docs.python.org/3/library/logging.html#logging.Logger.debug)). It introduces `extra` to the `Context` log methods, and routes the information through a `LogData` object, replacing the existing `message: str`.

- Proxies are handled correctly.
- Tests are updated to reflect the change, using the `caplog` fixture to assert the actual log message are written.
- The documentation is updated to reflect the changes to the `log_handler` on the client, showing how the MCP logs can be passed to the Python `logging` API.

```python
@mcp.tool
async def process_transaction(transaction_id: str, amount: float, ctx: Context):
    await ctx.info(
        f"Processing transaction {transaction_id}",
        extra={
            "transaction_id": transaction_id,
            "amount": amount,
            "currency": "USD"
        }
    )
    # ... processing logic ...
```

What do you think?

### DIFF
--- a/tests/client/test_logs.py
+++ b/tests/client/test_logs.py
@@ -6,18 +6,23 @@ from mcp import LoggingLevel
 from fastmcp import Client, Context, FastMCP
 from fastmcp.client.logging import LogMessage
 
-LOGGING_LEVEL_MAP = logging.getLevelNamesMapping()  # pyright: ignore [reportAttributeAccessIssue]
-
 
 class LogHandler:
     def __init__(self):
         self.logs: list[LogMessage] = []
         self.logger = logging.getLogger(__name__)
+        # Backwards-compatible way to get the log level mapping
+        if hasattr(logging, "getLevelNamesMapping"):
+            # For Python 3.11+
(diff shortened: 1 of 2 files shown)


tests/client/test_logs.py
```diff
@@ -1,17 +1,27 @@
+import logging
+
 import pytest
 from mcp import LoggingLevel
 
 from fastmcp import Client, Context, FastMCP
 from fastmcp.client.logging import LogMessage
 
+LOGGING_LEVEL_MAP = logging.getLevelNamesMapping()  # pyright: ignore [reportAttributeAccessIssue]
```
cjermain: This was introduced in Python 3.11. I can add a workaround for older versions.

jlowin:
This is great!

PR #1327, opened 2025-08-01
Add comprehensive OAuth 2.1 authentication system with WorkOS integration

### LINKED ISSUE #1273: TokenVerifier instead of OAuthProvider for auth
### Enhancement Description

Hi, in the new spec, the fastmcp server nolonger need to implement the full oauth2 spec, but only a token verifier . Is it possible to support that in fastmcp? Right now it requires me to implement OAuthProvider class for `auth` with verify_token method, which is similar to TokenVerifier, but the type is not compatible.

### Use Case

_No response_

### Proposed Implementation

```Python

```

jlowin:
FastMCP's (all of MCP's?) authentication story has been fragmented and difficult to navigate. Users needed deep OAuth knowledge to implement secure authentication, and the existing patterns didn't follow emerging MCP standards. This creates a unified, user-friendly authentication system that makes secure OAuth as simple as setting environment variables.

I worked with our partners at WorkOS to ensure that we had an incredibly smooth experience for AuthKit, and the patterns we're establishing here should make it easy to build integrations for other providers as well.

The new system introduces three clear authentication patterns that match real-world needs: token verification for internal systems, remote authentication for most production apps, and full OAuth servers for advanced use cases. Most users should use remote authentication with providers like WorkOS, which now works out of the box.

```python
# Before: Complex, manual OAuth setup required
# Users had to implement their own OAuth flows

# After: One-line authentication with WorkOS
from fastmcp.server.auth.providers.workos import AuthKitProvider

mcp = FastMCP("My App", auth=AuthKitProvider(
    authkit_domain="https://my-app.authkit.app",
    base_url="https://my-fastmcp-server.com"
))
```

The client-side OAuth flow now handles the complete browser-based authentication automatically, including dynamic client registration, callback server management, and token caching. Users get secure authentication without needing to understand OAuth internals.

**Key additions:**

- **AuthKitProvider**: WorkOS integration with automatic JWT verification and metadata forwarding
- **OAuth client**: Browser-based authentication with automatic token management and caching  
- **Unified documentation**: Clear guidance on choosing the right authentication pattern
- **Environment variable configuration**: Set `FASTMCP_SERVER_AUTH=AUTHKIT` and corresponding variables
- **Token caching**: Persistent OAuth token storage prevents repeated logins
- **Dynamic client registration**: Automatic OAuth client setup with supported providers

The refactored architecture moves from scattered verification patterns to a coherent provider system. Authentication providers now handle both token verification and metadata endpoints, making the server's capabilities discoverable to MCP clients following the latest protocol updates.

Closes #1273


tobinsouth:
This is incredibly cool and probably the best developer auth experience in MCP.

Will run some tests on this across clients later today.

taylorwilsdon:
Hell yeah I’ve got it spun up in https://github.com/taylorwilsdon/google_workspace_mcp/pull/137 and it is doing what I hoped with mcp inspector. Here’s to hoping broad client support follows. Big step in the right direction, kudos

PR #1340, opened 2025-08-02
feat: add component_filter for dynamic per-request filtering based on tags and request context

eyalr99:
## Add component_filter parameter for dynamic per-request component filtering

I built this because I needed per-user component filtering in my project and figured others might run into the same problem.
This PR adds a `component_filter` parameter to the FastMCP constructor that enables advanced, per-request filtering of components based on their tags and request context.

### What it enables

- **Per-request filtering**: Access current HTTP request via `get_http_request()` to filter components based on user roles, authentication headers, tenant information, or any request data
- **Complex tag logic**: Unlike with `include_tags` or `exclude_tags` which are per server instance, implement custom logic like requiring ALL tags to be included for a component to be enabled, conditional requirements, or version-based filtering  
- **Dynamic policies**: Filter components based on user permissions, feature flags, tenant isolation, or business logic
- **Async functions**: For complex operations like database permission lookups or external API calls to authorization services

### Example use cases

```python
# Role-based filtering using request headers
def permission_filter(tags: set[str]) -> bool:
    request = get_http_request()
    user_role = request.headers.get("X-User-Role", "guest")
    
    if user_role == "admin":
        return True
    elif user_role == "user":
        return "public" in tags
    return False

mcp = FastMCP(component_filter=permission_filter)
```

The filter works alongside existing `include_tags`, `exclude_tags` and `enabled` for layered filtering.

**Note**: I looked at solutions like permit.io but they require external service integrations. This provides a lightweight, built-in approach for component-level dynamic access control.

### DIFF
--- a/docs/servers/server.mdx
+++ b/docs/servers/server.mdx
@@ -64,6 +64,12 @@ The `FastMCP` constructor accepts several arguments:
   Hide components with any matching tag
 </ParamField>
 
+<ParamField body="component_filter" type="Callable[[set[str]], bool | Awaitable[bool]] | None">
+  <VersionBadge version="2.12.0" />
+  
+  A custom function to determine component availability based on tags. Receives a component's tags and returns `True` to enable or `False` to disable the component. Supports both sync and async functions. This filter is applied in addition to `include_tags` and `exclude_tags`, enabling dynamic per-request filtering based on user permissions, request context, or complex tag logic
+</ParamField>
+
 <ParamField body="on_duplicate_tools" type='Literal["error", "warn", "replace"]' default="error">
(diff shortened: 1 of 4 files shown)


jlowin:
Thanks for the contribution @eyalr99! Now that we support [Middleware](https://gofastmcp.com/servers/middleware#mcp-middleware), this feature would be more versatile (and easier to implement) as a new middleware that users can add to their servers. Perhaps the middleware could be parameterized by a function so users don't have to subclass it, just something like

```
mcp = FastMCP(middleware=(FilterMiddleware(filter_fn)])
```

I think we have some examples of *extremely* simple versions of this in the docs

PR #1343, opened 2025-08-02
Fix: memory leak in Context notification background tasks

chi2liu:
## 📋 Summary

  This PR fixes a memory leak in the Context notification system where `asyncio.create_task()` calls were creating
  fire-and-forget tasks that accumulated in the event loop without proper cleanup, leading to memory growth in
  long-running FastMCP servers.

  ## 🚨 Problem Description

  ### Memory Leak Pattern
  The current implementation uses a "fire-and-forget" pattern for notification flushing:

  ```python
  # Original problematic code
  def _try_flush_notifications(self) -> None:
      # ...
      # Schedule flush as a task (fire-and-forget)
      asyncio.create_task(self._flush_notifications())
```

  While this design provides good performance by not blocking the caller, it has a critical flaw: completed tasks are
  never cleaned up from the event loop, causing memory accumulation in production environments.

  ### Impact Analysis

  - Memory Growth: Each notification creates a task object that persists in memory even after completion
  - Long-running Servers: Production FastMCP servers with frequent tool/resource changes experience significant memory growth
  - High-frequency Scenarios: Applications with dynamic tool registration/deregistration are particularly affected
  - No Current Mitigation: The fire-and-forget pattern provides no mechanism for cleanup

  ### Reproduction Context

  This issue occurs during normal FastMCP operations:
  - Tool enable/disable operations (tool.enable(), tool.disable())
  - Dynamic resource changes
  - Prompt list modifications
  - Any operation that triggers _queue_*_list_changed() methods

  ## ✅ Solution

  ### Design Philosophy

  The fix maintains the original fire-and-forget performance benefits while adding proper resource lifecycle management:

  1. Preserve Async Performance: Tasks remain non-blocking and don't affect caller performance
  2. Add Memory Management: Track tasks using WeakSet for automatic cleanup
  3. Lifecycle Alignment: Clean up remaining tasks when Context exits
  4. Zero Breaking Changes: Maintains complete API compatibility


strawgate:
fire and forget tasks may never run due to immediate garbage collection but there would have to be something else wrong for them to hang out in memory forever

Is this a problem you've experienced? Can you open an issue with a minimal recreation of the issue?

chi2liu:
#1349

chi2liu:
> fire and forget tasks may never run due to immediate garbage collection but there would have to be something else wrong for them to hang out in memory forever
> 
> Is this a problem you've experienced? Can you open an issue with a minimal recreation of the issue?

@strawgate While your theoretical understanding is correct (tasks should be GC'd), practical testing shows measurable retention.
  The fix eliminates this without any downside, making it worthwhile for production FastMCP deployments. I created a benchmark and discovered this memory leak. For details, please refer to the associated issue. #1349

PR #1353, opened 2025-08-03
Move Keep Alive to Proxy Client

strawgate:
A breaking change appears!!!!

Tests are failing due to a bug that I will file and link shortly

### DIFF
--- a/docs/clients/transports.mdx
+++ b/docs/clients/transports.mdx
@@ -140,7 +140,7 @@ client = Client(transport)
 
 ### Session Persistence
 
-STDIO transports maintain sessions across multiple client contexts by default (`keep_alive=True`). This improves performance by reusing the same subprocess for multiple connections, but can be controlled when you need isolation.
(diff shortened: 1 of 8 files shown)


strawgate:
Blocked by https://github.com/jlowin/fastmcp/issues/1354

Could also just unwinding the change to flatten the mounted mcp servers but that has its own issues

strawgate:
I think part of the problem might be that we make a client per session so that we don't share initialized sessions?

strawgate:
Ill abandon this PR early this week and make a PR with a couple tests that I think reflect the issue:

```

    async def test_keep_alive_true_exit_scope_kills_server(self, stdio_script):
        pid: int | None = None

        async def test_server():
            transport = PythonStdioTransport(script_path=stdio_script, keep_alive=True)
            client = Client(transport=transport)
            assert client.transport.keep_alive is True
            async with client:
                result1 = await client.call_tool("pid")
                nonlocal pid
                pid = result1.data

        await test_server()

        gc.collect()

        await asyncio.sleep(2)

        with pytest.raises(psutil.NoSuchProcess):
            psutil.Process(pid)
```

Issue #1212, opened 2025-07-21
FastMCP Server should allow a configurable ping


DinoChiesa:
### Enhancement Description

With either the now-deprecated SSE or with HTTP Streaming, a [ping](https://modelcontextprotocol.io/specification/2025-06-18/basic/utilities/ping) from server to client is helpful in keeping the connection alive and healthy. 

The underlying MCP library includes [a send_ping method](https://github.com/modelcontextprotocol/python-sdk/blob/main/src/mcp/server/session.py#L287) for server-to-client ping. 

Some other frameworks, like [punkpeye/fastmcp](https://github.com/punkpeye/fastmcp?tab=readme-ov-file#configurable-ping-behavior) offer an easy way to configure the server's ping behavior with an interval. 

I have a case where I'd like to use THIS library and get configurable ping, easily.  

Ideally I will be able to use something like this to tell it to do the right thing:  
```py
    mcp.run(
        transport="http", host="0.0.0.0", port=port, path="/mcp", stateless_http=True, ping=5000
    )
```

This solves the problem in which a client drops a connection unnecessarily. 

The alternative would be for the client to behave well and ping the server and maintain the session. Not all clients can do that. This makes the session more resilient.

### Proposed Implementation

```Python

```

DinoChiesa:
Whoo! thanks

PR #1358, opened 2025-08-03
Feature/server side configurable ping

muhammadkhalid-03:
I added an optional configurable ping feature (issue #1212 )  that the server can send to the client. The feature is implemented through middleware that creates a background task for each client session and uses the mcp built-in 'send_ping()' method.

Note: At least one tool must be defined for the server to establish connections

EDIT: I'm not sure why the [Run tests: Python 3.10 on windows-latest] test is failing

### DIFF
--- a/src/fastmcp/server/server.py
+++ b/src/fastmcp/server/server.py
@@ -397,7 +397,7 @@ class FastMCP(Generic[LifespanResultT]):
                     # uses the built-in send_ping method from the mcp library
                     await session.send_ping()
                     logger.debug(
-                        f"Sent ping to client (interval: {self.ping_config}ms)"
+                        f"Sent ping to client (interval: {self.ping_config.interval_ms}ms)"
                     )
                 except Exception as e:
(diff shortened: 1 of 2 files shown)


jlowin:
Sorry for taking so long to review - this is a good enhancement but I don't think this is an effective implementation. It feels simultaneously overcomplicated and underinvested. 

- `PingConfig` seems like it would idiomatically be typed as `int | None` with None disabling the ping
- overloading `_apply_middleware()` to start these loops feels immediately like the wrong place, especially because I imagine the ping loop should only be started once and this method is called on every single interaction (including nested ones)
- storing ping tasks on the server also feels wrong and won't survive horizontal scaling, especially when ping tasks are session-scoped, not server scoped

In all this feels like an opportunity to either use middleware (though I'm not sure if it exactly supports this pattern) to attach a ping to the per-session context object OR use a low-level `lifecycle` object (which I think is intended for this purpose) - neither should require manipulating the server logic itself in this way.

Issue #1359, opened 2025-08-03
Registering instance methods as tool function requires `self`


richardkmichael:
### Description

Is there a regression in v2.11.0?

My test suite broke, reporting `self` as a required parameter.  There are no breaking changes related to this in v2.11.0.  I think I have a repro, but of course I could be wrong, even in the usage in the original code working in v2.10.6.

The repro is more or less straight from the FastMCP examples, but has an Annotation on the instance method (my usage).

### Example Code

```Python
"""
FastMCP 2.11.0 Bug Reproduction

Run with:
  uv run --with fastmcp==2.10.6 python type_annotation_test.py  # Works
  uv run --with fastmcp==2.11.0 python type_annotation_test.py  # Fails
"""

import asyncio
import sys
from typing import Annotated

try:
    from fastmcp import FastMCP
    from fastmcp.client import Client
    import fastmcp
except ImportError:
    print("ERROR: FastMCP not installed")
    sys.exit(1)


class MyTool:
    # Works without Annotation
    # def foo(self, name) -> str:

    # Fails
    def foo(self, name: Annotated[str, "A parameter"]) -> str:
        """This method breaks in FastMCP 2.11.0"""
        return f"Hello {name}"


async def main():
    print(f"FastMCP version: {fastmcp.__version__}")

    mcp = FastMCP("test")
    obj = MyTool()
    mcp.tool(obj.foo)

    try:
        async with Client(mcp) as client:
            result = await client.call_tool("foo", {"name": "world"})
            print(f"✅ SUCCESS: {result.content[0].text}")
    except Exception as e:
        print(f"❌ FAILED: {e}")


if __name__ == "__main__":
    asyncio.run(main())



$ uv run --with fastmcp==2.10.6 python type_annotation_test.py
FastMCP version: 2.10.6
Testing instance method with Annotated parameter...
✅ SUCCESS: Hello world

$ uv run --with fastmcp==2.11.0 python type_annotation_test.py
FastMCP version: 2.11.0
Testing instance method with Annotated parameter...
❌ FAILED: Input validation error: 'self' is a required property
```

### Version Information

```Text
FastMCP version:                                                                                                      2.11.0
MCP version:                                                                                                          1.12.3
Python version:                                                                                                       3.13.2
Platform:                                                                                macOS-14.7.6-arm64-arm-64bit-Mach-O
[...]

PR #1360, opened 2025-08-04
Fix method-bound FunctionTool schemas

strawgate:
Fixes: https://github.com/jlowin/fastmcp/issues/1359


src/fastmcp/utilities/types.py
```diff
@@ -339,7 +344,7 @@ def to_resource_content(
             raise ValueError("No resource data available")
 
         mime = mime_type or self._mime_type
-        UriType = Annotated[AnyUrl, UrlConstraints(host_required=False)]
```
strawgate: Oops

jlowin:
Thank you!

Issue #1286, opened 2025-07-29
Support async execution in client_factory for FastMCPProxy class


bianning:
### Enhancement Description

The current implementation of FastMCPProxy allows passing a client_factory to create Client instances. However, this factory currently only supports synchronous execution. Since the proxy needs to fetch information asynchronously (e.g., from remote servers), we should enable async support in client_factory.

### Use Case

When fetching data from a remote server to create a client, async operations would improve performance by avoiding blocking the event loop.

### Proposed Implementation

```Python
1.Modify the client_factory parameter to optionally accept an async version (e.g., Union[Callable[[], Client], AsyncCallable[[], Client]]).
2.Maintain backward compatibility by allowing the existing synchronous factory.
3.Add clear documentation about async usage patterns.
```

jlowin:
That feels a little heavyweight, the factory is really just intended to permit flexible reuse of clients (e.g. create a new client or load a cached one) and is usually a lambda - if you are loading clients from a remote config, why not just recreate the entire proxy? Can you say more about a specific use case you have in mind?

bianning:
Makes sense to keep client_factory lightweight. That said, our use case involves some async requirements — here’s the context:

We're using FastMCPProxy as a gateway layer in a larger internal system, where it needs to dynamically initialize the appropriate client and server for downstream services. There are a few reasons why async support in the client_factory is essential:

We only receive a service ID during communication, not the actual address or OpenAPI schema.

To initialize ProxyClient(transport=FastMCPTransport(mcp_server)), we need to configure mcp_server with its base URL and OpenAPI spec.

This metadata has to be fetched dynamically at runtime from a central service registry — and that’s inherently an async operation.

The data isn’t pre-cached and may change often (especially across environments), so we need to fetch it live per request.

Because of all this, we can't construct the client synchronously without blocking the event loop.

jlowin:
I see -- I think my general recommendation would be to create the entire server in an async function but I do take your point, and we can easily make client factory be async or sync

bianning:
I've implemented async client_factory support and opened a PR for you to review.
Feel free to take a look whenever it’s convenient — totally open to suggestions if there’s a cleaner or better way to approach it.

jlowin:
Closed in #1375

PR #1375, opened 2025-08-05
Add async support to client_factory in FastMCPProxy  (#1286)

bianning:
This adds async support for client_factory in FastMCPProxy.

Related Issue https://github.com/jlowin/fastmcp/issues/1286

The change:

Accepts either a sync or async factory
Keeps existing sync behavior unchanged

### DIFF
--- a/tests/server/proxy/test_proxy_server.py
+++ b/tests/server/proxy/test_proxy_server.py
@@ -1,3 +1,4 @@
+import inspect
 import json
 from typing import Any, cast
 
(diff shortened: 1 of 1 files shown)


src/fastmcp/server/proxy.py
```diff
@@ -48,11 +49,27 @@
...
-class ProxyToolManager(ToolManager):
+
+class ProxyManagerMixin:
+    """A mixin for proxy managers to provide a unified client retrieval method."""
+
+    client_factory: ClientFactoryT
+
+    async def _get_client(self) -> Client:
+        """Gets a client instance by calling the sync or async factory."""
+        if inspect.iscoroutinefunction(self.client_factory):
```
jlowin: Minor nit - the pattern we prefer elsewhere is I think 
```python
client = self.client_factory()
if inspect.isawaitable(client):
    client = await client
return client
```
This way the factory can be any function that returns an awaitable (e.g. a sync-decorated coroutine, or a class).

tests/server/proxy/test_proxy_server.py
```diff
@@ -109,8 +109,9 @@ def test_as_proxy_with_url():
     """FastMCP.as_proxy should accept a URL without connecting."""
     proxy = FastMCP.as_proxy("http://example.com/mcp/")
     assert isinstance(proxy, FastMCPProxy)
-    assert isinstance(proxy.client_factory().transport, StreamableHttpTransport)
-    assert proxy.client_factory().transport.url == "http://example.com/mcp/"  # type: ignore[attr-defined]
+    client = cast(Client, proxy.client_factory())
+    assert isinstance(client.transport, StreamableHttpTransport)
+    assert client.transport.url == "http://example.com/mcp/"  # type: ignore[attr-defined]
```
jlowin, line 116: Could you add a simple test to ensure that async factories are executed?

src/fastmcp/server/proxy.py
bianning: Updated _get_client to use the isawaitable pattern as suggested.

tests/server/proxy/test_proxy_server.py
bianning: Added a test for async factory usage as well.

PR #1399, opened 2025-08-07
fixes #1398: Add JWT claims to AccessToken

### LINKED ISSUE #1398: Allow the MCP server to access JWT "claims"
### Enhancement Description

The new (as of version 2.11.0) **Remote OAuth Authentication** functionality for servers allows verifying JWT tokens with an external identity provider such as Keycloak. In many use cases, the resource server then needs to access information provided in the JWT `claims` to authorize different actions. However, the current implementation of `JWTVerifier.load_access_token` discards most JWT `claims` after verifying the token and returns only the following:

https://github.com/jlowin/fastmcp/blob/77bdc9de200a45149c09cbfabfe82de0744d53a7/src/fastmcp/server/auth/providers/jwt.py#L446-L451

As a result, it is impossible to access information such as the `preferred_username` claim without decoding the token from scratch.

### Use Case

As a developer using the new Remote OAuth Authentication feature (introduced in version 2.11.0) for my MCP server, I need to access the authenticated user's username—specifically, the value provided in the `preferred_username` or `sub` claims of the JWT bearer token issued by my identity provider (e.g., Keycloak). This is necessary so that, after the server verifies the JWT, I can query my backend for resources accessible to the current user.

More broadly, I also want to access all other claims provided in the JWT so I can perform associated functions.

### Proposed Implementation

```Python
I suggest sub-classing the `AccessToken` class to also store a dictionary of the claims.


class AccessTokenWithClaims(AccessToken):
    """Extended AccessToken to include claims."""

    claims: dict = {}


Then, we can utilize the `get_access_token()` method to obtain the `AccessTokenWithClaims` and read the claims when needed within the MCP server.
```

panargirakis:
resolves issue #1398

This pull request refactors the authentication token handling in the FastMCP server to use a new `AccessTokenWithClaims` model that extends the existing `AccessToken` type by including all JWT claims. This change ensures that downstream components and tests have access to the full set of token claims, improving flexibility and correctness in authentication workflows.

### Model and Type Refactoring

* Introduced `AccessTokenWithClaims`, a subclass of `AccessToken` that adds a `claims` dictionary to hold all JWT claims.
* Updated all authentication provider methods (`verify_token`, `load_access_token`) to return `AccessTokenWithClaims` instead of `AccessToken`, ensuring claims are always available. [[1]](diffhunk://#diff-0c4995439b98a89be0d3e44d65ed8979f2726fec15fa69ce0965491f9753dec1L110-R112) [[2]](diffhunk://#diff-0c4995439b98a89be0d3e44d65ed8979f2726fec15fa69ce0965491f9753dec1L147-R149) [[3]](diffhunk://#diff-f3f1e259d65b79f6ff14e1629fd41ec26e16076bde366276aab6bd3c86182c21L359-R367) [[4]](diffhunk://#diff-f3f1e259d65b79f6ff14e1629fd41ec26e16076bde366276aab6bd3c86182c21L446-R451) [[5]](diffhunk://#diff-f3f1e259d65b79f6ff14e1629fd41ec26e16076bde366276aab6bd3c86182c21L460-R461) [[6]](diffhunk://#diff-f3f1e259d65b79f6ff14e1629fd41ec26e16076bde366276aab6bd3c86182c21L471-R472) [[7]](diffhunk://#diff-f3f1e259d65b79f6ff14e1629fd41ec26e16076bde366276aab6bd3c86182c21L510-R511) [[8]](diffhunk://#diff-f3f1e259d65b79f6ff14e1629fd41ec26e16076bde366276aab6bd3c86182c21L533-R539) [[9]](diffhunk://#diff-b2e3c7de57df67d59d89a8fc692caa7be58d215b57a6a6483b5421377ab1a681L3-R17)

### Provider Implementation Updates

* Modified JWT and static token providers to instantiate and return `AccessTokenWithClaims`, populating the `claims` field with the raw JWT claims or token data. [[1]](diffhunk://#diff-f3f1e259d65b79f6ff14e1629fd41ec26e16076bde366276aab6bd3c86182c21L446-R451) [[2]](diffhunk://#diff-f3f1e259d65b79f6ff14e1629fd41ec26e16076bde366276aab6bd3c86182c21L533-R539)

### Test Suite Enhancements

* Updated all related tests to use and validate `AccessTokenWithClaims`, including checks for the presence and correctness of claims in returned tokens. [[1]](diffhunk://#diff-4e41c9e2bb164dac05f6fbea5133d9eca9d4d589dd77ab202df11e188f649adeR144-R162) [[2]](diffhunk://#diff-b2e3c7de57df67d59d89a8fc692caa7be58d215b57a6a6483b5421377ab1a681L41-R41) [[3]](diffhunk://#diff-b2e3c7de57df67d59d89a8fc692caa7be58d215b57a6a6483b5421377ab1a681L268-R268) [[4]](diffhunk://#diff-b2e3c7de57df67d59d89a8fc692caa7be58d215b57a6a6483b5421377ab1a681L288-R288) [[5]](diffhunk://#diff-8e5cf019cc806c3d7613b461484786e1d7f307760651ef26b4a7fb4d8f059443R39) [[6]](diffhunk://#diff-8e5cf019cc806c3d7613b461484786e1d7f307760651ef26b4a7fb4d8f059443R48)

These changes collectively improve the reliability and extensibility of token-based authentication by ensuring all relevant claims are available throughout the FastMCP authentication flow.


src/fastmcp/server/auth/models.py
```diff
@@ -0,0 +1,9 @@
+from typing import Any
+
+from mcp.server.auth.provider import AccessToken
+
+
+class AccessTokenWithClaims(AccessToken):
```
jlowin: Please move this class into src/server/auth/auth.py rather than a new file. 

I'm also ok (even encouraging) if you want to name this class `AccessToken` since it will be the canonical access token class for the FastMCP framework, and subclass `from mcp.server.auth.provider import AccessToken as _SDKAccessToken` or similar to avoid name conflicts

jlowin (changes requested):
Thanks for this enhancement, it's great! I left a comment about name/location for the main class. I think you may also need to update dependencies.py, which currently imports the SDK's get_access_token() function and puts it in the `__all__` list.  The only real reason is because it's typed as returning an SDK AccessToken, not a FastMCP AccessToken, so for convenience, wrapping it will let FastMCP users see the `claims` attribute in their IDE hints.

```python
from mcp.server.auth.middleware.auth_context import get_access_token as _sdk_get_access_token

def get_access_token() → AccessToken:
    return _sdk_get_access_token()
```

src/fastmcp/server/auth/models.py
panargirakis: Thanks the comments are addressed, including moving, renaming, and updating `dependencies.py`! I had to add some more logic to `get_access_token()` to satisfy the linter and ensure type safety.

jlowin: Ah sorry I didn't anticipate that - there may be ways to `cast()` or otherwise convince it, we can look at that later. this is great, thank you

jlowin (approved):

Issue #1390, opened 2025-08-07
in list_resource_templates results, mcp server name is just recorded in ResourceTemplate.uriTemplate then danger in mis-operation at next action.


itaru2622:
### Description

this issue is observed MCP client, in the both of below two cases.

case1) MCP servers <=> client(fastmcp | inspector)
case2) MCP servers <=> FastMCP Proxy <=>  client(fastmcp | inspector)

assume client or FastMCP Proxy is configured as below:

```yaml
mcpServers:

   mcpServer1: { ... }

   mcpServer2: { ... }

```

in both cases,  MCP client recivesd list of mcp.types.ResourceTemplate.
it includes name of mcpServer1 or 2 in ResourceTemplate.uriTemplate but no other field.

as the result, human cannot distinguish which item comes from which mcpServer, like attached inspector UI window.
then, it causes mis-operation in next operation(read_resource) and send request to the wrong server.

comparing the results of list_tools and list_prompts, the name of mcp server is shown as prefix, 
and detectable by human like attached inspector UI window.
below screenshots were taken in the above case2 senario.

<img width="755" height="317" alt="Image" src="https://github.com/user-attachments/assets/def2318e-542d-475f-9496-0ac72f2c2f85" />

<img width="575" height="326" alt="Image" src="https://github.com/user-attachments/assets/9652fde6-2ec0-4bb8-8666-01f25a4d93f6" />

### Example Code

```yaml
# exact configuration:


mcpServers:
  github-full:
    transport: stdio
    command: docker
    args:
    - run
    - -i
    - --rm
    - -e
    - GITHUB_PERSONAL_ACCESS_TOKEN
    - ghcr.io/github/github-mcp-server
    - stdio
    env:
      GITHUB_PERSONAL_ACCESS_TOKEN: anything_is_ok_just_for_list_xxx
  github-readonly:
    transport: stdio
    command: docker
    args:
    - run
    - -i
    - --rm
    - -e
    - GITHUB_PERSONAL_ACCESS_TOKEN
    - ghcr.io/github/github-mcp-server
    - stdio
    - --read-only
    env:
      GITHUB_PERSONAL_ACCESS_TOKEN: anything_is_ok_just_for_list_xxx

```

```python

# snippet code:
from fastmcp import FastMCP, Client
client = Client(conf)
async with client:
   res1 = await client.list_resource_templates()
   res2 = await client.list_prompts()

print(res1,res2)
```

### Version Information

```Text
[...]

PR #1407, opened 2025-08-08
fix: Fix missing server name prefixes in resource/template names (#1390)

chi2liu:
#1390 In multi-server configurations, resource and resource template names were not getting server prefixes, making it impossible for users to distinguish which server owns which resource. Only the URI/uriTemplate fields had prefixes.

This change adds server name prefixes to the 'name' field of both resources and resource templates when they are mounted, matching the existing behavior of tools and prompts.

Now all MCP object types (tools, prompts, resources, resource templates) consistently show server prefixes in their names when mounted, preventing user confusion and mis-operations in multi-server setups.

### DIFF
--- a/src/fastmcp/resources/resource_manager.py
+++ b/src/fastmcp/resources/resource_manager.py
@@ -103,6 +103,10 @@ class ResourceManager:
                         )
                         # Create a copy of the resource with the prefixed key
                         prefixed_resource = resource.with_key(prefixed_uri)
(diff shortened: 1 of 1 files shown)


itaru2622:
@chi2liu your fix is case1 of #1390 but not covered case2.

and I fixed all cases in #1411

jlowin:
Thank you! Superseded by #1423 which slightly modifies this approach.

Issue #1372, opened 2025-08-05
Nested $refs pruned from $defs in Tools from OpenAPI spec with v2.11.1 experimental


mariaTestCode:
### Description

I'm seeing nested $refs do not appear in $defs when creating server via 
```
FastMCP.from_openapi(
 openapi_spec=openapi_spec,
    client=client,
    route_maps=ROUTE_MAPPINGS,
)
```
with FASTMCP_EXPERIMENTAL_ENABLE_NEW_OPENAPI_PARSER=true on 2.11.1

Example input:
```
openapi: 3.0.3
info:
  contact:
    name: Ads API Team
    url: http://go/ads-api
    email: selfserve-infra-api@spotify.com
  description: The Spotify Ads API.
  title: Ads API
  version: 3.0.0
servers:
  - url: https://api-partner.spotify.com/ads-sandbox/v3
    description: Sandbox server.
paths:
  /users:
    post:
      summary: Create a user
      requestBody:
        required: true
        content:
          application/json:
            schema:
              $ref: '#/components/schemas/User'
      responses:
        '201':
          description: User created

components:
  schemas:
    User:
      type: object
      properties:
        id:
          type: string
        profile:
          $ref: '#/components/schemas/Profile'
      required:
        - id
        - profile

    Profile:
      type: object
      properties:
        name:
          type: string
        address:
          $ref: '#/components/schemas/Address'
      required:
        - name
        - address

    Address:
      type: object
      properties:
        street:
          type: string
        city:
          type: string
        zipcode:
          type: string
      required:
        - street
        - city
        - zipcode
```
Resulting tool response:
```
{
  "jsonrpc": "2.0",
  "id": 1,
  "result": {
    "tools": [
      {
        "name": "Create_a_user",
        "description": "Create a user\n\n\n**Responses:**\n\n- **201** (Success): User created",
        "inputSchema": {
          "type": "object",
          "properties": {
            "id": {
              "type": "string"
            },
            "profile": {
              "$ref": "#/$defs/Profile"
            }
          },
          "required": [
            "id",
            "profile"
          ],
          "$defs": {
            "Profile": {
              "required": [
                "name",
                "address"
              ],
              "type": "object",
              "properties": {
                "name": {
                  "type": "string"
                },
                "address": {
                  "$ref": "#/$defs/Address"
                }
              }
            }
          }
        },
        "_meta": {
          "_fastmcp": {
            "tags": []
          }
        }
      }
    ]
  }
}
```

[...]

PR #1386, opened 2025-08-06
fix(openapi): resolve transitive  dependencies in schema combination

displague:
(🤖 Generated in GitHub CodeSpaces with Copilot (Claude Sonnet 4))

This fix addresses issue #1372 where nested schema references were incorrectly pruned from the  section during schema combination.

Problem:
- When combining OpenAPI schemas with transitive dependencies (e.g., User → Profile → Address)
- Only direct references were preserved in
- Transitive dependencies like Address were incorrectly removed
- This affected both stable and experimental OpenAPI parsers

Solution:
- Implemented transitive closure algorithm in _combine_schemas functions
- Recursively searches through  to find all transitive dependencies
- Continues iteration until no new references are discovered
- Preserves all transitively referenced schemas while still removing truly unused ones

Changes:
- src/fastmcp/utilities/openapi.py: Enhanced _combine_schemas with transitive dependency resolution
- src/fastmcp/experimental/utilities/openapi/schemas.py: Enhanced _combine_schemas_and_map_params with same fix
- tests/experimental/utilities/openapi/test_schemas.py: Added comprehensive test coverage for:
  - Basic transitive dependencies (User → Profile → Address)
  - Deep nesting (4+ levels of references)
  - Mixed with parameter mapping functionality
- tests/server/openapi/test_optional_parameters.py: Added test for stable parser

Test coverage includes:
- Exact reproduction of issue #1372 scenario
- Deep nesting scenarios (4+ levels)
- Circular reference handling
- Unused schema removal validation
- Both _combine_schemas and _combine_schemas_and_map_params functions

The fix ensures MCP tool schemas generated from OpenAPI specs correctly include all necessary schema definitions, preventing validation errors when tools are invoked with nested object structures.

Fixes #1372

PR #1408, opened 2025-08-08
fix: Fix nested schema references missing from $defs in OpenAPI parser (#1372)

chi2liu:
#1372 The experimental OpenAPI parser was only collecting directly referenced schemas but missing transitively referenced ones (e.g., User->Profile->Address would include Profile but not Address).

This fix recursively collects all referenced schemas by:
1. Finding direct references in the main schema
2. Iteratively checking each found reference for nested references
3. Including all transitively referenced schemas in the final $defs

This ensures complete schema definitions are available for validation and prevents 'PointerToNowhere' errors at runtime.

### DIFF
--- a/src/fastmcp/experimental/utilities/openapi/schemas.py
+++ b/src/fastmcp/experimental/utilities/openapi/schemas.py
@@ -320,7 +320,7 @@ def _combine_schemas_and_map_params(
     if result.get("additionalProperties") is False:
         result.pop("additionalProperties")
 
-    # Remove unused definitions (lightweight approach - just check direct $ref usage)
+    # Remove unused definitions - recursively find all referenced schemas
     if "$defs" in result:
         used_refs = set()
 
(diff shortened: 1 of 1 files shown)


TJReinert:
Hey! Thanks so much for working on this - I really appreciate the effort. I’ve checked out your branch and gave it a try, but I’m still seeing the same behavior. I’ve attached a minimal reproducible OpenAPI document that triggers the problem for me.

The symptoms appear unchanged, but I’m going to keep digging into the code to see if I can help identify what’s going on.
[parser-error.json](https://github.com/user-attachments/files/21694894/parser-error.json)


Edit: So I think this same bug persists on response types. Your changes fix inputs.

jlowin:
@TJReinert are you using the experimental parser? The test file you shared (`tests/utilities/openapi/test_openapi_advanced.py`) uses the legacy parser.

jlowin:
Never mind @TJReinert -- I think I see where the issue is, though it does highlight that this PR has no tests -- I am going to open a new PR that combines this PR and some of the approach in #1386 with a full test suite. Thanks for the contribution @chi2liu!

chi2liu:
@jlowin @TJReinert Thanks

Issue #972, opened 2025-06-27
OAuth works with MCP Inspector but not with Claude Integrations or FastMCP Client


shahrear33:
### Description

I'm trying to implement OAuth for my MCP server by following [this documentation](https://gofastmcp.com/servers/auth/bearer).

It's a remote MCP instance. Example URL: `http://localhost:8000/mcp/sse`.

I can connect successfully using [**MCP Inspector**](https://github.com/modelcontextprotocol/inspector), but when I try to connect using **Claude Integration** or the **FastMCP Client**, it fails with the following error:
```
{
  "error": "invalid_client",
  "error_description": "Client authentication failed (e.g., unknown client, no client authentication included, or unsupported authentication method). The requested OAuth 2.0 Client does not exist."
}
```
I've created a repository here: [https://github.com/shahrear33/mcp-oauth](https://github.com/shahrear33/mcp-oauth).

I'm using **Clerk** as the authentication provider with Dynamic Client Registration (DCR) Enabled

### Example Code

```Python
server.py

"""
    MCP OAuth Server Example

This module demonstrates a FastMCP server with Bearer token authentication mounted on FastAPI,
including OAuth 2.1 endpoints for dynamic client registration and metadata discovery.
"""

import time
from fastapi.responses import JSONResponse
import os

from fastapi import FastAPI, HTTPException, Request
from fastmcp import FastMCP
from fastmcp.server.auth import BearerAuthProvider
from loguru import logger
from starlette.responses import PlainTextResponse
from fastapi.middleware.cors import CORSMiddleware
from datetime import datetime
import dotenv

dotenv.load_dotenv()

# Configuration - use environment variables or defaults
ISSUER = os.getenv("CLERK_ISSUER", "https://dev.example.com")
AUDIENCE = os.getenv("CLERK_AUDIENCE", "my-mcp-server")
CLIENT_SECRET = os.getenv("CLERK_CLIENT_SECRET")
BASE_URL = os.getenv("BASE_URL", "http://127.0.0.1:8000")

logger.info(f"ISSUER: {ISSUER}")
logger.info(f"AUDIENCE: {AUDIENCE}")
logger.info(f"CLIENT_SECRET: {CLIENT_SECRET}")
logger.info(f"BASE_URL: {BASE_URL}")


def mcp_factory() -> FastMCP:
    """Create and configure the FastMCP server."""
    logger.info("Initializing MCP server")

    auth = BearerAuthProvider(
[...]

Issue #914, opened 2025-06-22
oauth-protected-resource => more docs on testing oauth via mcp inspector


nick-youngblut:
### Enhancement Description

Example:

```python
from fastmcp import FastMCP
from fastmcp.server.auth import BearerAuthProvider

auth = BearerAuthProvider(
    jwks_uri="https://www.googleapis.com/oauth2/v3/certs",
    issuer="https://accounts.google.com",
    audience="my_unique_creds.apps.googleusercontent.com",
)

mcp = FastMCP(name="My MCP Server", auth=auth)

mcp.run(
    transport="sse",
    host="0.0.0.0",
    port=8080,
    path="/mcp",
)
```

* Run the server via `python example/server_oauth.py`
* Run the mcp inspector via `npx @modelcontextprotocol/inspector`
* In the Inspector app:
  * Transport Type = `SSE`
  * URL = `http://0.0.0.0:8080/mcp`
  * Click `Connect`

The inspector logs:

```
SSE transport: url=http://0.0.0.0:8080/mcp, headers={"Accept":"text/event-stream"}
Received 401 Unauthorized from MCP server. Authentication failure.
```

The mcp server logs:

```
[06/22/25 11:59:31] INFO     Starting MCP server 'My MCP Server' with transport 'sse' on http://0.0.0.0:8080/mcp       server.py:1030
INFO:     Started server process [82256]
INFO:     Waiting for application startup.
INFO:     Application startup complete.
INFO:     Uvicorn running on http://0.0.0.0:8080 (Press CTRL+C to quit)
INFO:     127.0.0.1:52437 - "GET /mcp HTTP/1.1" 401 Unauthorized
INFO:     127.0.0.1:52438 - "OPTIONS /.well-known/oauth-protected-resource HTTP/1.1" 404 Not Found
INFO:     127.0.0.1:52438 - "GET /.well-known/oauth-protected-resource HTTP/1.1" 404 Not Found
```

If I change `mcp = FastMCP(name="My MCP Server", auth=auth)` to `mcp = FastMCP(name="My MCP Server")`, then the inspector connects to the server.

It would help to have more info in the docs on how to test authentication via the mcp inspector.

### Proposed Implementation

```Python

```

nick-youngblut:
In particular, it appears that the mcp inspector is looking for endpoints that do not exist: `/.well-known/oauth-protected-resource`

Explaining this in the docs would be helpful

nick-youngblut:
[...]

PR #1409, opened 2025-08-08
Automatic Metadata Endpoint Configuration

toharm:
Related issues:

https://github.com/jlowin/fastmcp/issues/972
https://github.com/jlowin/fastmcp/issues/914

Rather than mounting the FastMCP server within a FastAPI server, FastMCP will automatically configure metadata endpoints for you. Includes testing.

- Sanitized response
- DCR if it is supported
- Persistent storage using SQLite for client/token validation
- Caching JWKs for the issuer
- Complete functionality testing, all previous tests passing too (tested endpoints, with OAuth2, without OAuth2, sanitization, bad requests, token storage and validation...)


jlowin:
This is a massive PR that goes well beyond the description - does this implement a full OAuth server in FastMCP? This is not a responsibility that's appropriate for this project (and wouldn't be in server.py regardless). Much of the functionality here is already available elsewhere in the project, and while FastMCP provides interfaces for creating such auth servers, we are not prepared to a full implementation. Thank you for contributing however!

PR #1427, opened 2025-08-10
Dereference json schema

ibash:
Claude has trouble calling tools that have use refs in their jsonschema. In particular when fastmcp is used with claude code only mcp tools with primitive paramters work, mcp tools that use pydantic types break.

This should be fixed in claude/claude code, but in the meantime we can ensure wide compatability by inlining type definitions in the json schema.

Ref: https://github.com/pydantic/pydantic/issues/889
Ref: https://github.com/pydantic/pydantic/issues/12023

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -16,6 +16,7 @@ dependencies = [
     "pyperclip>=1.9.0",
     "openapi-core>=0.19.5",
     "msgspec>=0.19.0",
+    "jsonref>=1.1.0",
 ]
 requires-python = ">=3.10"
 readme = "README.md"
--- a/src/fastmcp/tools/tool.py
+++ b/src/fastmcp/tools/tool.py
@@ -10,9 +10,11 @@ from typing import (
     Generic,
     Literal,
     TypeVar,
+    cast,
     get_type_hints,
 )
 
+import jsonref
 import mcp.types
 import pydantic_core
 from mcp.types import ContentBlock, TextContent, ToolAnnotations
@@ -388,6 +390,11 @@ class ParsedFunction:
 
         input_type_adapter = get_cached_typeadapter(fn)
         input_schema = input_type_adapter.json_schema()
+
+        # dereference definitions so tool calling in downstream agents (e.g.
+        # claude code) work. do this before compressing the schema so defs get
+        # cleaned up.
+        input_schema = cast(dict[str, Any], jsonref.replace_refs(input_schema, proxies=False))
         input_schema = compress_schema(input_schema, prune_params=prune_params)
 
         output_schema = None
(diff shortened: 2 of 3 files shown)


jlowin:
Thanks @ibash - I do not want to implement such a heavy-handed change just because of a broken client, as the server is doing the right thing here but the client is sending malformed arguments. (This is tracked in Claude Code at https://github.com/anthropics/claude-code/issues/3084). I've confirmed the issue and it is unfortunate. 

However global dereferencing will result in massive schemas, tons of deepcopy operations, and generally harm clients that *do* implement the protocol correctly. 

Have you tested that your approach works? I think leaving jsonref objects as jsonrefs and not pruning the schema will result in errors. Rather than force all tools to behave this way, you can do this yourself easily:

```python
import json

import jsonref

from fastmcp import FastMCP
from fastmcp.utilities.json_schema import compress_schema


class Person(BaseModel):
    name: str
    age: int


app = FastMCP()


@app.tool()
def my_tool(person: Person) -> str:
    """send input as a json object"""
    return f"Hello, {person.name}! You are {person.age} years old."


my_tool.parameters = compress_schema(
    json.loads(jsonref.dumps(jsonref.replace_refs(my_tool.parameters)))
)
```

However, in my tests, Claude Code failed to call the above tool correctly, so I think this is a more deeply rooted bug in how it prepares objects.

jlowin:
This may be worth opening as a formal bug with Claude Code

ibash:
Ah, it seems like #3084 is this bug. The approach in this PR does work, I'm using it right now.

I understand the problem with dereferencing the schema but the tradeoff is that every user of fastmcp will have a broken integration with claude code.

What do you think about adding this in temporarily until claude code lands a patch?

As it stands the examples in the fastmcp documentation simply don't work (both this version and v1 in the mcp sdk).

apiwat-chantawibul:
- I'm encountering similar problem with gemini-cli where its MCP client [expects json schema to always have `type` property](https://github.com/google-gemini/gemini-cli/blob/a31830a3cb16584ec0448d6a23da27947ac9e72c/packages/core/src/tools/mcp-client.ts#L576) which is incompatible with how pydantic and fastmcp generate `$ref` in json schema.
- I do think MCP server-client communication should allow `$ref` which is standard in jsonschema.
- I think it is responsibility of MCP-client to convert standard json schema received to whatever format that model actually needs.

---

That said. I found the [workaround](https://github.com/jlowin/fastmcp/pull/1427#issuecomment-3172602843) posted by @jlowin useful but needs correction / simplifications. Use this:

```python
my_tool.parameters = compress_schema(jsonref.replace_refs(my_tool.parameters, proxies = False))
```

instead of

```python
my_tool.parameters = compress_schema(
    json.loads(jsonref.dumps(jsonref.replace_refs(my_tool.parameters)))
)
```

PR #1429, opened 2025-08-10
Fix TransformedTool silent failure with missing return annotations #1369

### LINKED ISSUE #1369: Transformed Tools silently fail if original function doesn't have a type annotation
### Description

I'm not sure if it can be considered a bug or not but I was a bit puzzled when trying to transform a tool based on the search example on https://gofastmcp.com/integrations/chatgpt 
```

@mcp.tool   
def search(query: str):    
  """Searches for items in the database. Hardcoded results for testing""" 
   return {"ids": ["O1", "O2"]}                    
                                                        | 
search_tool_test = Tool.from_tool(
  search,
  name="test_search_tool",
  description=""" Search for .... """
)      
```     

that a call to the transformed tool returned None rather than the expected call results, but realised it's because the original search method as documented on the page above didn't provide a return value type annotation. Adding that in and it all works, 

```
@mcp.tool                   
def search(query: str) -> dict[str, list[str]]:    
  """Searches for items in the database."""     
  return {"ids": ["O1", "O2"]}      
```                  

but given the original method call worked without it, should transform_tool calls work without needing it, or perhaps complain instead of silently failing ?

### Version Information

```Text
FastMCP version:                                                        2.11.0
MCP version:                                                            1.12.3
Python version:                                                         3.12.3
```

### Additional Context

_No response_

chi2liu:
## Summary
- Fixes TransformedTool silently failing when transform functions lack return type annotations
- Adds validation to raise clear error when return annotation is missing
- Prevents incorrect inheritance of parent tool's output schema

## Problem
When creating a TransformedTool with a custom transform function that lacks a return type annotation, the tool would silently fail by incorrectly inheriting the parent tool's output schema. This led to unexpected behavior that was hard to debug.

Fixes #1369


jlowin:
Requiring a return annotation shouldn't be necessary - looking at the original issue it's not clear what the bug is, let's take the conversation back to the issue to better understand it.

PR #1339, opened 2025-08-02
Fix : subprocess termination in MCPConfigTransport | Fixes #1311

radheradhe01:
Fix subprocess termination in MCPConfigTransport 

Description :
Fixes subprocess termination issue where FastMCP Client failed to properly terminate stdio MCP subprocesses after calling [client.close()](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html). This caused resource leaks as subprocesses remained running even after the client was closed.

Related Issue
Closes #1311

Type of Change

<input checked="" disabled="" type="checkbox"> Bug fix (non-breaking change which fixes an issue)
<input disabled="" type="checkbox"> New feature (non-breaking change which adds functionality)
<input disabled="" type="checkbox"> Breaking change (fix or feature that would cause existing functionality to not work as expected)
<input disabled="" type="checkbox"> Documentation update

Root Cause
The [MCPConfigTransport](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) class wraps other transports but didn't implement a [close()](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) method to delegate cleanup to the underlying transport. It inherited the base class's empty [close()](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) method instead of calling the wrapped transport's proper cleanup logic.

Testing
Automated Tests
<input checked="" disabled="" type="checkbox"> All existing stdio tests pass (8/8)
<input checked="" disabled="" type="checkbox"> All existing client tests pass (72/72)
<input checked="" disabled="" type="checkbox"> No regressions introduced

Manual Verification
<input checked="" disabled="" type="checkbox"> MCPConfig Dict format subprocess termination ✅
<input checked="" disabled="" type="checkbox"> Direct StdioTransport subprocess termination ✅
<input checked="" disabled="" type="checkbox"> PythonStdioTransport subprocess termination ✅

PR #1430, opened 2025-08-10
Fix stdio subprocess cleanup issue

### LINKED ISSUE #1311: FastMCP Client doesn't properly close stdio mcp after client.close()
### Description

The FastMCP Client fails to properly terminate the stdio mcp after calling `client.close()`, while the official MCP Client handles subprocess cleanup correctly.

## Steps to Reproduce

1. Create a test script with the following code:

### add_mcp.py
```python
from fastmcp import FastMCP

mcp = FastMCP("Demo 🚀")

@mcp.tool
def add(a: int, b: int) -> int:
    """Add two numbers"""
    return a + b

if __name__ == "__main__":
    mcp.run()
```

### test.py
```python
from mcp import ClientSession, StdioServerParameters
from mcp.client.stdio import stdio_client
from fastmcp import Client
import asyncio

server_params = StdioServerParameters(command="python", args=["add_mcp.py"])

async def run():
    # Test official MCP Client
    async with stdio_client(server_params) as (read, write):
        async with ClientSession(read, write) as session:
            # Initialize the connection
            await session.initialize()
            await asyncio.sleep(1)
            tools = await session.list_tools()
            print(tools)
            
    print("Official MCP Client finished")
    await asyncio.sleep(10)  # <- python add_mcp.py closed successfully 
    
    # Test FastMCP Client
    client = Client({"add-mcp": {"command": "python", "args": ["add_mcp.py"]}})
    await client._connect()
    print(tools)
    await client.close()
    print("FastMCP Client finished")
    
    await asyncio.sleep(10)  # <- python add_mcp.py NOT closed

def main():
    asyncio.run(run())

if __name__ == "__main__":
    main()
```

2. run `python test.py`
3. use `htop` and `filter: add_mcp.py` to inspect.

### Version Information

```Text
FastMCP version:                                                                     2.10.5
MCP version:                                                                         1.11.0
Python version:                                                                      3.12.9
Platform:                                     Linux-5.4.0-125-generic-x86_64-with-glibc2.31
```

### Additional Context

_No response_

chi2liu:
MCPConfigTransport was missing close() method, causing stdio subprocesses to remain running after Client.close().

Fixes #1311


jlowin:
@chi2liu please review open PRs before opening new ones as it splits the conversation - there is an ongoing discussion in #1339 and #1311 about why the strategy you have here is not sufficient. Going to close this in favor of 1339, if the hasattr check is an important part of the fix I think it would be best as a comment there, thanks!

PR #1432, opened 2025-08-10
fix: resolve transitive schema dependencies in OpenAPI output schema hoisting

abhishekbhakat:
Fixes OpenAPI schema hoisting to include transitive dependencies, preventing PointerToNowhere errors when clients load tool output schemas with nested $ref chains (e.g., A→B→C). 
Replaces naive pruning with recursive dependency collection in both legacy and experimental parsers.

### DIFF
--- a/src/fastmcp/experimental/utilities/openapi/schemas.py
+++ b/src/fastmcp/experimental/utilities/openapi/schemas.py
@@ -516,14 +516,8 @@ def extract_output_schema_from_responses(
     # Convert refs if needed
     output_schema = _ensure_refs_converted(schema)
 
-    # If schema has a $ref, resolve it first before processing nullable fields
-    if "$ref" in output_schema and schema_definitions:
-        ref_path = output_schema["$ref"]
-        if ref_path.startswith("#/$defs/"):
-            schema_name = ref_path.split("/")[-1]
-            if schema_name in schema_definitions:
-                # Replace $ref with the actual schema definition
-                output_schema = _ensure_refs_converted(schema_definitions[schema_name])
+    # Don't resolve top-level $refs inline - let the dependency collector handle them
+    # This ensures referenced schemas remain in $defs rather than being inlined
 
     # Convert OpenAPI schema to JSON Schema format
     # Only needed for OpenAPI 3.0 - 3.1 uses standard JSON Schema null types


abhishekbhakat:
## Real-world impact: Apache Airflow integration

This issue surfaced while integrating FastMCP with [Apache Airflow](https://github.com/apache/airflow)'s OpenAPI endpoints. Airflow is a widely-adopted workflow orchestration platform, and getting FastMCP to work seamlessly with established projects like this is crucial for broader adoption.

### Example failure case

When exposing Airflow's `GET /api/v2/plugins` endpoint as a FastMCP tool, clients receive this error:

```json
{
    "content": [
        {
            "type": "text",
            "text": "PointerToNowhere: '/$defs/FastAPIAppResponse' does not exist within {'properties': {'plugins': {'items': {'$ref': '#/$defs/PluginResponse'}, 'type': 'array', 'title': 'Plugins'}, 'total_entries': {'type': 'integer', 'title': 'Total Entries'}}, 'type': 'object', 'required': ['plugins', 'total_entries'], 'title': 'PluginCollectionResponse', 'description': 'Plugin Collection serializer.', '$defs': {'PluginResponse': {'properties': {'name': {'type': 'string', 'title': 'Name'}, 'macros': {'items': {'type': 'string'}, 'type': 'array', 'title': 'Macros'}, 'flask_blueprints': {'items': {'type': 'string'}, 'type': 'array', 'title': 'Flask Blueprints'}, 'fastapi_apps': {'items': {'$ref': '#/$defs/FastAPIAppResponse'}, 'type': 'array', 'title': 'Fastapi Apps'}, 'fastapi_root_middlewares': {'items': {'$ref': '#/$defs/FastAPIRootMiddlewareResponse'}, 'type': 'array', 'title': 'Fastapi Root Middlewares'}, 'appbuilder_views': {'items': {'$ref': '#/$defs/AppBuilderViewResponse'}, 'type': 'array', 'title': 'Appbuilder Views'}, 'appbuilder_menu_items': {'items': {'$ref': '#/$defs/AppBuilderMenuItemResponse'}, 'type': 'array', 'title': 'Appbuilder Menu Items'}, 'global_operator_extra_links': {'items': {'type': 'string'}, 'type': 'array', 'title': 'Global Operator Extra Links'}, 'operator_extra_links': {'items': {'type': 'string'}, 'type': 'array', 'title': 'Operator Extra Links'}, 'source': {'type': 'string', 'title': 'Source'}, 'listeners': {'items': {'type': 'string'}, 'type': 'array', 'title': 'Listeners'}, 'timetables': {'items': {'type': 'string'}, 'type': 'array', 'title': 'Timetables'}}, 'type': 'object', 'required': ['name', 'macros', 'flask_blueprints', 'fastapi_apps', 'fastapi_root_middlewares', 'appbuilder_views', 'appbuilder_menu_items', 'global_operator_extra_links', 'operator_extra_links', 'source', 'listeners', 'timetables'], 'title': 'PluginResponse', 'description': 'Plugin serializer.'}}}"
        }
    ],
    "isError": true
}
```

As you can see, the schema includes `PluginResponse` in `$defs`, but `FastAPIAppResponse` (referenced by `PluginResponse.fastapi_apps`) is missing, causing the client to fail when trying to resolve the reference chain.

### Why this matters

For FastMCP to gain wider adoption, it needs to work reliably with real-world APIs from established projects. Airflow's OpenAPI spec follows standard patterns with nested schema references, and this fix ensures FastMCP can handle such schemas correctly.

The patch addresses this by implementing proper transitive dependency collection, ensuring all referenced schemas are included in the hoisted `$defs` block.

jlowin:
Is this different from the discussion in #1372 and #1426?

abhishekbhakat:
> Is this different from the discussion in #1372 and #1426?

#1372 #1426 was about the experimental parser's input schema processing - when creating tools from OpenAPI specs, the inputSchema was missing transitive dependencies.

This is about output schema processing - when tools return responses, the output_schema was missing transitive dependencies, causing client-side validation failures.

So yes, this PR addresses a separate but related issue that wasn't covered by the previous fixes! The Airflow example demonstrates this perfectly - it's an output schema validation failure that occurs when clients try to load the tool response schema.

abhishekbhakat:
@jlowin, I added tests which will fail without the fixes. The tests should explain what the PR aims to fix in an easy way.

jlowin:
Ok thanks! I’ll be back later this afternoon and have a chance to look more closely then

jlowin:
@abhishekbhakat all of the tests in this PR pass on the main branch. In fact, isn't [this unit test](https://github.com/jlowin/fastmcp/blob/18533951ad3a3f420d23c295862d6b8696d70cbe/tests/experimental/utilities/openapi/test_transitive_references.py#L59-L115) already checking the described situation of ensuring transitive dependencies are present in an output schema?

Please make sure to test against `main` to avoid duplicated work!

One other thing - your actual code appears to deal only with pruning unused `$defs` and not with transitive dependencies at all. If there's another issue lurking here we can definitely look into it - please open an issue with an MRE!

abhishekbhakat:
I can still reproduce is on the legacy parser: its prune only scans the main schema and misses $defs→$defs refs, so in an A→B→C chain it keeps B but drops C.

Planing to cut it out in the next release ?
Until then it kinda blocks any dev having dependencies on fastmcp.

jlowin:
We're not focusing on improvements to the legacy parser, it is only present to ensure continuity and a baseline while we collect test cases for the new parser. These fixes will go out in 2.11.3 soon.

PR #1434, opened 2025-08-11
Add OAuth proxy that allows authentication with social IDPs without DCR support

jlowin:


src/fastmcp/server/auth/proxy.py
```diff
@@ -0,0 +1,1059 @@
...
+        else:
+            logger.debug("No upstream revocation endpoint configured")
+
+        logger.info("Token revoked successfully")
+
+    # -------------------------------------------------------------------------
+    # Custom Route Handling
+    # -------------------------------------------------------------------------
+
+    async def _handle_proxy_token_request(self, request: Request) -> JSONResponse:
```
AaronPowell96, line 742: Is this actually being used and forwarding the verifier or is `_handle_idp_callback` being used instead?

PR #1444, opened 2025-08-11
Improve stdio and mcp_config clean-up

strawgate:
Improve stdio transport clean-up:
1. Ensure stdio tasks can be cleaned up by removing the `_connect_task` from the class to avoid the `self` reference in the background task that prevents transport from being garbage collected.
2. Send a stop signal to the connection task if the transport is being garbage collected.

Improve mcp_config clean-up:
1. The MCPConfigTransport now tracks its underlying transports so a call to close() will close underlying transports
2. Note: The stdio clean-up above allows the transport to be automatically garage collected

Change `fastmcp` on the `Context` object to be a weakref. Context objects floating around currently hold strong references to the `fastmcp` and proxy objects preventing garbage collection of related servers and clients.

Add a property to Context to ensure that users can continue to access `ctx.fastmcp` without having to evaluate the weakref

/marvin take a look


strawgate:
I was able to get the transport to correctly exit when it falls out of scope by moving the nested connect task function but I am unable to get the client + transport to correctly exit when they both fall out of scope, will continue poking at this

strawgate:
@jlowin moving the connect task out of the class seems to resolve the clean-up issue for the stdio transport (by ensuring that the background task doesn't hold a `self` reference to the transport) but does not resolve the clean-up issue for MCPConfig composite servers, so i'm taking a look at that now.

Anyio 4.10 includes a new AsyncContextManagerMixin that I want to look into more and try out and see if that might help us with clean-up on these objects

src/fastmcp/mcp_config.py
```diff
@@ -104,7 +105,11 @@ def to_transport(self) -> FastMCPTransport:
             exclude_tags=self.exclude_tags,
         )
 
-        return FastMCPTransport(wrapped_mcp_server)
+        return wrapped_mcp_server, transport
```
jlowin, line 108: Is this wrapped_mcp_server used? could we get away with just returning the transport as to_transport()?

Oh I see below where it's used -- maybe rather than have the conditional you could just create the server there, and always hit `to_transport()`? Just a thought

jlowin (approved):

src/fastmcp/mcp_config.py
strawgate: this is a bug actually, `to_transport` is supposed to wrap the server in a FastMCPTransport and return it, not directly return the underlying transport. Will update quick

tests/client/test_stdio.py
```diff
@@ -56,6 +74,77 @@ async def test_keep_alive_maintains_session_across_multiple_calls(
...
+        await test_server()
+
+        gc_collect_harder()
+
+        # This test may fail/hang while debugging because the debugger holds a reference to the underlying transport
+
+        with pytest.raises(psutil.NoSuchProcess):
+            while True:
+                psutil.Process(pid)
+                await asyncio.sleep(0.1)
```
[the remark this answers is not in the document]
strawgate: wrong

src/fastmcp/mcp_config.py
strawgate: if we just return the normal transport than it wont be a transforming server, itll just be a normal server

strawgate:
/marvin i want your best impression of an expert developer who offers real advice to make the pull request even better

jlowin:
No idea why the static error is popping up now, seems unrelated?

strawgate:
> No idea why the static error is popping up now, seems unrelated?

I think it has been happening the whole time, will investigate though

PR #1498, opened 2025-08-14
Fix oauth pyright type checking

strawgate:
Dont know why this is only an issue on some branches


strawgate:
this is just https://github.com/jlowin/fastmcp/pull/1490 pr but not on my fork, will merge as I had approval there

PR #1499, opened 2025-08-14
docs: Fix type annotation in return value documentation

MaikelVeen:
This PR fixes a small documentation issue I came across whilst reading the docs. When using this code directly from the example, one would get  the following type checker error:

```
error: Expected class but received "(content: ContentBlock | str, role: Role | None = None, **kwargs: Any) -> PromptMessage" (reportGeneralTypeIssues)
```

The return type annotation should be `PromptMessage` here I believe. Thanks for the work on the package!

### DIFF
--- a/docs/servers/prompts.mdx
+++ b/docs/servers/prompts.mdx
@@ -188,10 +188,10 @@ FastMCP intelligently handles different return types from your prompt function:
 -   **`Any`**: If the return type is not one of the above, the return value is attempted to be converted to a string and used as a `PromptMessage`.
 
 ```python
-from fastmcp.prompts.prompt import Message
+from fastmcp.prompts.prompt import PromptMessage
 
 @mcp.prompt
-def roleplay_scenario(character: str, situation: str) -> list[Message]:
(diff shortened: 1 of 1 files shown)


jlowin:
Thanks for calling this out! I think actually it should be `PromptResult` to be complete but will confirm

jlowin:
/marvin is it possible for you to update this to reflect `PromptResult` as the type?

jlowin:
Ah Marvin can't open a fork. @MaikelVeen I apologize I'm going to hijack your PR to see if we can get Marvin working on forks

jlowin:
Hm this appears to be an upstream problem in the claude code action's setup and not something we can fix here. (https://github.com/anthropics/claude-code-action/issues/223, https://github.com/anthropics/claude-code-action/issues/46). Let's merge this to get you in the repo and I'll adjust after!

jlowin (approved):

MaikelVeen:
Ah yes, using `PromptResult` makes even more sense 🙌🏻

PR #1507, opened 2025-08-15
[I18n]: I18n docs support - add zh_CN

Gu-f:
Internationalized document support, with additional simplified Chinese content added.
Translation in progress...  
https://fastmcp.wiki/  
image


Gu-f:
```text
  # Check if docs.json content changed (ignoring formatting)
  if ! jq --sort-keys . docs/docs.json.orig | diff -q - <(jq --sort-keys . docs/docs.json) > /dev/null 2>&1; then
    echo "❌ docs.json content has changed!"
    echo ""
    echo "Run `just api-ref-all` to regenerate the SDK docs, then commit the changes. Note: you may need to install `just` (https://github.com/casey/just)"
    exit 1
  fi
```  
I modified the docs.json file. However, this check failed due to the imposed content restrictions.    
It seems that the docs.json.orig needs to be adjusted.

jlowin:
Hi @Gu-f -- I really appreciate the effort here, but as a small maintainer team that doesn't speak Chinese, we are not going to be able to keep a localization of FastMCP that will be up to date and mirror the EN content appropriately. We also have no way to ensure the quality of the documentation over time. Therefore, unfortunately, we can't accept this PR now. I hope that devices like llms.txt will make it possible for users to dynamically access the English documentation via LLMs in their preferred language. Thanks!

PR #1517, opened 2025-08-16
feat: introduce fastmcp.json configuration system

### LINKED ISSUE #1480: Introduce fastmcp.json
### Enhancement

currently a ton of information is overloaded on the `FastMCP` server instance, such as dependencies, host, port, etc. *some* of this can be supplied via CLI or at runtime, but not in a structured way. Also exposing the correct entrypoint (e.g. which FastMCP instance in which file) is not always clear.

We will add a `fastmcp.json` file that provides a structured, versioned schema for specifying:
- entrypoint (path:obj or maybe path and obj seperately? tbd. should factory functions be handled differently than instances?)
- dependencies
- version (python version)
- transport (and also host / post)
- etc

Some open questions: should there be one json file per server? or one json file with multiple server entries? inclined to say one file per server with `fastmcp.json` as the canonical default.

The idea is now `fastmcp run fastmcp.json` (or just `fastmcp run`) is a valid, fully specified way to spin up a server

`fastmcp init` can scaffold a mcp project

jlowin:
## Summary

This PR introduces `fastmcp.json`, a declarative configuration system for FastMCP servers that simplifies deployment and dependency management.

Closes #1480

## What is fastmcp.json?

A configuration file that replaces command-line arguments and programmatic configuration with a single declarative format:

```json
{
  "$schema": "https://gofastmcp.com/schemas/fastmcp_config/v1.json",
  "entrypoint": {
    "file": "server.py",
    "object": "mcp"
  },
  "environment": {
    "python": "3.12",
    "dependencies": ["pandas", "requests"]
  },
  "deployment": {
    "transport": "http",
    "port": 8000,
    "env": {
      "API_URL": "https://api.${BASE_URL}/v1"
    }
  }
}
```

## Key Features

**Auto-detection**: Running `fastmcp run` without arguments automatically finds and uses `fastmcp.json` in the current directory.

**Flexible entrypoint syntax**: Supports `"server.py:mcp"` string format that auto-parses to object format.

**Environment variable interpolation**: Use `${VAR_NAME}` syntax in deployment.env for runtime substitution.

**Isolated environments**: Automatically creates isolated Python environments with specified dependencies using `uv`.

**Full CLI integration**: Works with `fastmcp run`, `dev`, `install`, and `inspect` commands.

## Migration from Dependencies Parameter

The `dependencies` parameter in FastMCP servers is now deprecated in favor of fastmcp.json. Existing code continues to work with deprecation warnings guiding users to the new approach.

## Testing

Comprehensive test coverage including configuration parsing, environment variable interpolation, schema validation, and deprecation warnings.


strawgate:
🦮 Guiding questions/ideas 
Base64 encoded json config via CLI 👀👀👀
Mount MCP servers in json config? 
Transform tools in json config?
Call tools on server startup?

Maybe worth noodling on a separate fastmcp-cli package at some point that provides some of this? 

Id love to have a great way to define a FastMCP client entirely declaratively inline in a cursor/Claude config and base64 + tool transform + Mount MCP would be killer

jlowin:
Oooh interesting. My main objective with this is to make the server's *own* config portable (e.g. my server lives in this file and requires this environment and expects to be run with this transport and this log level) without having to first run any code or import the server or serialize a CLI command -- I think a lot of your enhancements are sort of related to either building the server (transform tools from json config) or interacting with the server (call tools) and I'd view that as a different domain than this file (though I fully endorse the utility). This is more agnostic of what the server is or does (at least at this time) and really just tells you how to run it.

PR #1538, opened 2025-08-18
Fixed wrong import path in new docs page

KaliszS:
Just fixed one import from an example in new 2.12 docs page.


jlowin (approved):
Thanks!

PR #1545, opened 2025-08-19
switch from `pyright` to `ty`

zzstoatzz:
## the problem

`pyright` is slow and lenient. on the `fastmcp` codebase, it takes over 5 seconds to run type checking, which adds up

## the solution  

switch to [ty](https://github.com/astral-sh/ty), astral's new rust-based type checker. while `ty` is still preview software, it's been getting increasingly stable and i've heard the beta phase is near. its also like 20x faster and catches more problems

### performance comparison

| type checker | time | improvement |
|--------------|------|-------------|
| pyright | 5.19s | baseline |
| **ty** | **0.27s** | **19x faster** 🚀 |

### what this pr does

- replaces `pyright` with `ty` in all tooling (`pre-commit`, ci, `justfile`)
- configures `ty` to ~match `pyright`'s behavior for backwards compatibility
- fixes 10 type safety issues that `ty` caught but `pyright` missed
- documents which rules need more work (currently ignored to maintain parity)

### type coverage

`ty` is stricter than `pyright` and found 352 potential type issues. to maintain backwards compatibility, i've configured `ty` to ignore the same categories of errors that `pyright` was missing. the low-hanging fruit (10 errors) have been fixed:

- `unsupported-operator`: 2 errors - fixed with proper none checks
- `invalid-return-type`: 3 errors - fixed match statement exhaustiveness  
- `invalid-assignment`: 5 errors - added type ignores for contextvars

the remaining rules have significantly more violations (40-126 each) and are documented in `pyproject.toml` for future cleanup

### DIFF
--- a/src/fastmcp/experimental/utilities/openapi/parser.py
+++ b/src/fastmcp/experimental/utilities/openapi/parser.py
@@ -550,7 +550,7 @@ class OpenAPIParser(
 
                     # Add this schema and recursively find its dependencies
                     if (
-                        collected
+                        collected is not None
                         and schema_name not in collected
                         and schema_name in all_schemas
                     ):


KaliszS:
Yeah, I can confirm. In my solutions "unstable" ty usually works better and is more stable than pyright or mypy.

.github/workflows/run-static.yml
```diff
@@ -36,12 +36,9 @@ jobs:
...
           enable-cache: true
           cache-dependency-glob: "uv.lock"
-      - name: Set up Python
-        uses: actions/setup-python@v5
-        with:
-          python-version: "3.12"
+
       - name: Install dependencies
-        run: uv sync --dev
+        run: uv sync
```
zzstoatzz, line 41: `--dev` is installed implicitly with `uv sync`

pyproject.toml
```diff
@@ -114,18 +114,23 @@ python_files = ["test_*.py", "*_test.py"]
...
+[tool.ty.rules]
+# Rules with too many errors to fix right now (40+ each)
+invalid-argument-type = "ignore" # 40 errors
+no-matching-overload = "ignore"  # 126 errors  
+unknown-argument = "ignore"      # 61 errors
+unresolved-attribute = "ignore"  # 60 errors
+
+# Rules with moderate errors that need more investigation
+call-non-callable = "ignore" # 7 errors
+missing-argument = "ignore"  # 23 errors
```
zzstoatzz, line 133: the significant diff

jlowin (commented):
Not all heroes wear capes

PR #1546, opened 2025-08-19
Support `initialize` requests in middleware

jlowin:
This PR adapts part of #1215 to support initialize requests in FastMCP middleware. In order to do this, we have to introduce a new low-level `Session` object because initialize is handled outside the normal `handler` approach that FastMCP generally relies on.


BrandonShar:
Hey @jlowin this library has been amazing to use! 

I'm working on figuring out a solution for horizontally scaling while still being able to log usage across sessions and my current plan is:

1. set the server to stateless
2. use starlette middleware to read/set mcp-session-id
3. grab the initialize request and associate it with the session id.

This middleware looks perfect for step 3 (and I'd much rather stay within the FastMCP ecosystem), so I wanted to see if I could help with getting merged or if you were considering other approaches that might solve the same problem.

rishid:
Any help you need to get this merged? This is a feature I have been waiting for.

PR #1556, opened 2025-08-20
update aproto server to handle rich links

zzstoatzz:


jlowin:
This is purely a test. /marvin if you can, say hi

PR #1568, opened 2025-08-21
Added import of AuthProxy to auth __init__

KaliszS:
I fixed import path in docs, but this time I've added import to auth `__init__.py`, so next time it won't be required to update docs, if something will be updated again ;)


jlowin:
Good catch, missed this one!

PR #1575, opened 2025-08-21
Add setting to disable logging configuration

leameow:
Most production setup have their own logging configuration.

### DIFF
--- a/src/fastmcp/__init__.py
+++ b/src/fastmcp/__init__.py
@@ -6,11 +6,11 @@ from fastmcp.settings import Settings
 from fastmcp.utilities.logging import configure_logging as _configure_logging
 
 settings = Settings()
-# if False:
-_configure_logging(
-    level=settings.log_level,
-    enable_rich_tracebacks=settings.enable_rich_tracebacks,
-)
+if settings.log_enabled:
+    _configure_logging(
+        level=settings.log_level,
+        enable_rich_tracebacks=settings.enable_rich_tracebacks,
+    )
 
 from fastmcp.server.server import FastMCP
 from fastmcp.server.context import Context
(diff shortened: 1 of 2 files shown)


leameow:
> My recommendation is to add a literal "NOTSET" value to the existing LOG_LEVEL, and modify _configure_logging() to return early if the requested level is NOTSET. That way users can disable fastmcp's log configuration by saying e.g. `FASTMCP_LOG_LEVEL=NOTSET`

Logger set to `NOTSET` just means to use the root level, that's somewhat adding a new semantic to an existing concept :/

jlowin:
Is that not what you're doing here, avoiding FastMCP specific configuration and inheriting whatever other settings you've introduced?

The PR disables FastMCP from configuring logging behavior at all, so it is exactly equivalent to if the log level setting didn't exist or were not set.

leameow:
> Is that not what you're doing here, avoiding FastMCP specific configuration and inheriting whatever other settings you've introduced?
> 
> The PR disables FastMCP from configuring logging behavior at all, so it is exactly equivalent to if the log level setting didn't exist or were not set.

My point is that semantically setting `NOT_SET` inherit only the log level, but keep the handler configuration. In this case this would change the behavior to also skip the handler. This really is a nitpick tho, I don't see anyone complaining about this. It's just not the behavior I would assume if looking at the configuration.

I can update the PR.

leameow:
Change done!

jlowin:
Oh sorry I see your point, I do agree. Hadn't had sufficient coffee to consider the full implication. I'm happy to accept the PR as-is or revert to the previous approach if you don't mind the extra commit.

leameow:
> Oh sorry I see your point, I do agree. Hadn't had sufficient coffee to consider the full implication. I'm happy to accept the PR as-is or revert to the previous approach if you don't mind the extra commit.

Haha, sure no worry, let me revert!

leameow:
Went back to first approach

PR #1599, opened 2025-08-23
Add Context Snapshot Support

shaun-mercor:
A common pattern I am implementing to test agentic AI is to 

1. Initialize an MCP server session
2. Client calls a `load_data`-type function that initializes the MCP session with a previous / external state
3. Agent / client calls tools that can adjust the session state
4. On session close or the end of agent execution trace, capture the new context state
5. For unit tests, compute the difference between the initial state and the end state and write assertions accordingly.

I wanted to create a series of functions to make getting and comparing context states easier to work with.

I think this could be the go-to way people are evaluating MCP agents for consistency 🚀

### DIFF
--- a/examples/state_snapshot_demo.py
+++ b/examples/state_snapshot_demo.py
@@ -8,6 +8,7 @@ from fastmcp.server.context import Context
 
 mcp = FastMCP("State Snapshot Demo")
 
+
 @mcp.tool
 def setup_user_data(username: str, theme: str, language: str, ctx: Context) -> str:
     """Set up some user preferences in the context state."""
@@ -15,13 +16,15 @@ def setup_user_data(username: str, theme: str, language: str, ctx: Context) -> s
     ctx.set_state("theme", theme)
     ctx.set_state("language", language)
     ctx.set_state("session_start", "2024-01-15T10:30:00Z")
-    
+
     return f"User data set up for {username} (theme: {theme}, language: {language})"
+
 
 @mcp.tool
 def create_state_snapshot(snapshot_name: str, ctx: Context) -> str:
     """Create a named snapshot of the current context state."""
     return ctx.create_snapshot(snapshot_name)
+
 
 @mcp.tool
 def show_current_state(ctx: Context) -> dict:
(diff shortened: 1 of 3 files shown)


shaun-mercor:
Happy to contribute to docs as well if helpful to get support.

jlowin:
Hi @shaun-mercor - I appreciate the thought but this is a lot of new methods on a very specialized object (`Context`), without a guarantee that they work in all situations (for example, this won't work in horizontally scaled servers and is likely not effective across multiple non-STDIO sessions). 

We already have a way to get/set state on the context object, as your example uses, so I'm not clear why the context object is taking responsibility for maintaining snapshots of itself - this feels like something that is better achieved by writing snapshots to external state and interacting with them through a vanilla tool call, not a core framework capability. Perhaps you could rewrite this as a `contrib` module that demonstrates how to take advantage of that pattern without any core framework modifications? Otherwise, could you open an issue to describe the specific issue and surface area you want to target and why the framework prohibits it?

shaun-mercor:
Fair feedback :) 

The primary use case for this pattern vs. storing in an external system is to support ephemeral MCP/agentic unit tests, where the test case is loaded into the session context, the agent runs, and then you want to test, without having any external dependencies or external systems.

What'd you have in mind for a `contrib` module? That sounds kind of cool, but I didn't see an example one to follow. 

If you have time, my only other question about your reply 
> This won't work in horizontally scaled servers and is likely not effective across multiple non-STDIO sessions

Are you able to horizontally scale servers and preserve `ctx` today? Maybe I missed some of the magic initially, cool if so! ✨

jlowin:
I understand your objective, but the functionality you're introducing does not inherently depend on any part of the framework nor does it change how the framework works. You could just as easily add a `ContextSnapshot` class to your unit tests with the exact same methods and a `_snapshots` attribute to store the serialized context objects, and use it the same way as the changes in this PR for the same purposes. That could be something to add in the `fastmcp.contrib` module, which is essentially useful utilities for working with FastMCP that are not part of the core framework (and are not maintained by the core team, but rather by their authors, so we can't guarantee they work with all versions of FastMCP). You can see examples in that directory.

PR #1636, opened 2025-08-26
fix: fix StructuredLoggingMiddleware payload serialization

vl-kp:
## Description

**Contributors Checklist**

- [ ] My change is related to issue #(issue number)

**Review Checklist**

---

### DIFF
--- a/src/fastmcp/server/middleware/logging.py
+++ b/src/fastmcp/server/middleware/logging.py
@@ -152,7 +152,7 @@ class StructuredLoggingMiddleware(Middleware):
         if self.methods and context.method not in self.methods:
             return await call_next(context)
 
-        self.logger.log(self.log_level, json.dumps(start_entry))
+        self.logger.log(self.log_level, json.dumps(start_entry, default=str))
 
         try:
             result = await call_next(context)
@@ -162,7 +162,7 @@ class StructuredLoggingMiddleware(Middleware):
                 "request_success",
                 result_type=type(result).__name__ if result else None,
             )
-            self.logger.log(self.log_level, json.dumps(success_entry))
+            self.logger.log(self.log_level, json.dumps(success_entry, default=str))
 
             return result
         except Exception as e:
@@ -172,5 +172,5 @@ class StructuredLoggingMiddleware(Middleware):
                 error_type=type(e).__name__,
                 error_message=str(e),
             )
-            self.logger.log(logging.ERROR, json.dumps(error_entry))
+            self.logger.log(logging.ERROR, json.dumps(error_entry, default=str))
             raise


strawgate:
For tools we have a configurable serializer, would it make sense for the structured logging middleware to have a configurable serializer?

vl-kp:
updated, @strawgate  please review again

strawgate:
Sorry, I should have indicated that my comment was more of a question, we'll need @jlowin for a review

jlowin:
Good by me!

src/fastmcp/server/middleware/logging.py
```diff
@@ -111,6 +114,7 @@ def __init__(
         log_level: int = logging.INFO,
         include_payloads: bool = False,
         methods: list[str] | None = None,
+        serializer: Callable[[Any], Any] | None = None,
```
strawgate: Should the callable return a str?

src/fastmcp/server/middleware/logging.py
```diff
@@ -152,7 +178,9 @@ async def on_message(self, context: MiddlewareContext, call_next: CallNext) -> A
         if self.methods and context.method not in self.methods:
             return await call_next(context)
 
-        self.logger.log(self.log_level, json.dumps(start_entry))
+        self.logger.log(
+            self.log_level, json.dumps(start_entry, default=self._json_default)
```
strawgate: The default serializer should take an object and return a str -- we should call json dumps in the default serializer

strawgate: Looking at this more, I think the serialization should happen in _create_log_message. Instead of placing the `__dict__` into the payload we should call the serializer and store the result into payload?

src/fastmcp/server/middleware/logging.py
vl-kp: it returns a json serializable object, not necessarily a str

strawgate: A serializer should return a string not something that still needs to be serialized

Is there a reason this can't be an actual serializer?

Check out the tool serializer implementation for a concrete example of what I was referring to

vl-kp: changed to use default_serializer

src/fastmcp/server/middleware/logging.py
vl-kp: moved into _create_log_message

strawgate:
This looks good, I may pull this locally tomorrow and make some small adjustments.. Thanks for the revisions. 

I assume after the latest changes this is still useful from your perspective and solves your use-case?

vl-kp:
> This looks good, I may pull this locally tomorrow and make some small adjustments.. Thanks for the revisions.
> 
> I assume after the latest changes this is still useful from your perspective and solves your use-case?

yes, it won't crash anymore

strawgate:
@vl-kp i dont think tests were passing after your most recent change and I got a little carried away with the "clean-up"

can you make sure the latest version of the PR still solves your use-case? 

note: i changed serializer to payload_serializer

vl-kp:
> @vl-kp i dont think tests were passing after your most recent change and I got a little carried away with the "clean-up"
> 
> can you make sure the latest version of the PR still solves your use-case?
> 
> note: i changed serializer to payload_serializer

thanks, latest code still works for me

PR #1641, opened 2025-08-26
feat: add schema dereference middleware

vl-kp:
## Description

**Contributors Checklist**

- [ ] My change is related to issue #(issue number)

**Review Checklist**

---

### DIFF
--- a/docs/servers/middleware.mdx
+++ b/docs/servers/middleware.mdx
@@ -420,6 +420,35 @@ mcp.add_middleware(StructuredLoggingMiddleware(include_payloads=True))
 
 The built-in versions include payload logging, structured JSON output, custom logger support, payload size limits, and operation-specific hooks for granular control.
 
+### Schema Dereference Middleware
+
+Some MCP clients have limited support for JSON Schema `$ref` references in tool and resource template schemas. This can lead to null parameters or validation failures when a client cannot resolve references during tool invocation.
+
+<Info>
+The schema dereference middleware flattens internal `$ref` references (those pointing to `#/$defs/...`) in listing responses, so clients receive inline schemas that are easier to consume.
+</Info>
+
+```python
+from fastmcp import FastMCP
+from fastmcp.server.middleware.schema_dereference import SchemaDereferenceMiddleware
+
+mcp = FastMCP("MyServer")
+mcp.add_middleware(SchemaDereferenceMiddleware())
+```
+
+What it does
+- Inlines internal `$ref` to `#/$defs/...` for:
+  - on_list_tools: `Tool.parameters`
+  - on_list_resource_templates: `ResourceTemplate.parameters`
+- Handles common schema shapes:
+  - Properties, arrays (`items`), maps (`additionalProperties`)
+  - Composition keywords (`allOf`, `oneOf`, `anyOf`)
+  - Transitive and reused definitions
+
+<Tip>
+Enable this middleware when serving tools or templates to clients that cannot resolve JSON Schema `$ref` internally. It is safe to use with clients that fully support `$ref` as well.
+</Tip>
+
 ### Rate Limiting Middleware
 
 Rate limiting is essential for protecting your server from abuse, ensuring fair resource usage, and maintaining performance under load. FastMCP includes sophisticated rate limiting middleware at `fastmcp.server.middleware.rate_limiting`. 
--- a/src/fastmcp/server/middleware/schema_dereference.py
+++ b/src/fastmcp/server/middleware/schema_dereference.py
@@ -1,0 +1,201 @@
+from copy import deepcopy
+from typing import Any
+
+import mcp.types as mt
+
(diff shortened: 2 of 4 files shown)


vl-kp:
Hi @jlowin could you review this PR

richardkmichael:
Since I have a similar use-case and have my own hack locally, I tested this PR.  I have comments on the PR below.

However, since I wanted slightly different behaviour, I also tried re-implementing it with `jsonref` (mentioned on #1192), and it's so simple that I'd be hesitant to add a dedicated middleware in repo.  Maybe in contrib, but as-is, I'm not sure it's flexible enough.

First, here it is with jsonref -- it removes tools which fail dereferencing (to avoid LLM misuse) but logging a warning (i.e., fix the server):

```python
from jsonref import JsonRefError, replace_refs

class DereferenceSchemaMiddleware(Middleware):
    def _dereference_tool(self, tool, remove_defs=True):
        """Dereference a tool's schema, returning the modified tool or None on failure.

        Args:
            tool: Tool object with parameters schema to dereference
            remove_defs: If True, remove $defs from schema after dereferencing (default: True)
        """
        try:
            # proxies=False returns plain dicts to ensure compatibility with FastMCP
            # lazy_load=False resolves immediately
            dereferenced = replace_refs(tool.parameters, proxies=False, lazy_load=False)

            # If requested, remove $defs (since all references have been resolved)
            if remove_defs and isinstance(dereferenced, dict) and "$defs" in dereferenced:
                dereferenced = {k: v for k, v in dereferenced.items() if k != "$defs"}

            tool.parameters = dereferenced
            return tool

        except JsonRefError as e:
            logger.warning(
                f"Failed to dereference schema for tool '{tool.name}': {e}. "
                "Tool will be hidden from client."
            )
            return None
        except Exception as e:
            logger.error(
                f"Unexpected error dereferencing schema for tool '{tool.name}': {e}. "
                "Tool will be hidden from client."
            )
            return None

    async def on_list_tools(self, context: MiddlewareContext, call_next):
        tools = await call_next(context)

        return [
            dereferenced_tool
            for tool in tools
            if (dereferenced_tool := self._dereference_tool(tool)) is not None
        ]
```

Even more minimal if tolerating a lingering `$defs` in the schema, and returning the unmodified tool if dereferencing fails (similar to the PR):

```python
from jsonref import JsonRefError, replace_refs

class DereferenceSchemaMiddleware(Middleware):
    def _dereference_tool(self, tool):
        """Dereference a tool's schema, returning the unmodified tool if dereferencing fails.

        Args:
            tool: Tool object with parameters schema to dereference
        """
        try:
            # proxies=False returns plain dicts to ensure compatibility with FastMCP
            # lazy_load=False resolves immediately
            tool.parameters = replace_refs(tool.parameters, proxies=False, lazy_load=False)

        except JsonRefError:
            pass

        return tool


    async def on_list_tools(self, context: MiddlewareContext, call_next):
        tools = await call_next(context)

        return [ self._dereference_tool(tool) for tool in tools ]
```

### Comments on the PR

- I want deferencing conditional upon the client (i.e., broken ones).

The middleware is not conditional, so I tried both subclass and separate class importing the deref function; neither seem great options.

Since the deref function is module-level, a subclass can't `self._deref(...)`, and needs to conditionally `super().on_list_tools(...)` -- this means for the `else`-branch (not deref'ing), it needs to know if the parent-class does anything in addition to deref and implement that part only.

If not subclassing, the deref function can be imported, but this essentially side-stepping this middleware entirely, and just using its "internals" as a utility function.

I think these could be helped by defining the deref function in the class, improving encapsulation and permitting a subclass to call it whenever needed.

Both mean re-implementing the hooks, adding the condition.

Conditionality could be provided by adding a callable to the constructor for simple cases, and also a template function (e.g., `def self._should_dereference(): return True`) to override for complex conditions (I needed this).

- The PR seems to implement "best effort" deref, I want all-or-nothing with a warning/error.

I'm curious about the thinking here.  The PR leaves alone refs in the corner-cases.  I think this means tools failing to deref all arguments will still be potentially misused by the LLM (if it uses those args) -- making the problem less frequent, but still present.  Also, there is no logging when this occurs.

- The deref is in specific hooks, not at the schema-generation level.

This is debatable, i.e., a general vs. pragmatic and sufficient solution.

How were the specific hooks needing deref determined?

I don't think the deref in `on_list_resource_templates()` is actually needed because no JSON Schema is returned to the client for this call.  I reviewed the MCP Schema, and I _think_ only list_tools can contain `$ref` (`input_schema` is a JSON Schema fragment), but it's not straight-forward to determine this from the schema.

So a simple solution, for now, only needs to deref in `on_list_tools()`.  Easy enough to add more later.

OTOH, a general approach would alter FastMCPs schema _generation_, and all future responses would be deref'd.  (Tempting, to avoid MCP Schema interpretation.)


So my thought is that these knobs and design-decisions seem unnecessary, given the simplicity of the jsonref implementation.  🤷 

I have other comments about the code itself, which I can add later if going forward.

jlowin:
Closing as stale

Issue #1156, opened 2025-07-15
Detect stateless_http=True and prevent bi-directional communication


jlowin:
### Enhancement Description

When running in stateless HTTP mode, bi-directional features like sampling and elicitation are not supported (this is enforced at the protocol / infrastructure level, not a FastMCP decision). However, if users attempt to invoke sampling, it will go through and permanently hang (see #678). 

Explore whether the FastMCP server can determine its deployment transport (in particular, stateless http) and raise errors accordingly when unsupported features are used. 

### Proposed Implementation

```Python

```

tonyxwz:
I feel that `stateless_http` is probably not the most intuitive name for this argument.

HTTP is a stateless protocol, this flag is stating the fact that the streamable-http server that processes the transport messages is stateless and hence cannot handle requests such as elicitation and sampling because the when the elicitation and sampling result is sent back by the client, the server is refreshed and does not remember anything.

In fact, the `stateless_http` argument becomes `stateless` [in the transport itself](https://github.com/modelcontextprotocol/python-sdk/blob/ca3466666310dbcb5c45690ac2571c574759984f/src/mcp/server/streamable_http_manager.py#L52), but users need to dig into the source code to understand this.


Maybe it is better to change `FastMCP.run` to

```python

class StreamableHttpCtorArgs(TypedDict):
    # figure out a way to share this with the streamable http constructor.
    stateless: NotRequired[bool]

StreamableHttpTransportRunParams = TypedDict(
    "StreamableHttpTransportRunParams",
    {"streamable-http": StreamableHttpCtorArgs},
)

Transport = Literal["streamble-http", "sse", "stdio"] | StreamableHttpTransportRunParams | ...

fastmcp.run(
    transport = {
        "streamable-http": { "stateless": True },
    }  
)
```

instead of using keyword arguments and suffix to distinguish the transport arguments.

dwaipayan05:
Can we explicitly mention that in the User Documentation that `stateless_http` needs to be disabled in pages like https://gofastmcp.com/servers/elicitation ?

PR #1665, opened 2025-08-29
chore: prevent bi-directional features for stateless_http

vl-kp:
## Description

**Contributors Checklist**

- [x] My change is related to issue https://github.com/jlowin/fastmcp/issues/1156

**Review Checklist**

---

### DIFF
--- a/src/fastmcp/server/context.py
+++ b/src/fastmcp/server/context.py
@@ -373,6 +373,13 @@ class Context:
         or the request will error.
         """
 
+        # Guard against unsupported transport capabilities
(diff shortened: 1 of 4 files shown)


jlowin:
Thanks @vl-kp - coincidentally, I was just having a conversation about needing to do this. However, I do not like the approach of storing transport-level information on the server instance as yet another attribute, especially when it effectively reintroduces a server-level attribute that is already deprecated. This will require a design that more broadly lets us share transport information with running servers that I have some thoughts on (and this will be one consequence of). Thanks for the PR though!

PR #1701, opened 2025-09-01
Make openai depedency optional

### LINKED ISSUE #1696: The `openai` dependency library is large and unnecessary
### Enhancement

The latest release added `openai` as a required dependency package for the project. It's large and unnecessary for most scenarios. Its size is even larger than `fastmcp` itself (793kb vs 305kb).

It is only used by the sampling submodule.
It is only used for fallback purposes.
It is only an experimental feature.
It is only for fallback with openai-compatible modules. (e.g. fallback may use local models)

In all above scenarios, downstream users need to explicitly import the related modules and classes.

It may be better to move the `openai` dependency package to an optional/extra dependency group instead of the required dependency group.

jlowin:
Closes #1696 by moving openai to an optional dependency, with an informative message.

### DIFF
--- a/examples/sampling_fallback.py
+++ b/examples/sampling_fallback.py
@@ -1,3 +1,7 @@
+# /// script
+# dependencies = ["openai", "fastmcp"]
+# ///
+
 import asyncio
 import os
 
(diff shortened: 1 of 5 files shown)


strawgate:
Should it be in a group so that you can uv add FastMCP[OpenAI]?

jlowin:
I think this does that no? For example I added “fastmcp[openai]” to the uv dev requirements. Or does uv need more than that

strawgate:
Oh okay, for some reason I thought it had to be in its own group to work that way

strawgate (approved):

PR #1704, opened 2025-09-01
Preserve order in mixed text/image tool returns

### LINKED ISSUE #1656: Mixing text and images in output, changes the return item order
### Description

``` 
        return [
            f"text1",
            img_obj1.to_image_content(),
            f"text2",
            img_obj2.to_image_content()
        ]
```

returns

```
[
   {
      "type":"text",
      "text":"[\"text1\",\"text2\"]",
      "uuid":"f113f6d3-be48-4ced-bf1a-6ef682b98429"
   },
   {
      "type":"image",
      "file_uuid":"2cab7d83-386c-4251-8e23-fd3afe6f3a32"
   },
   {
      "type":"image",
      "file_uuid":"53a4462d-7f73-467d-9428-301e2475fbc2"
   }
]
```

testing in claude desktop.

### Example Code

```Python

```

### Version Information

```Text
FastMCP version:                                                                             2.11.2
MCP version:                                                                                 1.12.2
Python version:                                                                              3.13.7
Platform:                                                                 Windows-11-10.0.26100-SP0
FastMCP root path: C:\Users\[username]\AppData\Local\Packages\PythonSoftwareFoundation.Python.3.13_qbz5…
```

strawgate:
Fix issue where returning mixed text and images like [text1, img1, text2, img2] would reorder items by aggregating text content. Now each item maintains its original position by creating separate TextContent blocks.

Fixes #1656


strawgate:
/marvin let's still combine adjacent non-mcp types. i.e. image, "x", "y", image should be 2 image blocks and 1 content block.

src/fastmcp/tools/tool.py
```diff
@@ -502,26 +502,59 @@ def _convert_to_content(
...
+                    # Single item - convert directly to avoid combining when not needed
+                    content_items.extend(
+                        _convert_to_content(
+                            non_mcp_batch[0],
+                            serializer=serializer,
+                            _process_as_single_item=True,
+                        )
+                    )
+                else:
+                    # Multiple items - combine into a single text block
```
jlowin, line 523: This feels like the complexity may not be worth the benefit of compressing them all into text content... maybe its less surprising to emit a new TextContent for each one, and users who really want control should return TextContent explicitly?

tests/server/test_server_interactions.py
```diff
@@ -426,7 +428,7 @@ async def test_tool_mixed_list_with_image(
         self, tool_server: FastMCP, tmp_path: Path
     ):
         """Test that lists containing Image objects and other types are handled
-        correctly. Note that the non-MCP content will be grouped together."""
+        correctly. Items now preserve their original order."""
```
jlowin, line 431: This is a weird Claude thing I've noticed where it frequently refers to essentially the outgoing diff, since this comment will be confusing once this is actually merged. The word "now" being the offender.

src/fastmcp/tools/tool.py
strawgate: Maybe if there's any native MCP type we keep everything separate but otherwise we compress? 

It's pretty annoying for an array to get turned into 100 text content blocks but once you're returning an MCP native type maybe we assume you wanted it to be that way

PR #1706, opened 2025-09-01
Fix OAuth resource URL handling and WWW-Authenticate header

### LINKED ISSUE #1685: www-authenticate returns well-known relative to resource_server_url rather than base URL
### Description

https://github.com/jlowin/fastmcp/pull/1682 introduced `resource_server_url` to OAuth proxy providers, to fix https://github.com/jlowin/fastmcp/issues/1348.

However, now the `www-authenticate` header returned from the initial unauthenticated request to `resource_server_url` returns a path relative to `resource_server_url` for the `well-known` endpoint - which does not exist. The actual path is relative to the base URL.

## Actual Behaviour

`www-authenticate` is returned relative to the resource server URL.

```sh
$ curl -sSI http://localhost:8000/mcp | grep www

www-authenticate: Bearer error="invalid_token", error_description="Authentication required", resource_metadata="http://localhost:8000/mcp/.well-known/oauth-protected-resource"
```

```sh
$ curl -sSI http://localhost:8000/mcp/.well-known/oauth-protected-resource

HTTP/1.1 404 Not Found
```

## Expected Behaviour

`www-authenticate` is returned relative to the base URL.

```sh
$ curl -sSI http://localhost:8000/mcp | grep www

www-authenticate: Bearer error="invalid_token", error_description="Authentication required", resource_metadata="http://localhost:8000/.well-known/oauth-protected-resource"
```

```sh
$ curl -sSI http://localhost:8000/.well-known/oauth-protected-resource

HTTP/1.1 200 OK
```

### Example Code

```Python
from fastmcp import FastMCP
from fastmcp.server.auth.providers.github import GitHubProvider

github_auth = GitHubProvider(
    client_id="...",
    client_secret="...",
    resource_server_url="http://localhost:8000/mcp"
)

mcp = FastMCP(name="My MCP", auth=github_auth)

if __name__ == "__main__":
    mcp.run(transport="http", port=8000)
```

### Version Information

```Text
Using `2.12.0rc1` from latest commit at time of reporting (`https://github.com/jlowin/fastmcp.git#1c6ba0a0119913e7999d06838d966e44175a75e7`).
```

jlowin:
This PR fixes OAuth authentication issues related to resource URL handling and the WWW-Authenticate header.

## Problem

When an MCP server is mounted at a non-root path (e.g., `/mcp` or `/api/v1/mcp`), there were two related issues:

1. **Issue #1685**: The WWW-Authenticate header incorrectly included the MCP path in the `.well-known` URL, returning `http://localhost:8000/mcp/.well-known/oauth-protected-resource` instead of `http://localhost:8000/.well-known/oauth-protected-resource`

2. **PR #1682 regression risk**: The resource URL in the `.well-known` metadata needs to automatically reflect the actual MCP endpoint path for RFC 8707 compliance

## Solution  

Refactored the authentication architecture to dynamically determine the resource URL based on where the MCP endpoint is actually mounted, while ensuring the `.well-known` endpoint always remains at the base URL.

Key changes:
- Auth providers now accept the MCP path dynamically via `get_routes(mcp_path, mcp_endpoint)`
- The resource URL is automatically set to the MCP endpoint path
- The `.well-known` endpoint always stays at the base URL
- Deprecated `resource_server_url` parameter (with backwards compatibility)

## Example

```python
from fastmcp import FastMCP
from fastmcp.server.auth.providers.github import GitHubProvider

auth = GitHubProvider(
    client_id="your-client-id",
    client_secret="your-secret",
    base_url="http://localhost:8000"  # Only need base URL now
)

app = mcp.http_app(path="/api/v1/mcp")  # Resource URL automatically captured
```

With this setup:
- WWW-Authenticate header points to: `http://localhost:8000/.well-known/oauth-protected-resource`
- Resource metadata shows: `"resource": "http://localhost:8000/api/v1/mcp"`

Fixes #1685

### DIFF
--- a/examples/auth/azure_oauth/server.py
+++ b/examples/auth/azure_oauth/server.py
@@ -23,7 +23,6 @@ auth = AzureProvider(
     tenant_id=os.getenv("AZURE_TENANT_ID")
     or "",  # Required for single-tenant apps - get from Azure Portal
     base_url="http://localhost:8000",
-    resource_server_url="http://localhost:8000/mcp",
     # redirect_path="/auth/callback",  # Default path - change if using a different callback URL
 )
 
--- a/examples/auth/github_oauth/server.py
+++ b/examples/auth/github_oauth/server.py
@@ -19,7 +19,6 @@ auth = GitHubProvider(
     client_id=os.getenv("FASTMCP_SERVER_AUTH_GITHUB_CLIENT_ID") or "",
     client_secret=os.getenv("FASTMCP_SERVER_AUTH_GITHUB_CLIENT_SECRET") or "",
     base_url="http://localhost:8000",
-    resource_server_url="http://localhost:8000/mcp",
     # redirect_path="/auth/callback",  # Default path - change if using a different callback URL
 )
 
--- a/examples/auth/google_oauth/server.py
+++ b/examples/auth/google_oauth/server.py
@@ -19,7 +19,6 @@ auth = GoogleProvider(
     client_id=os.getenv("FASTMCP_SERVER_AUTH_GOOGLE_CLIENT_ID") or "",
     client_secret=os.getenv("FASTMCP_SERVER_AUTH_GOOGLE_CLIENT_SECRET") or "",
     base_url="http://localhost:8000",
-    resource_server_url="http://localhost:8000/mcp",
     # redirect_path="/auth/callback",  # Default path - change if using a different callback URL
     # Optional: specify required scopes
     # required_scopes=["openid", "https://www.googleapis.com/auth/userinfo.email"],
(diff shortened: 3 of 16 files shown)


jlowin:
There's a lot of code here, especially to make it as backwards-compatible as possible, but the meat of this change is really https://github.com/jlowin/fastmcp/pull/1706/files#diff-0c4995439b98a89be0d3e44d65ed8979f2726fec15fa69ce0965491f9753dec1R95-R151 and related refactoring

jlowin:
Note - given how critical auth is and how hard it is to test without “real” deployments I’m considering relaxing FastMCP’s breaking change policy for server auth only, to allow auth to evolve rapidly toward a “correct” implementation. This would permit auth modules to have breaking changes at any time to prevent unsecure settings from being maintained. For example this PR would be dramatically smaller and easier. Obviously we would want to stop this practice as quickly as possible so maybe it’s only a 2.12.x policy? @strawgate what do you think?

strawgate:
Yes, should we flag auth modules in the docs as an area that is rapidly evolving and where breaking changes may necessitate pinning to a specific FastMCP version?

jlowin:
Ok i am going to update this PR to be NON backwards compatible and separately PR version guidelines

Issue #1716, opened 2025-09-02
OAuthProxy does not configure well-known routes to expose valid_scopes


dmikusa:
### Enhancement

OAuthProxy presently creates a ClientRegistrationOptions instance with `enabled=True`, but does not pass through any `valid_scopes`. Consequently, the metadata for OAuth well-known endpoints does not include valid scopes.

There are some clients, like Claude, which look at the advertised scopes and use this to set the list of scopes that will be requested for client registration and during authorization. If you want Claude (and possibly others) to request scopes, then these need to be set.

jlowin:
Closed in #1717

PR #1717, opened 2025-09-02
Expose valid scopes from OAuthProxy metadata

dmikusa:
## Description

OAuthProxy presently creates a ClientRegistrationOptions instance with `enabled=True`, but does not pass through and `valid_scopes`. Consequently, the metadata for OAuth well-known endpoints does not include valid scopes.

There are some clients, like Claude, which looks at the advertised scopes and use this to set the list of scopes that will be requested for client registration and during authorization. If you want Claude (and possibly others) to request scopes, then these need to be set.

This PR allows passing in a list of valid scopes when you create the OAuthProxy instance and attaches these to the ClientRegistrationOptions instance that is created. This is all that is required for the metadata to be exposed in the well-known routes.

I thought about setting `valid_scopes` equal to the token verifier's `required_scopes`, but I can think of scenarios where you might want to advertise a larger set of scopes than what is required. For example, if you have read & write as valid scopes. You might only require read scope. The MCP server might then only expose a subset of tools to a user with read-only scope. The way this PR is set up, `valid_scopes` and `required_scopes` are separate so it would accommodate that situation.

**Contributors Checklist**

- [x] My change is related to issue #1716 

**Review Checklist**

---


jlowin:
Thanks @dmikusa - should this default to `required_scopes` if not provided? (or error if the valid scopes are not >= the list of required scopes?)  Is it bad to not have any advertised at all?

dmikusa:
> should this default to required_scopes if not provided?

I had that thought after I submitted the PR. That would probably be a reasonable default, and require less individual configuration. I'll make that change.

> Is it bad to not have any advertised at all?

Hard to say. I think this will vary from client to client. 

I can only tell you what Claude does (from observation only). With Claude, when that list is empty, Claude will request the scope `claudeai`, which is not helpful. For Claude, I'd think having something would be better than nothing.

dmikusa:
Ok, that change has been made.

src/fastmcp/server/auth/oauth_proxy.py
```diff
@@ -263,9 +264,16 @@ def __init__(
...
                 If empty list, all redirect URIs are allowed (not recommended for production).
                 These are for MCP clients performing loopback redirects, NOT for the upstream OAuth app.
+            valid_scopes: List of all the possible valid scopes for a client.
+                These are advertised to clients through the `/.well-known` endpoints.
         """
         # Always enable DCR since we implement it locally for MCP clients
-        client_registration_options = ClientRegistrationOptions(enabled=True)
+        client_registration_options = ClientRegistrationOptions(
+            enabled=True,
+            valid_scopes=valid_scopes is not None
```
jlowin: might be a simpler construction, since it will automatically look through None or [] (which is only valid implicitly if token_verifier.required_scopes is also [])

```suggestion
            valid_scopes=valid_scopes or token_verifier.required_scopes
```

jlowin: (and remove next two lines, not sure how to annotate that as a code suggestion)

src/fastmcp/server/auth/oauth_proxy.py
```diff
@@ -263,9 +264,16 @@ def __init__(
                 If None (default), only localhost redirect URIs are allowed.
                 If empty list, all redirect URIs are allowed (not recommended for production).
                 These are for MCP clients performing loopback redirects, NOT for the upstream OAuth app.
+            valid_scopes: List of all the possible valid scopes for a client.
+                These are advertised to clients through the `/.well-known` endpoints.
```
jlowin:
```suggestion
                These are advertised to clients through the `/.well-known` endpoints. Defaults to `reuqired_scopes` if not provided.
```

jlowin:
Thanks for that update - I think this may need to be exposed on the various integrations (google, github, workos, azure) to be practical though. I'm hesitant to regurgitate all params though, so we can leave it for now if this satisfies your requirement.

src/fastmcp/server/auth/oauth_proxy.py
dmikusa: I did it this way because I believe Python considers an empty list falsey, which would mean you couldn't put an empty list in as the valid_scopes, since an empty list would fail the logic and use the required scopes instead. That was my concern anyway.

Idk if that's a legit requirement, to have an empty list. I guess maybe if you don't want to advertise any scopes? 🤔 

I don't have that need. I'm fine either way.

dmikusa:
> I'm hesitant to regurgitate all params though, so we can leave it for now if this satisfies your requirement.

I'm fine with just this bit. We're not using the other providers at the moment.

src/fastmcp/server/auth/oauth_proxy.py
jlowin: Oh, my thinking is if you have required scopes, then it doesn't make sense to advertise an empty list (so its appropriate to look through it) and if you don't have required scopes, then either the `None` or the `[]` on required_scopes is actually correct for valid scopes. In other words valid scopes must always contain at least your required scopes, no?

dmikusa: Sounds good to me. I'll make that change.

dmikusa: Change made.

jlowin (approved):
Thanks!

PR #1719, opened 2025-09-02
Add development docs for FastMCP

jlowin:
@strawgate would appreciate any input/edits/rewrites/etc here!


docs/development/tests.mdx
```diff
@@ -0,0 +1,320 @@
...
+
+```python
+# Basic assertion - minimal context on failure
+assert result.status == "success"
+
+# Better - explains what was expected
+assert result.status == "success", f"Expected successful operation, got {result.status}: {result.error}"
+```
+
+Try not to have too many assertions in a single test unless you truly need to check various aspects of the same behavior. In general, assertions of different behaviors should be in separate tests.
```
strawgate, line 210: Do we want to mention inline snapshots in here?

strawgate (approved):
Looks great, one comment about maybe documenting inline snapshots

jlowin:
/marvin it look slike we have two broken links 
```
Checking for broken links...

deployment/self-hosted.mdx
  - deployment/testing

development/contributing.mdx
  - development/testing

Found 2 broken links in 7808.02ms
```

I think those moved to development/tests.mdx but please confirm then fix

PR #1714, opened 2025-09-02
OAuthProxy forward PKCE parameters

AaronPowell96:
## Description

**Contributors Checklist**

- [x] My change is related to issue #1713
- [x] I have followed the repository's development workflow
- [x] I have tested my changes manually and by adding relevant tests
- [x] I have performed all required documentation updates

**Review Checklist**

- [x] I have self-reviewed my changes
- [x] My Pull Request is ready for review

---

Closes #1713

PR #1724, opened 2025-09-02
Revert "OAuthProxy forward PKCE parameters"

jlowin:
Reverts jlowin/fastmcp#1714

@AaronPowell96 unfortunately #1714 breaks our existing working integrations with the following error:

[09/02/25 16:38:16] ERROR    IdP token exchange failed: invalid_grant: A code_verifier was not  oauth_proxy.py:858
                             included, but the authorization request included a code_challenge.                   
                             Codes requested with a code_challenge MUST be redeemed with a                        
                             code_verifier.  

Unless/until we find a way to provide the verifier to the upstream (which I think would require rewriting the SDK's token handler, which I'm not sure we want to do), I think we have to revert the change.


AaronPowell96:
@jlowin Thanks for looking into this, sorry it caused a negative side effect! - Are you able to provide a test to reproduce the above issue that surfaced so that I can dig a little deeper? Thanks again

jlowin:
That's quite all right, I wanted #1714 to work because it makes sense! I just didn't realize it was so tightly coupled to the verifier.

There are a couple of server / client pairs for common IDPs in the examples folder (for instance, here is GitHub: https://github.com/jlowin/fastmcp/tree/main/examples/auth/github_oauth). I periodically run the server and client to do a round-trip integration test, which is what failed above. Unfortunately like all things OAuth these are difficult to test in CI, though we *do* have the secrets available in CI for a GitHub OAuth app. I recommend GitHub here because it is IMO the easiest of the providers to set up and test.

Issue #1741, opened 2025-09-03
Supported scopes not in sync across oauth-authorization-server and oauth-protected-resource endpoints


dmikusa:
### Description

With #1717, the `.well-known/oauth-authorization-server` was enabled to list a set of supported scopes, not just the scopes that are required (required scopes can be a "minimal" set, while supported scopes may be a larger set as some may be optional). 

However, in that PR I forgot to update `.well-known/oauth-protected-resource` which also contains supported scopes. Thus, the advertised supported scopes are different between the two endpoints. I believe these should be kept in sync, otherwise, it would be confusing to a client that looks at both endpoints.

### Example Code

```Python

```

### Version Information

```Text
main branch
```

dmikusa:
> appears to be a duplicate of itself, which suggests it might be a redundant filing.

Lol. This is a unique issue.

jlowin:
😂

Closed in #1742!

PR #1742, opened 2025-09-03
Protected resource advertises valid scopes, if available

dmikusa:
## Description

With #1717, the `.well-known/oauth-authorization-server` was enabled to list a set of supported scopes, not just the scopes that are required (required scopes can be a "minimal" set, while supported scopes may be a larger set as some may be optional).

This PR follows up on #1717 and enables the same support for `.well-known/oauth-protected-resource`. When creating this resource route, we pulls the valid_scopes from the ClientRegistrationOptions object, if it's set. If it's not set, we fall back to required_scopes, which was the previous behavior.

This keeps the supported scopes list in sync across the two endpoints.

**Contributors Checklist**
- [x] My change is related to issue #1741 

**Review Checklist**

---


jlowin (approved):
Thank you!

PR #1750, opened 2025-09-04
Fix settings module import issue in VS Code debugger

### LINKED ISSUE #1749: AttributeError when running FastMCP server with VS Code debugger
### Description

# AttributeError when running FastMCP server with VS Code debugger

## Bug Description

FastMCP server fails to start **ONLY when run through VS Code debugger** with `AttributeError: module 'fastmcp.settings' has no attribute 'mask_error_details'`, while working perfectly with direct Python execution.

## Environment

- **FastMCP Version**: 2.12.2
- **Python Version**: 3.13.7
- **OS**: macOS (Darwin)
- **IDE**: VS Code with Python extension
- **Debugger**: debugpy

### Code that triggers the issue:
```python
from fastmcp import FastMCP

auth_provider = GoogleProvider(
    client_id="your_client_id",
    client_secret="your_client_secret",
    required_scopes=[
        "openid",
        "https://www.googleapis.com/auth/userinfo.email",
    ],
)

mcp = FastMCP("My MCP Server", auth=auth_provider)  # ← Error occurs here ONLY during debugging
```

### ✅ Works (Direct execution)
```bash
python server.py
# Server starts successfully with GoogleProvider
```

### ❌ Fails (VS Code debugger only)
```json
// .vscode/launch.json
{
    "name": "FastMCP Debug Server",
    "type": "debugpy", 
    "request": "launch",
    "program": "${workspaceFolder}/server.py",
    "justMyCode": false
}
```

### Execution Path Differences

**Direct execution**:
```
python server.py → __main__ → sequential module imports → settings fully initialized
```

**VS Code debugger execution**:
```
debugpy/launcher → debugpy/__main__ → runpy.run_path() → server.py
```

### Technical Differences in Debugger Environment

1. **Import Hook Interference**: 
   - Debugger installs Python import hooks to collect debugging information
   - These hooks alter module loading timing and sequence
   - Can cause race conditions in module initialization

2. **Module Reloading Behavior**:
   - Debugger supports hot reloading and breakpoint injection
   - May reload or defer initialization of certain modules
   - Global objects like `settings` may not be fully initialized when accessed

3. **Execution Context Changes**:
   - `runpy.run_path()` creates different execution context than direct `__main__`
   - Module `__name__` resolution and import paths behave differently
   - Can affect circular import detection and resolution

4. **Dependency Loading Order**:
   - GoogleProvider imports additional OAuth dependencies
   - Debugger's import hooks may alter the order these dependencies are loaded
   - Creates timing windows where `settings` object is partially initialized

### Verified Technical Analysis

- ✅ `mask_error_details` attribute EXISTS in Settings class (settings.py:232-246)
- ✅ Direct import test works: `python -c "import fastmcp.settings; print(fastmcp.settings.mask_error_details)"` → `True`
- ✅ Normal execution: settings object fully initialized before ToolManager access
- ❌ **Debugger only**: settings object accessed before complete initialization due to altered import sequence

### Key Error Location
**File**: `fastmcp/tools/tool_manager.py:36`
**Code**: `self.mask_error_details = mask_error_details or settings.mask_error_details`

### Root Cause
The debugger environment's modified import behavior creates a **race condition** where:
1. `FastMCP.__init__` creates `ToolManager`
2. `ToolManager.__init__` tries to access `settings.mask_error_details`
3. But the global `settings` object (created in `fastmcp/__init__.py:8`) isn't fully initialized yet due to debugger's altered module loading sequence

## Impact

- **Severity**: High - Completely breaks debugging workflow for FastMCP servers using authentication
- **Scope**: All VS Code users debugging FastMCP servers with GoogleProvider (or potentially other auth providers)
- **Workaround**: Use direct Python execution instead of debugger (loses debugging capabilities)

## Additional Context

- Issue is **environment-specific**: only affects debugger execution
- Deprecation warning appears during debugging, suggesting debugger triggers deprecated code paths
- Problem likely affects other auth providers, not just GoogleProvider
- Significantly impacts developer experience for FastMCP authentication development

### Example Code

```Python
// .vscode/launch.json
{
    "version": "0.2.0",
    "configurations": [
        {
            "name": "FastMCP Debug",
            "type": "debugpy",
            "request": "launch",
            "program": "${workspaceFolder}/server.py"
        }
    ]
}



// server.py
from fastmcp import FastMCP

auth_provider = GoogleProvider(
    client_id="your_client_id",
    client_secret="your_client_secret",
    required_scopes=[
        "openid",
        "https://www.googleapis.com/auth/userinfo.email",
    ],
)

mcp = FastMCP("My MCP Server", auth=auth_provider)

@mcp.tool
def hello(): 
    return "hi"

if __name__ == "__main__":
    mcp.run(
        transport="http",
        host="127.0.0.1",
        port=8000,
    )
```

### Version Information

```Text
FastMCP version:                                                                 2.12.2
MCP version:                                                                     1.13.1
Python version:                                                                  3.13.7
Platform:                                             macOS-15.5-arm64-arm-64bit-Mach-O
FastMCP root path: /Users/user/test/.venv/lib/python3.13/site-packages
```

Jay-flow:
# Fix settings module import issue in VS Code debugger

This is the issue resolution PR below.
Fixes #1749

## Problem

When running FastMCP servers through VS Code debugger, the following error occurs:
```
AttributeError: module 'fastmcp.settings' has no attribute 'mask_error_details'
```

This error only happens in debugger environments, not during normal Python execution.

### The Name Conflict
Python allows the same name to exist as both:
- A module: `fastmcp/settings.py` 
- A variable: `settings = Settings()` in `fastmcp/__init__.py`

When code executes `from fastmcp import settings`, Python must resolve which one to return.

### Normal Execution Behavior
1. Python checks `fastmcp/__init__.py` namespace first
2. Finds the `settings` variable (Settings instance)
3. Returns the instance → Everything works ✅

### VS Code Debugger Behavior  
1. VS Code uses `debugpy` which installs import hooks for debugging features
2. These hooks modify Python's import resolution priority
3. The module path `fastmcp/settings.py` gets higher priority
4. Returns the module instead of the instance
5. Module has no `mask_error_details` attribute → AttributeError ❌

### Why This Happens
`debugpy` needs to intercept imports to:
- Set breakpoints in modules before they're imported
- Track variable values across module boundaries
- Enable step-through debugging

This legitimate debugging requirement inadvertently changes import resolution order.

### Implementation
```python
# src/fastmcp/__init__.py
from fastmcp.settings import Settings
import fastmcp.settings as settings_module

settings = Settings()

# Fix VS Code debugger compatibility
settings_module._instance = settings
```

```python
# src/fastmcp/settings.py
# Injected by __init__.py
_instance = None

def __getattr__(name: str):
    """
    Used to deprecate the module-level Image class; can be removed once it is no longer imported to root.
    """
    if name == "settings":
        # ... existing deprecation logic ...
        return settings
    
    # Proxy to injected instance (debugger fix)
    if _instance is not None and hasattr(_instance, name):
        return getattr(_instance, name)

    raise AttributeError(f"module '{__name__}' has no attribute '{name}'")
```

### How This Solves the Problem

1. **Normal Execution Path**
   - `from fastmcp import settings` → Gets Settings instance from `__init__.py`
   - Direct attribute access works normally

2. **Debugger Execution Path**  
   - `from fastmcp import settings` → Gets `settings.py` module
   - Module's `__getattr__` is called for `mask_error_details`
   - `__getattr__` proxies to the injected `_instance`
   - Returns the same value as normal execution

3. **Single Source of Truth**
   - Only one Settings instance exists (created in `__init__.py`)
   - Both import methods access the same instance
   - Consistent behavior regardless of execution environment

### Why This is the Best Solution

1. **No Breaking Changes**: All existing imports continue to work
2. **No Circular Dependencies**: Instance is injected after creation
3. **Minimal Code Changes**: Only touches two files with surgical precision
4. **Future Proof**: Automatically handles any new settings attributes
5. **Clean Architecture**: Clear initialization flow from `__init__.py`
6. **Performance**: No overhead in normal execution path

## Testing

Verified that both import methods now work correctly:
```python
# Both work in debugger now
import fastmcp
fastmcp.settings.mask_error_details  # ✅

import fastmcp.settings as settings_module  
settings_module.mask_error_details  # ✅
```

The fix ensures FastMCP servers can be debugged in VS Code without any AttributeError.

### DIFF
--- a/src/fastmcp/__init__.py
+++ b/src/fastmcp/__init__.py
@@ -3,9 +3,13 @@
 import warnings
 from importlib.metadata import version as _version
 from fastmcp.settings import Settings
+import fastmcp.settings as settings_module
 from fastmcp.utilities.logging import configure_logging as _configure_logging
 
 settings = Settings()
+
+settings_module._instance = settings
+
 if settings.log_enabled:
     _configure_logging(
         level=settings.log_level,
--- a/src/fastmcp/settings.py
+++ b/src/fastmcp/settings.py
@@ -344,6 +344,9 @@ class Settings(BaseSettings):
     ] = False
 
 
+_instance = None
+
+
 def __getattr__(name: str):
     """
     Used to deprecate the module-level Image class; can be removed once it is no longer imported to root.
@@ -360,5 +363,9 @@ def __getattr__(name: str):
                 stacklevel=2,
             )
         return settings
+
+    # Proxy to injected instance (debugger fix)
+    if _instance is not None and hasattr(_instance, name):
+        return getattr(_instance, name)
 
     raise AttributeError(f"module '{__name__}' has no attribute '{name}'")


jlowin:
Closing this until we have an MRE in #1749

Issue #1746, opened 2025-09-03
ToolCallResult: StructuredContent w/ AudioContent


strawgate:
### Description

Right now if a response includes an AudioContent the `content` field of the ToolCallResult is an array with an AudioContent in it

But the structuredcontent field and data fields are populated with a dict dump of the AudioContent object

```
CallToolResult(
    content=[
        AudioContent(
            type='audio',
            data='ZmFrZV9hdWRpb19kYXRh',
            mimeType='audio/mpeg',
            annotations=None,
            meta=None
        )
    ], 
    structured_content={
        'type': 'audio',
        'data': 'ZmFrZV9hdWRpb19kYXRh',
        'mimeType': 'audio/mpeg',
        'annotations': None,
        '_meta': None
    }, 
    data={
        'type': 'audio',
        'data': 'ZmFrZV9hdWRpb19kYXRh',
        'mimeType': 'audio/mpeg',
        'annotations': None,
        '_meta': None
    },
    is_error=False
)
```

I guess I would only expect structured content to be populated if the response is actually an object/json

### Example Code

```Python

```

### Version Information

```Text
latest
```

strawgate:
Boo

jlowin:
wait you thumbs down'd this, no?

/marvin look into the deduplication action and see if we made a mistake in coding it

strawgate:
I don't remember the exact order of events I may have thumbs downed it after Marvin closed the issue on me 😅

PR #1773, opened 2025-09-06
Cleanup Tool Content Conversion

strawgate:
## Description
re: https://github.com/jlowin/fastmcp/issues/1746 and https://github.com/jlowin/fastmcp/issues/1656

1. If the user returns any native MCP types, we do not merge content blocks, this means `["text", Image, "text", "text"]` will be an array of 4 content blocks.
2. If the user does not return any native MCP Types, we merge content blocks together, this means: `["text1", "text2"]` becomes `TextContent(text='["text1", "text2"])'`
3. Revamp the tests for content conversion
4. Prevent creating structured content out of MCP types if there is no output schema defined.

**Contributors Checklist**
- [x] My change is related to issue [#(issue number)](https://github.com/jlowin/fastmcp/issues/1656)

**Review Checklist**

---


strawgate:
@jlowin should be ready for review 😅

src/fastmcp/tools/tool.py
```diff
@@ -476,98 +486,69 @@ def from_function(
...
+
+
+def _convert_to_single_content_block(
+    item: Any,
+    serializer: ToolResultSerializerType | None = None,
+) -> ContentBlock:
+    if isinstance(item, ContentBlock):
+        return item
+
+    if isinstance(item, Image):
```
jlowin, line 514: It's not critical because who knows how many more we'll have but this does make me thing all of these should have inheritied from a common base class with `to_mcp_content()` as the only method. Would make detecting and converting easier. To be clear, just a thought, doesn't impact this PR

jlowin (approved):
LGTM!

src/fastmcp/tools/tool.py
strawgate: I had the same exact thought but had already touched more than I expected to with this PR

Issue #1670, opened 2025-08-29
Add completion/autocomplete support for resources and prompts


JayDoubleu:
## Summary

The MCP protocol supports completion requests via `CompleteRequest`/`CompleteResult` messages, allowing servers to provide autocomplete suggestions for resources and prompts. While FastMCP's client already supports sending completion requests via `complete()` and `complete_mcp()` methods, servers cannot currently register handlers to provide custom completion suggestions.

## Problem Statement

Currently, FastMCP servers have no way to:
- Register completion handlers for resources and prompts
- Provide autocomplete suggestions for resource templates with parameters
- Offer context-aware completions based on partial input

This is a missing piece of the MCP specification that would greatly enhance the developer experience when working with complex resource templates or prompts with multiple arguments.

### Resource Templates with Parameters
```python
@mcp.resource("devops://wiki/{organization}/{project}/{wiki_name}/{path*}")
async def wiki_resource(organization: str, project: str, wiki_name: str, path: str):
    # Resource implementation
    pass
```

Users could benefit from autocomplete suggestions for:
- `organization`: List of available organizations
- `project`: Projects filtered by selected organization  
- `wiki_name`: Wiki names within the selected project
- `path`: File paths within the wiki

### Dynamic File Path Completion
```python
@mcp.resource("file:///{path*}")
async def file_resource(path: str):
    # File resource implementation
    pass
```

The server could provide file path suggestions based on the partial path entered.

### Current State
- **Client Support**: ✅ Already implemented (`client.complete()`, `client.complete_mcp()`)
- **Server Handler Registration**: ❌ Missing
- **MCP Protocol Support**: ✅ Protocol defines `completion/complete` handler
- **Handler Pattern**: Exists for tools, resources, prompts in `_setup_handlers()`

### What's Missing
The `FastMCP._setup_handlers()` method registers handlers for:
- `list_tools()`
- `list_resources()`
- `list_resource_templates()`
- `list_prompts()`
- `call_tool()`
- `read_resource()`
- `get_prompt()`

But **not** for `completion()`.

### 1. Add Completion Handler Registration

In `src/fastmcp/server/server.py`, update `_setup_handlers()`:

```python
def _setup_handlers(self) -> None:
    """Set up core MCP protocol handlers."""
    self._mcp_server.list_tools()(self._mcp_list_tools)
    self._mcp_server.list_resources()(self._mcp_list_resources)
[...]

PR #1776, opened 2025-09-07
Add MCP completion utility support

rsp2k:
# Add MCP completion utility support

This PR implements the completion utility from the MCP specification, addressing issue #1670. The completion utility provides IDE-like autocomplete suggestions for resource templates and prompts.

While FastMCP's client already supported sending completion requests, servers had no way to register handlers to provide custom completion suggestions.

This implementation adds a clean `@server.completion()` decorator that makes it trivial to add autocomplete functionality to any FastMCP server.

## Usage Example

```python
from fastmcp import FastMCP

server = FastMCP("My Server")

@server.prompt
async def analyze_data(dataset: str, analysis_type: str) -> str:
    return f"Analyzing {dataset} with {analysis_type}"

# Add completions for each argument
@server.completion("prompt", "analyze_data", "dataset")
async def complete_dataset(partial: str) -> list[str]:
    datasets = ["customers", "products", "sales", "inventory"]
    return [d for d in datasets if d.startswith(partial)]

@server.completion("prompt", "analyze_data", "analysis_type") 
async def complete_analysis_type(partial: str) -> list[str]:
    types = ["statistical", "trend", "comparative", "predictive"]
    return [t for t in types if t.startswith(partial)]
```
## Demo
Demo: `examples/completion_demo.py`

## MCP SDK Approach

```python
# MCP SDK approach - complex manual routing
@mcp.completion()
async def handle_completion(ref, argument, context):
    if isinstance(ref, PromptReference):
        if ref.name == "analyze_data" and argument.name == "dataset":
            # handle dataset completion
    if isinstance(ref, ResourceTemplateReference):
        if ref.uri == "file:///{path}" and argument.name == "path":
            # handle path completion
    return None
```

### FastMCP's approach

```python
# FastMCP approach - automatic routing, focused handlers
@server.completion("prompt", "analyze_data", "dataset")
async def complete_dataset(partial: str) -> list[str]:
    # focused handler just for dataset completion

@server.completion("resource_template", "file:///{path}", "path")
async def complete_path(partial: str) -> list[str]:
    # focused handler just for path completion
```

## Test Coverage

All existing tests pass with no regressions.

### New Tests:
**Server tests** (11):
- Handler registration and routing
- Prompt and resource template completion
- 100-item limit enforcement
- Error handling (exceptions, null returns)
- Edge cases (unicode, long strings, empty results)
- Multiple handlers per server

**Client tests** (13):
- Both `client.complete()` and `client.complete_mcp()` methods
- Partial matching and filtering
- Non-existent handlers and arguments
- Case-insensitive matching
- Large result sets and unicode support

## Files Changed

- `src/fastmcp/server/server.py`: Core completion implementation (123 lines added)
- `tests/server/test_completion.py`: Server-side test suite
- `tests/client/test_completion.py`: Client-side test suite  
- `examples/completion_demo.py`: Working demonstration server

## Backward Compatibility

Fully backward compatible:
- No changes to existing APIs
- Completion support is optional
- Servers without completion handlers return empty results

### DIFF
--- a/examples/completion_demo.py
+++ b/examples/completion_demo.py
@@ -1,0 +1,203 @@
+#!/usr/bin/env python3
+"""
+Demo server showcasing FastMCP completion functionality.
+
+This example demonstrates how to use the @completion decorator to provide
+autocomplete suggestions for resource templates and prompts.
+"""
+
+import asyncio
+import os
+from typing import Any
+
+from fastmcp import FastMCP
+from fastmcp.resources import ResourceTemplate
+
+# Create the FastMCP server
+server = FastMCP("Completion Demo Server")
+
+
+# Example 1: File path completion for a resource template
+file_template = ResourceTemplate(
+    uri_template="file:///{path}",
+    name="File Resource",
+    description="Access files on the local system",
+    parameters={}
+)
+server.add_template(file_template)
+
+
+@server.completion("resource_template", "file:///{path}", "path")
+async def complete_file_path(partial: str) -> list[str]:
+    """Provide file path completions based on partial input."""
+    # In a real implementation, you'd scan the filesystem
+    # For demo purposes, we return some example files
+    example_files = [
+        "config.json",
+        "data.csv", 
+        "readme.txt",
+        "script.py",
+        "styles.css",
+        "index.html",
+        "package.json",
+        "requirements.txt"
+    ]
+    
+    if not partial:
+        return example_files
+    
+    # Return files that start with the partial input
+    return [f for f in example_files if f.lower().startswith(partial.lower())]
+
+
+# Example 2: Prompt with multiple completion handlers
+@server.prompt
+async def analyze_data(dataset: str, analysis_type: str, output_format: str = "json") -> str:
+    """Analyze a dataset with the specified analysis type and output format."""
+    return f"Analyzing {dataset} using {analysis_type} analysis, outputting as {output_format}"
+
+
+@server.completion("prompt", "analyze_data", "dataset")
+async def complete_dataset(partial: str) -> list[str]:
+    """Provide dataset name completions."""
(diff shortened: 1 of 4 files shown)


jlowin:
@rsp2k I appreciate the proactivity here but this needs to be designed, preferably in an issue, before opening a feature PR.

I don't think this is a great DX for users. It introduces a tight coupling between the string arguments to this decorator and the names of components and arguments elsewhere. For that reason alone I think it opens up a world of footguns without a significant improvement over what the core SDK already offers. 

I would expect a FastMCP implementation to introduce a first class, observable `Completion` component that reflects the utility of completions in the spec itself. Moreover, it has to be compatible with all other FastMCP first-class features; at a glance I don't think this would survive mounting a server (nor the prefixing manipulation that feature involves). 

Off the top of my head, I would embrace rather than rely implicitly on the tight coupling that Completions appear to have with prompts and resource templates. For example, are completions a new property of the resource/prompt itself? or of their respective managers? It seems that completions are an attempt to dodge the challenge that template and prompt arguments are always strings, so could we contemplate a DX like this:

```python
@mcp.prompt
def my_prompt(a: Annotated[Literal['foo', 'bar'], FuzzyCompletion()]):
    ...
```

To be clear, I have a lot of issues with the above and am not suggesting that we'd accept it, but at least it avoids the duplicity of users having to specify their arguments. Before we can think about this type of sugar, we need to understand the mechanism that binds completions to their respective components. To me that's the core difference between the SDK and a framework like FastMCP. The SDK is for adding completions in any manner you want; FastMCP has an opinion (here, that completions are attached directly to the arguments they reference).

In any case this is the sort of design, experimentation, and contemplation I would look for to be convinced this is a valuable addition to the framework, which is best done by making a strong case in the issue.

PR #1779, opened 2025-09-07
Automatically Create inline Snapshots

strawgate:
## Description
Agents and contributing docs both imply inline snapshots are created automatically when running pytest.

Let's make it so


jlowin (approved):
🙌

PR #1791, opened 2025-09-09
Use lowercase namespace for fastmcp logger

### LINKED ISSUE #1790: Change FastMCP logger namespace from "FastMCP" to "fastmcp"
### Enhancement

Currently the logging module uses "FastMCP.{name}" to namespace FastMCP loggers, which is nonstandard. It should match the package name and be lowercase: "fastmcp.{name}". 

This might constitute a breaking change so it should either be done in the 2.13 release OR we should create both (which could have adverse affects). (Note: investigate this, it might be considered a bugfix under all circumstances)

Note that the TimingMiddleware currently uses the lowercase namespace.

jlowin:
Closes #1790


src/fastmcp/utilities/logging.py
```diff
@@ -16,7 +16,7 @@ def get_logger(name: str) -> logging.Logger:
     Returns:
         a configured logger instance
     """
-    return logging.getLogger(f"FastMCP.{name}")
+    return logging.getLogger(f"fastmcp.{name}")
```
strawgate, line 22: Is there a reason to do this vs just doing getLogger(__name__) or whatever in each file?

jlowin: The only goal was to ensure that everything is namespaced if you use the fastmcp utility to get your logger no matter what you actually name it (__name__ being another way to do that by convention, but not guaranteed if someone wanted to use a fastmcp logger in another project for some reason).

strawgate: Do we want other projects to use a FastMCP namespaced logger?

I'm not super familiar with libraries offering this in general, I can imagine it being nice to have your logs look the same as the ones coming from the FastMCP server? I've certainly tried to do that in the past

For users who want their logging to look the same as fastmcp's logging would it make sense to just encourage them to use a configure_logger helper function or something similar?

jlowin: I think people who author servers *may* want their logs to look/feel like fastmcp logs (and be controlled by the same verbosity settings) which is why having a convenience utility function that ensures you get a properly namespaced/formatted logger is helpful

PR #1796, opened 2025-09-09
Allow configuring FastMCP transport setting in the same way as other configuration

### LINKED ISSUE #1795: Further allow configuring FastMCP through environment variables
### Enhancement

While many options are already configurable through environment variables, some are still missing.

Particularly 'transport' is unavailable, rendering the other options useless in some use cases: you can set FASTMCP_HOST and FASTMCP_PORT, but as the default transport setting is 'stdio' and there is no FASTMCP_TRANSPORT option, setting the host and port in environment variables is useless.

jvdmr:
\`run_async()\` hardcoded \`"stdio"\` as the fallback when no transport was specified, which meant users running third-party servers (like AWS MCP servers in ECS containers) had no way to change the default transport without modifying source code. The CLI and \`fastmcp.json\` provided workarounds, but a bare \`mcp.run()\` call was locked to stdio.

This adds a \`transport\` field to \`Settings\` so that \`FASTMCP_TRANSPORT=http\` (or any valid transport) controls the default. \`run_async\` reads from the setting instead of hardcoding \`"stdio"\`, while explicit \`transport=\` arguments and CLI \`--transport\` flags still override it. Fits naturally alongside the existing \`FASTMCP_HOST\` and \`FASTMCP_PORT\` settings.

\`\`\`python
# Third-party server code you can't modify:
mcp.run()  # reads FASTMCP_TRANSPORT from env

# Or in your own code:
import fastmcp
fastmcp.settings.transport = "http"
\`\`\`

Closes #1795


jlowin:
@jvdmr now that #3148 is in, we can get this merged - apologies for the significant delay. This will go out after 3.0 releases.

PR #1805, opened 2025-09-10
Refactor Logging and Structured Logging Middleware

### LINKED ISSUE #1801: Consolidate Logging and StructuredLogging Middleware
Logging and Structured Logging Middleware should only differ in how the actual message is formatted.

strawgate:
Consolidate the Logging and Structured Logging Middleware to share a common implementation and differ only in how the actual message is formatted.

Continue to maintain both classes to make this hopefully not a breaking change.

Also add payload size and payload token estimation (length // 4)

Fixes #1801

### DIFF
--- a/src/fastmcp/server/middleware/logging.py
+++ b/src/fastmcp/server/middleware/logging.py
@@ -43,6 +43,8 @@ class LoggingMiddleware(Middleware):
         max_payload_length: int = 1000,
         methods: list[str] | None = None,
         payload_serializer: Callable[[Any], str] | None = None,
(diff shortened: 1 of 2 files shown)


strawgate:
You're thinking too small

jlowin:
Do the middleware docs/examples need any updates, just realized

strawgate:
@jlowin nothing externally facing changes with this other than adding `include_payload_length`, i think the api docs will get auto updated? but i dont think there are any existing examples that would need to be updated

PR #1810, opened 2025-09-10
Fix Context docstring examples to use async syntax

### LINKED ISSUE #1809: Context is async or not ? Documentation and docstring in conflict
### Enhancement

Hi, 

While the documentation states that Context should be called in async ==> https://gofastmcp.com/servers/context

The docstring present sync examples ==> https://github.com/jlowin/fastmcp/blob/8eadaaa958b87457142c3b3e9a5163378523fa95/src/fastmcp/server/context.py#L88

Is there something I miss or is there a conflict in docs ?

strawgate:
Fixes #1809

Updates Context docstring examples to use async functions with proper await syntax. The previous sync examples were misleading since Context methods like `info()`, `read_resource()`, and `report_progress()` are async and need to be awaited.

### DIFF
--- a/src/fastmcp/server/context.py
+++ b/src/fastmcp/server/context.py
@@ -261,7 +261,7 @@ class Context:
         Example:
             ```python
             @server.tool
-            async def store_data(data: dict, ctx: Context) -> str:
+            def store_data(data: dict, ctx: Context) -> str:
                 session_id = ctx.session_id
                 redis_client.set(f"session:{session_id}:data", json.dumps(data))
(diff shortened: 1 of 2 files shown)


src/fastmcp/server/context.py
```diff
@@ -261,7 +261,7 @@ def session_id(self) -> str:
         Example:
             ```python
             @server.tool
-            def store_data(data: dict, ctx: Context) -> str:
+            async def store_data(data: dict, ctx: Context) -> str:
```
strawgate: I don't think this needs to be async

Issue #1811, opened 2025-09-11
get http header from AuthProvider


akkuman:
### Enhancement

The current server code cannot meet the following requirements:
1. The client includes the x-endpoint-url header as the authentication address when making a request (the server will verify whether the address is in the whitelist, so security issues can be ignored)
2. Get the token from the Authorization header and then use x-endpoint-url to make an authentication request

First, since the auth middleware is before entering the mcp.tool handler, fastmcp.server.dependencies.get_http_headers() cannot be used
I looked at the code and the current implementation is here and found that there is a request context

https://github.com/jlowin/fastmcp/blob/8eadaaa958b87457142c3b3e9a5163378523fa95/src/fastmcp/server/http.py#L61-L64

However, since this middleware is placed after the auth middleware, I cannot get the request information through _current_http_request in AuthProvider, so this requirement cannot be met.

https://github.com/jlowin/fastmcp/blob/8eadaaa958b87457142c3b3e9a5163378523fa95/src/fastmcp/server/http.py#L74-L118

At the same time, since the middleware passed to FastMCP (middleware=) will be placed after the Auth middleware, I cannot inject my own middleware to meet this requirement.

I currently use monkey patching to achieve this requirement.

```python

current_http_request: ContextVar[Request | None] = ContextVar(
    "current_http_request",
    default=None,
)


class PlatformBaseURLException(Exception):
    pass


@contextmanager
def set_http_request(request: Request) -> Generator[Request, None, None]:
    token = current_http_request.set(request)
    try:
        yield request
    finally:
        current_http_request.reset(token)


class CustomRequestContextMiddleware:
    """
    ref：fastmcp.server.http.RequestContextMiddleware
    """

    def __init__(self, app: ASGIApp):
        self.app = app

    async def __call__(self, scope: Scope, receive: Receive, send: Send):
        if scope["type"] == "http":
            with set_http_request(Request(scope)):
                await self.app(scope, receive, send)
        else:
            await self.app(scope, receive, send)


def create_base_app(
    routes: list[BaseRoute],
    middleware: list[Middleware],
    debug: bool = False,
    lifespan: Callable | None = None,
) -> StarletteWithLifespan:
    # Always add RequestContextMiddleware as the outermost middleware
    middleware.append(Middleware(RequestContextMiddleware))

    middleware.insert(0, Middleware(CustomRequestContextMiddleware))

    return StarletteWithLifespan(
        routes=routes,
        middleware=middleware,
[...]

PR #1812, opened 2025-09-11
feat: Move the Starlette context middleware to the front

akkuman:
Move the Starlette context middleware to the front to facilitate subsequent middleware to obtain request information

related issue: https://github.com/jlowin/fastmcp/issues/1811

Closes 1811

### DIFF
--- a/src/fastmcp/server/http.py
+++ b/src/fastmcp/server/http.py
@@ -3,7 +3,7 @@ from __future__ import annotations
 from collections.abc import AsyncGenerator, Callable, Generator
 from contextlib import asynccontextmanager, contextmanager
 from contextvars import ContextVar
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Union
 
 from mcp.server.auth.middleware.bearer_auth import RequireAuthMiddleware
 from mcp.server.lowlevel.server import LifespanResultT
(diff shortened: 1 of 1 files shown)


src/fastmcp/server/http.py
```diff
@@ -79,6 +79,13 @@ def set_http_request(request: Request) -> Generator[Request, None, None]:
         _current_http_request.reset(token)
 
 
+def get_current_http_request() -> Union[Request, None]:
```
jlowin: This can be deleted I think

akkuman: The main reason for adding this function is that `_current_http_request` starts with an underscore, which generally indicates that it is a private variable. It may be more appropriate to expose it through a getter function.

jlowin: Right but this function isn't actually used anywhere, so adding it here doesn't assist in any functionality and just muddies the space of this file (which is not intended to be user accessed). It could replace the existing implementation in dependencies.py (which is user facing) but I'd say that could be a separate PR since its a different enhancement.

akkuman: Ok, I will delete this function

jlowin (approved):
Thanks!

Issue #1816, opened 2025-09-12
Add support for OIDC configuration based authentication providers


ruhulio:
### Enhancement

`OAuthProxy` configuration has many parameters that can be automatically set by using the dynamic values provided by the OIDC configuration URL that many providers support, e.g. `https://provider.com/.well-known/openid-configuration`.

Adding support for this can help simplify integrating more providers.

jlowin:
Closed in #1817!

PR #1817, opened 2025-09-12
feat: Add support for OIDC configuration

ruhulio:
## Description

Add support for OIDC configuration to simplify integrating providers that support it.

The `OIDCProxy` class wraps `OAuthProxy` to use the dynamic values available from provider's configuration URL for the client application.

This also includes an `Auth0` implementation that uses `OIDCProxy`.

I did my best to follow existing patterns in the codebase and am happy to make any suggested changes.

I did not update the `Python SDK` docs as I believe that is done automagically from the committed code. I also set the `VersionBadge` values to `2.13.0` as I am assuming if this is accepted it would be for a larger release.

**Contributors Checklist**

- [x] My change is related to issue #1816 

**Review Checklist**

---

### DIFF
--- a/docs/integrations/auth0.mdx
+++ b/docs/integrations/auth0.mdx
@@ -8,7 +8,7 @@ tag: NEW
 
 import { VersionBadge } from "/snippets/version-badge.mdx"
 
-<VersionBadge version="2.13.0" />
+<VersionBadge version="2.12.4" />
 
 This guide shows you how to secure your FastMCP server using **Auth0 OAuth**. While Auth0 does have support for Dynamic Client Registration, it is not enabled by default so this integration uses the [**OIDC Proxy**](/servers/auth/oidc-proxy) pattern to bridge Auth0's dynamic OIDC configuration with MCP's authentication requirements.
 
@@ -155,8 +155,6 @@ The client caches tokens locally, so you won't need to re-authenticate for subse
 
 ## Environment Variables
 
-<VersionBadge version="2.13.0" />
-
 For production deployments, use environment variables instead of hardcoding credentials.
 
 ### Provider Selection
@@ -175,15 +173,15 @@ These environment variables provide default values for the Auth0 provider, wheth
 
 <Card>
 <ParamField path="FASTMCP_SERVER_AUTH_AUTH0_CONFIG_URL" required>
-Your Auth0 Application Client ID (e.g., `Ov23liAbcDefGhiJkLmN`)
+Your Auth0 Application Configuration URL (e.g., `https://.../.well-known/openid-configuration`)
 </ParamField>
 
 <ParamField path="FASTMCP_SERVER_AUTH_AUTH0_CLIENT_ID" required>
-Your Auth0 Application Client ID (e.g., `Ov23liAbcDefGhiJkLmN`)
+Your Auth0 Application Client ID (e.g., `tv2ObNgaZAWWhhycr7Bz1LU2mxlnsmsB`)
 </ParamField>
 
 <ParamField path="FASTMCP_SERVER_AUTH_AUTH0_CLIENT_SECRET" required>
-Your Auth0 Application Client Secret
+Your Auth0 Application Client Secret (e.g., `vPYqbjemq...`)
 </ParamField>
 
 <ParamField path="FASTMCP_SERVER_AUTH_AUTH0_AUDIENCE" required>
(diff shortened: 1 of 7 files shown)


src/fastmcp/server/auth/providers/auth0.py
```diff
@@ -0,0 +1,171 @@
...
+            "config_url": settings.config_url,
+            "client_id": settings.client_id,
+            "client_secret": settings.client_secret.get_secret_value(),
+            "audience": settings.audience,
+            "base_url": settings.base_url,
+            "required_scopes": auth0_required_scopes,
+            "allowed_client_redirect_uris": settings.allowed_client_redirect_uris,
+        }
+
+        if settings.redirect_path:
```
jlowin, line 162: I think this might not be exactly correct but I'm not sure of the intended outcome. Could it go in the init_kwargs dict above as `"redirect_path": settings.redirect_path`? Why if there's a redirect path in settings are we using the passed one?

docs/servers/auth/oidc-proxy.mdx
```diff
@@ -0,0 +1,185 @@
...
+  in OAuth Proxy, update your provider's redirect URI accordingly.
+</Warning>
+
+### Basic Setup
+
+Here's how to implement OIDC Proxy with any provider:
+
+```python
+from fastmcp import FastMCP
+from fastmcp.server.auth import OIDCProxy
```
jlowin:
```suggestion
from fastmcp.server.auth.oidc_proxy import OIDCProxy
```

docs/integrations/auth0.mdx
```diff
@@ -0,0 +1,233 @@
...
+</ParamField>
+</Card>
+
+### Auth0-Specific Configuration
+
+These environment variables provide default values for the Auth0 provider, whether it's instantiated manually or configured via `FASTMCP_SERVER_AUTH`.
+
+<Card>
+<ParamField path="FASTMCP_SERVER_AUTH_AUTH0_CONFIG_URL" required>
+Your Auth0 Application Client ID (e.g., `Ov23liAbcDefGhiJkLmN`)
```
jlowin: I think this is a copy/paste issue from below

docs/integrations/auth0.mdx
```diff
@@ -0,0 +1,233 @@
...
+title: Auth0 OAuth 🤝 FastMCP
+sidebarTitle: Auth0
+description: Secure your FastMCP server with Auth0 OAuth
+icon: shield-check
+tag: NEW
+---
+
+import { VersionBadge } from "/snippets/version-badge.mdx"
+
+<VersionBadge version="2.13.0" />
```
jlowin:
```suggestion
<VersionBadge version="2.12.4" />
```

docs/integrations/auth0.mdx
```diff
@@ -0,0 +1,233 @@
...
+2. After you authorize the app, you'll be redirected back
+3. The client receives the token and can make authenticated requests
+
+<Info>
+The client caches tokens locally, so you won't need to re-authenticate for subsequent runs unless the token expires or you explicitly clear the cache.
+</Info>
+
+## Environment Variables
+
+<VersionBadge version="2.13.0" />
```
jlowin:
```suggestion
```

The whole document is already versioned, so i don't think this is necessary here.

src/fastmcp/server/auth/providers/auth0.py
ruhulio: The line below is a bug. It should be:
```
init_kwargs["redirect_path"] = settings.redirect_path
```

This was necessary because the `OAuthProxy` super class sets the default value for `redirect_path` in the parameter argument instead of function code. As such, since `redirect_path` is optional here, it can inadvertently clobber the super class default if passed through without this check.

I considered changing `OAuthProxy`'s handling of `redirect_path` initialization, but that would be a better fit for a separate and targeted PR.

ruhulio:
> Thanks @ruhulio! I don't know very much about OIDC details, so I'll rely on your judgement here that the configuration options exposed are correct, comprehensive, and work correctly. You've tested this with Auth0?

@jlowin I included the links to the OIDC specs and related RFCs that I based the code upon. I added the `strict` flag support since there are providers out there that are not fully OIDC compliant but can be massaged a bit to work (e.g. `GitHub`). In the future™, for providers with broader support of metadata fields, this codebase could leverage that to do richer upfront validation.

And  I've tested this against multiple `Auth0` accounts and it has been working well.

ruhulio:
@jlowin Also wanted to note, if you do a release candidate for `2.12.4`, I'm happy to test out these changes holistically with all other changes in the release.

jlowin:
Thanks @ruhulio, this is great! No need for an `rc`, I have no worries about interactions here.

jlowin (approved):

PR #1820, opened 2025-09-13
Delay import of Provider classes until FastMCP Server Creation

strawgate:
## Description
Fix issues with Server/Settings creation with environment sourced Auth Providers as described at https://github.com/jlowin/fastmcp/issues/1749#issuecomment-3288790161 by not delaying import of the provider until server creation

**Contributors Checklist**
- [x] My change is related to issue #1749

**Review Checklist**

### DIFF
--- a/src/fastmcp/settings.py
+++ b/src/fastmcp/settings.py
@@ -3,9 +3,9 @@ from __future__ import annotations as _annotations
 import inspect
 import warnings
 from pathlib import Path
-from typing import Annotated, Any, Literal
+from typing import TYPE_CHECKING, Annotated, Any, Literal
 
-from pydantic import Field, ImportString, field_validator
+from pydantic import BaseModel, Field, ImportString, field_validator
 from pydantic.fields import FieldInfo
 from pydantic_settings import (
     BaseSettings,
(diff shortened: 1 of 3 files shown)


src/fastmcp/settings.py
```diff
@@ -357,6 +360,22 @@ def normalize_log_level(cls, v):
...
+        # https://github.com/jlowin/fastmcp/issues/1749
+        # Pydantic imports the module in an ImportString during model validation, but we don't want the server
+        # auth module imported during settings creation as it imports dependencies we aren't ready for yet.
+        # To fix this while limiting breaking changes, we delay the import by creating a temporary model
+        # with ImportString only when the class is actually needed
+
+        class ServerAuthModule(BaseModel):
+            auth_module: ImportString
+
+        return ServerAuthModule(auth_module=self.server_auth).auth_module
```
jlowin: Rather than creating this model inside the property, my recommendation is to create a TypeAdapter (either in this module's root scope or with our `get_cached_typeadapter()` utility to avoid repeated creation overhead) and use it directly:

```python
from pydantic import TypeAdapater, ImportString

t = TypeAdapter(ImportString)

# in function
return t.validate_python(self.server_auth)
```

jlowin (changes requested):

src/fastmcp/settings.py
strawgate: I'll try this, I didn't try it before because I had convinced myself that import is triggered on model validation and not field validation / type instantiation

If I can't do it directly with an importstring I'll move the model out

Issue #1762, opened 2025-09-05
OAuthProxy on 2.12.2 doesn't find dynamic client_id upon `authorize`


sornon1:
### Description

With a very basic setup for the OAuthProxy, 2.12.2 doesn't find the UUID of the dynamic client:
`{"error":"invalid_request","error_description":"Client ID '754ae90a-e262-4020-b55a-8a390f0b4eba' not found"}` 

Using 2.12.0 this works fine. 

Implementation is very basic, no special adjustments: 
```
oauth_proxy_provider = OAuthProxy(
    upstream_authorization_endpoint=f"{base_snowflake_url}/oauth/authorize",
    upstream_token_endpoint=f"{base_snowflake_url}/oauth/token-request",
    upstream_client_id=client_id,
    upstream_client_secret=client_secret,
    token_verifier=snowflake_token_verifier,
    base_url=mcp_server_url,
    redirect_path="/auth/callback"
)
```

### Example Code

```Python

```

### Version Information

```Text
FastMCP 2.12.2
MCP 1.13.1 
Python 3.12.9
Windows 11
```

dmikusa:
You can reproduce this with the following steps:

1. Run your FastMCP server.
2. Start MCP Inspector, go to the Authentication Settings, click Quick OAuth flow. This will succeed.
3. Restart your FastMCP server.
4. Back in MCP Inspector, look at the Preparing Authorization section. It'll display the `/authorize` link. Click that link. It'll open a new tab and your browser will display the error above.

The way I've been seeing this is with Claude though. My MCP connector is set up in Claude, I connect it and it works fine. I do some code changes, redeploy my FastMCP service, reconnect Claude and this problem happens. This process was working for me prior to upgrading.

dmikusa:
I think this is [the issue](https://github.com/jlowin/fastmcp/commit/d11e98fe4ad00994054adc5c525a130461add135#diff-cf09d314d237a65eacca854af3cdbc04ff7abdd0febc97ea7d7d37aefc742026).

When you restart the FastMCP server, it wipes out the clients list because that's just in memory. There used to be logic in there to handle the case where the client was not found, but [that logic was removed](https://github.com/jlowin/fastmcp/commit/d11e98fe4ad00994054adc5c525a130461add135#diff-cf09d314d237a65eacca854af3cdbc04ff7abdd0febc97ea7d7d37aefc742026L338-L355).

[...]

PR #1843, opened 2025-09-16
oauth_proxy: create client if client doesn't exist

mkschreder:
## Description

This reverts the changes that caused client to not be found. Temporarily fixing critical problem which completely prevented client authorization when using a proxy provider. 

**Contributors Checklist**

Fixes https://github.com/jlowin/fastmcp/issues/1762

**Review Checklist**

---

### DIFF
--- a/src/fastmcp/server/auth/oauth_proxy.py
+++ b/src/fastmcp/server/auth/oauth_proxy.py
@@ -389,6 +389,23 @@ class OAuthProxy(OAuthProvider):
         """
         client = self._clients.get(client_id)
 
+        if client is None:
+            # For unregistered DCR clients, create a permissive client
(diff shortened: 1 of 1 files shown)


jlowin:
I do not think that generating this client is a viable solution. We removed the code because it appears to be a security flaw: any client who shows up with an unrecognized ID is rewarded with a generic client object that seems to satisfy some (but not all?) auth providers... but does not respect the correct scopes, redirect URIs, etc. etc.

mkschreder:
@jlowin I see the code is back in latest main but taking parameters from storage. So removing it was an error or was the error only in registering the new client?

jlowin:
The error was creating a new client without stored parameters

Issue #1844, opened 2025-09-17
Response Caching Middleware


strawgate:
### Enhancement

We should add a middleware for response caching.

We should provide an in-memory, disk, and maybe even an Elasticsearch ;) cache implementation.

strawgate:
Response caching could support:
1. an allowlist and blocklist of tools to enable/disable caching
2. cache TTL
3. client-id specific caching?

strawgate:
From @aaazzam max size of item for cache would be good too, maybe 1MB default

strawgate:
Caching list operations would be good too, might need a typeddict for setting different ttls for different methods

aaazzam:
> From [@aaazzam](https://github.com/aaazzam) max size of item for cache would be good too, maybe 1MB default

Hey, that's me!

strawgate:
will ship with 2.13

PR #1845, opened 2025-09-17
Add Response Caching Middleware

strawgate:
## Description

Add a response caching middleware that leverages our new key-value library, https://github.com/strawgate/py-key-value `py-key-value-aio`

py-key-value-aio has great store support for local, distributed, and secret stores and offers wrappers for namespacing collections, limiting item sizes, retries, timeouts, etc.

- [x] Cache list calls
- [x] Cache call/read tool/prompt/resource
- [x] Customize TTL by method
- [x] Bust caches on update notifications 

**Contributors Checklist**

- [x] My change closes https://github.com/jlowin/fastmcp/issues/1844

**Review Checklist**

---

### DIFF
--- a/uv.lock
+++ b/uv.lock
@@ -581,7 +581,6 @@ dev = [
     { name = "pyperclip" },
     { name = "pytest" },
     { name = "pytest-asyncio" },
-    { name = "pytest-benchmark" },
     { name = "pytest-cov" },
     { name = "pytest-env" },
     { name = "pytest-flakefinder" },
@@ -604,7 +603,7 @@ requires-dist = [
     { name = "openai", marker = "extra == 'openai'", specifier = ">=1.102.0" },
     { name = "openapi-core", specifier = ">=0.19.5" },
     { name = "openapi-pydantic", specifier = ">=0.5.1" },
-    { name = "py-key-value-aio", extras = ["disk", "memory"], specifier = ">=0.2.2" },
+    { name = "py-key-value-aio", extras = ["disk", "memory"], specifier = ">=0.2.2,<0.3.0" },
     { name = "pydantic", extras = ["email"], specifier = ">=2.11.7" },
     { name = "pyperclip", specifier = ">=1.9.0" },
     { name = "python-dotenv", specifier = ">=1.1.0" },
@@ -627,7 +626,6 @@ dev = [
     { name = "pyperclip", specifier = ">=1.9.0" },
     { name = "pytest", specifier = ">=8.3.3" },
     { name = "pytest-asyncio", specifier = ">=0.23.5" },
-    { name = "pytest-benchmark", specifier = ">=5.1.0" },
     { name = "pytest-cov", specifier = ">=6.1.1" },
     { name = "pytest-env", specifier = ">=1.1.5" },
     { name = "pytest-flakefinder" },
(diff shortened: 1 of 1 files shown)


strawgate:
TypedDicts are a lot less fun than I was expecting and I'm not sure the cache entry model is going to stick around 

The implementation for the cache entries is mostly for the benefit of distributed cache implementations that can't just pickle

jlowin:
Love this idea! The `CachedPrompt` (and similar) classes feel a little overwrought -- I'm not totally clear on what makes the call_next results inherently uncacheable already? Or perhaps I'm missing something you're designing around

strawgate:
> Love this idea! The `CachedPrompt` (and similar) classes feel a little overwrought -- I'm not totally clear on what makes the call_next results inherently uncacheable already? Or perhaps I'm missing something you're designing around

Prompt cannot be instantiated directly as its an ABC and so what comes through the middleware is a FunctionPrompt which can't be serialized/deserialized due to the function reference -- so the CachedPrompt offers a serializable/deserializable model 

Would love other ideas if you have them

jlowin:
Oh! what do you think about removing the ABC from all components and just raising NotImplementedError() instead? Less self-documenting, more compatible?

jlowin:
> Oh! what do you think about removing the ABC from all components and just raising NotImplementedError() instead? Less self-documenting, more compatible?

Wait I ran into this in a different place and made this change in #2031, hopefully that makes this simpler

strawgate:
@jlowin sounds good, i just made a big update to this PR to use the kv store adapters so it's a fair bit closer to being ready now

src/fastmcp/server/middleware/caching.py
```diff
@@ -0,0 +1,671 @@
...
+    structured_content_size = len(
+        json.dumps(
+            value.structured_content, sort_keys=True, separators=(",", ":")
+        ).encode("utf-8")
+    )
+
+    return content_size + structured_content_size
+
+
+# def get_size_of_one_value(value: BaseModel | ToolResult | ReadResourceContents) -> int:
```
jlowin: Some extraneous code here

src/fastmcp/server/middleware/caching.py
```diff
@@ -0,0 +1,671 @@
...
+        default_ttl: int = ONE_HOUR_IN_SECONDS,
+        max_item_size: int | None = None,
+    ):
+        """Initialize the response caching middleware.
+
+        Args:
+            cache_backend: The cache backend to use. If None, an in-memory cache is used.
+            method_settings: The settings for the middleware. If None, the default settings are used.
+            default_ttl: The default TTL for cached responses. Defaults to one hour.
+            max_item_size: The maximum size of an item to cache. Defaults to no size limit.
```
jlowin: It looks like this is only used in tool calls but not others - is that intentional? Maybe update this docstring if so

src/fastmcp/server/middleware/caching.py
```diff
@@ -0,0 +1,671 @@
...
+
+
+def _get_arguments_str(arguments: dict[str, Any] | None) -> str:
+    """Get a string representation of the arguments."""
+
+    if arguments is None:
+        return "null"
+
+    try:
+        return json.dumps(arguments, sort_keys=True, separators=(",", ":"))
```
jlowin: elsewhere we've adopted `pydantic_core.to_json()` with a `str` fallback for this type of try/except, it seems quite performant and also automatically handles pydantic-recognized arguments. 

See e.g. https://github.com/jlowin/fastmcp/blob/7b7e277d5ddcf1dd0c1dbf3eaa3461c1ca077efe/src/fastmcp/tools/tool.py#L63

strawgate:
I'm going to update the PydanticAdapter in the py-key-value library to support lists of basemodels (and transparently nest them in an `items` key).

I should then be able to get rid of the cache-able lists and most of the cachable entries

src/fastmcp/server/middleware/caching.py
strawgate: Im going to remove the key hashing so keys are readable dumps of the args -- `py-key-value` should transparently handle hashing keys if needed.

Will use to_json for the dump when I make this update

src/fastmcp/server/middleware/caching.py
strawgate: I'm going to make an item size limiter wrapper in py-key-value and leverage that here for consistent size limiting

strawgate:
I'm have a couple more updates pending, including size limiting via a new kv store wrapper https://github.com/strawgate/py-key-value/pull/50 and then this will be ready to go by end of day today hopefully

strawgate:
I think this is almost ready to merge -- I'm considering either:
1. Namespace keys using the fastmcp server name
2. Document how users can leverage the PrefixCollectionWrapper in py-key-value-aio to easily share a single distributed key-value store with multiple servers

strawgate:
@jlowin this is ready

I did change ToolResult to require list[ContentBlock] instead of taking Any and just figuring out what to do with it. I think this matches more closely our goal of "ToolResult is you saying you want the result to look like X"

Though I think if we add contentcompatibilitymiddleware later there will be a second change here

src/fastmcp/tools/tool.py
```diff
@@ -63,42 +69,45 @@ def default_serializer(data: Any) -> str:
...
-        structured_content: dict[str, Any] | Any | None = None,
-    ):
-        if content is None and structured_content is None:
-            raise ValueError("Either content or structured_content must be provided")
-        elif content is None:
-            content = structured_content
-
-        self.content: list[ContentBlock] = _convert_to_content(result=content)
+class ToolResult(BaseModel):
+    content: list[ContentBlock] = Field(default_factory=list)
```
jlowin: I think this change is too strict; `ToolResult` is a way for users to control the relationship between `content` and `structured_content` without having to return a raw SDK response (which is a tuple of `(list[ContentBlock], dict[str, Any])`). If ToolResult *does* require essentially that form with no further user assistance, I think it's become primarily a tuple-forming utility, not a DX enhancement over the SDK. In general I think only advanced users should have to use the raw SDK types.

tests/server/test_server_interactions.py
```diff
@@ -975,7 +975,8 @@ async def test_manual_structured_content(self):
         @mcp.tool
         def f() -> ToolResult:
             return ToolResult(
-                content="Hello, world!", structured_content={"message": "Hello, world!"}
+                content=[TextContent(type="text", text="Hello, world!")],
```
jlowin: Here for illustration, instead of making it easy for a user to enhance their "normal" result with a piece of structured data, they would have to load and learn an MCP type.

jlowin:
Really like this -- I think the ToolResult behavior change is a step too far but the middleware is 👍

strawgate:
@jlowin I rolled back the ToolResult change and merging!

PR #1864, opened 2025-09-19
Fix Azure AD v2.0 compatibility by filtering unsupported 'resource' parameter

### LINKED ISSUE #1846: Microsoft Entra OAuth Error: The 'resource' request parameter is not supported.
### Description

Hello, I am trying the new **Microsoft Entra Oauth intrgration** with FastMCP as described [here](https://gofastmcp.com/integrations/azure) . Im very excited about this, but I'm encountering a strange error when trying to authenticate. 

### Error
In the **OAuth authorization URL** im recieving a parameter called **resource** (My url looks like http://localhost:3080/authorize? .... &resource=http%3A%2F%2Flocalhost%3A3080". 

As a result, the sign in fail (see image) but when I remove the resource from the url it suddenly works perfectly and I'm able to authenticate to my MCP server and get some info out of the token claims.

Image

It is entirely possible that I have made a mistake somewhere, and if so please let me know where, otherwise this may be a bug.

### Example Code

```Python
# -------- Server -------------
from fastapi.responses import JSONResponse
from fastmcp import FastMCP

load_dotenv()

    client_id=os.getenv("FASTMCP_SERVER_AUTH_AZURE_CLIENT_ID"),
    client_secret=os.getenv("FASTMCP_SERVER_AUTH_AZURE_CLIENT_SECRET"),
    tenant_id=os.getenv("FASTMCP_SERVER_AUTH_AZURE_TENANT_ID"),
    base_url=os.getenv("FASTMCP_SERVER_AUTH_AZURE_BASE_URL"),
    required_scopes=["User.Read"],
    redirect_path="/oauth/openid/callback",
)


mcp = FastMCP(name="Azure Secured App", auth=auth_provider)

@mcp.tool()
def read_user_info() -> dict:
    """Read user information from the access token."""
    token = get_access_token()
    return {
        "name": token.claims.get("name"),
        "job_title": token.claims.get("job_title")
    }

if __name__ == "__main__":
    logger.info("Starting MCP server with streamable-http transport")
    mcp.run(
        transport="streamable-http",
        host=os.getenv("BIND_ADDRESS", "0.0.0.0"),
        port=int(os.getenv("BIND_PORT", 3080)),
        stateless_http=True,
    )

# ---------- Client --------------

from fastmcp import Client
import asyncio

async def main():
    # The client will automatically handle Azure OAuth
    async with Client("http://localhost:3080/mcp/", auth="oauth") as client:
        # First-time connection will open Azure login in your browser
        print("✓ Authenticated with Azure!")
        
        # Test the protected tool
        result = await client.call_tool("read_user_info")
        print(result)


if __name__ == "__main__":
    asyncio.run(main())
```

### Version Information

```Text
FastMCP version:                                                                    2.12.3
MCP version:                                                                        1.14.0
Python version:                                                                    3.12.10
Platform:                                                     macOS-14.7.8-arm64-arm-64bit
```

nbaju1:
### Problem
The Azure provider fails when MCP clients send the `resource` parameter (from RFC 8707 - OAuth 2.0 Resource Indicators) because Azure AD v2.0 endpoints don't support this parameter. Users get the error:
```
The "resource" request parameter is not supported
```
### Root Cause
The `OAuthProxy` base class forwards all OAuth parameters including `resource` if present:
```python
# oauth_proxy.py line ~444
if params.resource:
    query_params["resource"] = params.resource
```

While this is correct per OAuth 2.0 specs, Azure AD v2.0 specifically rejects this parameter. Azure AD v2.0 uses scopes (like `api://app-id/.default`) to determine the resource/audience instead.

### Solution
Override the `authorize` method in `AzureProvider` to filter out the `resource` parameter before forwarding to Azure AD v2.0. This maintains compatibility with both:
- MCP clients that send the `resource` parameter (following RFC 8707)
- Azure AD v2.0 which doesn't support it

### Changes
- Added `authorize` method override in `AzureProvider` class
- Added imports for `AuthorizationParams` and `OAuthClientInformationFull`
- Added debug logging when filtering out the resource parameter

### Backwards Compatibility
This change is fully backwards compatible. Existing Azure provider users will see no change in behavior unless they were experiencing the "resource parameter not supported" error.

**AI generated code and PR description with Claude Opus 4.1**

**Contributors Checklist**
- [x] My change closes #1846

**Review Checklist**

---

### DIFF
--- a/src/fastmcp/server/auth/providers/azure.py
+++ b/src/fastmcp/server/auth/providers/azure.py
@@ -287,15 +287,16 @@ class AzureProvider(OAuthProxy):
         # Clear the resource parameter that Azure AD v2.0 doesn't support
         # This parameter comes from RFC 8707 (OAuth 2.0 Resource Indicators)
         # but Azure AD v2.0 uses scopes instead to determine the audience
+        params_to_use = params
         if hasattr(params, "resource"):
-            original_resource = params.resource
-            params.resource = None
-
-            if original_resource:
-                logger.debug(
-                    "Filtering out 'resource' parameter '%s' for Azure AD v2.0 (use scopes instead)",
-                    original_resource,
-                )
+            original_resource = getattr(params, "resource", None)
+            if original_resource is not None:
+                params_to_use = params.model_copy(update={"resource": None})
+                if original_resource:
+                    logger.debug(
+                        "Filtering out 'resource' parameter '%s' for Azure AD v2.0 (use scopes instead)",
+                        original_resource,
+                    )
 
         # Call parent's authorize method with the filtered parameters
-        return await super().authorize(client, params)
+        return await super().authorize(client, params_to_use)


nbaju1:
Failed test seems unrelated.

jlowin (changes requested):
Thanks for raising @nbaju1 and your approach makes sense -- my only hesitation is mutating the `params` object could have other consequences that would be difficult to diagnose. Would it be possible to instead create a shallow copy of `params` with `resource=None` (model_copy() might do it easily) and pass *that* to the super method instead?

nbaju1:
I added this fix to #1891. Can either close this PR, or fix the eventual merge conflict in that PR if this gets merged.

jlowin:
Ok, let's focus on #1891 and close this PR. thank you!

Issue #1865, opened 2025-09-19
[Investigation] Performance degradation on Windows platform causing test timeouts


dacamposol:
### Description

# Summary

While I was checking why the pipeline was failing so often, I discovered that the test `tests/server/http/test_http_dependencies.py::test_http_headers_resource_shttp` is flaky and frequently fails on the `windows-latest` runner by exceeding the default 3-second timeout. 

After further investigation, this appears to be a symptom of a broader performance degradation on the Windows platform, as the same related tests run significantly faster on `ubuntu-latest`. 

# Observed Behaviour

The test suite should pass reliably and quickly on all platforms, with individual test times being broadly similar, but the `windows-latest` runner is consistently an order of magnitude slower on specific tests, leading to timeouts:

| Test File                | ✅ ubuntu-latest Time | ⚠️ windows-latest Time  | Slowdown Factor / Notes                                               |
|----------------------------------|-----------------------|-------------------------|------------------------------------------------------------------------|
| test_http_dependencies.py        | ~0.124 s              | ~3.126 s (Timeout)      | ~25.2x slower                                                          |
| test_streamable_http.py          | ~0.353 s              | ~1.575 s                | ~4.5x slower                                                           |
| test_stdio.py (Process Mgmt)     | ~0.659 s              | ~1.921 s                | ~2.9x slower                                                           |
| **Total Suite Time**             | ~85 s (Full Suite)    | >83 s (Partial, Failed) | Ubuntu completes the full suite faster than Windows reaches failure.   |

The consistent slowdown across tests with a common theme points away from a bug in a single test. The issue is more likely located in a lower-level component that interacts differently with the operating system.

A primary candidate for investigation is the `StreamableHttpTransport` (as it's involved in the flaky test) and its underlying dependencies (`httpx`, `anyio`), as their interaction with the Windows networking and async I/O stack could be the source of the latency.

# Proposed Solution

For this project, just increasing the test time to 5s would solve the issue of the flaky CI/CD, but maybe it would be worth to investigate and link an issue to the involved libraries, so it doesn't get translated on a bad user experience for Windows users while using MCP clients developed with this library.

### Example Code

```Python

```

### Version Information

```Text
v2.12.3
```

jlowin:
Closed in #1866, thank you!

PR #1866, opened 2025-09-19
fix: Increase default 3s timeout on Pytest

dacamposol:
Prior to further investigation, let's get the CI/CD running properly.

Related to #1865


nbaju1:
Perhaps modify line [103](https://github.com/jlowin/fastmcp/blob/39a1e59bfd9a665fd961d18418c5addde94c3019/pyproject.toml#L103) in `pyproject.toml` instead to fix this for the entire project, not just in the workflows.

jlowin:
I have to admit I spent the last two days trying to find the technical cause of this (via #1856) and this is just... 🤦 . Nice work.

PR #1786, opened 2025-09-08
Prune titles from tool input and output schemas

jlowin:
Enables title pruning for both input and output schemas in tool creation by adding prune_titles=True to compress_schema calls. This completes the work mentioned in issue #412.

Closes #412

Changes:
- Modified tool.py to prune titles from input schemas (line 416)
- Modified tool.py to prune titles from output schemas (line 476)
- Added comprehensive tests for title pruning behavior
- Updated existing test snapshots to reflect title removal

Generated with [Claude Code](https://claude.ai/code)

PR #1872, opened 2025-09-19
Fix: Remove JSON schema title metadata while preserving parameters named 'title'

### LINKED ISSUE #1782: Prune titles from OutputSchema
### Enhancement

In https://github.com/jlowin/fastmcp/issues/412 we pruned titles from tool schemas, but the OutputSchemas still have Titles, we should prune them

jlowin:
Fixes #1782

## Problem

The original implementation of title pruning was overly aggressive - it removed ALL fields named "title" throughout the entire schema, including legitimate parameter properties. This broke tools that had parameters named "title".

## Solution

Made the title pruning logic context-aware by tracking the parent key during schema traversal. Now only removes "title" fields that are schema metadata (appear alongside `type`, `properties`, etc.), not property names under `properties`.

## Key Changes

1. **Fixed `compress_schema` function**: Added parent_key tracking to distinguish between:
   - Schema metadata titles: `{"title": "MySchema", "type": "object"}` ✅ removed
   - Parameter properties named "title": `{"properties": {"title": {...}}}` ✅ preserved

2. **Added comprehensive tests**: Created `test_json_schema_title_pruning.py` with edge cases including parameters named "title"

3. **Updated tool.py**: Applied title pruning to both input and output schemas

## Test Coverage

All tests pass including the specific test case that was failing with parameters named "title":
- `tests/server/test_server_interactions.py::TestToolParameters::test_parameter_descriptions_with_field_annotations`

This is a proper fix that addresses the root cause rather than the incorrect timeout-based approach suggested for PR #1786.


jlowin:
@strawgate just want to run this by you since you opened the original issue

PR #1891, opened 2025-09-22
Azure (Entra) OAuth: Validate token against FastMCP app, not Graph.

### LINKED ISSUE #1663: Azure scope fails
### Description

Hi!
Thanks for a great package. First off, sorry for a messy issue, but it's almost 1am and I've troubleshooted this for way too long to admit. 😊 I'll clean up tomorrow. 
Disclaimer: I'm the author of [fastapi-azure-auth](https://github.com/Intility/fastapi-azure-auth), so I've unfortunately dealt with Azure's quirks quite a bit. 

I've got some auth issues I just can't get right, and I think it's because of how scopes works in Azure, and how you deal with them. 

FastMCP version: `main`, commit `6d9088704ed4bbada9264f9a27dbf7b04cf3381c`

### The problem with using graph as verify_token

The Azure provider's `verify_token` [function](https://github.com/jlowin/fastmcp/blob/main/src/fastmcp/server/auth/providers/azure.py#L74-L79) verifies whether a token is valid based on whether graph responds with OK or not. This won't work for MCPs implementing OBO flow, as the audience must be the appreg, and not graph. 

Example implementation:
```python
azure = AzureProvider(
    client_id=settings.MCP_SERVER_CLIENT_ID,
    client_secret=settings.MCP_SERVER_CLIENT_SECRET.get_secret_value(),
    tenant_id=settings.TENANT_ID,
    required_scopes=[f"api://{settings.MCP_SERVER_CLIENT_ID}/user_impersonation"],
)
```
will yield a token, which audience is equal to `settings.MCP_SERVER_CLIENT_ID`, so graph will respond with a 401:
`2025-08-29 00:31:50,912 - httpx - INFO - HTTP Request: GET https://graph.microsoft.com/v1.0/me "HTTP/1.1 401 Unauthorized"`

If I'd like to have a graph token, I'd have to set `User.Read` as the required scope. How ever, with a graph token, I cannot perform OBO flow.  

### The alternative

So, since using Graph as a token validator is wrong for Azure, I thought I'd just use the OauthProxy directly: 

```python
azure = OAuthProxy(
    upstream_authorization_endpoint=f"https://login.microsoftonline.com/{settings.TENANT_ID}/oauth2/v2.0/authorize",
    upstream_token_endpoint=f"https://login.microsoftonline.com/{settings.TENANT_ID}/oauth2/v2.0/token",
    upstream_client_id=settings.MCP_SERVER_CLIENT_ID,
    upstream_client_secret=settings.MCP_SERVER_CLIENT_SECRET.get_secret_value(),
    token_verifier=JWTVerifier(
        issuer=f"https://login.microsoftonline.com/{settings.TENANT_ID}/v2.0",
        audience=settings.MCP_SERVER_CLIENT_ID,
        required_scopes=[f"api://{settings.MCP_SERVER_CLIENT_ID}/user_impersonation"],
        jwks_uri=f"https://login.microsoftonline.com/{settings.TENANT_ID}/discovery/v2.0/keys",
    ),
    redirect_path="/auth/callback",
)
```
This works - except for the fact that `required_scopes` is so tightly coupled all around the app. If I log in with Claude Code I can sign in:

Image

But I cannot access any APIs. This is because the `scp` claim is `user_impersonation` in the token, not `api://{settings.MCP_SERVER_CLIENT_ID}/user_impersonation` as I wrote in `required_scopes`. 

So, the solution you might think is to do this:
```diff
-required_scopes=[f"api://{settings.MCP_SERVER_CLIENT_ID}/user_impersonation"],
+required_scopes=["user_impersonation"],
```
but no, this will fail authentication from Azure, as the requested scope must be `api://...`, as seen here:

Image

and the constructed sign-in URL to Azure won't work: 
```
The site at http://localhost:8000/auth/callback?error=invalid_client&error_description=AADSTS650053: The application 'mcp-server-demo' asked for scope 'user_impersonation' that doesn't exist on the resource '00000003-0000-0000-c000-000000000000'.
```
`00000003-0000-0000-c000-000000000000` = Microsoft Graph, and the audience is now wrong, since I haven't explicitly stated I want to get a token for my `settings.MCP_SERVER_CLIENT_ID`.

### Summary

- The Azure implementation does not work properly.
- I cannot use OAuthProxy with JWTVerifier, since the scope expected in the _token_ differs from the scope required in the sign-in URL to Azure
  - And I can't really override this with my own version of `JWTVerifier`, since the `required_scopes` also is inherited in the [auth middleware](https://github.com/jlowin/fastmcp/blob/main/src/fastmcp/server/http.py#L305-L320)

I'd love to be wrong here. If I'm right, and you want more information or a full demo, I don't mind joining a call tomorrow. 

### Example Code

```Python

```

### Version Information

```Text
FastMCP version:                                                                        2.11.4.dev129+6d90887
MCP version:                                                                                           1.13.0
Python version:                                                                                        3.13.2
Platform:                                                                   macOS-15.6-arm64-arm-64bit-Mach-O
FastMCP root path: /Users/jonas/Documents/code/personal/<project>/.venv/lib/python3.13/site-packages
```

### LINKED ISSUE #1846: Microsoft Entra OAuth Error: The 'resource' request parameter is not supported.

### LINKED ISSUE #1925: Add OpenID Connect prompt parameter support to OAuth providers
### Problem
  Users with multiple Microsoft/Azure AD accounts experience authentication failures when the OAuth flow auto-selects the wrong account. This commonly occurs in enterprises where users have multiple
  accounts (work/personal, production/test, cross-tenant guest accounts).

  ### Proposed Solution
  Add support for the OpenID Connect `prompt` parameter in OAuth providers. Specifically, `prompt=select_account` forces the account picker to appear, preventing wrong account selection.

  ### Current Workaround
  ```python
  class AzureProviderWithPrompt(AzureProvider):
      async def authorize(self, client, params) -> str:
          auth_url = await super().authorize(client, params)
          separator = "&" if "?" in auth_url else "?"
          return f"{auth_url}{separator}prompt=select_account"
```

#### Benefits

  - Prevents authentication failures from wrong account selection
  - Follows OpenID Connect standards
  - Essential for multi-tenant scenarios

####  References

  - https://openid.net/specs/openid-connect-core-1_0.html#AuthRequest
  - https://docs.microsoft.com/en-us/azure/active-directory/develop/v2-oauth2-auth-code-flow#request-an-authorization-code

### PR DESCRIPTION
**Pull request by nbaju1**
## Description
Many Azure integrations target a custom API rather than Microsoft Graph. The existing provider always validates tokens via Graph and forwards the RFC 8707 resource parameter, which v2.0 endpoints ignore. This change adds first-class support for custom API audiences and aligns authorization requests with Azure v2.0 semantics.

- New options: `audience` (your API’s Application ID URI) and `api_client_id` (the API app’s GUID used as aud). When `audience` is set, the provider uses `JWTVerifier` with the tenant-scoped issuer and JWKS (`/v2.0` issuer, `/discovery/v2.0/keys`) and validates `aud` against `api_client_id`. It also validates configuration: `api_client_id` is required when `audience` is set, and `required_scopes` must not be prefixed with the audience. The authorization flow drops the unsupported `resource` parameter and prefixes non-OpenID scopes with `<audience>/`, always ensuring `openid` is included. Microsoft Graph behavior remains the default and unchanged when `audience` is not provided.

```python

# Custom API audience (tokens validated via JWKS + issuer)
auth = AzureProvider(
    client_id="your-client-id",
    client_secret="your-client-secret",
    tenant_id="your-tenant-id",
    audience="api://your-api-id",
    api_client_id="00000000-0000-0000-0000-000000000000",
    required_scopes=["read", "write"],  # sent as: openid api://your-api-id/read api://your-api-id/write
)
```

Backwards compatibility: existing Graph-based setups continue to use Graph verification and the previous default scopes. Tests cover the new JWT path (issuer, JWKS, audience, default scopes) and the updated `authorize` behavior (resource removal and scope prefixing).

**Contributors Checklist**
- [x] My change closes #1663 & #1846

**Review Checklist**

---

**Claude Opus 4.1 and GPT-5 used in generating code and PR description**

Closes #1663
Closes #1846
Closes #1925

### DIFF
--- a/.github/workflows/martian-issue-triage.yml
+++ b/.github/workflows/martian-issue-triage.yml
@@ -1,0 +1,172 @@
+name: Martian Issue Triage
+
+on:
+  issues:
+    types: [opened, labeled]
+
+concurrency:
+  group: triage-martian-${{ github.event.issue.number }}
+  cancel-in-progress: true
+
+jobs:
+  martian-issue-triage:
+    if: |
+      (github.event.action == 'opened' && github.actor == 'strawgate') ||
+      (github.event.action == 'labeled' && github.event.label.name == 'triage-martian')
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    permissions:
+      contents: read
+      issues: write
+      pull-requests: read
+      id-token: write
+
+    steps:
+      - name: Checkout base repository
+        uses: actions/checkout@v5
+        with:
+          repository: ${{ github.repository }}
+          ref: ${{ github.event.repository.default_branch }}
+
+      # Install UV package manager
+      - name: Install UV
+        uses: astral-sh/setup-uv@v6
+        with:
+          enable-cache: true
+          cache-dependency-glob: "uv.lock"
+
+      - name: Generate Marvin App token
+        id: marvin-token
+        uses: actions/create-github-app-token@v2
+        with:
+          app-id: ${{ secrets.MARVIN_APP_ID }}
+          private-key: ${{ secrets.MARVIN_APP_PRIVATE_KEY }}
+
+      - name: Set triage prompt
+        id: triage-prompt
+        run: |
+          cat >> $GITHUB_OUTPUT << 'EOF'
+          PROMPT<<PROMPT_END
+          You're an issue triage assistant for FastMCP, a Python framework for building Model Context Protocol servers and clients.
+
+          # IMPORTANT RULES
+          1. You will not make branches or pull requests. Your ONLY action will be investigating the issue, locating related issues,
+          pull requests, and files in the repository and reporting your findings. 
+          2. You will identify the issue type (bug/feature/question) up front and tailor the Recommendation (e.g., for questions: answer directly + links; for bugs: point to failing tests/lines).
+          3. You will avoid speculation and only assert facts that are deeply rooted (traceable) to the codebase, language/framework conventions, related issues, related pull requests, etc.
+          4. The main branch of the repository has been cloned locally, but changes will not be accepted and you are not allowed to make pull requests or other changes. You can search the local repository for relevant code. You will use the available MCP Server tools identify related issues and pull requests (search_issues and search_pull_requests) and you can use search_code to look at the code in relevant dependent packages. For example, you can use search_code to look at the underlying SDK `https://github.com/modelcontextprotocol/python-sdk` to see how it implements a certain class or function relevant to the issue at hand.
+
+          # Getting Started
+          1. Call the generate_agents_md tool to get a high-level summary of the project you're working in
+          2. Get the issue ${{ github.event.issue.number }} in the GitHub repository: ${{ github.repository }}.
+          3. Use the search_issues and search_pull_requests tools to scour the repository for actually related issues and pull requests
+          4. Call the search_code, get_files, etc. tools to search the repository to identify the related classes, methods, docs, tests, etc that are relevant to the issue.
+
+          # Providing a Great Response
+          Your number one priority is to provide a great response to the issue. A great response is a response that is clear, concise, accurate, and actionable. You will avoid long paragraphs, flowery language, and overly verbose responses. Your readers have limited time and attention, so you will be concise and to the point.
+
+          In priority order your goal is to:
+            1. Provide context about the request or issue (related issues, pull requests, files, etc.)
+            2. Layout a single high-quality and actionable recommendation for how to address the issue based on your knowledge of the project, codebase, and issue
+            3. Provide an high quality and detailed plan that a junior developer could follow to implement the recommendation
+
+          Populate the following sections in your response:
+            Recommendation (or “No recommendation” with reason)
+            Findings
+            Detailed Action Plan
+            Related Items
+            Related Files
+            Related Webpages
+
+          You may not be able to do all of these things, sometimes you may find that all you can do is provide in-depth context of the issue and related items. That's perfectly acceptable and expected. Your performance is judged by how accurate your findings are, do the investigation required to have high confidence in your findings and recommendations. "I don't know" or "I'm unable to recommend a course of action" is better than a bad or wrong answer.
+
+          When formulating your response, you will never "bury the lede", you will always provide a clear and concise tl;dr as the first thing in your response. As your response grows in length you can organize the more detailed parts of your response collapsible sections using <details> and <summary> tags. You shouldn't put everything in collapsible sections, especially if the response is short. Use your discretion to determine when to use collapsible sections to avoid overwhelming the reader with too much detail -- think of them like an appendix that can be expanded if the reader is interested.
+
+          # Example output for "Recommendation" part of the response
+          PR #654 already implements the requested feature but is incomplete. The Pull Request is not in a mergeable state yet, the remaining work should be completed: 1) update the Calculator.divide method to utilize the new DivisionByZeroError or the safe_divide function, and 2) update the tests to ensure that the Calculator.divide method raises the new DivisionByZeroError when the divisor is 0.
+
+          <details>
+          <summary>Findings</summary>
+          ...details from the code analysis that are relevant to the issue and the recommendation...
+          </details>
+
+          <details>
+          <summary>Detailed Action Plan</summary>
+          ...a detailed plan that a junior developer could follow to implement the recommendation...
+          </details>
+
+          # Example Output for "Related Items" part of the response
+
+          <details>
+          <summary>Related Issues and Pull Requests</summary>
+
+          | Repository | Issue or PR | Relevance |
+          | --- | --- | --- |
+          | jlowin/fastmcp | [Add matrix operations support](https://github.com/jlowin/fastmcp/pull/680) | This pull request directly addresses the feature request for adding matrix operations to the calculator. |
+          | jlowin/fastmcp | [Add matrix operations support](https://github.com/jlowin/fastmcp/issues/681) | This issue directly addresses the feature request for adding matrix operations to the calculator. |
+          </details>
+
+          <details>
+          <summary>Related Files</summary>
+
+          | Repository | File | Relevance | Sections |
+          | --- | --- | --- | --- |
+          | modelcontextprotocol/python-sdk | [test_calculator.py](https://github.com/modelcontextprotocol/python-sdk/blob/main/test_calculator.py) | This file contains the test cases for the Calculator class, including a test that specifically asserts a ValueError is raised for division by zero, confirming the current intended behavior. | [25-27](https://github.com/modelcontextprotocol/python-sdk/blob/main/test_calculator.py#L25-L27) |
+          | modelcontextprotocol/python-sdk | [calculator.py](https://github.com/modelcontextprotocol/python-sdk/blob/main/calculator.py) | This file contains the implementation of the Calculator class, specifically the `divide` method which raises the ValueError when dividing by zero, matching the bug report. | [29-32](https://github.com/modelcontextprotocol/python-sdk/blob/main/calculator.py#L29-L32) |
+          </details>
+
+          <details>
+          <summary>Related Webpages</summary>
+          | Name | URL | Relevance |
+          | --- | --- | --- |
+          | Handling Division by Zero Best Practices | https://my-blog-about-division-by-zero.com/handling+division+by+zero+in+calculator | This webpage provides general best practices for handling division by zero in calculator applications and in Python, which is directly relevant to the issue and potential solutions. |
+          </details>
+
+          PROMPT_END
+          EOF
+
+      - name: Setup GitHub MCP Server
+        run: |
+          mkdir -p /tmp/mcp-config
+          cat > /tmp/mcp-config/mcp-servers.json << 'EOF'
+          {
+            "mcpServers": {
+              "repository-summary": {
+                "type": "http",
+                "url": "https://agents-md-generator.fastmcp.app/mcp"
+              },
+              "code-search": {
+                "type": "http",
+                "url": "https://public-code-search.fastmcp.app/mcp"
+              },
+              "github-research": {
+                "type": "stdio",
+                "command": "uvx",
+                "args": [
+                  "github-research-mcp"
+                ],
+                "env": {
+                  "DISABLE_SUMMARIES": "true",
+                  "GITHUB_PERSONAL_ACCESS_TOKEN": "${{ steps.marvin-token.outputs.token }}"
+                }
+              }
+            }
+          }
+          EOF
+
+      - name: Run Martian for Issue Triage
+        uses: anthropics/claude-code-action@v1
+        with:
+          github_token: ${{ steps.marvin-token.outputs.token }}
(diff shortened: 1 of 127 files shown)


JonasKs:
> Backwards compatibility: existing Graph-based setups continue to use Graph verification and the previous default scopes.

I'd prefer this to be removed, as the current implementation   
1) uses external requests to graph instead of validating the tokens signature 
2) is flawed. The MCP server is not graph, so it shouldn't accept a graph token. 

I'll read through code tomorrow.

nbaju1:
> > Backwards compatibility: existing Graph-based setups continue to use Graph verification and the previous default scopes.
> 
> I'd prefer this to be removed, as the current implementation
> 
> 1. uses external requests to graph instead of validating the tokens signature
> 2. is flawed. The MCP server is not graph, so it shouldn't accept a graph token.
> 
> I'll read through code tomorrow.

Agree, just wasn't sure if removing it was too much of a breaking change.

jlowin:
We've explicitly excluded the auth modules from our breaking change policy for exactly this reason, so I am totally comfortable with the breaking change if it is one. @JonasKs @nbaju1 happy to defer to your thoughts here.

nbaju1:
I've updated the PR to remove validating against Graph entirely and only validate against the MCP server app it self. Want me to update the PR description to reflect this? 

I believe the `identifier_uri` and custom scope will become redundant once the MCP SDK also exposes the ID token in the `OauthToken` model (already done in the TypeScript SDK), although I haven't looked into the full details of that.

JonasKs:
Looks great! I'm not _that_ involved in the protocol (especially DCR), so maybe you guys could fill me in: How does AI clients, e.g. Claude, Claude Code or even MCP inspector deal with refreshing of tokens? Requesting an `offline_access` scope would give the client a refresh token, but I suspect this is highly Azure Entra ID opinionated.

jlowin:
@JonasKs refresh flows should be identical to normal OAuth flow; DCR only adjusts how clients are registered. The [dynamic] client receives a "normal" token + refresh token and from then on uses standard flows.

JonasKs:
@jlowin & @nbaju1 , then I believe this solution will still have flaws, and we need add a parameter `requested_scopes` (which can default to the client app reg), in order to support requesting refresh tokens (by appending `offline_access` as a requested scope I believe? This **won’t** return `scp: offline_access` as far as I know).

nbaju1:
@JonasKs we can add special handling for the Graph scopes. Add a new parameter `required_graph_scopes: list[str] | None = None` to `AzureProvider` which extends the `prefixed_scopes` list (un-prefixed of course) in the `authorize` method. Thoughts?

JonasKs:
Hmm.. 
From a devex perspective I think we should consider the API and how it can map to other providers. 

I like:
- `upstream_scopes` - what scopes to request from upstream IDP
- `validation_scopes`(aka `required_scopes`) - which scopes to validate 


I’ve mentioned this in my issue, but I think scope validation (and user role checks) should be on a per tool/resource/prompt/route basis, like in FastAPI with dependency injection.    
That’s not the scope of this PR, but I think `upstream_scopes`(or a better name) would make sense if that was ever implemented.

nbaju1:
I'm not familiar with other IDPs, so if this is a pattern that is standard, i.e. requesting scopes from the IDP that isn't returned in the access token, I agree that the name should be more general. 

I vote for getting the provider to work in this PR, then see about separating out the scope validation in a subsequent PR.

JonasKs:
I have no idea either to be honest. I'm on board with this!

whitehat101:
I was able to get this branch working with my Azure tenant.  I was initially getting token rejections: "Bearer token rejected for client"  I enabled debug logs, and enhanced the `JWTVerifier` with more verbose debug logs, I was able to figure out what was wrong.

```python
# src/fastmcp/server/auth/providers/jwt.py
            if self.issuer:
                if claims.get("iss") != self.issuer:
                    self.logger.debug(
                        "Token validation failed: issuer mismatch for client %s. Issuer expected: %s, Issuer received: %s",
                        client_id,
                        self.issuer,
                        claims.get("iss")
                    )
                    self.logger.info("Bearer token rejected for client %s.", client_id)
                    return None
```

It was complaining about the issuer:

> Issuer expected: https://login.microsoftonline.com/TENANT-UUID/v2.0
> Issuer received: https://sts.windows.net/TENANT-UUID/

I was getting a V1 token instead of a V2 token.

To fix this, I needed to edit my Azure App Registration's manifest to include:
```json
{
	"accessTokenAcceptedVersion": 2
}
```

I read that this may be changing to `requestedAccessTokenVersion`, but my manifest wanted the "accepted" version.

JonasKs:
Yeah, this is Azure in a nut shell. The manifests has two versions, and which version you see can vary from app reg to app reg. 

https://intility.github.io/fastapi-azure-auth/single-tenant/azure_setup#step-2---change-token-version-to-v2

sakhan-rio:
@JonasKs I faced similar issue when using a different app reg. Can we also parametrize issuer in AzureProvider class?

> Yeah, this is Azure in a nut shell. The manifests has two versions, and which version you see can vary from app reg to app reg.
> 
> https://intility.github.io/fastapi-azure-auth/single-tenant/azure_setup#step-2---change-token-version-to-v2

JonasKs:
I would strongly suggest _not_ supporting v1 tokens, and rather document how to enforce v2. V1 is **deperecated**, and has a whole lot of extra problems to deal with. The code will end up with lots of `if v1: else: ..`, and bugs will happen. 
My code base for fastapi-azure-auth got so much simpler when I removed v1 tokens support.

nbaju1:
Updated the parameter name to `additional_authorize_scopes` (think this is a bit more clearer) and added the feature request in #1925. @jlowin ready for review from my end now.

src/fastmcp/server/auth/providers/azure.py
```diff
@@ -123,16 +62,17 @@ class AzureProvider(OAuthProxy):
...
-    1. Register an application in Azure Portal (portal.azure.com)
-    2. Configure redirect URI as: http://localhost:8000/auth/callback
-    3. Note your Application (client) ID and create a client secret
-    4. Optionally note your Directory (tenant) ID for single-tenant apps
+    - OAuth proxy to Azure/Microsoft identity platform
+    - JWT validation using tenant issuer and JWKS
+    - Supports tenant configurations: specific tenant ID, "organizations", or "consumers"
+
+    Setup:
+    1. Create an App registration in Azure Portal
```
JonasKs, line 70: Can we add a step to change token version to `2`?

https://intility.github.io/fastapi-azure-auth/single-tenant/azure_setup#step-2---change-token-version-to-v2

marrobi:
> @JonasKs I faced similar issue when using a different app reg. Can we also parametrize issuer in AzureProvider class?
> 
> > Yeah, this is Azure in a nut shell. The manifests has two versions, and which version you see can vary from app reg to app reg.
> > https://intility.github.io/fastapi-azure-auth/single-tenant/azure_setup#step-2---change-token-version-to-v2

@nbaju1 @JonasKs This is looking great. I've just been testing this PR as I'm doing some MCP work with Azure Health Data Services. The OAuth flow seems to be working up until querying my the AHDS API. It's giving back `https://sts.windows.net/<tenant_id>` as the issuer - which fails. I don't have control to change this.

https://learn.microsoft.com/en-us/azure/healthcare-apis/authentication-authorization#access-token

Maybe issuer does need to be configurable?

JonasKs:
Request v2 tokens, its described above. V1 is deprecated.

marrobi:
> Request v2 tokens, its described above. V1 is deprecated.

Thanks for the quick response. I might be trying to do something that's not supported, and don't want to derail the PR discussion, so if its not obvious adon't let me get in the way.

I'm using "on behalf of flow" ( https://learn.microsoft.com/en-us/entra/identity-platform/v2-oauth2-on-behalf-of-flow )

I have a client application, which as per the PR docs has:
```
"api": {
            "requestedAccessTokenVersion": 2
        }
  ```      
   
This then has API permissions on a resource app (the Azure service) .
 
I have the resource app scopes configured under `FASTMCP_SERVER_AUTH_AZURE_REQUIRED_SCOPES`.

It's the resource app token that is coming back as the v1 token.

marrobi:
@JonasKs ignore that, sorted it, I've set `FASTMCP_SERVER_AUTH_AZURE_REQUIRED_SCOPES` to blank, and using the user token to get the resource app token. For some reason (user error) FastMCP was trying to validate the resource token, it shouldn;t have been.

Be great to see this PR released. Thanks @nbaju1!

jlowin (approved):
Thank you very much @nbaju1 and @JonasKs for working on this! This is a huge relief to see validated with such rigor

PR #1947, opened 2025-09-29
Ease editing restrictions in jsonschema to allow strings-as-ints

### LINKED ISSUE #1946: Type Coercion - Discussion #1381
### Description

See discussion #1381.  When faced with an LLM that provides string values for integer items (e.g. top_n="5"), FastMCP fails for validation and presents a decent but not great response error message.

I've traced the challenge to overly stringent validation of integer datatypes in the jsonschema library used by the underlying MCP server.  In the interests of managing the change in the FastMCP codebase, I'm posting the issue here and have a follow-up PR to this issue coming shortly.

### Example Code

```Python

```

### Version Information

```Text
FastMCP version:                                          0.0.1.dev2157+66221ed
MCP version:                                                             1.12.4
Python version:                                                          3.12.3
Platform:          Linux-6.6.87.2-microsoft-standard-WSL2-x86_64-with-glibc2.39
FastMCP root path:                          /home/jonpspri/Projects/fastmcp/src

(running off a local pull from main as of 28 September)
```

jonpspri:
## Description

**Contributors Checklist**

- [X] My change closes #1946 - Discussion #1381
- [ ] I have performed all required documentation updates (suggestions welcome)

**Review Checklist**

---

### DIFF
--- a/src/fastmcp/server/__init__.py
+++ b/src/fastmcp/server/__init__.py
@@ -1,6 +1,6 @@
 from .server import FastMCP
 from .context import Context
 from . import dependencies
-
+from ..utilities import json_schema_validate
 
 __all__ = ["FastMCP", "Context"]
--- a/src/fastmcp/tools/tool.py
+++ b/src/fastmcp/tools/tool.py
@@ -44,6 +44,8 @@ logger = get_logger(__name__)
 
 T = TypeVar("T", default=Any)
 
+json_schema_uri = "https://json-schema.org/draft/2020-12/schema"
+
 
 @dataclass
 class _WrappedResult(Generic[T]):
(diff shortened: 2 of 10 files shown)


jlowin:
I don't like the idea of relaxing type-safety in the server. We went down this road in FastMCP 1.0 and it led to even more confusion and difficulty. This is a client failure IMO and should put pressure on clients to improve rather than on servers to compromise.

jonpspri:
Looking at it with fresh eyes this morning and I think I agree.  I _thought_ I had read somewhere that baseline JavaScript treats stringified numbers and numbers the same, but I can't find a citation this morning.

Happy to just keep the branch in case we need it later, but I am not going to die on this hill.

jonpspri:
Closing based on conversation above.

PR #1974, opened 2025-10-01
Updates to Logging Middleware

### LINKED ISSUE #1973: Add request duration to logging middleware
### Enhancement

We have timing middleware but the logging middleware logs the request name and args, having it log duration on completion of the request would be very nice

strawgate:
* Logging middleware now tracks and logs request duration
* Logger is now namespaced under fastmcp.middleware.*
* An error message type is now printed when an exception occurs
* Timestamp has been removed

Cleaned up the tests a little bit with a helper function for extracting logs from caplog 

Closes #1973


strawgate:
We should probably also remove the timestamp (or make it optional)

jlowin (approved):

Issue #978, opened 2025-06-27
Improvements to Stdio Handling


strawgate:
TL;dr

1) We should do something like having `Client.__aenter__` throw if the session task is completed after connecting so that errors appear during the connection instead of later when you try to use it.

2) It would be nice if we could redirect stderr from mounted MCP servers to the python logging subsystem.

3) We should also consider checking client state before backgrounding the stdio client

#### Logging stderr
stdio_client current dumps errors to stderr, it would be nice if we could capture them and log them as errors via `logging` instead of dumping them to stderr
https://github.com/jlowin/fastmcp/blob/85cca273def16ea1326f16e32febd8876fc5a57d/src/fastmcp/client/transports.py#L369-L371

#### Startup issues

Right now exceptions during startup are presented as something like `you must use async with` which points to a line of code which is using `async with`.

I think this is because here https://github.com/jlowin/fastmcp/blob/85cca273def16ea1326f16e32febd8876fc5a57d/src/fastmcp/client/client.py#L242-L257

`self._session.initialize()` is throwing an `MCPError("Connection Closed")` but because the error occurs before we yield, the error changes from an MCPError to a RuntimeError in ContextLib:

```
    async def __aenter__(self):
        # do not keep args and kwds alive unnecessarily
        # they are only needed for recreation, which is not possible anymore
        del self.args, self.kwds, self.func
        try:
            return await anext(self.gen)
        except StopAsyncIteration:
            raise RuntimeError("generator didn't yield") from None
```

Which means that when [we land back up in the ProxyToolManager](https://github.com/jlowin/fastmcp/blob/85cca273def16ea1326f16e32febd8876fc5a57d/src/fastmcp/server/proxy.py#L43-L63)

```
    async def get_tools(self) -> dict[str, Tool]:
        """Gets the unfiltered tool inventory including local, mounted, and proxy tools."""
        # First get local and mounted tools from parent
        all_tools = await super().get_tools()

        # Then add proxy tools, but don't overwrite existing ones
        try:
            async with self.client:
                client_tools = await self.client.list_tools()
                for tool in client_tools:
                    if tool.name not in all_tools:
                        all_tools[tool.name] = ProxyTool.from_mcp_tool(
                            self.client, tool
                        )
        except McpError as e:
            if e.error.code == METHOD_NOT_FOUND:
                pass  # No tools available from proxy
            else:
                raise e
        return all_tools
```

We no longer have an McpError, we have a RuntimeError which gets raised back up to the Client
[...]

PR #1991, opened 2025-10-03
feat: expose errlog on stdio transport

### LINKED ISSUE #1990: Expose `errlog` to Stdio Transport
### Enhancement

# Enhancement
The MCP stdio_client supports an errlog parameter to redirect subprocess stderr to a specific file handle, but StdioTransport didn't expose this capability. Users had no way to capture or redirect stderr from MCP servers run via stdio.

mpuhacz:
## Description

The MCP stdio_client supports an errlog parameter to redirect subprocess stderr to a specific file handle, but StdioTransport didn't expose this capability. This PR provides that missing parameter across all STDIO transport variants.

The errlog parameter is now available on StdioTransport, PythonStdioTransport, NodeStdioTransport, and FastMCPStdioTransport.

**Contributors Checklist**

- [x] My change closes #1990 

**Review Checklist**

---

### DIFF
--- a/src/fastmcp/client/transports.py
+++ b/src/fastmcp/client/transports.py
@@ -8,7 +8,7 @@ import sys
 import warnings
 from collections.abc import AsyncIterator
 from pathlib import Path
-from typing import Any, Literal, TypeVar, cast, overload
+from typing import Any, Literal, TextIO, TypeVar, cast, overload
 
 import anyio
 import httpx
@@ -313,6 +313,7 @@ class StdioTransport(ClientTransport):
         env: dict[str, str] | None = None,
         cwd: str | None = None,
         keep_alive: bool | None = None,
+        errlog: TextIO | None = None,
     ):
         """
         Initialize a Stdio transport.
@@ -326,6 +327,8 @@ class StdioTransport(ClientTransport):
                        Defaults to True. When True, the subprocess remains active
                        after the connection context exits, allowing reuse in
                        subsequent connections.
+            errlog: Optional file-like object where stderr output from the subprocess
+                   should be written. If not provided, stderr is captured internally.
         """
         self.command = command
         self.args = args
@@ -334,6 +337,7 @@ class StdioTransport(ClientTransport):
         if keep_alive is None:
             keep_alive = True
         self.keep_alive = keep_alive
+        self.errlog = errlog
 
         self._session: ClientSession | None = None
         self._connect_task: asyncio.Task | None = None
@@ -368,6 +372,7 @@ class StdioTransport(ClientTransport):
                 args=self.args,
                 env=self.env,
                 cwd=self.cwd,
+                errlog=self.errlog,
                 session_kwargs=session_kwargs,
                 ready_event=self._ready_event,
                 stop_event=self._stop_event,
@@ -421,6 +426,7 @@ async def _stdio_transport_connect_task(
     args: list[str],
     env: dict[str, str] | None,
     cwd: str | None,
+    errlog: TextIO | None,
     session_kwargs: SessionKwargs,
     ready_event: anyio.Event,
     stop_event: anyio.Event,
@@ -438,7 +444,10 @@ async def _stdio_transport_connect_task(
                     env=env,
                     cwd=cwd,
                 )
-                transport = await stack.enter_async_context(stdio_client(server_params))
+                errlog_kwargs = {"errlog": errlog} if errlog is not None else {}
+                transport = await stack.enter_async_context(
+                    stdio_client(server_params, **errlog_kwargs)
+                )
                 read_stream, write_stream = transport
                 session_future.set_result(
                     await stack.enter_async_context(
@@ -471,6 +480,7 @@ class PythonStdioTransport(StdioTransport):
         cwd: str | None = None,
         python_cmd: str = sys.executable,
         keep_alive: bool | None = None,
+        errlog: TextIO | None = None,
     ):
         """
         Initialize a Python transport.
@@ -485,6 +495,8 @@ class PythonStdioTransport(StdioTransport):
                        Defaults to True. When True, the subprocess remains active
                        after the connection context exits, allowing reuse in
                        subsequent connections.
+            errlog: Optional file-like object where stderr output from the subprocess
+                   should be written. If not provided, stderr is captured internally.
         """
         script_path = Path(script_path).resolve()
         if not script_path.is_file():
@@ -502,6 +514,7 @@ class PythonStdioTransport(StdioTransport):
             env=env,
             cwd=cwd,
             keep_alive=keep_alive,
+            errlog=errlog,
         )
         self.script_path = script_path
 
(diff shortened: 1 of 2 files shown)


src/fastmcp/client/transports.py
```diff
@@ -485,6 +495,8 @@ def __init__(
                        Defaults to True. When True, the subprocess remains active
                        after the connection context exits, allowing reuse in
                        subsequent connections.
+            errlog: Optional file-like object where stderr output from the subprocess
+                   should be written. If not provided, stderr is captured internally.
```
strawgate: what does captured internally mean? its not captured at all, right?

src/fastmcp/client/transports.py
```diff
@@ -438,7 +444,10 @@ async def _stdio_transport_connect_task(
                     env=env,
                     cwd=cwd,
                 )
-                transport = await stack.enter_async_context(stdio_client(server_params))
+                errlog_kwargs = {"errlog": errlog} if errlog is not None else {}
+                transport = await stack.enter_async_context(
```
strawgate, line 460: `errlog or sys.stderr`?

strawgate:
one of the problems i had when i tried to do this in the past is that if you pass anything apart from a file path you get deadlocks, so you cant actually pass lots of things that would normally be just fine to pass into something that takes a TextIO

src/fastmcp/client/transports.py
mpuhacz: good callout - updated the docstring

src/fastmcp/client/transports.py
mpuhacz: errlog https://github.com/modelcontextprotocol/python-sdk/blob/main/src/mcp/client/stdio/__init__.py#L239

mpuhacz:
@strawgate are you able to replicate the issues you had previously using  this PR?

src/fastmcp/client/transports.py
strawgate: I'm saying instead of making this kwargs we could do `errlog or sys.stderr`

strawgate:
I had tried this exact same thing back during this https://github.com/jlowin/fastmcp/issues/978#issuecomment-3014802287

I don't think it's a huge deal but I think we should document it so that it's clear that this only supports a file path

src/fastmcp/client/transports.py
mpuhacz: I'm avoiding setting an explicit default to maintain forward compatibility. This way, any future changes to the MCP library's defaults will be inherited automatically, rather than requiring manual updates to stay in sync.

strawgate:
Also probably worth noting that the moment you do this you will no longer see server errors in the console

mpuhacz:
@strawgate I noticed you were using `subprocess.PIPE`, which is ~65KB, and i believe this might have been causing the deadlock when handling larger err logs.

strawgate:
> @strawgate I noticed you were using `subprocess.PIPE`, which is ~65KB, and i believe this might have been causing the deadlock when handling larger err logs.

The one in that link works just fine,

I had attempted to use the textio parameter but couldn't get it not to deadlock with non file destinations

Something like https://github.com/modelcontextprotocol/python-sdk/issues/645

mpuhacz:
@strawgate docstring updated - let me know your thoughts. 

Also, it would be great if you could share the code which would allow replicating the issue.

src/fastmcp/client/transports.py
strawgate: I don't think that's a concern for this

strawgate:
> @strawgate docstring updated - let me know your thoughts.
> 
> Also, it would be great if you could share the code which would allow replicating the issue.

I can try -- if we don't add a caveat here then we need a test for something other than a file destination anyway as we're now exposing this as part of our API

src/fastmcp/client/transports.py
mpuhacz: Can you help me understand why?

strawgate: Stderr is the defacto destination for logs written to console, them changing it would be a large breaking change, and if they were to change it we would not want to have that change apply implicitly anyway

mpuhacz: Updated (although I do not agree this is the right approach, defaulting to parent library defaults feels less coupled and follows the principle of least surprise - users would expect our wrapper to behave like the underlying library). But I understand your point about stderr being unlikely to change, so I've made the adjustment.

src/fastmcp/client/transports.py
```diff
@@ -8,7 +8,7 @@
 import warnings
 from collections.abc import AsyncIterator
 from pathlib import Path
-from typing import Any, Literal, TypeVar, cast, overload
+from typing import Any, Literal, TextIO, TypeVar, cast, overload
```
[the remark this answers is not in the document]
mpuhacz: There's no import like that

strawgate:
Were you able to get anything other than a file working with textio? I would love love love to find a way to forward MCP logs to the logger

mpuhacz:
@strawgate Unfortunately no, I've only tried it with files so far. Forwarding MCP logs to the logger would be awesome though - I agree that would be really useful.

Maybe we could explore logger integration in a follow-up PR? Could model it after the python-sdk approach (https://github.com/modelcontextprotocol/python-sdk/blob/main/src/mcp/client/stdio/__init__.py#L138-L167).

For this PR, any changes you'd like before merging?

strawgate:
The problem I have in general is that we're exposing textio in our API while we think that it doesn't actually work for the things you might expect it to work for.

E.x. In our API maybe we should just expose a log file argument

mpuhacz:
@strawgate just to confirm - are you suggesting renaming `errlog` to something like `output_stderror_file` for clarity around what’s being exposed? I'm okay with this, let me know if I should update it.

strawgate:
Or just `log_file`?

mpuhacz:
@strawgate updated -- let me know your thoughts!

mpuhacz:
@strawgate just a gentle bump - do you think the current PR version can be released as is?

strawgate:
My thought on `log_file` was that we would take a `Path` and make a TextIO for it. If in the future we want to expose a `logger` or something similar, we can add a parameter for that.

Does that make sense to you? Happy to discuss.

If that's good with you, I'm good with getting this in for 2.13

mpuhacz:
@strawgate sure, that makes sense - I updated the implementation accordingly, with the minor adjustment to keep it `Path | TextIO | None`. The Path will be automatically opened in append mode and properly closed, while TextIO is used directly for users who need custom file handling. Let me know your thoughts!

strawgate:
Can you give an example of what custom file handling we're supporting with this?

The custom file handling is what I'm nervous about as the issues we are going to get are deadlocks when using textio which go away if we only support a file path

mpuhacz:
@strawgate One example could be using `SpooledTemporaryFile` where data is spooled in memory. 
```py
log_file=tempfile.SpooledTemporaryFile(
    mode="w+", max_size=1024 * 1024, encoding="utf-8"
)
```
Anything which follow similar protocol to `SpooledTemporaryFile` can be essentially used there.

mpuhacz:
@strawgate gentle bump 😊 - let me know your thoughts!

jlowin:
Thanks @mpuhacz and @strawgate !

PR #2001, opened 2025-10-04
[Draft] Document OpenTelemetry integration

### LINKED ISSUE #1998: Document logging with OpenTelemetry
### Enhancement

We should probably document emitting logs with OpenTelemetry, we may also be able to emit spans via middleware

strawgate:
Add comprehensive documentation for integrating OpenTelemetry with FastMCP for distributed tracing and observability.

## Changes

- New integration guide at `docs/integrations/opentelemetry.mdx`
- Covers logging integration with LoggingHandler
- Demonstrates span creation via custom middleware
- Includes production OTLP export configuration
- Provides complete working example
- Updated related docs with OpenTelemetry references
- Registered in docs.json under new Observability section
- Added working example in `examples/opentelemetry_example.py`

Closes #1998

### DIFF
--- a/docs/integrations/opentelemetry.mdx
+++ b/docs/integrations/opentelemetry.mdx
@@ -1,0 +1,517 @@
+---
+title: OpenTelemetry Integration
+description: Instrument your FastMCP server with OpenTelemetry for distributed tracing and observability
+icon: chart-line
+---
+
+import { VersionBadge } from "/snippets/version-badge.mdx"
+
+OpenTelemetry provides comprehensive observability for your FastMCP servers through distributed tracing, logging, and metrics. FastMCP's existing logging infrastructure and middleware system integrate seamlessly with OpenTelemetry without requiring any changes to FastMCP itself.
+
+## Why OpenTelemetry?
+
+OpenTelemetry is the industry-standard observability framework that provides:
+
+- **Distributed Tracing**: Track MCP operations across your system with spans
+- **Structured Logging**: Export FastMCP logs to observability backends
+- **Metrics Collection**: Monitor performance and usage patterns
+- **Vendor Agnostic**: Works with Jaeger, Zipkin, Grafana, Datadog, and more
+- **Production Ready**: Battle-tested with stable APIs for tracing and metrics
+
+## Prerequisites
+
+Install OpenTelemetry packages for Python:
+
+```bash
+pip install opentelemetry-api opentelemetry-sdk
+```
+
+For production deployments with OTLP export:
+
+```bash
+pip install opentelemetry-exporter-otlp-proto-grpc
+```
+
+<Note>
+OpenTelemetry supports Python 3.9 and higher. Tracing and metrics are stable, while logging is in active development.
+</Note>
+
+## Logging Integration
+
+FastMCP uses Python's standard `logging` module, which OpenTelemetry can instrument directly using `LoggingHandler`. This sends your FastMCP logs to any OpenTelemetry-compatible backend.
+
+### Basic Setup
+
+```python
+from opentelemetry import trace
+from opentelemetry.sdk.trace import TracerProvider
+from opentelemetry.sdk.resources import Resource
+from opentelemetry.sdk.trace.export import BatchSpanProcessor, ConsoleSpanExporter
+from opentelemetry._logs import set_logger_provider
+from opentelemetry.sdk._logs import LoggerProvider, LoggingHandler
+from opentelemetry.sdk._logs.export import BatchLogRecordProcessor, ConsoleLogExporter
+
+from fastmcp import FastMCP
+from fastmcp.utilities.logging import get_logger
+
+# Configure OpenTelemetry
+resource = Resource(attributes={
+    "service.name": "my-fastmcp-server",
+    "service.version": "1.0.0",
+})
+
+# Set up tracing
+trace_provider = TracerProvider(resource=resource)
+trace_provider.add_span_processor(BatchSpanProcessor(ConsoleSpanExporter()))
+trace.set_tracer_provider(trace_provider)
+
+# Set up logging
+logger_provider = LoggerProvider(resource=resource)
+logger_provider.add_log_record_processor(BatchLogRecordProcessor(ConsoleLogExporter()))
+set_logger_provider(logger_provider)
+
+# Attach OpenTelemetry to FastMCP's logger
+fastmcp_logger = get_logger("my_server")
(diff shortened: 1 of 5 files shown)


docs/integrations/opentelemetry.mdx
```diff
@@ -0,0 +1,517 @@
...
+trace.set_tracer_provider(trace_provider)
+
+# Set up logging
+logger_provider = LoggerProvider(resource=resource)
+logger_provider.add_log_record_processor(BatchLogRecordProcessor(ConsoleLogExporter()))
+set_logger_provider(logger_provider)
+
+# Attach OpenTelemetry to FastMCP's logger
+fastmcp_logger = get_logger("my_server")
+fastmcp_logger.addHandler(LoggingHandler(logger_provider=logger_provider))
```
chrisguidry, line 95: I think this approach would be considered "manual" instrumentation, which is always an option, but wouldn't most folks want something like the "Zero Code" version: https://opentelemetry.io/docs/zero-code/python/

I think we should at least point folks to this as the default, and then we can explain in more detail the manual approach.

I do think there are some additional envvars or CLI switches required to get logging setup with the zero code approach.

docs/integrations/opentelemetry.mdx
```diff
@@ -0,0 +1,517 @@
...
+
+The structured logs include metadata like request timestamps, method names, token estimates, and payload sizes - perfect for observability platforms.
+
+## Spans via Middleware
+
+FastMCP's middleware system provides the perfect foundation for creating OpenTelemetry spans that track MCP operations.
+
+### Basic Tracing Middleware
+
+Create a middleware that emits spans for all MCP requests:
```
chrisguidry: Couldn't we just ship this middleware with FastMCP?  I'd argue it should be on by default, even.

chrisguidry (commented):
I'm wondering why we wouldn't just bake the instrumentation directly into FastMCP and enable it by default?  It's not harmful to have the no-op/non-recording spans if you're not exporting them anywhere.  There may be a very slight performance effect, but I think it's pretty minimal.  Worst case, we could give people an option to _disable_ it if they didn't want OTEL.

strawgate:
@chrisguidry I had Claude generate this, I was mostly going to stick with logging for now 

I thought the problem with emitting spans was going to be that we don't have a way to accept or propagate the trace id through MCP calls so the spans would all be uncorrelated

I'd be very happy to do more if we thought it was useful and sensible

docs/integrations/opentelemetry.mdx
strawgate: Can you explain this a little more? What we would have to do to give folks the zero code experience?

chrisguidry: Oh sure!  The OTEL libraries (at least for Python, but I think for most languages) make a pretty strong distinction between producing telemetry (`opentelemetry-api`) and exporting it (`opentelemetry-sdk`) and they offer these zero-code approaches where you are essentially wrapping your entrypoint with `opentelemetry-instrument` (in Python) so that your code doesn't need to set up any exporters.  Their [canonical example](https://opentelemetry.io/docs/zero-code/python/#configuring-the-agent) looks like this:

```bash
opentelemetry-instrument \
    --traces_exporter console,otlp \
    --metrics_exporter console \
    --service_name your-service-name \
    --exporter_otlp_endpoint 0.0.0.0:4317 \
    python myapp.py
```

What would happen here is that your `myapp.py` might be using `opentelemetry-api` to make spans and metrics, and then when you _run_ the program this way, the spans get exported to whichever ones you've configured there on the CLI.  This example is very similar in spirit to what you laid out here, setting up OTLP exporters and attaching them to the providers, it's just handled by `opentelemetry-instrument`.

So my proposal is just to point people to that link for how they'd run their fastmcp servers by, say, wrapping them like:

```bash
opentelemetry-instrument \
    --traces_exporter console,otlp \
    --metrics_exporter console \
    --service_name my-mcp-server \
    --exporter_otlp_endpoint 0.0.0.0:4317 \
    fastmcp run mymcpserver.py
```

chrisguidry:
> we don't have a way to accept or propagate the trace id through MCP calls

Yep, that's a great call, and I think it _might_ need to be pushed lower into the base SDK?  I would make an argument that even without trace propagation, there's still some value because you'd see all of the things your MCP server called out to, even if each tool call was a separate trace.  Also, as soon as the community figures out client-side trace propagation, it would all "just work" (LOL).

Didn't mean to block your efforts to get great documentation in for sure!

strawgate:
> > we don't have a way to accept or propagate the trace id through MCP calls
> 
> Yep, that's a great call, and I think it _might_ need to be pushed lower into the base SDK? I would make an argument that even without trace propagation, there's still some value because you'd see all of the things your MCP server called out to, even if each tool call was a separate trace. Also, as soon as the community figures out client-side trace propagation, it would all "just work" (LOL).
> 
> Didn't mean to block your efforts to get great documentation in for sure!

I'm ok adding it to our client and server as experimental until the lower level sdk gets it together, would that be interesting for you?

pamelafox:
I found this PR when trying to add otel instrumentation to a FastMCP server. I notice no mention of this package:
https://github.com/traceloop/openllmetry/blob/main/packages/opentelemetry-instrumentation-mcp/opentelemetry/instrumentation/mcp/fastmcp_instrumentation.py

Would that package not be recommended? Is that something you'd add to FastMCP itself?

pamelafox:
Update: I brought in the middleware from your PR and got it working with two OTel-compliant platforms. I also added in StarletteMiddleware on the FastMCP's Starlette app instance, for route-level tracing too.
You can see my usage here:
https://github.com/Azure-Samples/python-mcp-demos/pull/10/files#diff-2f9b1ef748487a1f4ade8aee44108cafa9db539ad34132e96912588d96e609de

I think it'd be great if this was added to FastMCP itself.
You may want to add options to turn off sending tool arguments in the spans, for PII/size reasons.

strawgate:
@pamelafox im not sure using middleware makes the most sense here but im glad it worked for your use-case

strawgate:
/marvin let's update the PR to bake the instrumentation right in

strawgate:
/marvin while creating spans in middleware is great and all, we also want to push the context information into response _meta fields and then when we receive a result with _meta trace id or whatever, we can use that to set the trace id for the context. That way someone calling us over a weird protocol like SSE can still propagate context for distributed traces

okamototk:
I'm looking forwared to this request.
@strawgate do you have any update?

PR #2006, opened 2025-10-05
Replace subprocess tests with in-process async servers

jlowin:
Replaces subprocess-based server tests with in-process async servers using asyncio context managers. This dramatically improves test performance and provides proper tracebacks when errors occur, making debugging significantly easier.

Migrates 19 fixtures across 10 test files from `run_server_in_process()` to the simpler `run_server_async()` pattern, eliminating the overhead of spawning separate processes while maintaining all existing functionality.

### DIFF
--- a/tests/integration_tests/auth/test_github_provider_integration.py
+++ b/tests/integration_tests/auth/test_github_provider_integration.py
@@ -161,19 +161,25 @@ def create_github_server_with_mock_callback(base_url: str) -> FastMCP:
 
 @pytest.fixture
 async def github_server() -> AsyncGenerator[str, None]:
-    """Start GitHub OAuth server on fixed port 9100."""
-    base_url = "http://127.0.0.1:9100"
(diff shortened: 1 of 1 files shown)


strawgate:
Do we actually need to use anyio for this? When I did this in my projects, removing asyncio_default_fixture_loop_scope = "session" and asyncio_default_test_loop_scope = "session" was enough to be able to create, yield, and connect to running servers.

Anyway, looks great

jlowin:
@strawgate I'll investigate, would love if that was the case

jlowin:
I learned Anthropic uses anyio extensively internally so got remotivated to look at this again

jlowin:
@strawgate you were right, got everything working with asyncio, no need for anyio in tests! I will however take a look at replacing some asyncio primitives with anyio ones in the library in the future just for compatibility, but this feels great.

PR #2013, opened 2025-10-06
[Draft] Add server_lifespan module to Contrib

strawgate:
## Description
Add a contrib module which provides a FastMCP which provides a per-server lifespan.

This looks simple enough that we might be able to just do this directly in FastMCP with a `lifespan_per_server` argument?

### DIFF
--- a/src/fastmcp/contrib/server_lifespan/README.md
+++ b/src/fastmcp/contrib/server_lifespan/README.md
@@ -1,0 +1,9 @@
+# Server Lifespan
+
+This module provides a `FastMCP` class which enters its Lifespan once on server startup and exits its lifespan on server exit.
+
+This allows developers to easily define depedencies that will start with the server and be cleaned up on server exit.
+
+## Usage
+
+See the [example implementation](./example.py) for how to use the Server Lifespan.
--- a/src/fastmcp/contrib/server_lifespan/__init__.py
+++ b/src/fastmcp/contrib/server_lifespan/__init__.py
@@ -1,0 +1,3 @@
+from .base import FastMCP
+
+__all__ = ["FastMCP"]
--- a/src/fastmcp/contrib/server_lifespan/base.py
+++ b/src/fastmcp/contrib/server_lifespan/base.py
@@ -1,0 +1,132 @@
+"""
+Server-level lifespan management for FastMCP.
+
+This module provides a FastMCP subclass that manages server-wide resources
+(like database connection pools) that should persist across multiple client sessions.
+
+Note: This is experimental. The upstream MCP SDK's lifespan is designed to be
+per-session. This contrib module works around that limitation for specific use cases.
+"""
+
+from collections.abc import AsyncIterator, Callable
+from contextlib import AbstractAsyncContextManager, AsyncExitStack, asynccontextmanager
+from typing import Any
+
+from mcp.server.lowlevel.server import LifespanResultT
+from typing_extensions import override
+
+from fastmcp import FastMCP as _FastMCP
+
+
+class FastMCP(_FastMCP[LifespanResultT]):
+    """
+    FastMCP subclass with server-wide lifespan management.
+
+    Use this when you need to initialize expensive resources (like database
+    connection pools) once when the server starts, rather than for each session.
+
+    Example:
+        ```python
+        @asynccontextmanager
+        async def server_lifespan(mcp):
+            db_pool = await create_db_pool()
+            try:
+                yield db_pool
+            finally:
+                await db_pool.close()
+
+        mcp = FastMCP("My Server", lifespan=lifespan)
+
+        @mcp.tool()
+        def query(ctx: Context) -> str:
+            # ctx.request_context.lifespan_context is the db_pool
+            return ctx.request_context.lifespan_context.query()
+
+        mcp.run()
+        ```
+
+    Limitations:
+        - Cannot use mount() or import_server() - raises NotImplementedError
+        - Server lifespan is entered once per run() call, not per session
+    """
+
+    def __init__(self, *args: Any, **kwargs: Any):
+        """
+        Initialize with server-wide lifespan.
+
+        Args:
+            name: Server name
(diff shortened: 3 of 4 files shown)


src/fastmcp/client/transports.py
```diff
@@ -827,9 +829,16 @@ async def connect_session(
...
                         yield client_session
+                except Exception as e:
+                    if isinstance(self.server, FastMCP):
+                        _ = await self.server.__aexit__(type(e), e, None)
+                    raise e
                 finally:
                     tg.cancel_scope.cancel()
 
+                if isinstance(self.server, FastMCP):
+                    _ = await self.server.__aexit__(None, None, None)
```
jlowin: This pattern might (will?) result in the __aexit__ being called twice when an error is raised - first in the except block and again here.

src/fastmcp/server/server.py
```diff
@@ -268,6 +276,42 @@ def __init__(
     def __repr__(self) -> str:
         return f"{type(self).__name__}({self.name!r})"
 
+    def _handle_lifespan_settings(
```
jlowin: Suggest moving this to a private function in this module rather than private method but thats purely a organization preference and doesnt impact functionality

src/fastmcp/server/server.py
```diff
@@ -334,6 +378,41 @@ def instructions(self, value: str | None) -> None:
...
+        if self._server_lifespan is None:
+            return
+
+        if self._server_lifespan_stack is not None:
+            await self._server_lifespan_stack.aclose()
+            self._server_lifespan_stack = None
+            self._server_lifespan_result = None
+
+        for server in self._mounted_servers:
+            await server.server.__aexit__(exc_type, exc_val, exc_tb)
```
jlowin: Isn't aexit here also handled by closing the lifespan stack?

src/fastmcp/server/server.py
```diff
@@ -345,24 +424,25 @@ async def run_async(
         Args:
             transport: Transport protocol to use ("stdio", "sse", or "streamable-http")
         """
-        if transport is None:
-            transport = "stdio"
-        if transport not in {"stdio", "http", "sse", "streamable-http"}:
-            raise ValueError(f"Unknown transport: {transport}")
+        async with self:
```
jlowin: This is probably still too high-level to ensure its use; I think `run_stdio_async` and `run_http_async` may be the methods that need this context manager so that users who invoke them directly (which is relatively common based on issue MREs) also get this functionality. Alternatively it could be safe to make the enter/exit behavior reentrant with a no-op so that users can freely enter the context manager themselves without worrying about whether the run method will do it for them.

strawgate:
@jlowin the current way users access the lifespan_context is via the RequestContext object from the underlying SDK

```
@dataclass
class RequestContext(Generic[SessionT, LifespanContextT, RequestT]):
    request_id: RequestId
    meta: RequestParams.Meta | None
    session: SessionT
    lifespan_context: LifespanContextT
    request: RequestT | None = None
```

So the way it's currently implemented is that you're picking whether `lifespan_context` is a session-bound or server-bound. So the `server_lifespan `gets started when the server starts and then we pass a "proxy" to the underlying SDK which gets called by the SDK with every session but just returns the server lifespan.

An alternative implementation could be that we expose the server_lifespan on the FastMCP Context object so that server_lifespan is available via the FastMCP Context and the session_lifespan is available via the MCP request context.

I am a bit biased here -- I think the session_lifespan is useless and so was mostly trying to push this to work the way I think most people think it should work.

jlowin:
I see I see -- in that case I'm tempted to say lets just remove the `session_lifespan` entirely (we should solve that with middleware) and go with the correct implementation of `server_lifespan` (and in that case call it `lifespan`?) And then, in an opinionated way, we *do not* set low-level SDK lifespans for FastMCP users, but rather implement our own sane version. In other words I'm game to lean fully into your opinon that it session lifespan is useless and just lose it. 

I think in this case we'd make the lifespan object availabl eon the FastMCP context rather than messing with the SDK object.

strawgate:
@jlowin I think this version should be a lot easier to reason about. Switching from session to server lifespan is a breaking change though

src/fastmcp/server/server.py
```diff
@@ -89,35 +89,40 @@
...
 ]:
     @asynccontextmanager
     async def wrap(
-        s: LowLevelServer[LifespanResultT],
+        low_level_server: LowLevelServer[LifespanResultT],
     ) -> AsyncIterator[LifespanResultT]:
-        async with AsyncExitStack() as stack:
-            context = await stack.enter_async_context(lifespan(app))
-            yield context
+        if fastmcp_server._lifespan == default_lifespan:
```
jlowin: does this equality work for context managers? does it need to be `is`?

jlowin:
This looks good to me -- since this is a behavior change on the same kwarg, I don't see a great way to make it deprecate gracefully and I think calling it out as a breaking behavioral change (but not an error) is the right way to handle over the 2.13 boundary.

Hyperclaw79:
As an end user, absolutely love this change! @strawgate Would it be possible to also update the docs as part of this PR itself in case anyone wants to try it right away? Since it's a breaking change, people will scratch their heads if there's no docs to specify the new behavior.

strawgate:
Feel free to try this branch, the current code on the branch actually only uses `lifespan` and just makes it server lifespan instead of session lifespan.

I'll update docs but the note about it being a breaking change will be in the release notes

Hyperclaw79:
Already uv synced the branch and it worked like a charm! 🔥
Thank you for the sweet change.

src/fastmcp/server/server.py
strawgate: I checked by passing `default_lifespan` and running the debugger, this does work with a context manager passed (which is also what `test_server_lifespan_basic` does) and this "if" branch is hit when `default_lifespan` is passed

strawgate:
@jlowin i think this is ready

Issue #2024, opened 2025-10-08
Check if `refresh_token` is None for refresh token rotation


Kyle-Dinh:
### Enhancement

This slight enhance is to make the FastMCP more flexible to Authorization Server responses when exchanging a refresh_token for a new access_token. The Authentication Server will return a new refresh token if refresh token rotation is supported or no `refresh_token` key if it isn't. I suggestion that instead of setting the condition to just the absense/presence of the key, we also check the value is set and not to `None`.

Kyle-Dinh:
Link to suggested changes here: https://github.com/jlowin/fastmcp/pull/2025

jlowin:
Closed in #2025

PR #2025, opened 2025-10-08
Check if refresh_token is None

Kyle-Dinh:
## Description
When exchanging a refresh_token for another refresh_token, there are a couple options:
Option 1: `refresh_token` is not set, implying that refresh token rotation is not present
Option 2: `refresh_token` is set to a value, implying a new refresh token is given

The changes I'm suggesting is more of an edge case, which is Option 3: `refresh_token` is set to None, implying that refresh token rotation is not present. While this is similar to Option 1, this goes farther beyond just checking if the key is present.

**Contributors Checklist**

- [x] My change closes [#2024](https://github.com/jlowin/fastmcp/issues/2024)

**Review Checklist**

---


jlowin:
I think this is a reasonable protection but did you see this edge case in the wild?

Kyle-Dinh:
Not specifically in the wild just was a case I ran into while dev-ing

jlowin:
Ok cool. I think the idiomatic way to write this is

```python
if new_refresh_token := token_response.get("refresh_token"):
    ...
```

Kyle-Dinh:
> Ok cool. I think the idiomatic way to write this is
> 
> ```python
> if new_refresh_token := token_response.get("refresh_token"):
>     ...
> ```

Ahh yep that makes sense just committed the change

Kyle-Dinh:
Hey @jlowin was just curious when would I expect a new release with this change?

jlowin:
The next release will be 2.13, which is quite large and requires a lot of testing. Unfortunately due to the number of changes it's difficult to create a patch release ahead of it. It will probably go out next week or the following week at latest. In the meantime, you can install from main!

(usually we try to release every 1-2 weeks but this is a big one)

PR #1997, opened 2025-10-03
init

aaazzam:
Description

  Add Supabase authentication provider for FastMCP following the same pattern as existing commercial auth providers (Auth0, Descope, WorkOS).

  The SupabaseProvider implements RemoteAuthProvider to verify JWTs issued by Supabase Auth using their JWKS endpoint. It supports Supabase projects using
  asymmetric ES256 signing keys (standard for projects created after May 2025, or migrated projects).

  Key Features:
  - JWT verification using Supabase's JWKS endpoint (/auth/v1/.well-known/jwks.json)
  - ES256 algorithm support (Elliptic Curve Digital Signature)
  - Environment variable configuration (FASTMCP_SERVER_AUTH_SUPABASE_*)
  - Authorization server metadata forwarding
  - Follows established FastMCP auth provider patterns

  Testing:
  - Validated end-to-end with real Supabase project
  - Successfully authenticated users and verified JWTs
  - Protected MCP endpoints working correctly

  🤖 Generated with https://claude.com/claude-code

  Contributors Checklist

  - My change closes #(issue number)
  - I have followed the repository's development workflow
  - I have tested my changes manually and by adding relevant tests
  - I have performed all required documentation updates

  Review Checklist

  - I have self-reviewed my changes
  - My Pull Request is ready for review

  ---
  Files Changed:
  - src/fastmcp/server/auth/providers/supabase.py - Provider implementation
  - tests/server/auth/providers/test_supabase.py - Comprehensive test suite (9 tests)

PR #2030, opened 2025-10-09
Add supabase docs

jlowin:
Fast follow to #1997 to add integration docs


Arjun-A-I:
Hey guys when can we expect the integration to be out...super excited to use supabase in ma mcp auth

jlowin:
I think we will hold this until we either confirm #1997 works or it comes out of beta, see https://github.com/orgs/supabase/discussions/38022

PR #2052, opened 2025-10-10
Add warnings regarding memory store

strawgate:
## Description
Add a warning when oAuth is used with memory store

### DIFF
--- a/src/fastmcp/server/middleware/middleware.py
+++ b/src/fastmcp/server/middleware/middleware.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 import logging
-from collections.abc import Awaitable
+from collections.abc import Awaitable, Sequence
 from dataclasses import dataclass, field, replace
 from datetime import datetime, timezone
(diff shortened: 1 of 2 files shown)


src/fastmcp/client/auth/oauth.py
```diff
@@ -191,6 +191,14 @@ def __init__(
         # Create server-specific token storage
         token_storage = token_storage or MemoryStore()
 
+        if isinstance(token_storage, MemoryStore):
```
jlowin, line 194: You know what, for clients I actually think filesystem storage could be a saner default. Clients themselves *are* expected to be ephemeral, but I think its a reasonable assumption that auth tokens survive any one client instance, for the same server at least. I think almost everyone would want to swap this to filesystem for local caching (but not any distributed storage, as that's "too" shared)

strawgate: We discussed out of band and I'd like to figure out secret storage before writing these to disk

PR #2056, opened 2025-10-10
Prevent confused deputy attacks in OAuth proxy

jlowin:
FastMCP's OAuth proxy uses a static client ID when forwarding authorization requests to upstream providers. This creates a confused deputy vulnerability: an attacker can register a malicious client via the open DCR endpoint, then trick a victim into authorizing it. Because the victim already has a consent cookie for the static client ID, the upstream authorization server bypasses the consent screen and issues an authorization code that redirects to the attacker's server.

This PR implements the mitigation required by the [MCP specification](https://modelcontextprotocol.io/specification/2025-06-18/basic/security_best_practices#confused-deputy-problem): the proxy now obtains explicit user consent for each dynamically registered client before forwarding to the upstream authorization server. It also updates relevant documentation.

image

The consent flow:
- User initiates authorization → redirected to `/consent` endpoint
- Consent page displays client details and requested scopes with CSRF protection
- User approves/denies → decision stored in signed cookie
- On approval, user is forwarded to upstream authorization server
- Subsequent requests from the same client auto-approve via signed cookie

Server-side storage ensures OAuth transactions and authorization codes persist across requests, with full support for custom storage backends via \`AsyncKeyValue\`.

Thanks to @localden for identifying this vulnerability and providing detailed analysis on this topic in his [blog post](https://den.dev/blog/mcp-confused-deputy-api-management/).

### DIFF
--- a/tests/integration_tests/auth/test_github_provider_integration.py
+++ b/tests/integration_tests/auth/test_github_provider_integration.py
@@ -265,7 +265,7 @@ async def test_github_oauth_authorization_redirect(github_server: str):
 
         # Step 3: Visit consent page to get CSRF token
         consent_response = await http_client.get(
-            f"{base_url}{consent_location}", follow_redirects=False
+            consent_location, follow_redirects=False
         )
         assert consent_response.status_code == 200
 


jlowin:
@strawgate this is a great test case for the new OAuth storage mechanisms

src/fastmcp/server/auth/oauth_proxy.py
```diff
@@ -112,12 +184,110 @@ def validate_redirect_uri(self, redirect_uri: AnyUrl | None) -> AnyUrl:
...
+
+    # Additional styles needed for this page
+    additional_styles = (
+        INFO_BOX_STYLES + DETAIL_BOX_STYLES + BUTTON_STYLES + TOOLTIP_STYLES
+    )
+
+    # Need to allow form-action for form submission
+    csp_policy = "default-src 'none'; style-src 'unsafe-inline'; img-src https:; base-uri 'none'; form-action *"
+
+    return create_page(
```
billyshambrook, line 280: Super excited to see this and with remote storage for all state! Sorry if I am commenting too early but would be great if the consent page could be customizable to show more of the brand of the MCP server, and the very least, be able to replace "FastMCP" with the server name and the logo. Or maybe just be able to pass in an entire custom HTML - Thanks!

jlowin: Agree! At the moment auth modules don't receive the server (other way around) so I punted on e.g. server name for this draft because it required more changes than I wanted in this PR

strawgate:
Generated Image October 10, 2025 - 10_23PM (1)

jlowin:
Feeling good about the state of this and want to continue iterating on some token storage concerns without polluting it so I'm going to merge this as we continue pressing on state / storage for sensitive items.

Issue #2059, opened 2025-10-11
Push client logging to logger


strawgate:
### Enhancement

We should consider a setting on the server for teeing client logging via ctx to the servers python logger or maybe just an argument on the logging methods for also logging to the server log

strawgate:
/marvin maybe we do a literal of log_to that can be either CLIENT (default) or CLIENT_AND_SERVER? 

Or maybe something else? It should be clear that it's going to both but ideally shorter somehow?

Also should we do the server logging before we send it to the client?

PR #2063, opened 2025-10-11
Also push client messages (info/warn/debug) to server debug log

strawgate:
## Summary

Log client messages sent via `ctx.info|debug|warning|etc`. Client messages are logged with their level mapped to Python levels and then "clamped" to `debug`.

Users can enable these logs without having to enable debug logging on the server (as described in the docs) via:
```python
import logging

from fastmcp.utilities.logging import get_logger

to_client_logger = get_logger(name="fastmcp.server.context.to_client")
to_client_logger.setLevel(level=logging.DEBUG)
```

We can evaluate letting users control how these log levels map to server levels, but for now users who want them logged at their original level can call `_unclamp_logger(logger=to_client_logger)` on the logger as a private workaround. Or if they would prefer to log these as `INFO`, they can apply `_clamp_logger(logger=to_client_logger,min_level=INFO,max_level=INFO)`

Fixes: https://github.com/jlowin/fastmcp/issues/2059 and https://github.com/jlowin/fastmcp/issues/2078

### DIFF
--- a/src/fastmcp/client/auth/oauth.py
+++ b/src/fastmcp/client/auth/oauth.py
@@ -191,6 +191,14 @@ class OAuth(OAuthClientProvider):
         # Create server-specific token storage
         token_storage = token_storage or MemoryStore()
 
+        if isinstance(token_storage, MemoryStore):
+            from warnings import warn
+
+            warn(
+                message="Using in-memory token storage is not recommended for production use -- "
+                + "tokens will be lost on server restart."
+            )
+
         self.token_storage_adapter: TokenStorageAdapter = TokenStorageAdapter(
             async_key_value=token_storage, server_url=server_base_url
         )
--- a/src/fastmcp/server/auth/oauth_proxy.py
+++ b/src/fastmcp/server/auth/oauth_proxy.py
@@ -577,7 +577,7 @@ class OAuthProxy(OAuthProvider):
         self._client_storage: AsyncKeyValue = client_storage or MemoryStore()
 
         # Warn if using MemoryStore in production
-        if client_storage is None or isinstance(client_storage, MemoryStore):
+        if isinstance(client_storage, MemoryStore):
             logger.warning(
                 "Using in-memory storage - all OAuth state will be lost on restart. "
                 "For production, configure persistent storage (Redis, PostgreSQL, etc.)."


jlowin:
I think the substance of this PR addresses a real pain point ("I logged but I don't see it!") where users don't realize that an MCP log is sent to clients, not visible on the server. 

But I've been hemming and hawing a little bit on implementatino. On the one hand I think visibility on the server is a huge observability benefit so I love this to the point I'm wondering if it should be default ON. However, I am worried that if a user sends e.g. a CRITICAL log to a client (for some reason) then we do *not* necessarily want to freak out their platform team by presenting a critical log on the server. 

Therefore my suggestion is to split the baby and **always** emit a server-side DEBUG log for every ctx.log() call (and have no new kwargs):
```
server_logger.debug(
      f"MCP log sent to client with level {level}: {message}",
      extra={...}
  )
```

This provides an audit trail of all client communications without polluting operational logs. It can be clearly documented and doesn't risk users starting to think that the MCP log is actually a server convenience. I think it would be actively bad if users used ctx.log() to log e.g. information they think is staying on the server.

strawgate:
We could do something like add a global setting which acts as a clamp on the log level for the logs teed to the server. That would allow someone to force all to debug or errors/critical to just warnings, etc

Maybe we could log at the level the user picks but to a namespaced logger that's off by default and we document how to turn on the logger 

We can include a comment about the server logging enablement in the docstrings on ctx

I'll noodle on this a little more with your comments in mind

strawgate:
I still need to update the tests and noodle on this a little more.

We don't currently show the logger names in the logs (perhaps this should be configurable), but I do like prefixing the messages with something like [To Client] so it doesn't quite feel like directly logging to the server but instead that we're logging what was sent to the client.

We can have the utility function that enables the logger take a "max_level" which can be set to debug, etc which can act as a clamp on the logger level for logs teed to the server.

Or maybe as you said everything goes to debug by default, the utility function can take a log_level_map which users can use to push to other log levels (and we provide a default map which maps them to be corresponding levels)

jlowin:
I like the idea of letting users map stuff -- feels like a good balance of very universal default with new debug verbosity, and letting them opt in to elevating

strawgate:
@jlowin updated the PR body with the current behavior

strawgate:
@jlowin i think this is ready

PR #2074, opened 2025-10-11
Add RFC 7662 token introspection provider

### LINKED ISSUE #2048: Token verification for Opaque token (RFC 7662)
### Enhancement

Currently, JWT token verification is supported but there is no support for opaque token. A lot of auth systems (along with ours) are still using Opaque token. To verify the token we connect to token introspection endpoint (available via auth server metadata) and check access and scopes.

[RFC 7662](https://datatracker.ietf.org/doc/html/rfc7662) outlines the details.

Potential interface might look like this:

```
class IntrospectionTokenVerifier(TokenVerifier):
    def __init__(
        self,
        *,
        introspection_url: AnyHttpUrl | str,
        client_id: str,
        client_secret: str,
        timeout_seconds: int = 10,
        required_scopes: list[str] | None = None,
    ):

        pass
```

jlowin:
Closes #2048

Note: would love some community validation this works with their various auth providers, hard to test as just one person!

Many OAuth providers issue opaque tokens that can't be validated locally like JWTs. These tokens require calling the provider's introspection endpoint to check if they're active and extract their scopes. FastMCP previously only supported JWT verification.

This PR adds an `IntrospectionTokenVerifier` implementing RFC 7662 OAuth 2.0 Token Introspection. The verifier authenticates to the introspection endpoint using HTTP Basic Auth and validates tokens in real-time, enabling immediate revocation and provider-managed authorization logic.

```python
from fastmcp import FastMCP
from fastmcp.server.auth.providers.introspection import IntrospectionTokenVerifier

verifier = IntrospectionTokenVerifier(
    introspection_url="https://auth.example.com/oauth/introspect",
    client_id="mcp-resource-server",
    client_secret="your-secret",
    required_scopes=["api:read", "api:write"]
)

mcp = FastMCP("Protected API", auth=verifier)
```

### DIFF
--- a/tests/client/test_openapi_experimental.py
+++ b/tests/client/test_openapi_experimental.py
@@ -191,6 +191,7 @@ async def test_client_with_excluded_header_is_ignored(sse_server: str):
         assert headers["host"] == "fastapi"
 
 
+@pytest.mark.flaky(retries=2, delay=1)
 async def test_client_headers_proxy(proxy_server: str):
     """
     Test that client headers are passed through the proxy to the remove server.
--- a/tests/server/auth/test_oauth_consent_flow.py
+++ b/tests/server/auth/test_oauth_consent_flow.py
@@ -231,10 +231,12 @@ class TestServerSideStorage:
                 csrf_token = secrets.token_urlsafe(32)
 
             # Approve consent with CSRF token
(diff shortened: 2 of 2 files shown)


jlowin:
Got 👍 that it's working in #2048

PR #2092, opened 2025-10-14
Fix experimental parser missing top-level response schemas (#2087)

### LINKED ISSUE #2087: `from_openapi()` fails to convert nested schema definitions, causing PointerToNowhere errors
## Title
`from_openapi()` fails to convert nested schema definitions, causing PointerToNowhere errors

## Description

When using `FastMCP.from_openapi()` to create an MCP server from an OpenAPI 3.0 specification, nested schema definitions are not properly converted from `components/schemas` to the `$defs` section. This causes `PointerToNowhere` errors when any endpoint returns objects that reference these nested schemas.

The conversion process successfully changes references from `#/components/schemas/X` to `#/$defs/X`, but fails to actually move the nested schema definitions into the `$defs` section, resulting in broken references.

## Expected Behavior

When converting an OpenAPI 3.0 spec with nested schemas like:
```json
{
  "components": {
    "schemas": {
      "Person": {
        "properties": {
          "Name": {"$ref": "#/components/schemas/Name"},
          "Job": {"$ref": "#/components/schemas/Job"}
        }
      },
      "Name": {
        "type": "object",
        "properties": {
          "First": {"type": "string"},
          "Last": {"type": "string"}
        }
      },
      "Job": {
        "type": "object",
        "properties": {
          "Company": {"type": "string"}
        }
      }
    }
  }
}
```

**All schemas** (Person, Name, Job) should be moved to the `$defs` section in the converted JSON Schema format.

## Actual Behavior

Only the top-level schemas directly referenced by API endpoints are moved to `$defs`. Nested schemas that are only referenced by other schemas are missing, causing the following error:

```
PointerToNowhere: '/$defs/Name' does not exist within {'type': 'object', 'properties': {'Count': {'type': 'integer', 'format': 'int32'}, 'Items': {'type': ['array', 'null'], 'items': {'$ref': '#/$defs/Person'}}}, '$defs': {'Person': {'type': 'object', 'properties': {'LastUpdate': {...}, 'Id': {...}, 'Name': {'$ref': '#/$defs/Name'}, 'Job': {'$ref': '#/$defs/Job'}, ...}}}}
```

Note that `Person` exists in `$defs` but `Name` and `Job` do not, even though they are referenced.

## Steps to Reproduce

1. Create an OpenAPI 3.0 spec with nested schema references (see example below)
2. Use `FastMCP.from_openapi()` to create an MCP server
3. Call any endpoint that returns objects with nested schemas
4. Observe `PointerToNowhere` error

## Minimal Reproducible Code

```python
import httpx
import asyncio
from fastmcp import FastMCP

async def test_nested_schemas():
    # OpenAPI spec with nested schemas
    openapi_spec = {
        "openapi": "3.0.1",
        "info": {"title": "Test API", "version": "v1"},
        "paths": {
            "/persons": {
                "get": {
                    "operationId": "GetPersons",
                    "responses": {
                        "200": {
                            "description": "OK",
                            "content": {
                                "application/json": {
                                    "schema": {"$ref": "#/components/schemas/PersonList"}
                                }
                            }
                        }
                    }
                }
            }
        },
        "components": {
            "schemas": {
                "PersonList": {
                    "type": "object",
                    "properties": {
                        "Count": {"type": "integer"},
                        "Items": {
                            "type": "array",
                            "items": {"$ref": "#/components/schemas/Person"}
                        }
                    }
                },
                "Person": {
                    "type": "object",
                    "properties": {
                        "Id": {"type": "string"},
                        "Name": {"$ref": "#/components/schemas/Name"},
                        "Job": {"$ref": "#/components/schemas/Job"}
                    }
                },
                "Name": {
                    "type": "object",
                    "properties": {
                        "First": {"type": "string"},
                        "Last": {"type": "string"}
                    }
                },
                "Job": {
                    "type": "object",
                    "properties": {
                        "Company": {"type": "string"},
                        "Title": {"type": "string"}
                    }
                }
            }
        }
    }
    
    client = httpx.AsyncClient(base_url="https://api.example.com")
    mcp = FastMCP.from_openapi(openapi_spec=openapi_spec, client=client)
    # This will fail when trying to call GetPersons

asyncio.run(test_nested_schemas())
```

## Real-World Example

This issue occurs with the SlideCrew API. The full OpenAPI spec is available at:
https://app.slidecrew.com/api/v1.IO/swagger.json

Specifically, the `/v1/persons` endpoint returns a `PersonList` containing `Person` objects, which reference `Name`, `Job`, and `PhoneNumber` schemas. All three nested schemas fail to convert properly.

### Code that triggers the error:

```python
import httpx
import asyncio
from fastmcp import FastMCP

async def load_openapi_spec():
    async with httpx.AsyncClient() as client:
        response = await client.get("https://app.slidecrew.com/api/v1.IO/swagger.json")
        response.raise_for_status()
        return response.json()

auth_token = "YOUR_API_KEY"

client = httpx.AsyncClient(
    base_url="https://app.slidecrew.com/",
    headers={"X-Api-Key": auth_token}
)

async def create_server():
    openapi_spec = await load_openapi_spec()
    return FastMCP.from_openapi(
        openapi_spec=openapi_spec,
        client=client,
        name="slidecrew-import-export"
    )

mcp = asyncio.run(create_server())
```

Calling the `GetPersons` operation results in:
```
PointerToNowhere: '/$defs/Name' does not exist within {...}
```

## Affected Schemas in SlideCrew API

Missing from `$defs`:
- `Name` (referenced by `Person`)
- `Job` (referenced by `Person`)
- `PhoneNumber` (referenced by `Person`)
- `CustomDataConnection` (referenced by `CustomField` and `Event`)
- `CustomDataInputTypes` (referenced by `CustomField`)
- `TimerMode` (referenced by `TimerData` and `Room`)
- Many others

## Environment

- **FastMCP version**: 2.12.4
- **Python version**: Python 3.12.1
- **OS**: [ Windows 11]
- **Installation method**: UV

## Root Cause Analysis

The schema conversion appears to only walk the tree of schemas starting from endpoint responses/requests, but doesn't recursively follow all `$ref` references to ensure **all** schemas are included in `$defs`. 

When a schema like `PersonList` references `Person`, which in turn references `Name`, the conversion process:
1. ✅ Converts `PersonList` → included in `$defs`
2. ✅ Converts `Person` → included in `$defs`  
3. ❌ Fails to convert `Name` → **missing** from `$defs`
4. ✅ Updates reference from `#/components/schemas/Name` to `#/$defs/Name`
5. ❌ Results in broken reference

## Suggested Fix

The schema conversion logic should recursively traverse **all** `$ref` references within schemas to ensure every referenced schema is included in the `$defs` section. A recursive depth-first or breadth-first traversal starting from endpoint schemas should collect all transitively referenced schemas.

Pseudocode:
```python
def collect_all_schemas(schema, collected=None):
    if collected is None:
        collected = set()
    
    if '$ref' in schema:
        ref_name = extract_schema_name(schema['$ref'])
        if ref_name not in collected:
            collected.add(ref_name)
            # Recursively collect from the referenced schema
            collect_all_schemas(all_schemas[ref_name], collected)
    
    # Check properties, items, allOf, anyOf, oneOf, etc.
    for key in ['properties', 'items', 'allOf', 'anyOf', 'oneOf']:
        if key in schema:
            # Recurse into nested structures
            ...
    
    return collected
```

## Impact

This bug makes FastMCP incompatible with any OpenAPI specification that uses nested schema definitions, which is a common pattern in well-structured APIs. Workarounds require either:
1. Manually patching the generated schema (fragile and requires maintenance)
2. Not using FastMCP and writing custom API clients
3. Flattening the OpenAPI spec to avoid nested schemas (not always feasible)

## Attachments

Full SlideCrew swagger.json available at:
https://app.slidecrew.com/api/v1.IO/swagger.json

---

## Additional Notes

Thank you for FastMCP - it's a great tool! This bug affects real-world APIs with properly structured schemas. A fix would make FastMCP compatible with a much wider range of OpenAPI specifications.

### Example Code

```Python

```

### Version Information

```Text
## Environment

- **FastMCP version**: 2.12.4
- **Python version**: Python 3.12.1
- **OS**: Windows 11
- **Installation method**: UV
```

jlowin:
Fixes #2087

When a response schema is a top-level `$ref` (e.g., `{"$ref": "#/components/schemas/PersonList"}`), the parser now correctly includes that schema in `response_schemas` alongside its nested dependencies.

**Before:** Only transitive dependencies were included (Person, Name, Job), but the top-level schema (PersonList) was missing.

**After:** All schemas are included (PersonList, Person, Name, Job).

**Changes:**
- Track schema name before `$ref` resolution in `_extract_responses`
- Check for marker in `_extract_output_schema_dependencies` to include top-level schema
- Added test case in `test_transitive_references.py`

### DIFF
--- a/src/fastmcp/experimental/utilities/openapi/parser.py
+++ b/src/fastmcp/experimental/utilities/openapi/parser.py
@@ -474,9 +474,22 @@ class OpenAPIParser(
                         and media_type_obj.media_type_schema
                     ):
                         try:
-                            schema_dict = self._extract_schema_as_dict(
-                                media_type_obj.media_type_schema
-                            )
+                            # Track if this is a top-level $ref before resolution
+                            top_level_schema_name = None
+                            media_schema = media_type_obj.media_type_schema
+                            if isinstance(media_schema, self.reference_cls):
+                                ref_str = media_schema.ref
+                                if isinstance(ref_str, str) and ref_str.startswith(
+                                    "#/components/schemas/"
+                                ):
+                                    top_level_schema_name = ref_str.split("/")[-1]
+
+                            schema_dict = self._extract_schema_as_dict(media_schema)
+                            # Add marker for top-level schema if it was a ref
+                            if top_level_schema_name:
+                                schema_dict["x-fastmcp-top-level-schema"] = (
+                                    top_level_schema_name
+                                )
                             resp_info.content_schema[media_type_str] = schema_dict
                         except ValueError as e:
                             # Re-raise ValueError for external reference errors
@@ -625,6 +638,13 @@ class OpenAPIParser(
         for response in responses.values():
             if response.content_schema:
                 for content_schema in response.content_schema.values():
+                    # Check if this schema was originally a top-level $ref
+                    if "x-fastmcp-top-level-schema" in content_schema:
+                        schema_name = content_schema["x-fastmcp-top-level-schema"]
+                        if schema_name in all_schemas:
+                            needed_schemas.add(schema_name)
+
+                    # Extract all dependencies (transitive refs within the schema)
                     deps = self._extract_schema_dependencies(
                         content_schema, all_schemas
                     )
--- a/tests/experimental/openapi_parser/utilities/openapi/test_transitive_references.py
+++ b/tests/experimental/openapi_parser/utilities/openapi/test_transitive_references.py
@@ -786,3 +786,57 @@ class TestTransitiveAndNestedReferences:
         assert len(overlap) == 0, (
             f"Found overlapping schemas between input and output: {overlap}"
         )
+
+    def test_issue_2087_top_level_response_ref_includes_all_nested_schemas(self):
+        """Issue #2087: Top-level response $ref must include itself in response_schemas."""
+        spec = {
+            "openapi": "3.0.1",
+            "info": {"title": "Test", "version": "1.0"},
+            "paths": {
+                "/persons": {
+                    "get": {
+                        "responses": {
+                            "200": {
+                                "description": "OK",
+                                "content": {
+                                    "application/json": {
+                                        "schema": {
+                                            "$ref": "#/components/schemas/PersonList"
+                                        }
+                                    }
+                                },
+                            }
+                        }
+                    }
+                }
+            },
+            "components": {
+                "schemas": {
+                    "PersonList": {
+                        "properties": {
+                            "Items": {
+                                "type": "array",
+                                "items": {"$ref": "#/components/schemas/Person"},
+                            }
+                        }
+                    },
+                    "Person": {
+                        "properties": {
+                            "Name": {"$ref": "#/components/schemas/Name"},
+                            "Job": {"$ref": "#/components/schemas/Job"},
+                        }
+                    },
+                    "Name": {"properties": {"First": {"type": "string"}}},
+                    "Job": {"properties": {"Company": {"type": "string"}}},
+                }
+            },
+        }
+
+        routes = parse_openapi_to_http_routes(spec)
+        route = routes[0]
+
+        # Bug was: PersonList missing from response_schemas despite being top-level ref
+        assert "PersonList" in route.response_schemas
+        assert "Person" in route.response_schemas
+        assert "Name" in route.response_schemas
+        assert "Job" in route.response_schemas


yesankur:
Is this fix available in new version. When is this going to be released

laxuankhanh1902:
Is it available in new version?

PR #2128, opened 2025-10-17
Disable parallel test execution on Windows

jlowin:
Windows CI tests have been failing when using pytest workers. The parallel test execution (`--numprocesses auto --maxprocesses 4 --dist worksteal`) causes test processes to die unexpectedly on Windows runners.

This PR conditionally disables parallel execution on Windows while maintaining it on Linux:

```yaml
# Before: always parallel
run: uv run pytest ... --numprocesses auto --maxprocesses 4 --dist worksteal

# After: conditional based on OS
run: uv run pytest ... ${{ matrix.os == 'windows-latest' && '' || '--numprocesses auto --maxprocesses 4 --dist worksteal' }}
```

This should resolve the Windows test failures while keeping faster test execution on Linux runners.

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -48,7 +48,7 @@ jobs:
         run: uv sync --upgrade
 
       - name: Run tests (excluding integration and client_process)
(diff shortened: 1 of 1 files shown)


jlowin:
This syntax appeared to have failed, experimenting with alternative in #2119

PR #2137, opened 2025-10-18
Add storage backend documentation

jlowin:
Adds a new document with an overview of storage backends

cc @strawgate

### DIFF
--- a/docs/servers/middleware.mdx
+++ b/docs/servers/middleware.mdx
@@ -461,7 +461,7 @@ from fastmcp.server.middleware.caching import ResponseCachingMiddleware
 mcp.add_middleware(ResponseCachingMiddleware())
 ```
 
-Out of the box, it caches call/list tool, resources, and prompts to an in-memory cache with TTL-based expiration. Sending a notification of a tool/resource/prompt change will invalidate the cache for the affected method. List calls are stored under global keys—when sharing a storage backend across multiple servers, consider namespacing collections to prevent conflicts. See [Storage Backends](/servers/storage-backends) for advanced configuration options.
+Out of the box, it caches call/list tool, resources, and prompts to an in-memory cache with TTL-based expiration. Cache entries expire based on their TTL; there is no event-based cache invalidation. List calls are stored under global keys—when sharing a storage backend across multiple servers, consider namespacing collections to prevent conflicts. See [Storage Backends](/servers/storage-backends) for advanced configuration options.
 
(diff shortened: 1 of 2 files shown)


docs/servers/middleware.mdx
```diff
@@ -461,9 +461,9 @@ from fastmcp.server.middleware.caching import ResponseCachingMiddleware
 mcp.add_middleware(ResponseCachingMiddleware())
 ```
 
-Out of the box, it caches call/list tool, resources, and prompts to an in-memory cache. Sending a notification of a tool/resource/prompt change will invalidate the cache for the affected method. List calls are stored under global keys, if you share a key_value backend across servers, keep this in mind and consider using the PrefixCollectionsWrapper in py-key-value-aio to namespace collections by server.
+Out of the box, it caches call/list tool, resources, and prompts to an in-memory cache with TTL-based expiration. Sending a notification of a tool/resource/prompt change will invalidate the cache for the affected method. List calls are stored under global keys—when sharing a storage backend across multiple servers, consider namespacing collections to prevent conflicts. See [Storage Backends](/servers/storage-backends) for advanced configuration options.
```
strawgate: No invalidation, not sure how I missed this doc

docs/servers/storage-backends.mdx
```diff
@@ -0,0 +1,259 @@
...
+- **RocksDB** - Embedded high-performance key-value store
+- **Valkey** - Redis-compatible server
+
+For configuration details on these backends, consult the [py-key-value-aio documentation](https://github.com/strawgate/py-key-value-aio).
+
+## Use Cases in FastMCP
+
+### Server-Side OAuth Token Storage
+
+The [OAuth Proxy](/servers/auth/oauth-proxy) and OAuth auth providers use storage for persisting OAuth client registrations and encrypted upstream tokens. By default, clients are stored on disk:
```
strawgate: Is this still true?

jlowin: hm no

strawgate:
At some point we could also recommend the wrappers like the passthrough cache wrapper which allows you to combine an in memory store with a remote store like redis `PassthroughCacheWrapper(primary_key_value=redis_store, cache_key_value=memory_store)`

(Or we could directly just wrap the user provided store in a passthrough cache store)

You can also use a ttl clamp on the memory store part which would keep entries in memory with a very short ttl (like 5 minutes) while the redis store uses the intended ttl.

I'll spend some time adding additional docs to the key value library about wrappers too that does a better job describing the options

PR #2142, opened 2025-10-19
Add ToolInjectionMiddleware + Tools for Read/List Resource/Prompt for Client Compat

### LINKED ISSUE #2140: Contrib Module: List/Get Resource and Prompt Tools
### Enhancement

We should offer a contrib module which adds four tools to the server: get_prompt, list_prompts, get_resource, list_resources which act as compatibility shims for clients that only support tools

strawgate:
Implements "Tool Injection" Middleware which allows inserting tools via middleware.

Adds two middleware components based on Tool Injection middleware: `ResourceToolMiddleware` and `PromptToolMiddleware` which add tools that expose list/read resources and list/read prompts as callable tools for clients that only support the tools capability.

## Features

- `list_resources`: List all available resources
- `get_resource`: Read a resource by URI
- `list_prompts`: List all available prompts
- `get_prompt`: Get a prompt with optional arguments

Closes #2140

### DIFF
--- a/.github/workflows/martian-issue-triage.yml
+++ b/.github/workflows/martian-issue-triage.yml
@@ -4,15 +4,16 @@ on:
   issues:
     types: [opened, labeled]
 
-concurrency:
-  group: triage-martian-${{ github.event.issue.number }}
-  cancel-in-progress: true
-
 jobs:
   martian-issue-triage:
     if: |
       (github.event.action == 'opened' && github.actor == 'strawgate') ||
       (github.event.action == 'labeled' && github.event.label.name == 'triage-martian')
+
+    concurrency:
+      group: triage-martian-${{ github.event.issue.number }}
+      cancel-in-progress: true
+
     runs-on: ubuntu-latest
     timeout-minutes: 10
     permissions:
--- a/.github/workflows/marvin-dedupe-issues.yml
+++ b/.github/workflows/marvin-dedupe-issues.yml
@@ -83,7 +83,6 @@ jobs:
           allowed_non_write_users: "*"
           claude_args: |
             --allowedTools Bash(gh issue view:*),Bash(gh search:*),Bash(gh issue list:*),Bash(gh api:*),Bash(gh issue comment:*),Task
-            --mcp-config /tmp/mcp-config/mcp-servers.json
           settings: |
             {
               "model": "claude-sonnet-4-5-20250929",


strawgate:
/marvin let's just return lists of prompts, list resources, and get prompt result and read resource contents objects, basically just return what the client returns? Why not just do that?

jlowin:
Love this -- to the point that I think we should soon consider some sort of "easy to add" toolkit like this and graduate from contrib.

jlowin:
I was thinking today these could be fantastic middleware. That way users don't have to worry about adding tools etc they do something like

```python
mcp = FastMCP(
    "my example server",
    middleware=[ResourceToolMiddleware(), PromptToolMiddleware()],
)
```

names TBD. Has the advantage of removing these tools from more typical tool management, tag inclusion, etc. At least I think so

strawgate:
@jlowin I was also considering adding these via middleware; I'll try it out this week.

strawgate:
needs docs

src/fastmcp/server/middleware/tool_injection.py
```diff
@@ -0,0 +1,125 @@
...
+)
+
+
+async def read_resource(
+    context: Context,
+    uri: Annotated[AnyUrl | str, "The URI of the resource to read."],
+) -> list[mcp.types.TextResourceContents | mcp.types.BlobResourceContents]:
+    """Read a resource available on the server."""
+
+    async with Client[FastMCPTransport](context.fastmcp) as client:
```
jlowin: this could be replaced with `await context.read_resource(uri)` which will still invoke all middleware and everything but not need to start a new session and do an initialize handshake, which could be an advantage

I notice we don't have equivalents for all methods (like there's no `context.list_resources()`) but we could add it if it makes this simpler

strawgate: done for this one

strawgate:
> This is really great. I think these survive e.g. include_tags, right? like if the user does FastMCP(include_tags='abc', middleware=[ResourcesMiddleware()]) we expose the resources tools properly? If it doesn't I'd view it as a vote to remove include_tags / exclude_tags and replace THEM with middleware!

@jlowin yes all injected tools survive include_tags -- i've added a test to cover this

jlowin (approved):

PR #2158, opened 2025-10-20
Rename OAuth providers to include DCR suffix

jlowin:
FastMCP currently provides the `OAuthProxy` class to bridge MCP's Dynamic Client Registration (DCR) expectations to OAuth providers that don't natively support DCR. This works by proxying the MCP DCR flow to traditional OAuth flows.

With [SEP 991](https://github.com/modelcontextprotocol/specification/pull/991) introducing a new OAuth mechanism where `client_id` is an HTTPS URL for metadata fetching, we'll likely need new proxy implementations in the future. To prepare for this, we need clear naming that distinguishes which OAuth pattern each provider uses.

We are going to introduce this naming as part of 2.13.0.

This PR renames all OAuth provider classes to include a "DCR" suffix, making it explicit they proxy to Dynamic Client Registration:

```python
from fastmcp.server.auth.providers.github import GitHubDCRProvider

auth = GitHubDCRProvider(
    client_id="your-client-id",
    client_secret="your-client-secret",
    base_url="https://your-server.com"
)
```

## Changes

- Renamed 6 provider classes: GitHub, Google, Azure, WorkOS, Auth0, AWS Cognito
- Added backwards-compatible aliases with deprecation warnings
- Updated environment variables to include `_DCR_` with fallback to old names
- Updated all documentation and examples

Old names still work with deprecation warnings when `settings.deprecation_warnings` is enabled. However, this PR should be marked as a `breaking change` to call attention to it.


jlowin:
On further examination, this PR is actually a bad outcome which would have us implement different providers for every possible auth flow. SEP-991 is quite clear that it is a third approach for registering clients and fits otherwise neatly into the oauth flow we already expose. Creating DCR-specific proxies and "SEP-991"-specific proxies would introduce massive code duplication and user confusion. So it was a (valuable? interesting? distracting?) exercise but at this time there's no need to rename our auth objects. Once SEP-991 is adopted we can simply add the apporpriate logic for client ID detection.

PR #2202, opened 2025-10-22
ci: fix test workflow triggers and acknowledge test warnings

zzstoatzz:
only run unit/integration tests on library/test/ci changes and [fix warnings](https://github.com/jlowin/fastmcp/actions/runs/18702622228/job/53334362692?pr=2202#step:5:68)

## ai yap summary

- Fix CI workflow to only run tests when relevant files change
- Explicitly acknowledge expected warnings in tests instead of suppressing them

### CI Workflow
Added `paths` filter to `pull_request` trigger in `.github/workflows/run-tests.yml` so tests only run when:
- `src/**`
- `tests/**`
- `uv.lock`
- `pyproject.toml`
- `.github/workflows/**`

This prevents unnecessary test runs on PRs that only touch examples or documentation.

### Test Warnings
Replaced warning suppression with explicit assertions using `pytest.warns()`:

1. **Deprecation warning** for `server.sse_app()` in `tests/client/test_sse.py`
   - Validates that the deprecation warning is correctly shown to users

2. **UserWarning for in-memory token storage** in OAuth tests:
   - `tests/client/auth/test_oauth_client.py`
   - `tests/client/test_client.py`
   - `tests/test_mcp_config.py`
   - Validates that users get warned about production usage

These changes ensure warnings are intentionally tested rather than silently suppressed, confirming they work as designed.

### DIFF
--- a/tests/client/test_client.py
+++ b/tests/client/test_client.py
@@ -968,26 +968,52 @@ class TestAuth:
             Client(transport=StdioTransport("echo", ["hello"]), auth="oauth")
 
     def test_oauth_literal_sets_up_oauth_shttp(self):
-        client = Client(
-            transport=StreamableHttpTransport("http://localhost:8000"), auth="oauth"
-        )
+        # Expect warning about in-memory token storage
+        with pytest.warns(
+            UserWarning,
+            match="Using in-memory token storage is not recommended for production use",
+        ):
+            client = Client(
+                transport=StreamableHttpTransport("http://localhost:8000"), auth="oauth"
+            )
         assert isinstance(client.transport, StreamableHttpTransport)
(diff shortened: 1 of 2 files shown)


.github/workflows/run-tests.yml
```diff
@@ -14,8 +14,13 @@ on:
       - "pyproject.toml"
       - ".github/workflows/**"
 
-  # run on all pull requests because these checks are required and will block merges otherwise
```
jlowin, line 17: @zzstoatzz the issue is that if someone opens a doc PR, these tests wont run and they're required to pass in order to merge (I can override, and do, but i wish there was a better way)

PR #2204, opened 2025-10-22
fix: apply prefix to tool name when mounting servers

zzstoatzz:
## problem

when mounting a server with a prefix, the tool's `key` field gets prefixed but the `name` field doesn't. this creates an inconsistency where:

```json
{
  "key": "hue_read_all_lights",
  "name": "read_all_lights"
}
```

this breaks chatmcp UI and other clients that expect `key` and `name` to be consistent when using tools.

## solution

apply the prefix to both `key` and `name` fields when mounting, matching how resources and templates already work:

```python
tool = tool.model_copy(
    key=key,
    update={"name": f"{mounted.prefix}_{tool.name}"}
)
```

this makes tool name prefixing consistent across all component types (tools, resources, templates, prompts).

## example

```python
from fastmcp import FastMCP

# child server
lights = FastMCP("Lights")

@lights.tool
def read_all_lights() -> list[str]:
    return ["light1", "light2"]

# parent server mounts with prefix
hub = FastMCP("Hub")
hub.mount(lights, prefix="hue")

# after fix: both key and name are prefixed
tools = await hub.get_tools()
tool = tools["hue_read_all_lights"]
assert tool.key == "hue_read_all_lights"
assert tool.name == "hue_read_all_lights"  # ✅ now consistent
```

### DIFF
--- a/src/fastmcp/server/server.py
+++ b/src/fastmcp/server/server.py
@@ -464,7 +464,12 @@ class FastMCP(Generic[LifespanResultT]):
                 child_tools = await mounted.server.get_tools()
                 for key, tool in child_tools.items():
                     new_key = f"{mounted.prefix}_{key}" if mounted.prefix else key
-                    all_tools[new_key] = tool.model_copy(key=new_key)
+                    all_tools[new_key] = tool.model_copy(
+                        key=new_key,
+                        update={"name": f"{mounted.prefix}_{tool.name}"}
+                        if mounted.prefix
+                        else {},
+                    )
             except Exception as e:
                 logger.warning(
                     f"Failed to get tools from mounted server {mounted.server.name!r}: {e}"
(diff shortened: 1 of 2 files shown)


jlowin:
Are you sure this inconsistency is the problem? I think `key` is an internal thing we created for FastMCP bookkeeping of mounted servers, it doesn't get revealed through the protocol. On a given tool, `key` is how it tracks its prefixes without actually changing the tool name (since it might be referenced in multiple mounted servers including its own parent server). in  `FastMCP._list_tools_mcp`, the `key` is turned into the tool's name

zzstoatzz:
After tracing through the full flow, I believe this fix is correct. Here's why:

1. **FastMCP Cloud uses `fastmcp inspect -f fastmcp`** (not MCP format) - see infrastructure/lambdas/mcp_builder/codebuild.py:499
2. **The fastmcp format includes BOTH `key` AND `name`** - see src/fastmcp/utilities/inspect.py:122-126
3. **FastMCP Cloud reads `tool.name`** from the manifest - see web/src/server/functions/build.ts:511

The flow:
- Mounting prefixes `tool.key` but not `tool.name`
- `fastmcp inspect` creates manifest with `{key: 'hue_read_all_lights', name: 'read_all_lights'}`
- FastMCP Cloud UI uses `tool.name` → tries to call 'read_all_lights'
- MCP protocol only knows about 'hue_read_all_lights' → fails

While you're correct that `_list_tools_mcp()` uses `tool.key` as the MCP protocol name, the fastmcp format manifest (which is what gets stored) has separate `key` and `name` fields, and only `key` is being prefixed.

The fix: prefix both `key` AND `name` when mounting with a prefix.

PR #2206, opened 2025-10-22
Add meta to call tool

### LINKED ISSUE #1315: Allow clients to send _meta to server
### Enhancement Description

Hello, guys!

First, thanks (and congrats) for the great work on FastMCP!

Since @jlowin’s [PR #1294](https://github.com/jlowin/fastmcp/pull/1294), servers are now allowed to send metadata in the _meta field to clients.

I’d like to propose (and possibly implement) support for the opposite direction: allowing clients to send _meta arguments to the server.

### Use Case
Agents sometimes need to send data during tool calls that don’t qualify as function arguments or HTTP headers. This information could be, for example, user preferences and might determine how tools (and other resources such as prompts) are processed in the server.

In any case, the protocol specification does not forbid MCP clients from sending _meta to MCP servers, so I believe this wouldn't be a bad design decision.

**Are there any alternative solutions you've considered?**
I’ve considered using HTTP request headers, but I don’t think that would cover all use cases and security necessities.

### Proposed Implementation

Since FastMCP relies on SDK's `ClientSession.call_tool` interface, it would be required to directly use `ClientSession.send_request` ([unless `_meta` is also exposed in the SDK](https://github.com/modelcontextprotocol/python-sdk/issues/1216))

```Python
    async def call_tool_mcp(
        self,
        name: str,
        arguments: dict[str, Any],
        progress_handler: ProgressHandler | None = None,
        timeout: datetime.timedelta | float | int | None = None,
        meta: dict | None = None # <-- Expose meta in the public interface
    ) -> mcp.types.CallToolResult:
        if isinstance(timeout, int | float):
            timeout = datetime.timedelta(seconds=timeout)

        # Since self.session.call_tool doesn't expose meta, it would be required to directly use self.session.send_request
        result = await self.session.send_request(
            types.ClientRequest(
                types.CallToolRequest(
                    method="tools/call",
                    params=types.CallToolRequestParams(
                        name=name,
                        arguments=arguments,
                        meta=meta <-- Pass metadata down to the MCP SDK
                    ),
                )
            ),
            types.CallToolResult,
            request_read_timeout_seconds=read_timeout_seconds,
            progress_callback=progress_callback,
        )
        return result
```

aiorga-sherpas:
## Description
Minor changes to add the meta argument to the call_tool method.
https://github.com/modelcontextprotocol/python-sdk/pull/1231 has been merged and will be in the next mcp release. With that and this minor changes we fix https://github.com/jlowin/fastmcp/issues/1315

- [x] My change closes #1315 

**Review Checklist**

---

### DIFF
--- a/.github/workflows/marvin.yml
+++ b/.github/workflows/marvin.yml
@@ -44,11 +44,10 @@ jobs:
       - name: Install dependencies
         run: uv sync --python 3.12
 
-      # Install pre-commit hooks automatically
-      - name: Install pre-commit hooks
-        run: |
-          uv run pre-commit install
-          echo "✅ Pre-commit hooks installed"
+      - name: Run prek
+        uses: j178/prek-action@v1
+        env:
+          SKIP: no-commit-to-branch
 
       - name: Generate Marvin App token
         id: marvin-token
--- a/.github/workflows/run-static.yml
+++ b/.github/workflows/run-static.yml
@@ -48,7 +48,7 @@ jobs:
             exit 1
           fi
           echo "✅ Lockfile is up to date"
-      - name: Run pre-commit
-        run: uv run pre-commit run --all-files
+      - name: Run prek
+        uses: j178/prek-action@v1
         env:
           SKIP: no-commit-to-branch
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -79,10 +79,10 @@ jobs:
         run: uv sync --resolution lowest-direct
 
       - name: Run tests (excluding integration and client_process)
-        run: uv run pytest --inline-snapshot=disable tests -m "not integration and not client_process" --numprocesses auto --maxprocesses 4 --dist worksteal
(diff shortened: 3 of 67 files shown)


jlowin:
Thanks @aiorga-sherpas -- can you confirm whether this requires `mcp>=1.19`? 

On the one hand I believe that the SDK generally ignores extra inputs in which case this does *not* require raising the pin to 1.19. On the other hand, users who provide this on `mcp<1.19` may be surprised when it doesn't work. 

I am comfortable raising the minimum in `pyproject.toml` as generally we won't always be able to support older versions as we stay on the cutting edge of functionality, though for non-breaking changes like this it is an option to raise an error if users pass `meta` when calling a tool and their MCP version is < 1.19

aiorga-sherpas:
If meta argument is used with `mcp<1.19` the following warning is printed: `Warning: Could not call tool sum: ClientSession.call_tool() got an unexpected keyword argument 'meta'` and the tool execution is skipped.
I could add an additional warning about having to update `mcp>=1.19` and try to execute it without sending the meta.

PR #2210, opened 2025-10-22
first class support for apps sdk

### LINKED ISSUE #2014: Support for OpenAI Apps SDK
### Enhancement

OpenAI announced the Apps SDK, an extension of mcp-ui. Let’s support this as a first class FastMCP extension.

aaazzam:
# Add native OpenAI widget support

FastMCP now provides first-class support for OpenAI's ChatGPT widgets through the `@app.ui.openai.widget` decorator. Previously, developers had to manually manage widget HTML resources, construct OpenAI-specific metadata, and transform return values to match OpenAI's `{"content": [...], "structuredContent": {...}}` wire format. This PR eliminates that boilerplate entirely—just decorate your function and return a `str`, `dict`, or `tuple[str, dict]`, and FastMCP handles the rest.

The decorator automatically registers widget HTML as an MCP resource, injects OpenAI metadata (CSP policies, widget templates, invocation status messages), and wraps your function to auto-transform its return value. You write clean Python functions that return natural data types, and the framework translates them to OpenAI's format behind the scenes.

```python
from fastmcp import FastMCP

app = FastMCP("PizzaTracker")

@app.ui.openai.widget(
    name="pizza-map",
    template_uri="ui://widget/pizza-map.html",
    html='<div id="root"></div><script src="https://cdn.com/widget.js"></script>',
    invoking="Hand-tossing a map",
    invoked="Served a fresh map"
)
def show_pizza_map(topping: str) -> dict:
    """Show an interactive pizza map."""
    return {"topping": topping}  # Auto-transformed to OpenAI format!
```

The decorator follows FastMCP's existing conventions, using `name=` (consistent with `@app.tool(name=...)`) and exposing all customization points: `widget_description`, `widget_csp_resources`, `widget_csp_connect`, `widget_prefers_border`, and more. Smart defaults handle the common cases—for instance, `widget_description` defaults to `"{title} widget UI."` when not specified.

This makes OpenAI widgets feel like a natural extension of FastMCP's tool system rather than a separate SDK layer. See `examples/widget_test/` for three working patterns showing dict-only, string-only, and combined return types.

Resolves #2014 

### DIFF
--- a/examples/widget_test/README.md
+++ b/examples/widget_test/README.md
@@ -75,7 +75,7 @@ Returns structured data only, which the widget renders:
 
 ```python
 @app.ui.openai.widget(
-    identifier="pizza-map",
+    name="pizza-map",
     template_uri="ui://widget/pizza-map.html",
     html=PIZZAZ_HTML,
     invoking="Hand-tossing a map",
@@ -91,7 +91,7 @@ Returns both narrative text and structured data:
 
 ```python
 @app.ui.openai.widget(
-    identifier="pizza-tracker",
+    name="pizza-tracker",
     template_uri="ui://widget/pizza-tracker.html",
     html=PIZZAZ_HTML,
     invoking="Tracking your pizza",
@@ -109,7 +109,7 @@ Returns text only, no structured data:
 
 ```python
 @app.ui.openai.widget(
-    identifier="pizza-status",
+    name="pizza-status",
     template_uri="ui://widget/pizza-status.html",
     html=PIZZAZ_HTML,
 )
--- a/examples/widget_test/pizzaz_server.py
+++ b/examples/widget_test/pizzaz_server.py
@@ -1,0 +1,100 @@
+"""Pizzaz Widget Demo Server
+
+This demonstrates FastMCP's OpenAI widget support using the Pizzaz mapping library.
+The widget renders an interactive map when you ask about pizza toppings.
+
+Example usage:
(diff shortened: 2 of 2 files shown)


src/fastmcp/ui/openai/response.py
```diff
@@ -0,0 +1,135 @@
+"""Response transformation utilities for OpenAI widgets."""
+
+from __future__ import annotations
+
+from typing import Any
+
+WidgetToolResponse = dict[str, Any]
```
BrandonShar, line 7: Would it make sense to introduce this as a dataclass/pydantic from the beginning? Support for `_meta` fields will land in the python sdk tomorrow and that's part of what's necessary to support openai's full protocol. I think `string|dict|WidgetToolResponse` could make a lot of sense instead of extending the tuple or supporting different return types. Something like a simpler version of the `CallToolResult` like:

```python
class WidgetToolResponse(BaseModel)
   content: str
   structured_content: dict[str, Any] | None = None
   meta: dict[str, Any] | None = None

aaazzam:
Nice! Welcome your feedback. Lay it on me!

BrandonShar:
Just PR'd https://github.com/jlowin/fastmcp/pull/2283 to hopefully make meta usage simpler!

jlowin:
#2283 is merged!

jlowin:
Closing as stale

PR #2223, opened 2025-10-23
Derive `jwt_signing_key` from Client Secret, default to Encrypted Disk Store

strawgate:
## Description
Derive `jwt_signing_key` from the Client Secret if not provided for OAuth Providers that issue JWTs. This signing key should be a key provided by the developer, obtained via the environment or externally, or should be a string passed by the developer that we use to derive a JWT signing key.

If the developer does not pass client_storage, we will initialize an encrypted disk store for the data with a key based off the jwt_signing_key. Changing the signing key will invalidate tokens and storage which will require special consideration by the developer.

If the developer provides client_storage, they are on the hook for securing the data including the decision to use an encrypted store.

**Contributors Checklist**

**Review Checklist**

---

### DIFF
--- a/docs/deployment/http.mdx
+++ b/docs/deployment/http.mdx
@@ -566,7 +566,7 @@ Production requires explicit key management to ensure tokens survive restarts an
 
 Add two parameters to your auth provider:
 
-```python {4-11}
+```python {8-12}
 from key_value.aio.stores.redis import RedisStore
 from key_value.aio.wrappers.encryption import FernetEncryptionWrapper
 from cryptography.fernet import Fernet
@@ -576,7 +576,7 @@ auth = GitHubProvider(
     client_secret=os.environ["GITHUB_CLIENT_SECRET"],
     jwt_signing_key=os.environ["JWT_SIGNING_KEY"],
     client_storage=FernetEncryptionWrapper(
-        key_value=RedisStore(host="redis.example.com", ...),
+        key_value=RedisStore(host="redis.example.com", port=6379),
         fernet=Fernet(os.environ["STORAGE_ENCRYPTION_KEY"])
     ),
     base_url="https://your-server.com"  # use HTTPS


strawgate:
HKDF is a fast key derivation function for scenarios with high-entropy inputs, like a pre-shared key from a Diffie-Hellman exchange. We should not be using HKDF for deriving keys from passwords or short strings.

We should use PBKDF2HMAC as it's designed for password-based key derivation.

I'm going to move this into py-key-value as a function of a new FernetEncryptionWrapper once https://github.com/strawgate/py-key-value/pull/77 is merged tonight/tomrrow

src/fastmcp/server/auth/jwt_issuer.py
```diff
@@ -40,51 +39,51 @@ def derive_jwt_key(upstream_secret: str, server_salt: str) -> bytes:
...
-    return HKDF(
-        algorithm=hashes.SHA256(),
-        length=32,
-        salt=salt.encode(),
-        info=info,
-    ).derive(secret_bytes)
+    ).derive(from_secret.encode())
+
+
+# def derive_encryption_key(from_secret: str, salt: str) -> bytes:
```
strawgate: remove

docs/deployment/http.mdx
```diff
@@ -557,30 +557,26 @@ This automatic approach is convenient for development but not suitable for produ
...
 3. **Persistent network-accessible storage** for encrypted upstream tokens
 
-The keys can be any secret strings (environment variables, secret manager, etc.) and should be different from each other. FastMCP derives proper cryptographic keys from whatever you provide using HKDF.
-
 **Configuration:**
 
-Add three parameters to your auth provider:
+Add two parameters to your auth provider:
 
 ```python {4-7}
```
jlowin:
```suggestion
```python {4-5}
```

PR #2227, opened 2025-10-23
Added to_data_uri method for Image class.

### LINKED ISSUE #2225: FEAT: Utility function for converting local images to Data URIs.
### Enhancement

With the introduction of the support for Icons, in the server, we could have a helper function which can convert local images into base64 data URIs since `Icon.src` supports it. This would be especially useful for serving images during development, when there's no domain available and avoid any CORS restrictions if implemented on Client's end.

Hyperclaw79:
## Description

* Added a handy `to_data_uri` method to the `Image` class to provide ability to serve Images as icons.
    Example:
    ```py
    Icon(src=Image(path="/local/path.img.png").to_data_uri())
    ```
* Upgraded the `_get_mime_type` function from relying on hardcoded dictionary to using the `mimetypes.guess_type` function, making it more aligned with other models like `File`, etc.
* Parametrized the `test_get_mime_type_from_path` unit test to align it with pytest best practices.

**Contributors Checklist**

- [x] My change closes #2225 

**Review Checklist**

---

## Summary by CodeRabbit

* **New Features**
  * Added image-to-data-URI conversion and improved handling of local image paths (expands env vars/home).
  * Image content generation now uniformly supports embedded image data.

* **Bug Fixes**
  * Improved MIME type detection, including better WebP support and safer fallbacks.

* **Documentation**
  * Added example showing generating icons from local image files using data URIs.

* **Tests**
  * Broadened test coverage with parameterized tests for MIME types and data-URI generation.

### DIFF
--- a/docs/servers/icons.mdx
+++ b/docs/servers/icons.mdx
@@ -115,6 +115,7 @@ For small icons or when you want to embed the icon directly, use data URIs:
 
 ```python
 from mcp.types import Icon
+from fastmcp.utilities.types import Image
 
 # SVG icon as data URI
 svg_icon = Icon(
@@ -125,5 +126,14 @@ svg_icon = Icon(
 @mcp.tool(icons=[svg_icon])
 def my_tool() -> str:
     """A tool with an embedded SVG icon."""
+    return "result"
+
+# Generating a data URI from a local image file.
+img = Image(path="./assets/brand/favicon.png")
+icon = Icon(src=img.to_data_uri())
+
+@mcp.tool(icons=[icon])
+def file_icon_tool() -> str:
+    """A tool with an icon generated from a local file."""
     return "result"
 ```


src/fastmcp/utilities/types.py
```diff
@@ -233,6 +243,21 @@ def to_image_content(
             annotations=annotations or self.annotations,
         )
 
+    def to_data_uri(self, mime_type: str | None = None) -> str:
+        """Get image as a data URI."""
+        data = self._get_data()
+        return f"data:{mime_type or self._mime_type};base64,{data}"
+
+    @classmethod
+    def path_to_data_uri(cls, path: str | Path, mime_type: str | None = None) -> str:
```
jlowin: This classmethod feels a little awkward to me -- typically in the codebase we would create a utility function for something like this. I think e.g. `icons=[Icon[src=Image.path_to_data_uri(path)]]` isn't a helpful or intuitive grouping. In fact I'd prefer `icons=[Icon[src=Image(path=path).to_data_uri()]]` under all circumstances as its much more self documenting!

I'm also surprised that this classmethod doesn't actually use the to_data_uri() method? Feels like there are two separate logic paths -- I'd reccomend removing it.

Hyperclaw79: I had added it in order to increase the support for default mimetypes for images without touching the existing `get_mime_type` function. 
But yeah, based on the way you used it inside Icon definition, it does feel awkward to have it. Will get rid of it. 
While on the topic, are you open to replacing the get_mimetype function having the hardcoded subset of images with the `mimetypes.guess_type` function?

jlowin: Yup, it looks like `File` already does this, but `Image` and `Audio` could be updated: https://github.com/jlowin/fastmcp/blob/main/src/fastmcp/utilities/types.py#L328

Hyperclaw79: Fixed in the latest commit. (Haven't touched Audio since that would stray away from the purpose of this PR. In the future, we could probably standardize it across the different types instead of per class implementation.)

src/fastmcp/utilities/types.py
```diff
@@ -190,41 +190,49 @@ def __init__(
...
-                ".gif": "image/gif",
-                ".webp": "image/webp",
-            }.get(suffix, "application/octet-stream")
+            # Workaround for WEBP in Py3.10
+            mimetypes.add_type("image/webp", ".webp")
+            resp = mimetypes.guess_type(self.path, strict=False)
+            if resp and resp[0] is not None:
+                return resp[0]
+            return "application/octet-stream"
         return "image/png"  # default for raw binary data
```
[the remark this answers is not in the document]
Hyperclaw79: This mimetype registry requirement is local to the function and is currently a workaround. We can get rid of it after dropping support for Py3.10.

src/fastmcp/utilities/types.py
```diff
@@ -190,41 +190,49 @@ def __init__(
...
+        self.path = self._get_expanded_path(path)
         self.data = data
         self._format = format
         self._mime_type = self._get_mime_type()
         self.annotations = annotations
 
+    @staticmethod
+    def _get_expanded_path(path: str | Path | None) -> Path | None:
+        """Expand environment variables and user home in path."""
+        return Path(os.path.expandvars(str(path))).expanduser() if path else None
```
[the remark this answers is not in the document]
Hyperclaw79: Outside the scope of PR. Can be tackled in a different PR for type standardization.

tests/utilities/test_types.py
```diff
@@ -227,6 +228,27 @@ def test_to_image_content_error(self, monkeypatch):
...
+        test_data = b"fake image data"
+        img_path.write_bytes(test_data)
+
+        img = Image(path=img_path)
+        data_uri = img.to_data_uri(mime_type=mime_type)
+
+        expected_data_uri = (
+            f"data:{expected_mime};base64,{base64.b64encode(test_data).decode()}"
+        )
+        assert data_uri == expected_data_uri
```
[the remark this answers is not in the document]
Hyperclaw79: That's not the target use case for this particular function.

docs/servers/icons.mdx
```diff
@@ -126,4 +127,13 @@ svg_icon = Icon(
...
     return "result"
+
+# Generating a data URI from a local image file.
+img = Image(path="./assets/brand/favicon.png")
+icon = Icon(src=img.to_data_uri())
+
+@mcp.tool(icons=[icon])
+def file_icon_tool() -> str:
+    """A tool with an icon generated from a local file."""
+    return "result"
```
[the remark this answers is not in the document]
Hyperclaw79: The whole purpose for this function is to serve local files and that's up to the users. So the changes you're mentioned are out of context.

src/fastmcp/utilities/types.py
```diff
@@ -190,41 +190,49 @@ def __init__(
...
-            suffix = self.path.suffix.lower()
-            return {
-                ".png": "image/png",
-                ".jpg": "image/jpeg",
-                ".jpeg": "image/jpeg",
-                ".gif": "image/gif",
-                ".webp": "image/webp",
-            }.get(suffix, "application/octet-stream")
+            # Workaround for WEBP in Py3.10
+            mimetypes.add_type("image/webp", ".webp")
```
jlowin, line 211: In this implementation, `mimetypes.add_type()` is being called on every Image instantiation - wouldn't it be better once at module import?

jlowin (commented):
One minor question

src/fastmcp/utilities/types.py
Hyperclaw79: I've explicitly kept it here to isolate it from the other classes as webp is only related to images and the only mimetype (being used) which isn't supported in Python <=3.10. After EOL for 3.10, we can completely drop this line. Keeping it at a module level might make it easier to forget to clean it up. Moreover, `add_type` is just a dictionary manipulation under the hood, so it's an inexpensive operation even when called multiple times. Lmk if you still prefer it at the module level, I can quickly make that change.

jlowin: Ok, I'm comfortable with that - I wasn't sure if it was an idempotent operation.

Hyperclaw79:
Thank you for the merge! Glad to have contributed to this wonderful library 😄

PR #2232, opened 2025-10-24
Add exc_info=True to inspect command error logging

### LINKED ISSUE #2231: Add exc_info=True to inspect command error logging for better debugging
## Problem

When the `fastmcp inspect` command fails (e.g., missing dependencies, import errors, syntax errors in server code), the error logs only show the error message without the stack trace. This makes debugging significantly harder because you can't see WHERE in the code the error occurred.

## Solution

Add `exc_info=True` to the `logger.error()` call in the inspect command's exception handler (line 769 in `src/fastmcp/cli/cli.py`).

## Why This Matters

The inspect command loads server modules dynamically, involving complex:
- Import chains
- Dependency resolution  
- Dynamic module loading

Without the stack trace:
- Error message alone (e.g., "ModuleNotFoundError: No module named 'foo'") doesn't reveal which part of the server code is trying to import it
- No visibility into the call stack that led to the error
- Missing context for chained exceptions

With `exc_info=True`:
- Full stack trace showing exact file and line number where error occurred
- Complete call stack leading to the error
- Visibility into chained exceptions

## Current Code

```python
except Exception as e:
    logger.error(
        f"Failed to inspect server: {e}",
        extra={
            "server_spec": server_spec,
            "error": str(e),
        },
    )
```

## Proposed Change

```python
except Exception as e:
    logger.error(
        f"Failed to inspect server: {e}",
        extra={
            "server_spec": server_spec,
            "error": str(e),
        },
        exc_info=True,
    )
```

Reported by user feedback.

aaazzam:
When the `fastmcp inspect` command fails during server loading (missing dependencies, import errors, syntax errors), only the error message is logged without the stack trace. This makes debugging difficult because you can't see where in the code the error occurred.

Adding `exc_info=True` to the error logger includes the full stack trace, showing:
- Exact file and line number where the error occurred
- Complete call stack leading to the error  
- Chained exceptions

This is particularly valuable for inspect because it loads server modules dynamically through complex import chains and dependency resolution.

**Example:**

Without `exc_info=True`:
```
ERROR Failed to inspect server: ModuleNotFoundError: No module named 'foo'
```

With `exc_info=True`:
```
ERROR Failed to inspect server: ModuleNotFoundError: No module named 'foo'
Traceback (most recent call last):
  File "server.py", line 3, in <module>
    import foo
ModuleNotFoundError: No module named 'foo'
```

Fixes #2231

### DIFF
--- a/src/fastmcp/cli/cli.py
+++ b/src/fastmcp/cli/cli.py
@@ -772,6 +772,7 @@ async def inspect(
                 "server_spec": server_spec,
                 "error": str(e),
             },
+            exc_info=True,
         )
         console.print(f"[bold red]✗[/bold red] Failed to inspect server: {e}")
         sys.exit(1)


strawgate:
would `logger.exception` give the behavior you want?

PR #2305, opened 2025-10-29
Fix Chrome CSP blocking OAuth consent form with custom protocol redirects

### LINKED ISSUE #2302: Chrome Blocks OAuth Consent Form Due to CSP Redirect Chain Issue
### Description

The new OAuth consent page's "Allow Access" button does not work in Chrome (but works in Firefox) **when the user is already authorized with the OAuth provider** (e.g., GitHub) and when using MCP clients like Cursor that use custom protocol handlers (e.g., `cursor://`) for OAuth callbacks.

When clicking the "Allow Access" button, Chrome blocks the form submission with the following CSP error:

```
Refused to send form data to 'https://[server]/consent/submit' because it violates 
the following Content Security Policy directive: "form-action *". 
Note that '*' matches only URLs with network schemes ('http', 'https', 'ws', 'wss'), 
or URLs whose scheme matches `self`'s scheme. The scheme 'https:' must be added explicitly.
```

When the user is already authorized with the OAuth provider (e.g., GitHub), the OAuth flow includes a redirect chain that directly ends in a custom protocol scheme without intermediate authorization prompts:

```
POST /consent/submit (HTTPS)
  ↓ 302 redirect
https://github.com/login/oauth/authorize?...
  ↓ 302 redirect
cursor://anysphere.cursor-mcp/oauth/user-[app]/callback
```

The current CSP in the consent page HTML:

```html
<meta http-equiv="Content-Security-Policy" content="default-src 'none'; style-src 'unsafe-inline'; img-src https:; base-uri 'none'; form-action *" />
```

The `form-action *` wildcard only matches "network schemes" (http, https, ws, wss) according to the CSP specification. Since the redirect chain terminates in `cursor://` (a custom protocol scheme), Chrome seems to block the final redirect and the user is stuck on the authorization page.

### Version Information

```Text
FastMCP version:     2.13.0.2
MCP version:         1.19.0
Python version:      3.13.3
Chrome version:      141.0.7390.123
Firefox version:     144.0.2
```

jlowin:
Fixes #2302

Chrome blocks OAuth consent form submissions when redirect chains terminate in custom protocol schemes like `cursor://`. The CSP `form-action *` directive doesn't match custom protocols in Chrome's implementation, causing form submissions to be blocked even though the form itself submits to HTTPS.

The fix dynamically extracts the scheme from the redirect URI and includes it explicitly in the CSP directive. For a redirect URI like `cursor://anysphere.cursor-mcp/oauth/callback`, the CSP becomes `form-action https: http: cursor:` instead of `form-action *`, allowing Chrome to validate the entire redirect chain.

```python
# Parse redirect URI and include custom protocol scheme in CSP
parsed_redirect = urlparse(redirect_uri)
redirect_scheme = parsed_redirect.scheme.lower()
form_action_schemes = ["https:", "http:"]
if redirect_scheme and redirect_scheme not in ("http", "https"):
    form_action_schemes.append(f"{redirect_scheme}:")
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Strengthened consent page security by tightening form submission policies: form-action now allows only explicit schemes derived from the redirect target (plus standard http/https), removing the previous wildcard to reduce unsafe submissions and improve browser compatibility and redirect handling.

### DIFF
--- a/src/fastmcp/server/auth/oauth_proxy.py
+++ b/src/fastmcp/server/auth/oauth_proxy.py
@@ -365,9 +365,19 @@ def create_consent_html(
     )
 
     # Need to allow form-action for form submission
-    # Explicitly allow https: and http: schemes (not wildcard) to work with Chrome
-    # when redirect chains end in custom protocol schemes like cursor://
-    csp_policy = "default-src 'none'; style-src 'unsafe-inline'; img-src https:; base-uri 'none'; form-action https: http:"
+    # Chrome requires explicit scheme declarations in CSP form-action when redirect chains
+    # end in custom protocol schemes (e.g., cursor://). Parse redirect_uri to include its scheme.
+    parsed_redirect = urlparse(redirect_uri)
+    redirect_scheme = parsed_redirect.scheme.lower()
+
+    # Build form-action directive with standard schemes plus custom protocol if present
+    form_action_schemes = ["https:", "http:"]
+    if redirect_scheme and redirect_scheme not in ("http", "https"):
+        # Custom protocol scheme (e.g., cursor:, vscode:, etc.)
+        form_action_schemes.append(f"{redirect_scheme}:")
+
+    form_action_directive = " ".join(form_action_schemes)
+    csp_policy = f"default-src 'none'; style-src 'unsafe-inline'; img-src https:; base-uri 'none'; form-action {form_action_directive}"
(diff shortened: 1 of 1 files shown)


johanneswuerbach:
@jlowin any chance to get this fix released?

PR #2326, opened 2025-10-31
Add experimental in-process background tasks via Docket

chrisguidry:
Brings FastAPI-style background task execution to FastMCP as the next step toward [MCP SEP-1686](https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1686) support. 

When `FASTMCP_EXPERIMENTAL_ENABLE_DOCKET=true`, FastMCP automatically creates a Docket instance with an in-process Worker, enabling tools to schedule background tasks via dependency injection:

```python
from fastmcp.dependencies import CurrentDocket

@mcp.tool()
async def analyze_data(query: str, docket: Docket = CurrentDocket()) -> str:
    await docket.add(long_running_analysis)(query)
    return "Analysis started"
```

This provides Starlette-like background task capabilities without requiring distributed infrastructure. All tasks run in the same process via `memory://` broker. Distributed workers will come in a future phase.

Also adds `CurrentContext()` and `CurrentWorker()` dependency functions following Docket's pattern (no `Depends()` wrapper needed).

New `fastmcp.dependencies` module centralizes all dependency injection exports to keep the top-level namespace clean.

Related: https://linear.app/prefect/document/fastmcp-sep-1686-tasks-7c4711171324

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -52,6 +52,8 @@ openai = ["openai>=1.102.0"]
 [dependency-groups]
 dev = [
     "dirty-equals>=0.9.0",
+    # Dependency to support Docket memory:// brokers - only necessary until next fakeredis release
+    "fakeredis[lua] @ git+https://github.com/cunla/fakeredis-py.git@ad50a0de8d6dce554fb629ec284bc4ccbc6a7f12",
     "fastmcp[openai]",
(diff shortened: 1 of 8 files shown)


jlowin:
Holding temporarily as we sort out 2.14 roadmap offline; may be fine to include as an experimental feature regardless

PR #2357, opened 2025-11-03
fix: on_initialize is not using request params but the whole request

Maxi91f:
## Description
The type hint on `on_initialize` doesn't reflect the actual type being passed.

We can either:
1. Change what we send to the middleware method. I did this on [this commit](https://github.com/jlowin/fastmcp/commit/ecd3893785ec26512dc732457ae50872094b3ae9) on this branch. This will be a small breaking change for the users.
2. Change the type hint. But it will be different from the other methods on the middleware.

Before merging this we need to decide if we will take path 1 or path 2. I'm open to suggestions.

**Contributors Checklist**

- [ ] My change closes #(issue number) (Sorry, no issue, it's a quick change I found working on another issue)
- [x] I have tested my changes manually and by adding relevant tests (or changed the existing ones)

**Review Checklist**

---

### DIFF
--- a/.github/workflows/run-static.yml
+++ b/.github/workflows/run-static.yml
@@ -48,7 +48,7 @@ jobs:
             exit 1
           fi
           echo "✅ Lockfile is up to date"
-      - name: Run pre-commit
-        run: uv run pre-commit run --all-files
+      - name: Run prek
+        uses: j178/prek-action@v1
         env:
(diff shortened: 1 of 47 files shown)


jlowin:
Thank you - I think it's actually more correct to provide the entire request rather than just its params, even though by convention the params tend to be the interesting payload, in case the request gains other important fields.

PR #2371, opened 2025-11-04
Harden install commands against special characters

jlowin:
Apply shell escaping to server names in claude-code and gemini-cli install commands to prevent issues with special characters.


jlowin:
Superseded by #3522 which validates server names at the entry point instead of shell-escaping in list-mode subprocess calls.

PR #2400, opened 2025-11-08
Handle request_context availability during MCP initialization

### LINKED ISSUE #2393: Can't access `request_context` in `on_request` middleware methods in 2.13
### Description

We have some middleware that needs to access the request headers. In 2.12.4 this worked fine but with 2.13.0.2 this is raising a ValueError.

We can access this in `on_call_tool` and `on_read_resource` etc, but it'd be more DRY if we can continue doing this in a generic `on_request` handler.

### Example Code

```Python
import asyncio

from fastmcp.server.middleware import Middleware, MiddlewareContext


class RequestHostMiddleware(Middleware):
    async def on_request(self, context: MiddlewareContext, call_next):
        fastmcp_context = context.fastmcp_context

        # raises: ValueError: Context is not available outside of a request
        headers = fastmcp_context.request_context.request.headers

        fastmcp_context.request_host = headers["X-Forwarded-Host"]


mcp = FastMCP()
mcp.add_middleware(RequestHostMiddleware())


async def demo():
        client.list_tools()


if __name__ == "__main__":
    asyncio.run(demo())
```

### Version Information

```Text
FastMCP version:                                                           2.13.0.2
MCP version:                                                                 1.20.0
Python version:                                                             3.10.12
Platform:          Linux-6.17.4-orbstack-00308-g195e9689a04f-aarch64-with-glibc2.35
FastMCP root path:                               /venv/lib/python3.10/site-packages
```

jlowin:
In v2.13.0, the new `on_initialize` middleware hook runs during client connection before the MCP session is established. This causes `context.request_context` to be unavailable, leading to `ValueError` when middleware tries to access it during initialization.

This PR changes `request_context` to return `None` when the MCP session hasn't been established yet, rather than raising an error. Middleware can now check `if ctx.request_context:` to determine availability and use HTTP request helpers (like `get_http_headers()`) as a fallback when working with HTTP transports.

Documentation added to both `middleware.mdx` and `context.mdx` with examples showing the recommended pattern for handling this case.

Closes #2393

### DIFF
--- a/.github/workflows/martian-test-failure.yml
+++ b/.github/workflows/martian-test-failure.yml
@@ -1,0 +1,179 @@
+name: Marvin Test Failure Analysis
+
+on:
+  workflow_run:
+    workflows: ["Run Tests"]
+    types:
+      - completed
+
+concurrency:
+  group: marvin-test-failure-${{ github.event.workflow_run.head_branch }}
+  cancel-in-progress: true
+
+jobs:
+  martian-test-failure:
+    # Only run if the test workflow failed
+    if: ${{ github.event.workflow_run.conclusion == 'failure' }}
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      pull-requests: write
+      issues: read
+      id-token: write
+      actions: read # Required for Claude to read CI results
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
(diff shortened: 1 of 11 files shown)


jlowin:
Ignoring the integration test failure which is a separate issue

PR #2413, opened 2025-11-11
Scalekit provider updates

### LINKED ISSUE #2412: Scalekit Auth Provider Updates
### Enhancement

The following changes are being requested for Scalekit Provider:

1. Support for required_scopes as an optional parameter
2. Change `mcp_url` to `base_url` to align with the meaning of uri being passed in.
3. Remove `aud` check in the provider layer. if needed the implementing server can make use of the aud claim from token
4. Remove `client_id` from provider as it doesn't hold any value inside the provider

AkshayParihar33:
## Description
To provide more capabilities and cleaner integration between **FastMCP** and **Scalekit**, we are making the following updates:

### Additions & Improvements
- Added support for `required_scopes` (optional parameter) to enable fine-grained access control.
- Renamed `mcp_url` → `base_url` for better alignment with the meaning of the URI being passed in.
- Made `client_id` optional in the provider configuration. It continues to be handled internally but is planned for deprecation in a future release.

### ⚠️ Breaking Change
- **Removed `aud` (audience) validation** in the provider layer.  
  The provider no longer enforces the `aud` claim during token verification.  
  If your implementation depends on audience checks, you can manually validate the `aud` value from the token within your server logic.

> Co-Authored with Codex

**Contributors Checklist**

- [x] My change closes #2412 

**Review Checklist**

---

### DIFF
--- a/tests/server/auth/providers/test_scalekit.py
+++ b/tests/server/auth/providers/test_scalekit.py
@@ -34,10 +34,10 @@ class TestScalekitProvider:
         provider = ScalekitProvider(
             environment_url="https://legacy.scalekit.com",
             resource_id="sk_resource_legacy",
-            mcp_url="https://legacy-app.com/mcp/",
+            mcp_url="https://legacy-app.com/",
         )
 
-        assert str(provider.base_url) == "https://legacy-app.com/mcp/"
+        assert str(provider.base_url) == "https://legacy-app.com/"
 
     def test_init_prefers_base_url_over_mcp_url(self):
         """mcp_url should take precedence over base_url when both provided."""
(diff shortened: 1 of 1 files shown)


jlowin:
Thanks @AkshayParihar33. Two thoughts:

1. While you are welcome to deprecate features for backwards compatibility (and generally it is appreciated!) for auth modules we are not currently requiring this; breaking changes are permitted in the interest of always ensuring auth is as secure as possible. Happy to merge this as-is but in case you decide you don't want to support the old pattenr, that's ok too.
2. on audience validation -- there are some open security concerns that the MCP Protocol requires servers to verify that tokens were issued to that server, and so we are actively discussing other DCR implementations that don't do this are potentially technically violating MCP spec... but also it's unclear whether it is actually feasible to satisfy this restriction while also permitting the type of open, pseudonymous authentication that DCR permits. Curious for any thoughts you have (can discuss offline if you prefer)

AkshayParihar33:
hi @jlowin, thanks for reviewing the PR:

Regarding pt. 1 - As a good number of our customer's first choice is FastMCP, I wanted to keep the backward compatibility for a few versions and hence the deprecation log is in place to notify to stop using the old field and in subsequent release (maybe in a couple of months) will remove that field so that the transition is seamless for the devs.

Regarding pt. 2 - Is it okay if we can discuss this offline over slack (if you prefer that) or via email? Could you share your email-id that I can invite over slack?

PR #2426, opened 2025-11-15
Simplify _is_object_schema helper

### LINKED ISSUE #2421: Self-referencing types not supported correctly in structured output
### Description

If I annotate the return type for a tool with a Pydantic object or a dataclass that contains a field with a self-referencing type (e.g. it contains an instance of itself), then FastMCP is unable to correctly create `structured_content` for this return value, and instead treats it as if it were a primitive type, wrapping it in a `result` field. When running the client below, I would expect to get:

```python
{'value': 123, 'stuff': [{'value': 456, 'stuff': []}]}
```

But instead I get:

```python
{'result': {'value': 123, 'stuff': [{'value': 456, 'stuff': []}]}}
```

And, indeed, if I change the type annotation for `stuff` to just plain old `list` (i.e. `list[Any]`), then I do in fact get the expected result:

```python
{'value': 123, 'stuff': [{'value': 456, 'stuff': []}]}
```

### Example Code

```Python
import asyncio
from dataclasses import dataclass

@dataclass
class ReturnThing:
    value: int
    stuff: list["ReturnThing"]

mcp = FastMCP("thing")

@mcp.tool
def return_things() -> ReturnThing:
    return ReturnThing(value=123, stuff=[ReturnThing(value=456, stuff=[])])

async def main():
    client = Client("http://127.0.0.1:8999/mcp")
    async with client:
        result = await client.call_tool("return_things")
        print(result.structured_content)

if __name__ == '__main__':
    asyncio.run(main())
```

### Version Information

```Text
FastMCP version:                                                                2.13.0.2
MCP version:                                                                      1.21.1
Python version:                                                                  3.11.10
Platform:                                   Linux-6.8.0-87-generic-x86_64-with-glibc2.35
FastMCP root path: <REDACTED>/.venv/lib/python3.11/site-packages
```

jlowin:
Reduces the overcomplicated `$ref` resolution logic to a simple check. Self-referencing types always produce object schemas in our use case.

Note: this is what #2424 SHOULD have been except it was prematurely opened and merged by LLM. 

Actually closes #2421

### DIFF
--- a/tests/server/test_server.py
+++ b/tests/server/test_server.py
@@ -424,7 +424,7 @@ class TestToolDecorator:
         mcp = FastMCP()
 
         with pytest.raises(
-            ValueError, match='Output schemas must have "type" set to "object"'
+            ValueError, match="Output schemas must represent object types"
         ):
 
             @mcp.tool(output_schema={"type": "integer"})


dhdaines:
Excellent, thanks!

PR #2437, opened 2025-11-17
Bump py-key-value-aio to 0.3.0

strawgate:
## Description
Bump to latest py-key-value-aio https://github.com/strawgate/py-key-value/releases/tag/0.3.0

There are breaking changes to unstable stores but these shouldn't impact memory/disk/redis use-cases

Adding ignore clean-up errors as Windows doesn't always release the db right away on close()

### DIFF
--- a/.github/workflows/martian-test-failure.yml
+++ b/.github/workflows/martian-test-failure.yml
@@ -2,7 +2,7 @@ name: Marvin Test Failure Analysis
 
 on:
   workflow_run:
-    workflows: ["Run Tests"]
+    workflows: ["Tests", "Run static analysis"]
     types:
       - completed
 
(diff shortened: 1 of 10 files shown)


jlowin:
@strawgate going to give this a 2.14 milestone and merge it for that release (adding Dont Merge warning only as a signal that we're intentionally holding)

jlowin:
Oh wait sorry - you bumped the upper limit, not the lower, so this is not a breaking change. we can merge now!

strawgate:
I'm good to merge this whenever, do we need to hold?

jlowin:
merge away 🙏

jlowin (approved):

strawgate:
@jlowin I think I probably should have bumped the lower too but I'll do that in a separate pr that we can hold until closer to release

PR #2444, opened 2025-11-18
Add _meta support in resource content responses

calvingiles:
OpenAI's ChatGPT apps SDK uses _meta in resource content to pass widget domain information to clients (see https://developers.openai.com/apps-sdk/build/mcp-server/). FastMCP resources support metadata via the meta parameter, but this wasn't being included in the content returned to clients.

The MCP specification defines a _meta field on ResourceContents for this purpose. FastMCP now populates this field with the resource's metadata when serving content, and ensures resources created from templates inherit the template's metadata.

## Description

**Contributors Checklist**

**Review Checklist**

---

### DIFF
--- a/tests/resources/test_resource_content_meta.py
+++ b/tests/resources/test_resource_content_meta.py
@@ -33,9 +33,9 @@ class TestResourceContentMetaRawResponse:
             assert content.meta["widgetDomain"] == "example.com"
 
             # Verify serialization includes _meta
-            content_dict = content.model_dump()
-            assert "meta" in content_dict
-            assert content_dict["meta"]["widgetDomain"] == "example.com"
+            content_dict = content.model_dump(by_alias=True)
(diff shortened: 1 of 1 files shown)


jlowin:
I like the goal of supporting _meta in resources, but this PR feels like the wrong approach -- the `handlers` are low-level constructs; FastMCP should never have logic there.

It looks to me like you're accepting meta when the resource is defined and then returning it when the resource is read? This is atypical (and not how FastMCP tools work). The *tool* has metadata; and also the *tool result* has metadata. Similarly, here, I'd expect the *resource* (or template) to have metadata, and also the *resource contents* to have their own metadata. Forcibly passing the resource-definition metadata into the resource-read metadata feels like a very specific opinion. I think this needs an issue to explore the ramifications of this decision before we enforce it globally; in any case the implementation can not be in the handler function.

PR #2465, opened 2025-11-21
Fix MCPConfigTransport subprocess resource leak

chrisguidry:
Test `test_multi_client_lifespan` was timing out because MCPConfigTransport leaked subprocess resources.

**Root cause:** `MCPConfigTransport.connect_session()` never cleaned up underlying `StdioTransport` objects. When the client context exited, only the main `FastMCPTransport` session was closed, but the underlying stdio connections to actual MCP server subprocesses were left open. With `keep_alive=True` (default), these transports don't auto-disconnect, so subprocess stdin was never closed and processes stayed alive indefinitely.

**The fix:** Added finally block to `MCPConfigTransport.connect_session()` that explicitly calls `close()` on all underlying transports. This ensures subprocess stdin is closed and processes terminate cleanly.

**Evidence:**
- Before: 100% timeout rate (processes `status='sleeping'` after context exit)
- After: 15/15 consecutive passes, processes terminate immediately  
- Test time: 5.2s (timeout) → 3.2s (clean exit)

**Secondary fix:** Simplified test logic from infinite polling loop to direct `psutil.Process(pid).status()` check. The old while-True loop never raised `NoSuchProcess` because the constructor succeeds for existing processes.

All 3271 tests passing.

### DIFF
--- a/src/fastmcp/client/transports.py
+++ b/src/fastmcp/client/transports.py
@@ -979,12 +979,18 @@ class MCPConfigTransport(ClientTransport):
                 yield session
         finally:
             # Clean up underlying transports to ensure subprocesses terminate
-            for transport in self._underlying_transports:
-                await transport.close()
+            # Use gather with return_exceptions to ensure all transports close even if one fails
+            await asyncio.gather(
+                *(transport.close() for transport in self._underlying_transports),
+                return_exceptions=True,
+            )
 
     async def close(self):
-        for transport in self._underlying_transports:
-            await transport.close()
+        # Use gather with return_exceptions to ensure all transports close even if one fails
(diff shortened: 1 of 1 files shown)


strawgate:
This doesnt seem quite right to me -- ill try to take a look at this later today.

The test will timeout while debugging but should not be timing out when not debugging.

chrisguidry:
Converting to a draft to give @strawgate a chance to take a closer look.

jlowin:
this has been killing us thank you

chrisguidry:
@strawgate, you're totally right!  This was something new due to me not using a weakref when I needed it.

Issue #2479, opened 2025-11-25
OAuth Refresh Tokens Lost on Server Restart


kaayush71:
### Description

OAuth refresh tokens work correctly within a single server session but fail after server restart because FastMCP stores its own refresh tokens in memory-only dictionaries. This makes OAuth refresh tokens completely unusable in production environments with server restarts, auto-scaling, or serverless deployments (AWS Lambda, Google Cloud Functions, etc.).

### Environment
FastMCP Version: 2.13.1
Provider: GoogleProvider with DynamoDB storage
Deployment: AWS Lambda (but affects any stateless environment)

### Problem
When using OAuthProxy (which GoogleProvider extends), FastMCP refresh tokens are stored in in-memory dictionaries that are lost on server restart:
fastmcp/server/auth/oauth_proxy.py (Lines 877-883)

```# Local state for token bookkeeping only (no client caching)
self._access_tokens: dict[str, AccessToken] = {}
self._refresh_tokens: dict[str, RefreshToken] = {}  # ❌ Memory only!

# Token relation mappings for cleanup
self._access_to_refresh: dict[str, str] = {}
self._refresh_to_access: dict[str, str] = {}
```

This causes a disconnect between:
Upstream refresh tokens (Google's) → Stored in client_storage (DynamoDB) ✅
FastMCP refresh tokens → Stored in memory only ❌

### Proposed Solution
Store FastMCP refresh tokens in client_storage similar to how upstream tokens are stored:

```
async def load_refresh_token(
    self,
    client: OAuthClientInformationFull,
    refresh_token: str,
) -> RefreshToken | None:
    """Load refresh token from persistent storage."""
    if self._client_storage:
        return await self._client_storage.get(
            collection="mcp-refresh-tokens",
            key=refresh_token
        )
    return self._refresh_tokens.get(refresh_token)  # Fallback to memory
```

And update token issuance to persist refresh tokens:

```async def _issue_tokens(...):
    # ... existing code ...
    
    if self._client_storage:
        await self._client_storage.set(
            collection="mcp-refresh-tokens",
            key=refresh_token.token,
            value=refresh_token,
            ttl=refresh_token.expires_at
        )
    
    self._refresh_tokens[refresh_token.token] = refresh_token  # Also keep in memory
```

### Workaround
None - requires code change in FastMCP core.
### Additional Context
Upstream refresh tokens (Google) ARE correctly persisted to DynamoDB
[...]

PR #2481, opened 2025-11-25
Try to fix loading of Refresh Tokens on many containers

### LINKED ISSUE #2478: Refresh tokens don't work when running multiple containers
### Description

When running in multiple containers/instances, only the container that initiated the OAuth has the ability to refresh the token, even if we are using Redis as our `client_storage`.

### Example Code

```Python
TBD
```

### Version Information

```Text
FastMCP version:                                                                                   2.13.1
MCP version:                                                                                       1.21.0
Python version:                                                                                   3.12.11
Platform:                                                                    macOS-15.7.1-arm64-arm-64bit
FastMCP root path: /Users/giovanna/Projects/adroll-mcp/local-prototype/.venv/lib/python3.12/site-packages
```

guschnwg:
## Description

When working with multiple containers, just the one that issued the tokens have it kept in memory.
The goal here is to validate the JWT being passed in the load_access_token, and let `exchange_refresh_token` revalidate and get from the `client_storage`.

https://github.com/user-attachments/assets/4aff30c6-84cb-4cbc-a092-3a3d2fd33fb8

## Tests

Created using Claude Code

Fixes #2478 and #2479 possibly

## Workaround

While I wait for this to be release and/or fixed, I have the following code in my project

```python
class OurOAuthProxy(OAuthProxy):
        async def load_refresh_token(
            self,
            client: OAuthClientInformationFull,
            refresh_token: str,
        ) -> RefreshToken | None:
            """Load refresh token from local storage."""
            if local_refresh_token := self._refresh_tokens.get(refresh_token):
                return local_refresh_token

            # Verify then return
            try:
                refresh_payload = self._jwt_issuer.verify_token(refresh_token)
                return RefreshToken(
                    token=refresh_token,
                    client_id=cast(str, refresh_payload.get("client_id")),
                    scopes=refresh_payload.get("scope", "").split(" "),
                    expires_at=refresh_payload.get("exp", 0),
                )
            except Exception as e:
                logger.debug("FastMCP refresh token validation failed: %s", e)
                raise TokenError("invalid_grant", "Invalid refresh token") from e

    auth = OurOAuthProxy(
        ...
    )

mcp = FastMCP(
    ...
    auth=auth,
    ...
)
```

### DIFF
--- a/src/fastmcp/server/auth/oauth_proxy.py
+++ b/src/fastmcp/server/auth/oauth_proxy.py
@@ -1296,7 +1296,25 @@ class OAuthProxy(OAuthProvider):
         refresh_token: str,
     ) -> RefreshToken | None:
         """Load refresh token from local storage."""
-        return self._refresh_tokens.get(refresh_token)
+        if local_refresh_token := self._refresh_tokens.get(refresh_token):
+            return local_refresh_token
+
+        # Verify then return
+        try:
+            refresh_payload = self._jwt_issuer.verify_token(refresh_token)
+            client_id = refresh_payload.get("client_id", "")
+            if not client_id or client_id != client.client_id:
+                return None  # Belongs to different client
+
+            return RefreshToken(
+                token=refresh_token,
+                client_id=client_id,
+                scopes=refresh_payload.get("scope", "").split(" "),
+                expires_at=refresh_payload.get("exp", 0),
+            )
+        except Exception as e:
+            logger.debug("FastMCP refresh token validation failed: %s", e)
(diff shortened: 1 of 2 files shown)


jlowin:
Thanks @guschnwg - I've opened #2483 with a more holistic approach (e.g. entirely remove the in-memory dicts rather than introduce a verification fallback, which I think could also fail in some circumstances)

guschnwg:
> Thanks @guschnwg - I've opened #2483 with a more holistic approach (e.g. entirely remove the in-memory dicts rather than introduce a verification fallback, which I think could also fail in some circumstances)

Nice! Thank you!! Closing this one

PR #2483, opened 2025-11-25
Fix OAuth proxy refresh token storage for multi-instance deployments

### LINKED ISSUE #2478: Refresh tokens don't work when running multiple containers

jlowin:
Fixes #2478 and #2479

The OAuth proxy was using a local in-memory dict for refresh token storage, which broke multi-instance deployments. Container A would issue a refresh token, but Container B couldn't find it when the client tried to refresh.

**Changes:**
- Refresh tokens now use the pluggable `client_storage` backend (Redis, disk, etc.)
- Tokens stored by SHA-256 hash only - if storage is compromised, attackers get hashes they can't reverse
- Removed `_access_token_store` and relationship mappings (unused for validation; JWT + JTI mapping handles that)
- Simplified revocation logic

**Security improvement:** Previously refresh tokens were stored in plaintext. Now only metadata is stored, keyed by hash:

```python
# Storage: hash → metadata only
await self._refresh_token_store.put(
    key=_hash_token(refresh_token),  # SHA-256
    value=RefreshTokenMetadata(
        client_id=client.client_id,
        scopes=scopes,
        expires_at=None,
        created_at=time.time(),
    ),
)

# Lookup: hash incoming token, reconstruct object
metadata = await self._refresh_token_store.get(key=_hash_token(token))
return RefreshToken(token=token, **metadata)
```


kaayush71:
Hey @jlowin , any update on when this will be available in the new release?

jlowin:
This will go out in 2.13.2, probably today

PR #2520, opened 2025-12-02
add client kwargs to proxy clients and meta to proxy tool calls

### LINKED ISSUE #2519: Client Metadata is not passed to MCPConfig Servers
### Description

I noticed that when using the MCPConfig transport with more than one server, client side metadata is not accessible from the tools (ctx.request_context.meta does not contain any custom metadata added to the tool call). I found that this is because servers are set up as proxies when more than one is present and the  [ProxyTool implementation does not pass metadata from the original session to the proxy session when the proxy tool call is made.](https://github.com/jlowin/fastmcp/blob/ee5f465a82350e1c5a56c4a2b47cfdc4cd736e76/src/fastmcp/server/proxy.py#L288). This can easily be fixed by checking if there is any metadata in the context before the tool call is made and then injecting it into the proxy tool call.

Additionally, I noticed that the proxy clients created by the MCPConfig transport do not inherit any of the custom callbacks or handlers provided to the original client. Example:
```python
#mcp config
mcp_config = {
    "mcpServers": {
        "calculator": {
            "transport": "stdio",
            "command": "python",
            "args": ["server_1.py"]
        },
        "weather": {
            "transport": "stdio",
            "command": "python",
            "args": ["server_2.py"]
        }
    }   
}
# custom progress handler
async def my_progress_handler(progress: float, total: float | None, message: str | None ) -> None:
    print(f"Custom handler Progress: {progress:.2f}% - {message or ''}")
#client
mcp_client=Client(mcp_config, progress_handler=my_progress_handler)
```
In the above example, if the tools on either server include `ctx.report_progress(...)` the custom `my_progress_handler` will not be used because [they are never passed on to the proxy client](https://github.com/jlowin/fastmcp/blob/ee5f465a82350e1c5a56c4a2b47cfdc4cd736e76/src/fastmcp/utilities/mcp_config.py#L52C13-L52C63). Since ProxyClient accept the same arguments as regular ones, this can be fixed by passing the original clients arguments to the ProxyClient when it is created. Unfortunately since the ProxyClient creation occurs deep in the transport creation, the solution I have here adds more lines of code than I would have hoped for such a simple fix.

### Example Code

```Python

```

### Version Information

```Text
I tested this with the most up to date code in the main branch and mcp version 1.22
```

cegersdoerfer:
## Description
This PR solves 2 issues:
1. Passing tool call metadata to servers initialized via the MCPConfig transport. The fix checks if there is any metadata in the context before the tool call is made and then injecting it into the proxy tool call.
2. Passing client handlers and callbacks to proxy clients created when using the MCPConfig transport. Since ProxyClient accept the same arguments as regular ones, the fix passes the original client arguments to the ProxyClient instances when they are created. Admittedly my changes to solve this issue added more code than I hoped so I am hoping for some feedback to find a better solution

**Contributors Checklist**

- [x] My change closes #2519 

**Review Checklist**

---


strawgate:
MCPTransport is supposed to be a friendly "Helper" -- I think if you need more advanced capabilities then you should consider setting up the servers and transports yourself?

cegersdoerfer:
@strawgate I would agree for the second problem which the PR fixes (passing client callbacks to proxies) since it requires significantly more changes to the code and may not be needed by most users. But, for the first issue (passing metadata to proxy tool calls), it only adds minimal code contained to a single method in `src/fastmcp/server/proxy.py` and it seems like something anyone using an MCPConfig and tool call metadata would need and expect as the default behavior. It is also the primary reason I opened the PR at all. I'd have no problem with removing all of the changes pertaining to passing client callbacks to proxies from this PR if you agree

jlowin (changes requested):
@cegersdoerfer I agree with @strawgate here -- passing the kwargs through so many layers does work but IMO is a smell that we're overloading the functionality. However the meta-passing is an excellent enhancement that we should include - if you wouldn't mind removing the kwargs from the PR, we can get this merged!

cegersdoerfer:
@jlowin done! Let me know if anything else is needed

PR #2551, opened 2025-12-04
SEP-1577: Sampling with tools

jlowin:
Sampling with tools ([SEP-1577](https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1577)) allows servers to pass tools to `ctx.sample()` for agentic workflows where the LLM can call helper functions during generation. Another method, `ctx.sample_step()`, allows complete control over the agentic loop. 

### Basic Sampling

Request text generation from the client's LLM:

```python
result = await ctx.sample("What is 2 + 2?")
print(result.text)  # "4"
```

### Structured Output

Pass `result_type` to get a validated Pydantic model instead of raw text:

```python
class Weather(BaseModel):
    temperature: float
    conditions: str

result = await ctx.sample(
    "What's the weather in NYC?",
    result_type=Weather,
)
print(result.result.temperature)  # 72.0
```

### Sampling with Tools

Pass `tools` to let the LLM call functions during generation. The server automatically executes tool calls and feeds results back to the LLM until it produces a final response:

```python
def search_web(query: str) -> str:
    """Search the web."""
    return f"Results for: {query}"

result = await ctx.sample(
    "Find information about Python async",
    tools=[search_web],
)
```

### Loop Control with sample_step()

For fine-grained control, use `sample_step()` which makes a single LLM call and returns a `SampleStep`. It's stateless—you control the conversation by passing history each iteration:

```python
messages = [SamplingMessage(role="user", content=question)]

while True:
    step = await ctx.sample_step(messages, tools=[search])
    
    if step.is_text:
        return step.text
    
    # Log tool calls, add custom logic, etc.
    for call in step.tool_calls:
        print(f"Called: {call.name}")
    
    messages = step.history  # Continue with updated history
```

Set `execute_tools=False` to handle tool execution yourself for maximum control.

### DIFF
--- a/src/fastmcp/client/client.py
+++ b/src/fastmcp/client/client.py
@@ -310,7 +310,8 @@ class Client(Generic[ClientTransportT]):
             # Default to tools-enabled capabilities unless explicitly overridden
             self._session_kwargs["sampling_capabilities"] = (
                 sampling_capabilities
-                or mcp.types.SamplingCapability(
+                if sampling_capabilities is not None
+                else mcp.types.SamplingCapability(
                     tools=mcp.types.SamplingToolsCapability()
                 )
             )
@@ -377,7 +378,8 @@ class Client(Generic[ClientTransportT]):
         # Default to tools-enabled capabilities unless explicitly overridden
         self._session_kwargs["sampling_capabilities"] = (
             sampling_capabilities
-            or mcp.types.SamplingCapability(tools=mcp.types.SamplingToolsCapability())
+            if sampling_capabilities is not None
+            else mcp.types.SamplingCapability(tools=mcp.types.SamplingToolsCapability())
         )
 
     def set_elicitation_callback(


src/fastmcp/experimental/sampling/handlers/openai.py
```diff
@@ -121,23 +148,126 @@ def _convert_to_openai_messages(
...
+                                    function={
+                                        "name": item.name,
+                                        "arguments": json.dumps(item.input),
+                                    },
+                                )
+                            )
+                        elif isinstance(item, TextContent):
+                            text_parts.append(item.text)
+                        elif isinstance(item, ToolResultContent):
+                            # Tool results in a list - add as separate messages
```
strawgate: this appears to concatenate them?

strawgate:
/marvin can you fix the merge conflicts please

strawgate:
i think this is currently a breaking change to `ctx.sample` with the introduction of the `SamplingResult` response model?

src/fastmcp/server/context.py
```diff
@@ -557,6 +670,40 @@ async def sample(
...
+                    )
+
+        # Create synthetic final_response tool for structured output
+        final_response_tool: SamplingTool | None = None
+        if result_type is not None and result_type is not str:
+            final_response_tool = _create_final_response_tool(result_type)
+            sampling_tools.append(final_response_tool)
+
+            # Add hint to system prompt
+            hint = "Call final_response when you have completed the task."
```
strawgate: this is a bit smelly and will probably result in us having to do what pydantic does and have like 3 ways to do this

examples/advanced_sampling/tool_use.py
```diff
@@ -0,0 +1,85 @@
...
+mcp = FastMCP("Research Assistant")
+
+
+@mcp.tool
+async def research(question: str, ctx: Context) -> dict:
+    """Research a question using available tools and return a structured report."""
+    result = await ctx.sample(
+        messages=f"Research this question and provide a comprehensive answer:\n\n{question}",
+        system_prompt="You are a research assistant. Use the available tools to gather information, then call final_response with your structured report.",
+        tools=[search_web, get_word_count],
```
strawgate, line 61: the tool use example is actually testing functions not tools

src/fastmcp/server/context.py
```diff
@@ -568,43 +715,408 @@ async def sample(
...
+        # Extract text from content
+        text = _extract_text_from_content(result.content)
+
+        return SamplingResult(
+            text=text,
+            result=cast(ResultT, text),
+            history=list(sampling_messages),
+        )
+
+    async def _sample_with_fallback_handler(
```
strawgate: i think we have two implementations of an agent tool loop here between this and sample with client sampler that I think can be refactored into one loop that gets passed the sample function to use or something like that?

strawgate:
Does sampling tool do something for us that we couldnt do with FunctionTool?

PR #2558, opened 2025-12-05
Remove enable_docket setting; Docket is now always on

chrisguidry:
Simplifies configuration by removing the `enable_docket` flag. Every FastMCP server now starts a Docket instance and Worker automatically. Only `enable_tasks` remains to control SEP-1686 task protocol support.

```python
# Before: two settings needed
FASTMCP_ENABLE_DOCKET=true
FASTMCP_ENABLE_TASKS=true

# After: just one
FASTMCP_ENABLE_TASKS=true
```

Making Docket always-on uncovered some startup timing issues. The proxy test fixture was racing against Docket startup (~140ms), so we added a `wait_until_ready()` method backed by an `asyncio.Event` that signals when the server lifespan is fully initialized. Also updated the mount lifespan test expectation—with Docket keeping lifespans alive, mounted sub-servers no longer re-enter their lifespan on each proxy call.

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -48,13 +48,7 @@ jobs:
         run: uv sync --upgrade
 
       - name: Run tests (excluding integration and client_process)
-        run: |
-          if [ "$RUNNER_OS" == "Windows" ]; then
-            # Run without parallelism on Windows to debug xdist crashes
-            uv run pytest --inline-snapshot=disable tests -m "not integration and not client_process"
-          else
-            uv run pytest --inline-snapshot=disable tests -m "not integration and not client_process" --numprocesses auto --maxprocesses 4 --dist worksteal
-          fi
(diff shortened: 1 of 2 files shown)


src/fastmcp/server/server.py
```diff
@@ -504,11 +511,23 @@ async def _lifespan_manager(self) -> AsyncIterator[None]:
...
+                self._started.set()
+                try:
+                    yield
+                finally:
+                    self._started.clear()
 
         self._lifespan_result_set = False
         self._lifespan_result = None
 
+    async def wait_until_ready(self) -> None:
```
strawgate: private?

chrisguidry: Good call, I can do that

jordanhboxer:
this makes FastMCP incompatible with Trio now that Docket is required.

PR #2570, opened 2025-12-07
Add TaskConfig for SEP-1686 execution modes

jlowin:
Exposes the full MCP task execution modes via `TaskConfig` instead of just `task=True/False` (bool still works though!)

The primary user-facing impact of this change is the ability to create MCP components (e.g. tools) that *require* background task execution as opposed to whatever the client wants.

```python
from fastmcp import FastMCP, TaskConfig

mcp = FastMCP("MyServer")

@mcp.tool(task=TaskConfig(mode="optional"))   # supports both sync and background
async def flexible(): ...

@mcp.tool(task=TaskConfig(mode="required"))   # error if client doesn't request task  
async def must_be_background(): ...

@mcp.tool(task=TaskConfig(mode="forbidden"))  # error if client requests task
async def sync_only(): ...
```

Boolean shortcuts still work: `task=True` → `mode="optional"`, `task=False` → `mode="forbidden"`.

Key changes:
- New `TaskConfig` dataclass in `src/fastmcp/server/tasks/config.py`
- Model classes use `task_config: TaskConfig` field (decorators still use `task=` for ergonomics)
- Server enforces modes: returns `-32601` for `required` without task or `forbidden` with task
- `TaskConfig.validate_function()` centralizes the async function check

Internally, task details are stored on these new `TaskConfig` objects, anticipating both future MCP-task-specific config as well as the ability to pass Docket-specific config without accessing function dependencies.

cc @chrisguidry

### DIFF
--- a/docs/css/version-badge.css
+++ b/docs/css/version-badge.css
@@ -1,29 +1,11 @@
 /* Version badge -- display a badge with the current version of the documentation */
 .version-badge {
-  display: inline-block;
-  align-items: center;
-  gap: 0.3em;
-  font-size: 1em;
-  margin-top: 0px;
-  margin-bottom: 0px;
-  padding-top: 6px;
-  padding-bottom: 6px;
-  padding-left: 20px;
-  padding-right: 20px;
-  font-family: "Inter", sans-serif;
-  color: #ff5400;
-  background: #fef2f2;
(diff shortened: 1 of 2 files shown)


He-Pin:
Cool

olwethu-web:
Bravo

PR #2580, opened 2025-12-09
Remove TaskConfig and client from root exports

jlowin:
Removes `TaskConfig` and `client` from the root `__all__` exports to encourage more explicit imports.

`TaskConfig` should now be imported from its public API:
```python
from fastmcp.server.tasks import TaskConfig
```

The `client` module remains accessible via explicit import (`from fastmcp import client`) but is no longer included in `from fastmcp import *` exports.

All documentation and tests have been updated to use the correct import paths.


jlowin:
This is not a breaking change as TaskConfig has not been released and `client` remains accessible via import statement

PR #2581, opened 2025-12-09
Remove deprecated `from fastmcp.settings import settings`

### LINKED ISSUE #2183: Remove from_fastmcp.settings import settings (deprecated 2.10.2)
Remove support for `from fastmcp.settings import settings`.

Deprecated in: **2.10.2**
Migration: Use `fastmcp.settings` directly instead

Files affected:
- src/fastmcp/settings.py

jlowin:
Removes the deprecated `from fastmcp.settings import settings` import pattern (deprecated in 2.10.2) and the related `fastmcp.settings.settings` property access (deprecated in 2.8.0).

Users should access the global settings object directly via `fastmcp.settings`:

```python
import fastmcp

# Correct usage
fastmcp.settings.log_level = "DEBUG"
```

Closes #2183


reneleonhardt:
When will Python 3.14 support be released?

fastmcp 2.13.3 uses pydantic-core 2.33.2 (and pydantic 2.11.7) which aren't compatible with Python 3.14.

https://github.com/jlowin/fastmcp/blob/v2.13.3/uv.lock#L1209
https://github.com/jlowin/fastmcp/blob/v2.13.3/uv.lock#L1229

PR #2598, opened 2025-12-12
Refactor resource behavior and add meta support

jlowin:
IMPORTANT: this was merged and reverted. See #2609 for new PR

---

Resources can now include metadata via `ResourceContent.meta`, giving resource authors a way to pass structured information alongside content.

```python
@mcp.resource("data://report")
def get_report() -> ResourceContent:
    return ResourceContent(
        content="Report data...",
        mime_type="text/plain",
        meta={"generated_at": "2024-01-15", "version": 2}
    )
```

Internally, this represents a significant refactor of resource result handling, with `ResourceContent` now the canonical type for all resource reads. Custom resources returning `str` or `bytes` from `read()` continue to work with a deprecation warning when enabled.

### DIFF
--- a/src/fastmcp/server/proxy.py
+++ b/src/fastmcp/server/proxy.py
@@ -196,6 +196,10 @@ class ProxyResourceManager(ResourceManager, ProxyManagerMixin):
             client = await self._get_client()
             async with client:
                 result = await client.read_resource(uri)
+                if not result:
+                    raise ResourceError(
+                        f"Remote server returned empty content for {uri}"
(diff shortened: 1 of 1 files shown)


src/fastmcp/__init__.py
```diff
@@ -31,5 +33,6 @@
     "Client",
     "Context",
     "FastMCP",
+    "ResourceContent",
```
jlowin: ResourceContent should not be a root export

PR #2614, opened 2025-12-14
Clean up cancelled connection startup

### LINKED ISSUE #2613: Memory leak: cancelling `async with Client(...)` during connect leaves background session task running
### Description

Cancelling a `Client` context manager during connection startup can leak resources. If the task is cancelled during `__aenter__`, particularly inside of `_connect()` waiting for the background session runner to become ready, `__aexit__` is never invoked, so the partially-started background `session_task` can remain alive and keep transport/session resources referenced. 

### Example Code

```Python
import asyncio
import gc
import tracemalloc

from fastmcp import Client as FastMCPClient


URL = "http://127.0.0.1:8000/mcp"
ITERATIONS = 50
CANCEL_AFTER = 0.02


def snapshot(iteration: int) -> None:
    # Force garbage collection and log the number of FastMCPClient instances
    collected = gc.collect()

    count = sum(1 for obj in gc.get_objects() if isinstance(obj, FastMCPClient))

    current, peak = tracemalloc.get_traced_memory()

    print(
        f"iter={iteration} collected={collected} "
        f"traced_current_kb={current // 1024} traced_peak_kb={peak // 1024} "
        f"FastMCPClient_count={count}",
        flush=True,
    )


async def _tool_session() -> None:
    client = FastMCPClient(URL)
    async with client:
        await client.call_tool("tool", {})


async def main() -> None:
    tracemalloc.start(10)

    for i in range(1, ITERATIONS + 1):
        try:
            task = asyncio.create_task(_tool_session())
            await asyncio.sleep(CANCEL_AFTER)
            task.cancel()
            try:
                await task
                print(f"iteration={i} tool call cancelled", flush=True)
        except Exception as exc:
            print(f"iteration={i} got exception: {exc}", flush=True)
        snapshot(i)


if __name__ == "__main__":
    asyncio.run(main())
```

### Version Information

```Text
FastMCP version: 2.14.0
MCP version: 1.24.0
Python version: 3.14.0
Platform: macOS-15.7.1-arm64-arm-64bit-Mach-O
```

shawnthapa:
## Description
This change handles `CancelledError` inside `_connect()` only when the current call started a new session task (under the session lock), then tears down the partial session and resets state. It also adds regression tests for cancelled entry cleanup and for a cancelled concurrent waiter that must not close an already-active session.

**Contributors Checklist**

- [x] My change closes #2613 

**Review Checklist**

---

### DIFF
--- a/src/fastmcp/client/client.py
+++ b/src/fastmcp/client/client.py
@@ -6,7 +6,7 @@ import datetime
 import secrets
 import uuid
 import weakref
-from contextlib import AsyncExitStack, asynccontextmanager
+from contextlib import AsyncExitStack, asynccontextmanager, suppress
 from dataclasses import dataclass, field
 from pathlib import Path
 from typing import Any, Generic, Literal, TypeVar, cast, overload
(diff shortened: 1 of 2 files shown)


jlowin (approved):
Thank you!

PR #2632, opened 2025-12-16
Feature/supabase custom auth route

### LINKED ISSUE #2522: Allow custom supabase URL scheme
### Enhancement

The supabase auth provider accepts a `project_url` but enforces a scheme with f"{self.project_url}/auth/v1/...". This is fine for managed deployments but not for self-hosted ones where you may want to use a different URL.
I see two possible solutions to improve this:
* Replace `project_url` with `auth_url` and have the user include the full path (breaks the API, probably not ideal)
* Add an optional `auth_route` argument with "/auth/v1" as default value (probably better)

I would gladly open a PR with whichever solution is preferred.

EloiZalczer:
## Description
Added an auth_route argument to SupabaseProvider, according to the suggestion in https://github.com/jlowin/fastmcp/issues/2522.

This makes it possible to work with self-hosted supabase auth setups using custom routes.

**Contributors Checklist**

- [x] My change closes #2522 

**Review Checklist**

---


jlowin (approved):
@EloiZalczer Thanks! I made some minor changes to account for changes in #2752 and I think it's GTG. Since main is tracking the FastMCP 3.0 release, I will also cherry pick these commits into a 2.14.2 release we may do in the next week.

PR #2637, opened 2025-12-17
Don't fail when mounted proxy servers are unreachable

FaranIdo:
When a FastMCP server has mounted proxy servers (via `FastMCP.as_proxy()`), connection failures to any remote server would prevent local tools, resources, and prompts from being listed or called - even those on the main server or working remotes. This happened because the proxy managers only caught `METHOD_NOT_FOUND` errors, not connection failures which surface as `RuntimeError` from the Client.

The fix catches both `McpError` and `RuntimeError` in the proxy managers and server iteration loops, allowing the server to gracefully skip unavailable remotes while continuing to serve local and reachable components.

```python
# Local tools now work even when a mounted proxy is unreachable
main_app = FastMCP("MainApp")

@main_app.tool
def local_tool() -> str:
    return "works"

dead_proxy = FastMCP.as_proxy(
    Client(transport=SSETransport("http://unreachable:9999")),
    name="dead_remote"
)
main_app.mount(dead_proxy)

# Previously: listing/calling tools would fail
# Now: local_tool is accessible, dead remote is silently skipped
```


FaranIdo:
This fix is still needed after the Provider refactoring in https://github.com/jlowin/fastmcp/pull/2635.

  That PR handles exceptions for listing operations (`list_tools`, `list_resources`, `list_prompts`) - if a provider is unavailable, it gets skipped. But execution operations like
  `call_tool`, `read_resource`, and `get_prompt` don't have the same handling.

  The issue shows up when a dead proxy is mounted before a working server with the same prefix (or no prefix):

  ```python
  main_app.mount(dead_proxy, "shared")      # first, has no tools
  main_app.mount(working_server, "shared")  # second, has "my_tool"

  await client.list_tools()                  # works - returns ["shared_my_tool"]
  await client.call_tool("shared_my_tool")   # fails!
```

  Even though my_tool only exists in working_server, the call fails because both providers share the same prefix. dead_proxy gets checked first, MountedProvider.get_tool() tries to
  connect to see if the tool exists there, and the RuntimeError bubbles up before we ever check working_server.

  The fix catches RuntimeError in provider iteration loops and continues to the next provider, same as listing operations already do.

FaranIdo:
@jlowin, any chance you can please review this?

jlowin:
@FaranIdo we are in the middle of a comprehensive backend refactor that affects how both mounts and proxies work. please be patient.

jlowin:
I believe this is already handled automatically during provider iteration now; the results are gathered with exceptions suppressed so they can be logged without interrupting execution.

Issue #2300, opened 2025-10-29
ResponseCachingMiddleware breaks imported/mounted server prefixes


wipash:
### Description

When using `ResponseCachingMiddleware`, prefixes of imported/mounted MCP servers are not returned in response to `tools/list`.
The tools still require the prefix in order to be called though.

Everything works fine when not using `ResponseCachingMiddleware`.

```json
// List tools

curl -X POST http://localhost:8000/mcp \
                  -H "Content-Type: application/json" \
                  -H "Accept: text/event-stream, application/json" \
                  -d '{
                    "jsonrpc": "2.0",
                    "id": "1",
                    "method": "tools/list",
                    "params": {}
                  }' \
           | grep '^data: ' \
           | sed 's/^data: //' \
           | jq '.result'

// Response
{
  "tools": [
    {
      "name": "example_user_tool",
      "inputSchema": {
        "properties": {},
        "type": "object"
      },
      "_meta": {
        "_fastmcp": {
          "tags": []
        }
      }
    },
    {
      "name": "example_incident_tool",
      "inputSchema": {
        "properties": {},
        "type": "object"
      },
      "_meta": {
        "_fastmcp": {
          "tags": []
        }
      }
    },
    {
      "name": "example_knowledgebase_tool",
      "inputSchema": {
        "properties": {},
        "type": "object"
      },
      "_meta": {
        "_fastmcp": {
          "tags": []
        }
      }
    }
  ]
}


// Call example_user_tool
curl -X POST http://localhost:8000/mcp \
                  -H "Content-Type: application/json" \
                  -H "Accept: text/event-stream, application/json" \
                  -d '{
                    "jsonrpc": "2.0",
                    "id": "1",
                    "method": "tools/call",
                    "params": {
                      "name": "example_user_tool"
                    }
                  }' \
           | grep '^data: ' \
           | sed 's/^data: //' \
           | jq '.result.content[0].text'

// Response
"Unknown tool: example_user_tool"

// Call user_example_user_tool
curl -X POST http://localhost:8000/mcp \
                  -H "Content-Type: application/json" \
                  -H "Accept: text/event-stream, application/json" \
                  -d '{
                    "jsonrpc": "2.0",
                    "id": "1",
                    "method": "tools/call",
                    "params": {
                      "name": "user_example_user_tool"
                    }
                  }' \
           | grep '^data: ' \
           | sed 's/^data: //' \
           | jq '.result.content[0].text'

// Response
"This is an example user tool."
[...]

PR #2648, opened 2025-12-18
Simplify .key as computed property

jlowin:
The `.key` property now serves as a universal lookup interface for all components, implemented as a computed property rather than a stored field.

**Before:** Components had a `_key` private attribute and custom `model_copy(key=...)` for setting keys separately from names/URIs. This created complexity and confusion about when to use `.key` vs `.name`.

**After:** `.key` is computed from the underlying identifier:
- Tools/Prompts: `.key` returns `.name`
- Resources: `.key` returns `str(.uri)`  
- Templates: `.key` returns `.uri_template`

When prefixing components via `model_copy(update={"name": "prefixed_tool"})`, the `.key` automatically updates. Resource/template names are NOT prefixed—only their URIs are prefixed.

```python
@mcp.tool()
def greet(name: str) -> str:
    return f"Hello, {name}!"

tool = await mcp.get_tool("greet")
assert tool.key == "greet"  # .key returns .name

# After prefixing
prefixed = tool.model_copy(update={"name": "api_greet"})
assert prefixed.key == "api_greet"  # automatically updated
```

Also moves `import_server` tests to `tests/deprecated/` with warning suppression.


chrisguidry (approved):
I like it!

ajram23:
@jlowin I see this merged into 3.X branch. Any recommendations for folks running on 2.14.X where prefix is broken?

jlowin:
This wasn't a bug fix. What issue are you referring to?

ajram23:
@jlowin thanks for the prompt response - I came here through https://github.com/jlowin/fastmcp/issues/2300

jlowin:
Oh I see - that was an inadvertent outcome of this change. We're not planning to back port this, it would be complicated. I recommend using the 3.0 b1 if you need that middleware to work!

ajram23:
:( just upgraded to .14.X from .10.X

PR #2680, opened 2025-12-23
Unify component storage in LocalProvider

jlowin:
The three manager classes (ToolManager, ResourceManager, PromptManager) duplicated logic that already existed in the provider pattern. This PR consolidates them into a single `LocalProvider` that stores locally-defined components.

**Before:** FastMCP had separate managers for each component type, plus a provider chain. The managers stored decorator-registered components, while providers supplied dynamic components. This created two parallel paths for component lookup.

**After:** LocalProvider is a proper Provider that stores decorator-registered components. FastMCP's `_providers` list always starts with LocalProvider, so the server just iterates providers for all operations. Server decorators (`@tool`, `@resource`, `@prompt`) and `add_*` methods are thin passthroughs to LocalProvider—they inject server defaults and delegate. Single code path, simpler architecture.

```python
# LocalProvider is the first provider, stores local components
mcp = FastMCP("App")

@mcp.tool  # Delegates to mcp._local_provider.tool() with server defaults
def greet(name: str) -> str:
    return f"Hello, {name}!"

# Internally: mcp._local_provider stores the tool
# get_tools() iterates _providers (LocalProvider first)
```

**Breaking:** `ToolManager`, `ResourceManager`, `PromptManager` are deleted. Users importing these directly will get ImportError. These were internal implementation details, not public API.

### DIFF
--- a/tests/server/middleware/test_rate_limiting.py
+++ b/tests/server/middleware/test_rate_limiting.py
@@ -396,8 +396,8 @@ class TestRateLimitingMiddlewareIntegration:
 
         rate_limit_server.add_middleware(
             RateLimitingMiddleware(
-                max_requests_per_second=6.0,  # Accounting for initialization and list_tools calls
-                burst_capacity=4,
+                max_requests_per_second=1.0,  # Very slow refill to ensure rate limiting triggers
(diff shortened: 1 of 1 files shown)


jlowin:
@coderabbitai ignore

PR #2696, opened 2025-12-23
Fix test cleanup for uvicorn 0.39+ context isolation

jlowin:
Uvicorn 0.39.0 introduced a fix for ContextVars pollution ([encode/uvicorn#2742](https://github.com/encode/uvicorn/pull/2742)) that wraps ASGI tasks in empty contexts. This is a workaround for a CPython bug ([python/cpython#140947](https://github.com/python/cpython/issues/140947)) being fixed in Python 3.13+.

The context isolation breaks task cancellation propagation, causing test cleanup to hang when we `task.cancel()` and `await task` on uvicorn server tasks.

**Fix:** Use graceful shutdown (`server.should_exit = True`) for real uvicorn servers, and add timeouts for mocked server cleanup.

**Files changed:**
- `tests/server/test_logging.py` - Add timeout to mocked server cleanup  
- `tests/client/test_streamable_http.py` - Use graceful shutdown for nested_server fixture
- `tests/client/test_sse.py` - Use graceful shutdown for nested_sse_server fixture

### DIFF
--- a/tests/client/test_sse.py
+++ b/tests/client/test_sse.py
@@ -117,17 +117,16 @@ async def nested_sse_server():
         ws="websockets-sansio",
     )
 
-    server_task = asyncio.create_task(uvicorn.Server(config).serve())
+    uvicorn_server = uvicorn.Server(config)
+    server_task = asyncio.create_task(uvicorn_server.serve())
     await asyncio.sleep(0.1)
 
     try:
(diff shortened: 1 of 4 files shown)


Kludex:
I'm reverting that PR from Uvicorn, not sure how it impacts you. But please ping me if you have any questions.

PR #2697, opened 2025-12-23
Let FastMCPError propagate unchanged from managers

jlowin:
Previously, managers only re-raised their specific error types (e.g., `ToolError` in ToolManager, `ResourceError` in ResourceManager). This meant that other `FastMCPError` subclasses like `fastmcp.exceptions.ValidationError` would get wrapped, losing their original message.

Now all managers catch `FastMCPError` (the base class) and re-raise it unchanged, allowing any FastMCP error type to propagate with its original message intact.

```python
from fastmcp import FastMCP
from fastmcp.server import Context
from fastmcp.exceptions import ValidationError

mcp = FastMCP()

@mcp.tool
def validate(ctx: Context) -> str:
    raise ValidationError("Invalid input format")
    
# Error message is now "Invalid input format", not wrapped in "Error calling tool..."
```

### DIFF
--- a/src/fastmcp/prompts/prompt_manager.py
+++ b/src/fastmcp/prompts/prompt_manager.py
@@ -7,7 +7,7 @@ from typing import Any
 from mcp import GetPromptResult
 
 from fastmcp import settings
-from fastmcp.exceptions import NotFoundError, PromptError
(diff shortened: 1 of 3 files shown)


jlowin:
Merging, this needs the uvicorn 0.39 fixes in #2696

PR #2698, opened 2025-12-23
Let FastMCPError propagate unchanged from managers

jlowin:
Merge 2.x fixes to main, including #2697, which ensures FastMCP errors propogate. Some of these are redundant but the new commit history will simplify future merges from the release branch.


jlowin:
CLosing - too many conflicts. We will cherry pick instead

PR #2699, opened 2025-12-23
Fix uvicorn 0.39+ test timeouts and FastMCPError propagation

jlowin:
Uvicorn 0.39+ wraps ASGI tasks in empty contexts to prevent ContextVars pollution (see [uvicorn#2742](https://github.com/Kludex/uvicorn/pull/2742)). This breaks task cancellation propagation, causing test timeouts during fixture teardown.

This PR:
1. Fixes test cleanup by using graceful shutdown (`server.should_exit = True`) for real uvicorn servers and timeout-based cleanup for mocked servers
2. Updates managers to catch `FastMCPError` (base class) instead of specific error types, so all FastMCP exceptions propagate with their original messages intact

```python
from fastmcp import FastMCP
from fastmcp.exceptions import ValidationError

mcp = FastMCP()

@mcp.tool
def validate() -> str:
    raise ValidationError("Invalid input format")
    
# Error message is now "Invalid input format", not wrapped in "Error calling tool..."
```

Implements #2696 for main branch, not just v2

### DIFF
--- a/src/fastmcp/prompts/prompt_manager.py
+++ b/src/fastmcp/prompts/prompt_manager.py
@@ -7,7 +7,7 @@ from typing import Any
 import mcp.types
 
 from fastmcp import settings
-from fastmcp.exceptions import NotFoundError, PromptError
+from fastmcp.exceptions import FastMCPError, NotFoundError, PromptError
 from fastmcp.prompts.prompt import (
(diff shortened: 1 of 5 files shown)


Kludex:
I'm reverting Uvicorn's behavior.

PR #2701, opened 2025-12-24
Fix Prefect website URL in docs footer

mgoldsborough:
The link in your footer was broken.


jlowin:
Ha, thanks, we temporarily redirected that to the docs themselves!

PR #2750, opened 2025-12-26
Add task_meta parameter to read_resource() for explicit task control

jlowin:
Extends the `task_meta` parameter pattern to both tools and resources, providing explicit control over sync vs task execution with type-safe return types.

Previously, the return type of `read_resource()` and `call_tool()` was a union type requiring runtime `isinstance()` checks. Now, overloads narrow the type based on whether `task_meta` is provided:

```python
# Synchronous execution - type is ResourceResult
result = await server.read_resource("data://test")
content = result.contents[0].content  # No isinstance check needed

# Task execution - type is CreateTaskResult  
task_result = await server.read_resource("data://test", task_meta=TaskMeta())
task_id = task_result.task.taskId  # No isinstance check needed
```

Key changes:
- `read_resource()` and `call_tool()` accept optional `task_meta` parameter with overloads for type narrowing
- MCP routes (`_call_tool_mcp`, `_read_resource_mcp`) no longer set `fn_key` - components enrich it themselves
- Each component enriches `fn_key` in its `_run()`/`_read()` method using `self.key` as a fallback
- Provider wrappers enrich `fn_key` with the parent's namespaced key before delegating to child servers
- Uses `dataclasses.replace()` for immutable TaskMeta updates
- Added tests for mounted components with direct `task_meta` parameter usage

### DIFF
--- a/src/fastmcp/resources/resource.py
+++ b/src/fastmcp/resources/resource.py
@@ -5,6 +5,7 @@ from __future__ import annotations
 import base64
 import inspect
 from collections.abc import Callable
+from dataclasses import replace
 from typing import TYPE_CHECKING, Annotated, Any, ClassVar, overload
 
 import mcp.types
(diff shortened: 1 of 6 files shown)


jlowin:
> Use replace() instead of constructor for consistency
will fix in next PR

PR #2764, opened 2025-12-27
Fix OAuth Proxy resource parameter validation

jlowin:
The OAuth Proxy now validates the `resource` parameter at authorization time and correctly binds JWT audience to the MCP endpoint path.

**Changes:**
- Adds `set_mcp_path()` to `AuthProvider` for deferred resource URL configuration  
- Creates `JWTIssuer` with correct audience based on actual mount path
- Rejects authorization if client's requested resource doesn't match server's URL

**Impact:** This is an important security fix. Existing OAuth Proxy tokens will be invalidated after upgrade—users will need to re-authenticate once.


jlowin:
_resource_url can only be `None` if there is no base_url, or if set_mcp_path() was never called; both would be a misconfigured (in fact, completely broken) OAuthProxy. It could not complete flows at all.

PR #2776, opened 2025-12-29
Fix base_url fallback when url is not set

### LINKED ISSUE #2775: OpenAPITool base_url fallback not working
### Description

When httpx.AsyncClient is created without a base_url, it defaults to `httpx.URL("")`. The code checks if `self._client.base_url` to determine whether to use the fallback, but `httpx.URL("")` is truthy even though it stringifies to an empty string.

```python
   base_url = (
        str(self._client.base_url)
        if self._client.base_url
        else "http://localhost"
    )
```

### Example Code

```Python
import httpx

# httpx.URL("") is truthy but stringifies to ""
url = httpx.URL("")
bool(url)  # True
str(url)   # ""

# Before: returns "" because URL("") is truthy
base_url = str(url) if url else "http://localhost"

# After: returns "http://localhost" because "" is falsy
base_url = str(url) or "http://localhost"

# With base_url set - works correctly
client = httpx.AsyncClient(base_url="https://api.example.com")
str(client.base_url) or "http://localhost"  # "https://api.example.com"

# Without base_url - now falls back correctly
client = httpx.AsyncClient()
str(client.base_url) or "http://localhost"  # "http://localhost"

# httpx doesn't allow base_url=None
client = httpx.AsyncClient(base_url=None)
# => Invalid type for url. Expected str or httpx.URL, got <class 'NoneType'>: None
```

### Version Information

```Text
TBW
```

bhbs:
## Description

When httpx.AsyncClient is created without a base_url, it defaults to `httpx.URL("")`. The original code checked `if self._client.base_url` to determine whether to use the fallback, but `httpx.URL("")` is truthy even though it stringifies to an empty string. This caused requests to be made to `""` instead of `"http://localhost/"`.

**Contributors Checklist**

- [x] My change closes #2775 

**Review Checklist**

---

### DIFF
--- a/src/fastmcp/server/providers/openapi/components.py
+++ b/src/fastmcp/server/providers/openapi/components.py
@@ -108,11 +108,7 @@ class OpenAPITool(Tool):
     async def run(self, arguments: dict[str, Any]) -> ToolResult:
         """Execute the HTTP request using RequestDirector."""
         try:
-            base_url = (
-                str(self._client.base_url)
-                if self._client.base_url
-                else "http://localhost"
-            )
+            base_url = str(self._client.base_url) or "http://localhost"
 
             # Build the request using RequestDirector
             request = self._director.build(self._route, arguments, base_url)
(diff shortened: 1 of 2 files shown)


jlowin (approved):

jlowin:
Thank you!

PR #2781, opened 2025-12-29
Enhance documentation on tool transformation

### LINKED ISSUE #2780: Add additional use case for Tool Transformation to documentation
### Enhancement

The current Tool Transformation documentation has a section on use cases ( https://gofastmcp.com/patterns/tool-transformation#why-transform-tools ).  That list of use cases focused on transforming tools you did not make.

When reading the documentation, I did not read past that list of use cases because I currently only consume tools I author.

But even when consuming tools I author, the tool transformation capabilities can be very useful.  Especially for making environment-aware tools.

I think it would be useful to add a small note to cover this other use case so future devs (and LLMs) can be reminded of this other class of use case.

shea-parkes:
## Description

Updated documentation for tool transformations to mention environment-aware tools use case.

- [x] My change closes #2780
- [x] ~~I have tested my changes manually and by adding relevant tests~~

**Review Checklist**

### DIFF
--- a/docs/patterns/tool-transformation.mdx
+++ b/docs/patterns/tool-transformation.mdx
@@ -22,6 +22,8 @@ Often, an existing tool is *almost* perfect for your use case, but it might have
 
 Instead of rewriting the tool from scratch, you can **transform** it to fit your needs.
 
+Transformation is also powerful for **environment-aware tools**. You can dynamically update argument descriptions or `typing.Literal` enumerations to reflect the data actually available in your current environment.
+
 ## Basic Transformation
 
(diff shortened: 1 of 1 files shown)


shea-parkes:
My simple use case was to update a `table_name` parameter based on what tables were actually available in the execution environment.  I originally ~monkey-patched the Python function because I hadn't read down the documentation to learn about the `ArgTransform` class (because the use cases at the top didn't match my needs so I clicked "next" without reading further).

I eventually had a GenAI coding assistant question why I wasn't using `ArgTransform`.  Which then got me to learn about this powerful feature, and want to improve the documentation so others find it quicker.

Issue #2793, opened 2026-01-06
OAuth Discovery ignores RFC 8414 Section 3.1 path composition rules


CaliLuke:
# Bug: OAuth Discovery ignores RFC 8414 Section 3.1 path composition rules

## Problem Description

When using `FastMCP` with an `OAuthProxy` configured with an issuer that has a path component (e.g., `http://example.com/mcp`), the generated Authorization Server Metadata endpoint does not follow RFC 8414 Section 3.1.

The metadata is always served at the root `/.well-known/oauth-authorization-server`, regardless of the issuer path.

According to **RFC 8414 Section 3.1**:
> If the authorization server's issuer identifier includes a path component ... the well-known URI string would be inserted between the host and [the path], leading to a path such as `/.well-known/oauth-authorization-server/issuer1`

## Reproduction

1. Configure an MCP server with `OAuthProxy` and an issuer URL containing a path:

```python
from fastmcp import FastMCP
from fastmcp.server.auth import OAuthProxy

mcp = FastMCP("My Server")

# Configure OAuth with an issuer that has a path component (/my-prefix)
auth = OAuthProxy(
    upstream_authorization_endpoint="http://example.com/auth",
    upstream_token_endpoint="http://example.com/token",
    upstream_client_id="client-id",
    upstream_client_secret="client-secret",
    issuer_url="http://localhost:8000/my-prefix",
    base_url="http://localhost:8000/my-prefix"
)

# In a real app, you would mount this auth provider
# mcp = FastMCP("My Server", auth=auth)
```

2. Start the server.
3. Check the discovery URL for the path-aware issuer:
   `http://localhost:8000/.well-known/oauth-authorization-server/my-prefix`

**Actual Result:** `404 Not Found`
**Expected Result:** `200 OK` with metadata JSON

The server *does* serve metadata at `http://localhost:8000/.well-known/oauth-authorization-server`, but this is incorrect for an issuer of `.../my-prefix`.

## Technical Details

It appears `OAuthProxy.get_well_known_routes()` implements the correct logic to handle path components, but it is not called during the default `FastMCP` initialization flow.

Instead, `FastMCP` calls `auth.get_routes()`, which (via inheritance) calls `create_auth_routes()` from the SDK. The SDK's `create_auth_routes` function hardcodes the metadata path to `/.well-known/oauth-authorization-server`, ignoring the path component in the issuer URL.

CaliLuke:
[...]

PR #2794, opened 2026-01-06
fix: path-aware OAuth discovery (issue #2793)

CaliLuke:
Fix for issue #2793.

This PR updates `OAuthProvider.get_routes` to correctly handle `issuer_url`s that contain a path component (e.g., `http://example.com/api`). Previously, the metadata route was always mounted at the root `/.well-known/oauth-authorization-server`, ignoring the path component. This change ensures compliance with RFC 8414.

Added regression tests in `tests/server/test_path_aware_auth.py`.

Verified with `prek` and existing integration tests.

### DIFF
--- a/tests/server/auth/test_oauth_mounting.py
+++ b/tests/server/auth/test_oauth_mounting.py
@@ -306,7 +306,7 @@ class TestOAuthMounting:
 
         # The route should be path-aware (RFC 8414)
         assert (
-            auth_server_routes[0].path == "/.well-known/oauth-authorization-server/api"
+            auth_server_routes[0].path == "/api/.well-known/oauth-authorization-server"
         )
 
         # Find the protected resource metadata route for comparison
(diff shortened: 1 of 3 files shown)


CaliLuke:
@jlowin please let me know what I need to do to get this merged, this bug is driving me up the wall. The review by Marvin is actually incorrect, it would fail for subdomains, my solution is more general.

tests/server/auth/test_oauth_mounting.py
```diff
@@ -306,7 +306,7 @@ async def test_oauth_authorization_server_metadata_path_aware_discovery(
 
         # The route should be path-aware (RFC 8414)
         assert (
-            auth_server_routes[0].path == "/.well-known/oauth-authorization-server/api"
+            auth_server_routes[0].path == "/api/.well-known/oauth-authorization-server"
```
jlowin, line 309: These are not correct, the previous version was correct for path-aware RFC 8414.

jlowin:
@CaliLuke the way to solve this today is in the documentation I linked in your issue. I think Marvin is correct; well-known must be mounted under root with the path-prefixing at the *end*. Otherwise we would be able to mount well-known routes automatically without the manual steps in the documentation!

I think we can close this PR, but I will open one to add automatic path-prefixing based on base URL, which I think is a good enhancement.

PR #2798, opened 2026-01-06
Backport OAuth token storage TTL fix to release/2.x

jlowin:
Backport of the OAuth token storage TTL fixes from main (b980f87b, 843d4ac0). #2796 

The token storage was using access token expiry as the TTL, which caused both the access token and refresh token to be deleted when the short-lived access token expired. This prevented token refresh from working.

**Server-side fix:** Use `max(refresh_expires_in or 0, expires_in, 1)` to ensure the upstream token storage TTL is at least as long as the longest-lived token, with a minimum of 1 second to avoid `InvalidTTLError`.

**Client-side fix:** Use a 1-year fixed TTL instead of access token expiry, since refresh tokens are typically valid much longer than access tokens.

### DIFF
--- a/src/fastmcp/server/auth/oauth_proxy.py
+++ b/src/fastmcp/server/auth/oauth_proxy.py
@@ -1576,7 +1576,7 @@ class OAuthProxy(OAuthProvider):
         )
 
         # Store new refresh token JTI mapping with aligned expiry
-        refresh_ttl = new_refresh_expires_in or 60 * 60 * 24 * 30
+        # (reuse refresh_ttl calculated above for upstream token store)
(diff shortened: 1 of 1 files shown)


dexalex84:
Interesting 

That can fix the same issue I have with:
"Not stored tokens found" message in Cursor output, that I'm getting from time to time, currently I'm debugging why exactly it happens.

Is this planned planned for the next 2.14.x or  2.x.x ?

UPDATE:
I thought it should be back merged into v2.14.2, but could not find it, maybe not published to pypy repo yet

UPDATE:
Added in v2.14.3

Thanks !
@jlowin

Issue #2807, opened 2026-01-07
Generated schema from tools using enum is rejected by Claude Opus


pamelafox:
### Description

I am using this FastMCP server:
https://github.com/Azure-Samples/python-mcp-demos/blob/main/servers/basic_mcp_http.py

It uses enums like so:

```
class PaymentMethod(Enum):
    AMEX = "amex"
    VISA = "visa"
    CASH = "cash"


class Category(Enum):
    FOOD = "food"
    TRANSPORT = "transport"
    ENTERTAINMENT = "entertainment"
    SHOPPING = "shopping"
    GADGET = "gadget"
    OTHER = "other"


@mcp.tool
async def add_expense(
    date: Annotated[date, "Date of the expense in YYYY-MM-DD format"],
    amount: Annotated[float, "Positive numeric amount of the expense"],
    category: Annotated[Category, "Category label"],
    description: Annotated[str, "Human-readable description of the expense"],
    payment_method: Annotated[PaymentMethod, "Payment method used"],
):
```

When I use it with Opus 4.5 in GitHub Copilot, I get this error:

```
Reason: Request Failed: 400 {"error":{"message":"tools.0.custom.input_schema: Invalid $ref in schema: '#/$defs/Category' does not exist. Ensure all $ref targets are defined in the schema's definitions or $defs.. Learn more about tool use at https://docs.claude.com/en/docs/tool-use.","code":"invalid_request_body"}}
```

This is the schema that GitHub Copilot sends to Opus 4.5, presumably the same schema generated by FastMCP:

```
{
        "function": {
            "name": "mcp_expenses-mcp-2_add_user_expense",
            "description": "Add a new expense to Cosmos DB.",
            "parameters": {
                "$defs": {
                    "Category": {
                        "enum": [
                            "food",
                            "transport",
                            "entertainment",
                            "shopping",
                            "gadget",
                            "other"
                        ],
                        "type": "string"
                    },
                    "PaymentMethod": {
                        "enum": [
                            "amex",
                            "visa",
                            "cash"
                        ],
                        "type": "string"
                    }
                },
                "properties": {
                    "date": {
                        "description": "Date of the expense in YYYY-MM-DD format",
                        "format": "date",
                        "type": "string"
                    },
                    "amount": {
                        "description": "Positive numeric amount of the expense",
                        "type": "number"
                    },
                    "category": {
                        "$ref": "#/$defs/Category",
                        "description": "Category label"
                    },
                    "description": {
[...]

Issue #2236, opened 2025-10-24
Incompatibility between FastMCP output for Enum types versus MCP inspector expectations


pamelafox:
### Description

Currently, when you use an Enum type for a tool argument, the MCP inspector does not understand that type as an enum, due to FastMCP using refs/defs in the schame for them. This issue is discussed here in the inspector repo:
https://github.com/modelcontextprotocol/inspector/issues/755
The inspector only has support for enums via outputType/enum list.
I'm trying to figure out the path forward to be able to inspect FastMCP-generated tools.

Is there a way with FastMCP to output enums that does not use defs/refs and uses the simpler schema instead? Would it make sense to always do so?

### Example Code

```Python
import csv
import logging
from datetime import date
from enum import Enum
from pathlib import Path
from typing import Annotated

from fastmcp import FastMCP

logging.basicConfig(level=logging.INFO, format="%(asctime)s - %(message)s")
logger = logging.getLogger("ExpensesMCP")


SCRIPT_DIR = Path(__file__).parent
EXPENSES_FILE = SCRIPT_DIR / "expenses.csv"


mcp = FastMCP("Expenses Tracker")


class PaymentMethod(Enum):
  AMEX = "amex"
  VISA = "visa"
  CASH = "cash"


class Category(Enum):
  FOOD = "food"
  TRANSPORT = "transport"
  ENTERTAINMENT = "entertainment"
  SHOPPING = "shopping"
  GADGET = "gadget"
  OTHER = "other"


@mcp.tool
async def add_expense(
  date: Annotated[date, "Date of the expense in YYYY-MM-DD format"],
  amount: Annotated[float, "Positive numeric amount of the expense"],
  category: Annotated[Category, "Category label"],
  description: Annotated[str, "Human-readable description of the expense"],
  payment_method: Annotated[PaymentMethod, "Payment method used"],
):
  """Add a new expense to the expenses.csv file."""
  if amount <= 0:
    return "Error: Amount must be positive"

  date_iso = date.isoformat()
  logger.info(f"Adding expense: ${amount} for {description} on {date_iso}")

  try:
    file_exists = EXPENSES_FILE.exists()

    with open(EXPENSES_FILE, "a", newline="", encoding="utf-8") as file:
      writer = csv.writer(file)

      if not file_exists:
        writer.writerow(
          ["date", "amount", "category", "description", "payment_method"]
        )

      writer.writerow(
        [date_iso, amount, category.value, description, payment_method.name]
      )

[...]

PR #2814, opened 2026-01-09
Dereference $ref in tool schemas for MCP client compatibility

jlowin:
MCP clients like VS Code Copilot and Claude Desktop don't properly handle `$ref` in tool input schemas—they strip `$defs` before sending schemas to the LLM, leaving dangling references that break enum parameters and complex types.

FastMCP now automatically dereferences all `$ref` entries, inlining definitions directly into the schema. This happens in `compress_schema()` at tool creation time and again in `to_mcp_tool()` at serving time (to catch OpenAPI tools that bypass `compress_schema`).

```python
# Before: Schema with $ref (breaks in VS Code Copilot)
{
  "$defs": {"Status": {"enum": ["active", "inactive"], "type": "string"}},
  "properties": {"status": {"$ref": "#/$defs/Status"}}
}

# After: Dereferenced schema (works everywhere)
{
  "properties": {"status": {"enum": ["active", "inactive"], "type": "string"}}
}
```

For circular/self-referencing schemas that can't be fully dereferenced, FastMCP falls back to resolving only the root-level `$ref` while preserving `$defs` for nested references.

Follow-up to #2807, addresses #2236

### DIFF
--- a/src/fastmcp/utilities/json_schema.py
+++ b/src/fastmcp/utilities/json_schema.py
@@ -62,10 +62,11 @@ def dereference_refs(schema: dict[str, Any]) -> dict[str, Any]:
 
 
 def _merge_ref_siblings(
-    original: dict[str, Any] | list | Any,
-    dereferenced: dict[str, Any] | list | Any,
+    original: Any,
+    dereferenced: Any,
     defs: dict[str, Any],
-) -> dict[str, Any] | list | Any:
+    visited: set[str] | None = None,
+) -> Any:
     """Merge sibling keywords from original $ref nodes into dereferenced schema.
 
     When jsonref resolves $ref, it replaces the entire node with the referenced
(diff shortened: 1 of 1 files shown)


debnil-mercor:
@jlowin: thank you for your amazing work on this repo! Any chance this could get released sooner than v3? I keep running into this issue when parsing outputs from Gemini 3 Pro, and having this fix in FastMCP would be amazing. I upgraded to 2.14.3 but it didn't fix this. Let me know if I should do something else.

jlowin:
Hi @debnil-mercor - I've opened https://github.com/jlowin/fastmcp/pull/2861 and will try release in the next couple days

debnil-mercor:
@jlowin: thanks so much for the quick response! Let me know if I can help accelerate this.

PR #2815, opened 2026-01-09
Fix Claude Code installation lock contention in CI workflows

jlowin:
Adds a cleanup step before each `claude-code-action` invocation to remove stale lock files that can block installation retries after timeouts.

This addresses a known issue where Claude Code installation can time out and leave behind lock files at `~/.claude/.locks` or `~/.local/state/claude/locks`, causing subsequent retry attempts to fail with:

> "Could not install - another process is currently installing Claude. Please try again in a moment."

See: https://github.com/anthropics/claude-code-action/issues/709

**Affected workflows:** `marvin.yml`, `marvin-label-triage.yml`, `marvin-dedupe-issues.yml`, `martian-issue-triage.yml`, `martian-test-failure.yml`

### DIFF
--- a/.github/workflows/martian-issue-triage.yml
+++ b/.github/workflows/martian-issue-triage.yml
@@ -155,6 +155,9 @@ jobs:
           }
           EOF
 
+      - name: Clean up stale Claude locks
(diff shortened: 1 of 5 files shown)


jlowin:
/marvin just testing you can hear me

PR #2832, opened 2026-01-10
Add standalone decorators and eliminate fastmcp.fs module

jlowin:
Standalone decorators (`@tool`, `@resource`, `@prompt`) now create MCP objects without registering them to a server. This enables workflows where you create an object first, then register it—useful for tool transformations and FileSystemProvider discovery.

```python
from fastmcp.tools import tool
from fastmcp import FastMCP

@tool
def greet(name: str) -> str:
    return f"Hello, {name}!"

# greet is now a FunctionTool, can be registered to any server
mcp = FastMCP("MyServer")
mcp.add_tool(greet)
```

**Changes:**
- Standalone decorators in `fastmcp.tools`, `fastmcp.resources`, `fastmcp.prompts`
- `FileSystemProvider` moved to `fastmcp.server.providers`
- `LocalProvider.resource()` and `prompt()` delegate to standalone decorators
- Comprehensive `enabled=False` tests for all component types
- Eliminates `fastmcp.fs` module entirely

Note: this is also a significant refactor of the new FileSystemProvider (#2823) which had its own metadata-based decorators and was housed in the wrong submodule

### DIFF
--- a/src/fastmcp/server/providers/local_provider.py
+++ b/src/fastmcp/server/providers/local_provider.py
@@ -461,7 +461,7 @@ class LocalProvider(Provider):
             annotations = ToolAnnotations(**annotations)
 
         if isinstance(name_or_fn, classmethod):
-            raise ValueError(
+            raise TypeError(
                 inspect.cleandoc(
                     """
                     To decorate a classmethod, first define the method and then call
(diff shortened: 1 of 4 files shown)


jlowin:
@coderabbitai review

PR #2853, opened 2026-01-12
Fix timeout=None overriding create_mcp_http_client defaults

jlowin:
When `timeout` is `None`, passing it explicitly to `create_mcp_http_client()` overrides the function's built-in defaults (30s connect, 5min read). The factory path already uses conditional unpacking to avoid this—now the direct call path does too.

```python
# Before: always passes timeout, even when None
http_client = create_mcp_http_client(
    headers=headers,
    timeout=timeout,  # None overrides defaults
    auth=self.auth,
)

# After: only passes timeout when explicitly set
http_client = create_mcp_http_client(
    headers=headers,
    auth=self.auth,
    **({"timeout": timeout} if timeout is not None else {}),
)
```

### DIFF
--- a/src/fastmcp/client/transports.py
+++ b/src/fastmcp/client/transports.py
@@ -288,8 +288,8 @@ class StreamableHttpTransport(ClientTransport):
         else:
             http_client = create_mcp_http_client(
                 headers=headers,
-                timeout=timeout,
                 auth=self.auth,
+                **({"timeout": timeout} if timeout is not None else {}),
             )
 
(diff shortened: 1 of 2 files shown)


jlowin:
Closing - the fix is unnecessary. `create_mcp_http_client` explicitly handles `timeout=None` and applies MCP defaults internally, so passing `None` is equivalent to omitting the argument. The SSE pattern is needed because `sse_client` uses signature defaults, but that doesn't apply here.

PR #2854, opened 2026-01-12
Fix timeout=None overriding create_mcp_http_client defaults

jlowin:
Cherry-pick of #2853 for the 2.x release line.

When `timeout` is `None`, passing it explicitly to `create_mcp_http_client()` overrides the function's built-in defaults. Now uses conditional unpacking to preserve defaults when no timeout is specified.


jlowin:
Closing - see #2853.

PR #2856, opened 2026-01-12
Decorators return functions instead of component objects

### LINKED ISSUE #2292: Explore Tool Decorator Changes
### Enhancement

Let's explore this https://github.com/jlowin/fastmcp/issues/2257

> Totally unrelated to https://github.com/beartype and type-checking, but [FastMCP's "Decorating Methods" caveats doesn't seem quite right](https://gofastmcp.com/patterns/decorating-methods). On the face of it, it really does seem like a future version of FastMCP should be able to transparently support standard-issue decoration use cases like:
> from fastmcp import FastMCP
> mcp = FastMCP()
> 
> class MyClass:
>     @mcp.tool  # <-- Seems like FastMCP really should be able to make that work? Maybe?
>     def my_method(self, x: int) -> int:
>         return x * 2
> In theory, the @FastMCP.tool decorator should be able to support both instance and class methods by (in order):
> 
> Detecting when it's being applied to an instance or class method. @FastMCP.tool currently detects only the latter, but could detect the former as well by either:
> Trivial signature introspection. If the first parameter is self, it's almost certainly an instance method.
> Non-trivial call stack introspection. Technically, classes are lexical scopes. This means that @FastMCP.tool can iteratively crawl up the call stack until it either leaves a module boundary or finds a class scope. Definitely hard stuff. Definitely doable, too. Personally, I'd just wing it and introspect for self parameters until users complain. 😆
> If @FastMCP.tool detects it's being applied to an instance or class method, it returns a non-data descriptor (i.e., an object whose class satisfies the descriptor protocol by defining the __get__() dunder method) rather than doing what it currently does. That descriptor's __get__() dunder method then (in order):
> Binds the function it was instantiated with to either the desired instance or class object.
> DOES ALL OF THE STUFF THE @FastMCP.tool DECORATOR CURRENTLY DOES. Clearly, this is where the saucy magic sauce happens. 😋
> Returns the now-bound instance or class method.
> In theory, everything should just "work" from the end user perspective. You've almost certainly thought of all that, because you are FastMCP. There's a good reason why that simply doesn't work, I'm sure. Still. Descriptors are fun! I'm saying "fun" a lot. It's almost like that word means something different when I use it.

jlowin:
Decorators (`@tool`, `@resource`, `@prompt`) now return the original function unchanged instead of transforming it into a component object. This aligns FastMCP with how decorators work in Flask, FastAPI, and Typer.

```python
@mcp.tool
def greet(name: str) -> str:
    return f"Hello, {name}!"

# v3: greet is still your function
greet("World")  # "Hello, World!"

# v2: greet was a FunctionTool object
# isinstance(greet, FunctionTool)  # was True
```

**Why this matters:**
- Functions stay callable for testing and reuse
- Instance methods work naturally with `mcp.add_tool(obj.method)`
- More intuitive - decorators don't transform what you wrote

**Breaking change:** Code that treats decorated functions as `FunctionTool`/`FunctionResource`/`FunctionPrompt` objects will break. Set `FASTMCP_DECORATOR_MODE=object` for v2 behavior.

Closes #2292

### DIFF
--- a/src/fastmcp/prompts/function_prompt.py
+++ b/src/fastmcp/prompts/function_prompt.py
@@ -198,9 +198,13 @@ class FunctionPrompt(Prompt):
                                 arg_description = f"{arg_description}\n\n{schema_note}"
                             else:
                                 arg_description = schema_note
-                        except Exception:
+                        except Exception as e:
                             # If schema generation fails, skip enhancement
-                            pass
+                            logger.debug(
+                                "Failed to generate schema for prompt argument %s: %s",
+                                param_name,
+                                e,
+                            )
 
                 arguments.append(
                     PromptArgument(
--- a/src/fastmcp/tools/function_tool.py
+++ b/src/fastmcp/tools/function_tool.py
@@ -196,9 +196,9 @@ class ParsedFunction:
                 # include_extras=True preserves Annotated metadata
                 type_hints = get_type_hints(fn, include_extras=True)
                 output_type = type_hints.get("return", output_type)
-            except Exception:
+            except Exception as e:
                 # If resolution fails, keep the string annotation
-                pass
+                logger.debug("Failed to resolve type hint for return annotation: %s", e)
 
         if output_type not in (inspect._empty, None, Any, ...):
             # there are a variety of types that we don't want to attempt to


leycec:
**Wondrous.** Interestingly, the biggest gain here is interoperability with Python's existing decorator ecosystem. Previously, FastMCP decorators weren't safely chainable with other third-party decorators. Now they are.

This actually matters for @beartype. We explicitly support FastMCP by treating decorated callables as `FunctionTool/FunctionResource/FunctionPrompt` objects. We should stop doing that. I've made an internal note to myself to stop doing that. My older self will ignore that note, though. I know that guy. He never reads anything. In 2026, who's got the time? :joy:

PR #2869, opened 2026-01-13
Add OpenTelemetry tracing support

### LINKED ISSUE #2813: Add native OTEL integration
### Enhancement

Native OTEL integration -- @strawgate has a middleware-based implementation in #2001, but with the new Provider backend we can get much richer telemetry by adding tracing at that layer (especially as providers are now the universal way to access all components), for example when nested mounted servers or proxy servers are called. @chrisguidry is looking into it.

chrisguidry:
Adds OpenTelemetry tracing for observability into FastMCP server and client operations.

Server spans are created for tool calls, resource reads, and prompt renders with attributes like component key, component type, provider type, session ID, and auth context. Client spans wrap outgoing calls with trace context propagation via W3C headers in request meta.

```python
# Tracing is always active - to export traces, configure an OTel SDK
from opentelemetry.sdk.trace import TracerProvider
from opentelemetry.sdk.trace.export import BatchSpanProcessor
from opentelemetry.exporter.otlp.proto.grpc.trace_exporter import OTLPSpanExporter

provider = TracerProvider()
provider.add_span_processor(BatchSpanProcessor(OTLPSpanExporter()))
trace.set_tracer_provider(provider)

# Then use fastmcp normally - spans export to your configured backend
```

Components provide their own span attributes through a `get_span_attributes()` method that subclasses override - this lets LocalProvider, FastMCPProvider, and ProxyProvider each include relevant context (original names, backend URIs, etc).

image

Closes #2813

🤖 Generated with Claude Code

### DIFF
--- a/loq.toml
+++ b/loq.toml
@@ -1,319 +1,103 @@
 # loq configuration - file size enforcement
 # Run `loq baseline` to update when files exceed limits
 
-default_max_lines = 500
+default_max_lines = 1000
 respect_gitignore = true
 
-exclude = [
-    "**/uv.lock",
-    ".git/",
-    "docs/",
-]
+exclude = ["**/uv.lock", ".git/**", "docs/**"]
 
 [[rules]]
-path = "src/fastmcp/server/auth/oauth_proxy.py"
-max_lines = 2282
+path = "tests/**"
+max_lines = 1000
+
+[[rules]]
+path = "tests/server/providers/test_local_provider_tools.py"
+max_lines = 1554
 
 [[rules]]
 path = "tests/client/test_client.py"
 max_lines = 1438
 
 [[rules]]
-path = "src/fastmcp/client/tasks.py"
-max_lines = 551
-
-[[rules]]
 path = "tests/server/test_auth_integration.py"
 max_lines = 1242
-
-[[rules]]
-path = "tests/server/test_dependencies.py"
-max_lines = 1046
-
-[[rules]]
-path = "tests/prompts/test_prompt.py"
-max_lines = 552
-
-[[rules]]
-path = "tests/client/test_elicitation.py"
-max_lines = 1132
-
-[[rules]]
-path = "src/fastmcp/tools/tool_transform.py"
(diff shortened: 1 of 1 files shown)


docs/servers/telemetry.mdx
```diff
@@ -0,0 +1,302 @@
+---
+title: OpenTelemetry
+sidebarTitle: Telemetry
+description: Native OpenTelemetry instrumentation for distributed tracing.
+icon: chart-line
+---
+
+FastMCP includes native OpenTelemetry instrumentation for observability. Traces are automatically generated for all MCP operations, providing visibility into server behavior, request handling, and provider delegation chains.
```
chrisguidry: I'd say for all tool, prompt, resource, and resource template operations.  We're deferring the other lower-level stuff to the `mcp` lower-level SDK

docs/servers/telemetry.mdx
```diff
@@ -0,0 +1,302 @@
...
+
+## How It Works
+
+FastMCP uses the OpenTelemetry API (not the SDK) for instrumentation. This means:
```
chrisguidry: No need to say "not the SDK" in this

docs/servers/telemetry.mdx
```diff
@@ -0,0 +1,302 @@
...
+    raise ValueError("Something went wrong")
+
+# The span will have:
+# - status = ERROR
+# - exception event with stack trace
+```
+
+## Attributes Reference
+
+### Standard Semantic Conventions
```
chrisguidry, line 194: I think we've got some redundant sections here, let's consolidate the references into this one section and don't repeat the definitions of the attributes multiple times

examples/diagnostics/__init__.py
```diff
@@ -0,0 +1 @@
+"""FastMCP Diagnostics example - for testing tracing, errors, and observability."""
```
chrisguidry: I don't think we need this file, this isn't used as a package

PR #2898, opened 2026-01-17
Add robust signal handling for SIGINT and SIGTERM - closes #2837

### LINKED ISSUE #2837: Stdio transport requires multiple Ctrl-C to terminate
When running a FastMCP server with stdio transport directly via `python server.py`, terminating with Ctrl-C requires multiple presses and produces messy tracebacks.

## Reproduction

```python
from fastmcp import FastMCP

mcp = FastMCP("Test Server")

@mcp.tool
def hello(name: str) -> str:
    return f"Hello, {name}!"

if __name__ == "__main__":
    mcp.run()
```

Run with `python server.py`, then press Ctrl-C.

## Expected behavior

Single Ctrl-C cleanly terminates the server.

## Actual behavior

Requires 2-3 Ctrl-C presses and produces cascading `KeyboardInterrupt` tracebacks:

```
^C^CTraceback (most recent call last):
  File "..._asyncio.py", line 2365, in run
    return runner.run(wrapper())
  ...
  File ".../asyncio/runners.py", line 157, in _on_sigint
    raise KeyboardInterrupt()
KeyboardInterrupt

During handling of the above exception, another exception occurred:
...
```

## Root cause

The stdio transport blocks on `selector.select()` waiting for stdin. When SIGINT arrives:
1. The blocking call is interrupted
2. Async cleanup begins but gets interrupted by subsequent Ctrl-C
3. Threading shutdown also gets interrupted

## Notes

- HTTP transport handles Ctrl-C cleanly (uvicorn has proper signal handling)
- Stdio is designed to be controlled by a parent MCP client process, so interactive use is an edge case
- However, `python server.py` is a common way to test servers during development

Fern-Ali:
## Description

Stdio transport required pressing Ctrl+C twice to terminate the server. Signal handling for SIGINT/SIGTERM needs to be installed early (before entering the async runtime) to enable single Ctrl+C termination.

**Changes:**

- **New `run_stdio()` sync method**: Created dedicated sync entry point with signal handling
  - Wraps `run_stdio_async()` with proper signal handler installation/cleanup
  - Can be called directly by users or through `.run()`

- **Updated `.run()` method**: Now delegates to `run_stdio()` for stdio transport (default)
  - Simplified implementation - just routing logic
  - HTTP/SSE transports continue using `anyio.run()` directly (they have uvicorn's handlers)

- **Updated `run_stdio_async()` documentation**: 
  - Clarified this is the async implementation for advanced users
  - Recommends using `run_stdio()` or `.run()` for most use cases
  - Notes that direct callers should manage their own signal handling

- **Signal handler location**: 
  - Handlers in `run_stdio()` - the canonical sync entry point for stdio
  - Installs handlers BEFORE entering `anyio.run()` for single Ctrl+C to work
  - Only for main thread (transport-specific, thread-aware)
  - Includes `os._exit(0)` call for immediate termination
  - Full cleanup with signal handler restoration in finally block

- **Graceful shutdown documentation**:
  - Comprehensive comment on `os._exit(0)` usage in `run_stdio()`
  - Explains necessity: blocking stdio streams require immediate termination
  - Documents future path: refactor to graceful shutdown when MCP SDK v2 supports it

- **Test suite updates**:
  - Added test for new `run_stdio()` method
  - Updated `.run()` test to verify delegation to `run_stdio()`
  - All 9 tests passing
  - Cleaned up unused imports and variables for linting compliance

**Usage Patterns:**

**Pattern 1: Simple sync script**
```python
from fastmcp import FastMCP
mcp = FastMCP("My Server")

@mcp.tool()
def hello(name: str) -> str:
    return f"Hello, {name}!"

if __name__ == "__main__":
    mcp.run()  # ✅ Signal handling included, single Ctrl+C works
```

**Pattern 2: Direct stdio call**
```python
from fastmcp import FastMCP
mcp = FastMCP("My Server")

if __name__ == "__main__":
    mcp.run_stdio()  # ✅ Signal handling included, single Ctrl+C works
```

**Pattern 3: Pure async (advanced users managing own event loop)**
```python
import asyncio
from fastmcp import FastMCP

mcp = FastMCP("My Server")

async def main():
    # When already in async context, use run_stdio_async()
    await mcp.run_stdio_async()  # ⚠️ User must manage own signals, requires 2 Ctrl+C

asyncio.run(main())
```

**Key Benefits:**

**Single Ctrl+C**: Users only need to press Ctrl+C once (not twice)
**Clean User Experience**: No traceback on Ctrl+C, clean shutdown message
**Flexible API**: Multiple entry points for different use cases
**Well-Documented**: Clear guidance on which method to use when
**Transport-Specific**: Signal handling only for stdio (HTTP/SSE use uvicorn)
**Thread-Aware**: Handlers only in main thread
**Proper Cleanup**: Signal handlers restored in finally block
**Groundwork for Future**: Clear path for graceful shutdown when MCP SDK v2 supports it
**No Breaking Changes**: `run_stdio_async()` still exists and works as before, backward compatible

**Architecture:**

The solution creates a clear hierarchy:
- **`run_stdio()`** - Sync wrapper with signal handling (RECOMMENDED for most users)
- **`run_stdio_async()`** - Async implementation without signal handling (for advanced users managing own event loop)
- **`.run()`** - High-level convenience that delegates to appropriate method

This ensures signal handling is available regardless of entry point while maintaining clean separation of concerns.

**Files Changed:**
- `src/fastmcp/server/server.py` - Added `run_stdio()`, updated `.run()`, updated `run_stdio_async()` docs
- `tests/server/test_signal_handling.py` - Added test for `run_stdio()`, updated existing tests

**AI Tool Used:** GitHub Copilot

**Contributors Checklist**

**Review Checklist**

---

**Tests passing!** ✅ All 9 signal handling tests pass  
**Linting clean!** ✅ All ruff checks pass

### DIFF
--- a/src/fastmcp/server/server.py
+++ b/src/fastmcp/server/server.py
@@ -2636,8 +2636,9 @@ class FastMCP(Provider, Generic[LifespanResultT]):
         """Run the server using stdio transport with signal handling.
 
         This is the recommended way to run stdio servers as it provides
-        proper signal handling for single Ctrl+C termination. It can be
-        called from both sync and async contexts.
+        proper signal handling for single Ctrl+C termination.
+        Call from synchronous contexts. If you're already inside an event loop,
+        use `run_stdio_async()` (and manage signals yourself).
 
         Args:
             show_banner: Whether to display the server banner


Fern-Ali:
@jlowin Let me know when you decide! I can move the logic if necessary, or update as you see fit. Thanks!

jlowin:
I think there isn't any way to gracefully stop a server, so nothing's lost by being aggressive - that may change in MCP SDK v2, so maybe just add a comment on the os.exit() that once servers support graceful shutdown it should be refactored, and please move into run_stdio_async. Then I think we can get this in. Thanks!

Fern-Ali:
hi @jlowin, I spent some time trying to move the logic to run_stdio_async, but this brings back the undesireable multiple ctrl+c to stop the server. 🥲

Handlers must be installed before entering anyio.run() - otherwise the first Ctrl+C isn't caught and you need to press it twice.

The signal handling is still transport-specific (only stdio gets handlers, not HTTP/SSE), and the comment in the code explains why it needs to be in .run() for proper functionality. -> update incoming -> lmk if its unsatisfactory

jlowin:
My worry is that many invocations don't use run because it can't be called from an async app, so the functionality ends up being orphaned

jlowin:
Maybe it belongs in the CLI routes? But then python server.py doesn't benefit...

I'm all for the functionality just want to find the right place

Fern-Ali:
that is a good point @jlowin that i didnt consider. thinking about options... 🤔

Fern-Ali:
I'm back @jlowin ! 
Firstly, thank you so much for the feedback. I really appreciate it, and I've learned a lot about event loops in the process.

After exploring this issue further, the "async app" concern appears to be a fundamental constraint of Python’s signal model: signal handlers must be registered in the main thread before an event loop starts, and async applications already have a running loop.

This PR solves the canonical paths: 
- ✅ `mcp.run()` / `mcp.run(transport="stdio")` / `mcp.run_stdio()` → single Ctrl+C
- ⚠️ `await mcp.run_stdio_async()` → user manages signals (unavoidable)

The new `run_stdio()` method puts stdio-specific logic where it belongs (not in the generic `.run()` router), following the same pattern as `run_http_async()`.

This fixes issue #2837 for normal usage. It would seem Async apps are orphaned by Python's signal model, not FastMCP.

LMK your thoughts!

Issue #2877, opened 2026-01-14
Passing context to Background Task


shorbachuk:
### Description

Running FastMCP mounted under FastAPI:

  + Exception Group Traceback (most recent call last):
  |   File "C:\Users\xxxx\AppData\Local\Programs\Python\Python313\Lib\site-packages\docket\worker.py", line 889, in _execute
  |     raise ExceptionGroup(
  |     ...<8 lines>...
  |     )
  | ExceptionGroup: Failed to resolve dependencies for parameter(s): ctx (1 sub-exception)
  +-+---------------- 1 ----------------
    | Traceback (most recent call last):
  +-+---------------- 1 ----------------
    | Traceback (most recent call last):
    | Traceback (most recent call last):
    |   File "C:\Users\xxxx\AppData\Local\Programs\Python\Python313\Lib\site-packages\docket\dependencies.py", line 795, in resolved_dependencies
    |     arguments[parameter] = await stack.enter_async_context(
    |                            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |         dependency
    |         ^^^^^^^^^^
    |     )
    |     ^
    |   File "C:\Users\xxxx\AppData\Local\Programs\Python\Python313\Lib\contextlib.py", line 668, in enter_async_context
    |     result = await _enter(cm)
    |              ^^^^^^^^^^^^^^^^
    |   File "C:\Users\xxxx\AppData\Local\Programs\Python\Python313\Lib\site-packages\fastmcp\server\dependencies.py", line 273, in __aenter__
    |     return get_context()
    |   File "C:\Users\xxxx\AppData\Local\Programs\Python\Python313\Lib\site-packages\fastmcp\server\dependencies.py", line 265, in get_context
    |     raise RuntimeError("No active context found.")
    | RuntimeError: No active context found.

### Example Code

```Python
from typing import Any, Optional, Annotated
import anyio
from pydantic import Field
from fastmcp import Context, FastMCP
from fastmcp.server.tasks import TaskConfig
from mcp.types import ToolAnnotations
from config import settings
from fastmcp.dependencies import CurrentContext
from fastmcp.server.context import Context

router = FastMCP(name="my-mcp")

@router.tool(
    name="Sample Tool",
    title="Run a sample tool",
    description="Run a sample tool",
    meta={"version": settings["VERSION"], "author": settings["AUTHORS"]},
    annotations=ToolAnnotations(destructiveHint=True, readOnlyHint=False),
[...]

Issue #2568, opened 2025-12-07
Support `input_required` task status for elicitation and sampling during background tasks [SEP-1686]


jlowin:
Background tasks currently support basic lifecycle states (`working`, `completed`, `failed`, `cancelled`) but not the `input_required` status from SEP-1686.

The `input_required` status enables advanced flows where a background task can pause execution to request elicitation or sampling from the client, then resume after receiving input.

Spec reference: https://modelcontextprotocol.io/specification/2025-11-25/basic/utilities/tasks#message-flow

Current state:
- ✅ `related-task` metadata
- ✅ Progress reporting
- ❌ `input_required` status
- ❌ Elicitation during task execution
- ❌ Sampling during task execution

gfortaine:
## Reference Implementation Available

The official **MCP Python SDK** already has a complete implementation of `input_required` status for elicitation and sampling during tasks:

📁 [`modelcontextprotocol/python-sdk/src/mcp/server/experimental/task_context.py`](https://github.com/modelcontextprotocol/python-sdk/blob/main/src/mcp/server/experimental/task_context.py)

### Key Pattern

```python
async def elicit(self, message: str, requested_schema: ElicitRequestedSchema) -> ElicitResult:
    """Send an elicitation request via the task message queue.
    
    This method:
    1. Checks client capability
    2. Updates task status to "input_required"
    3. Queues the elicitation request
    4. Waits for the response (delivered via tasks/result round-trip)
    5. Updates task status back to "working"
    6. Returns the result
    """
    self._check_elicitation_capability()
    
    # Update status to input_required
    await self._store.update_task(self.task_id, status=TASK_STATUS_INPUT_REQUIRED)
    
    # Build request and queue it
    request = self._session._build_elicit_form_request(
        message=message,
        requested_schema=requested_schema,
        related_task_id=self.task_id,
    )
    
    resolver: Resolver[dict[str, Any]] = Resolver()
    queued = QueuedMessage(type="request", message=request, resolver=resolver, ...)
    await self._queue.enqueue(self.task_id, queued)
    
    # Wait for response (routed back via TaskResultHandler)
    response_data = await resolver.wait()
    await self._store.update_task(self.task_id, status=TASK_STATUS_WORKING)
    return ElicitResult.model_validate(response_data)
```

### Components Needed

| Component | Purpose |
|-----------|---------|
| `TaskMessageQueue` | Queue outbound requests to send when client polls |
[...]

PR #2905, opened 2026-01-18
feat(context): Add background task support for Context (SEP-1686)

gfortaine:
## Summary

This PR implements background task support for the `Context` class following the unified interface pattern suggested by @chrisguidry in code review.

**Key Changes:**
- Added `task_id` parameter to `Context.__init__()`
- Added `is_background_task` and `task_id` properties
- Enhanced `elicit()` to work transparently in background task mode
- Added `tasks/elicitation.py` module for Redis-based coordination

## How It Works

```python
@mcp.tool(task=True)
async def interactive_task(ctx: Context) -> str:
    # This works transparently in both request and background task modes
    result = await ctx.elicit("Please provide additional input", str)
    
    if isinstance(result, AcceptedElicitation):
        return f"You provided: {result.data}"
    else:
        return "Elicitation was declined or cancelled"
```

When running as a background task:
1. Task status is set to `input_required` in Redis
2. `notifications/tasks/updated` is sent with elicitation metadata
3. Task waits for client to send input via `tasks/sendInput`
4. Input is stored in Redis and task resumes

## Architecture

```
┌─────────────────────────────────────────────────────┐
│                    User Code                        │
│     await ctx.elicit("Need input", str)            │
└─────────────────────────────────────────────────────┘
                         │
                         ▼
┌─────────────────────────────────────────────────────┐
│              Context.elicit()                       │
│   if is_background_task:                           │
│       return _elicit_for_task(...)                 │
│   else:                                            │
│       return session.elicit(...)                   │
└─────────────────────────────────────────────────────┘
```

## Approach Change

Originally proposed `TaskContext` as a separate dependency class. Based on review feedback, refactored to enhance the existing `Context` class instead, following the v3.0 principle of unified interfaces.

## Tests

- 10 new tests covering background task support
- All 1857 server tests passing

## Related

- Implements SEP-1686 (`input_required` status)
- Addresses #2568, fixes #2877

## Breaking Changes

None. The new `task_id` parameter is optional and defaults to `None`.

PR #2906, opened 2026-01-18
feat: distributed notification queue + BLPOP elicitation for background tasks

### LINKED ISSUE #2904: Task polling statusMessage not forwarded to MCP client
## Problem

When using FastMCP with tasks enabled, the `statusMessage` field from task status updates is not being forwarded to the MCP client during task polling. This causes progress messages to not appear in the client UI (e.g., VS Code chat progress).

## Current Behavior

1. MCP server tool returns a task result
2. Tool updates `statusMessage` during execution (e.g., "Searching...", "Analyzing...")
3. Client polls `tasks/get` 
4. `statusMessage` is visible in server stderr logs but **not displayed in client UI**

## Expected Behavior

The `statusMessage` from task state should be included in `tasks/get` responses so clients can display progress to users.

## Environment

- FastMCP version: 2.5.1+
- MCP client: VS Code 1.108 (stable)
- Python: 3.12

## Workaround

Using the official MCP Python SDK (`mcp>=1.10.0` from `modelcontextprotocol/python-sdk`) instead of FastMCP correctly forwards `statusMessage` to clients.

## Investigation

The issue appears to be in how FastMCP's task infrastructure handles status updates. The official MCP SDK's `TaskSupport` class properly includes `statusMessage` in task get responses.

## Related

- Originally misattributed to VS Code: https://github.com/microsoft/vscode/issues/287893 (now closed)
- MCP Tasks specification: SEP-1732

gfortaine:
## Summary

Implements reliable notification delivery and efficient elicitation for background tasks (SEP-1686).

Fixes #2904 (`statusMessage` not forwarded), addresses feedback from #2877 and #2568.

### `notifications.py` (new — 256 lines)
Distributed notification queue using LPUSH/BRPOP pattern. Workers push to a Redis queue per session, MCP server process subscribes and forwards to the client. Retry up to 3x on delivery failure, TTL-based expiration for stale messages.

### `elicitation.py`
- Replaced polling (7,200 round-trips/hour) with BLPOP (single blocking call)
- Fail-fast: if notification push fails, return `cancel` immediately instead of blocking for 1 hour
- Session ID from `get_task_context()` (authoritative source) instead of session attribute
- Typed notifications via `TaskStatusNotification.model_validate()`

### `context.py`
- `report_progress()` falls back to Docket's `ExecutionProgress` when no `progressToken` (background tasks)
- Uses delta tracking (`_fastmcp_last_progress`) with `increment()` since Docket only exposes relative progress

### `handlers.py`
- Initial `notifications/tasks/status` notification before queueing (typed, not raw JSONRPC)
- Subscriber lifecycle management: starts on first task, cleans up on session disconnect

### Tests (zero mocks)
- `test_context_background_task.py`: 896 → 394 lines. `Client(mcp)` + `memory://` backend.
- `test_notifications.py`: 165 new lines. `NotificationCaptureHandler` verifies E2E notification metadata.
- Only surgical mock: `TestElicitFailFast` patches `push_notification` to test fail-fast behavior.

## Notes
- Rebased on v3.0.0b2 (4 commits)
- All 256 task tests pass, `ty` static analysis clean
- Backward compatible — foreground progress behavior unchanged

### DIFF
--- a/docs/python-sdk/fastmcp-server-providers-openapi-components.mdx
+++ b/docs/python-sdk/fastmcp-server-providers-openapi-components.mdx
@@ -27,7 +27,7 @@ run(self, arguments: dict[str, Any]) -> ToolResult
 Execute the HTTP request using RequestDirector.
 
 
-### `OpenAPIResource` <sup><a href="https://github.com/jlowin/fastmcp/blob/main/src/fastmcp/server/providers/openapi/components.py#L222" target="_blank"><Icon icon="github" style="width: 14px; height: 14px;" /></a></sup>
+### `OpenAPIResource` <sup><a href="https://github.com/jlowin/fastmcp/blob/main/src/fastmcp/server/providers/openapi/components.py#L233" target="_blank"><Icon icon="github" style="width: 14px; height: 14px;" /></a></sup>
 
 
 Resource implementation for OpenAPI endpoints.
@@ -35,7 +35,7 @@ Resource implementation for OpenAPI endpoints.
 
 **Methods:**
 
-#### `read` <sup><a href="https://github.com/jlowin/fastmcp/blob/main/src/fastmcp/server/providers/openapi/components.py#L252" target="_blank"><Icon icon="github" style="width: 14px; height: 14px;" /></a></sup>
+#### `read` <sup><a href="https://github.com/jlowin/fastmcp/blob/main/src/fastmcp/server/providers/openapi/components.py#L263" target="_blank"><Icon icon="github" style="width: 14px; height: 14px;" /></a></sup>
 
 ```python
 read(self) -> ResourceResult
@@ -44,7 +44,7 @@ read(self) -> ResourceResult
 Fetch the resource data by making an HTTP request.
 
 
-### `OpenAPIResourceTemplate` <sup><a href="https://github.com/jlowin/fastmcp/blob/main/src/fastmcp/server/providers/openapi/components.py#L336" target="_blank"><Icon icon="github" style="width: 14px; height: 14px;" /></a></sup>
+### `OpenAPIResourceTemplate` <sup><a href="https://github.com/jlowin/fastmcp/blob/main/src/fastmcp/server/providers/openapi/components.py#L347" target="_blank"><Icon icon="github" style="width: 14px; height: 14px;" /></a></sup>
 
 
 Resource template implementation for OpenAPI endpoints.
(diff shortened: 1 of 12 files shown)


gfortaine:
hey team 👋

follow-up to #2905 — this adds distributed notification support so elicitation actually works for background tasks running on remote workers.

ran a deep review loop and caught a few things worth mentioning:

**fixes included in this pr:**
- `set_current()` instead of `increment()` for progress — atomic, no race condition if the same tool reports from multiple contexts
- fail-fast on `push_notification` failure — no 1-hour hang waiting for a response that'll never come  
- cleanup registration via exit_stack — `stop_subscriber` actually gets called now
- `get_task_context().session_id` as single source of truth — removes closure capture mismatch risk

tests: 280 passed, 3 skipped. rebased on main.

happy to walk through any of the implementation details if that would help. is there anything specific you'd like me to adjust before this gets merged?

cc @jlowin @chrisguidry

chrisguidry:
This is looking really compelling so far, @gfortaine, thank you!  

However, similar to the last PR, I'm having trouble with the style of the tests here.  I think these heavily mocked tests are going to end up being difficult to maintain over time, because they are mocking many internal implementation details (including of redis itself).  Take a look at the surrounding tests task-related tests and you'll see a very stark difference in style.

Can we rework these tests to be simpler by just setting up example MCP servers with task support that make elicitations and then assert that the client sees those elicitations?  We should be able to write tests that just "do the thing" because we have redis available (via the `memory://` broker) while the tests are running.

In our `/python-tests` skill in the repo, we've got some good guidance here: https://github.com/jlowin/fastmcp/blob/main/.claude/skills/python-tests/SKILL.md#dont-mock-what-you-own

gfortaine:
@chrisguidry totally fair — mocking redis internals creates maintenance debt and the contrast with the surrounding task tests was stark. rewrote everything.

- `test_context_background_task.py`: 896 → 394 lines. `Client(mcp)` + `memory://`, zero mocks. tests just spin up a server, call tools with `task=True`, elicit, assert.
- `test_notifications.py`: added `NotificationCaptureHandler` that verifies the client sees `notifications/tasks/status` with `input_required` metadata — exactly the end-to-end assertion you asked for.
- only surgical mock: `TestElicitFailFast` patches `push_notification` to test fail-fast. everything else is real.

also switched to `TaskStatusNotification.model_validate()` for typed notification path and fixed the `ty` static analysis issues.

rebased on `v3.0.0b2` (4 commits). how does this land for you?

src/fastmcp/server/tasks/elicitation.py
```diff
@@ -5,7 +5,7 @@
 an active request context, so elicitation requires special handling:
 
 1. Set task status to "input_required" via Redis
-2. Send notifications/tasks/updated with elicitation metadata
+2. Send notifications/tasks/status with elicitation metadata
```
chrisguidry, line 8: Thank you! 😅

PR #2908, opened 2026-01-18
[Draft] Enhance debug settings with dual-purpose flag and granular control

### LINKED ISSUE #2860: Clarify debuug options
### Description

I just wanted to clarify the intent of the debug options; I will be happy to make PR if I've got this right

There is currently `FASTMCP_DEBUG` 

https://github.com/jlowin/fastmcp/blob/84a43da28951148b5114a23b55d5faa42ca94d42/src/fastmcp/server/server.py#L251

and `FASTMCP_LOG_LEVEL`

https://github.com/jlowin/fastmcp/blob/84a43da28951148b5114a23b55d5faa42ca94d42/src/fastmcp/server/server.py#L250

The `log_level` is documented [global setting](https://gofastmcp.com/servers/server#global-settings) but it doesn't mention the `debug` parameter.

AFAICT, `debug` only gets passed through to the `Starlette()` creation as the [debug argument](https://starlette.dev/applications/) ("Boolean indicating if debug tracebacks should be returned on errors").  Otherwise it has no effect on fastmcp, or if you're not using a HTTP-ish backend.

So I think we should document this behaviour of `FASTMCP_DEBUG` -- because it's a bit confusing a priori as it's not _really_ giving you _fastmcp_ debugging, as such, but the middleware error tracebacks.  

`FASTMCP_LOG_LEVEL` gets translated through to the `uvicorn` `log_level` argument automatically by `run()` which is fine, or can be overrideen with an argument.  That is documented.

I'm also wondering if calling `fastmcp.utilities.logging.configure_logging()` a second time is considered OK?  I know that the `fastmcp` wrapper is probably preferred, but I have a range of other settings that means starting the service from an entry-point in my `pyproject.toml` and calling `run()` directly works nicely ... but since the logging is set on import (https://github.com/jlowin/fastmcp/blob/main/src/fastmcp/__init__.py#L9) to implement a `--debug` argument in my entry-point script, I have been making it call `configure_logging()`.   Is there a better way to implement this?

### Example Code

```Python

```

### Version Information

```Text
2.14
```

strawgate:
The existing `debug` setting was confusing because it only affected Starlette's debug tracebacks for HTTP/SSE transports, not FastMCP's logging. Users naturally expected `FASTMCP_DEBUG` to enable FastMCP debugging, leading to confusion.

This PR implements a dual approach:

**The `debug` flag now does BOTH:**
- Sets `log_level` to DEBUG
- Enables Starlette debug tracebacks for HTTP/SSE transports

**New `starlette_debug` flag for granular control:**
- Only enables Starlette debug tracebacks
- No effect on log level
- Useful when you want HTTP error details without DEBUG logging

This provides both convenience (debug) and precision (starlette_debug).

**Example usage:**
```python
# Enable everything for debugging
export FASTMCP_DEBUG=true

# Or granular control
export FASTMCP_LOG_LEVEL=INFO
export FASTMCP_STARLETTE_DEBUG=true
```

Closes #2860

---

### DIFF
--- a/docs/servers/server.mdx
+++ b/docs/servers/server.mdx
@@ -362,8 +362,8 @@ print(fastmcp.settings.strict_input_validation)  # Default: False
 
 Common global settings include:
 - **`log_level`**: Logging level ("DEBUG", "INFO", "WARNING", "ERROR", "CRITICAL"), set with `FASTMCP_LOG_LEVEL`
-- **`debug`**: Global debug mode that sets log level to DEBUG and enables Starlette debug tracebacks for HTTP/SSE transports, set with `FASTMCP_DEBUG`. Provides a convenient way to enable comprehensive debugging. For granular control, use `log_level` and `starlette_debug` separately
-- **`starlette_debug`**: Enable Starlette debug mode for HTTP/SSE transports only, set with `FASTMCP_STARLETTE_DEBUG`. When enabled, detailed error tracebacks are returned in HTTP responses. Only affects HTTP/SSE transports; has no effect on stdio transport
+- **`debug`**: Global debug mode that sets log level to DEBUG and enables Starlette debug tracebacks for HTTP/SSE transports, set with `FASTMCP_DEBUG`. Provides a convenient way to enable comprehensive debugging. For granular control, use `log_level` and `starlette_debug` separately. **WARNING**: Only enable in development - debug tracebacks expose detailed internal information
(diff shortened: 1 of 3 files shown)


strawgate:
/marvin can you address feedback and fix the failing static analysis checks

jlowin:
I think the starlette debug thing is an accident (or LLM enthusiasm) -- I personally mainly use this feature just to have a way of detecting if I'm in a debugging session, not for any FastMCP functionality.

strawgate:
@jlowin should we just remove the debug flag then? Or what do you think we should do here

Issue #2899, opened 2026-01-17
Add unit tests that don't include dev dependencies


jlowin:
### Enhancement

Ensure that functionality doesn't inadvertently miss a critical dependency.

strawgate:
i think we'll need a "minimum-dev" or something as we won't even have pytest?

/marvin can you take a look at this and come up with the best plan and make a PR

strawgate:
/marvin make a PR where the new workflow is at the root of the repo and I'll move it into the right folder

PR #2909, opened 2026-01-18
[Draft]Add workflow to test minimal dependencies

strawgate:
This workflow addresses #2899 by testing FastMCP without dev dependencies to catch missing runtime dependencies.

The workflow installs only runtime dependencies (no dev group), then adds minimal test infrastructure (pytest and pytest-asyncio) to run import checks, smoke tests, and CLI verification. This ensures that all runtime functionality works with just the declared dependencies.

As requested, the workflow file is at the root of the repo for easy relocation.

### DIFF
--- a/test-minimal-deps.yml
+++ b/test-minimal-deps.yml
@@ -1,0 +1,78 @@
+name: Test Minimal Dependencies
+
+env:
+  PY_COLORS: 1
+
+on:
+  push:
+    branches: ["main"]
+    paths:
+      - "src/**"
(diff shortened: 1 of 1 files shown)


strawgate:
This looks not like what we want

jlowin:
I was trying to figure out how to choose the tests for this

strawgate:
@jlowin perhaps we just add some non-pytest smoke tests?

Ie uv run in a bash script and check the outputs or something?

Issue #2495, opened 2025-11-27
Expose the upstream token in the context when using OAuthProxy


nscendoni:
## Problem

When using `OAuthProxy` to authenticate users, there's no way for tools or downstream services to access the original OAuth token from the upstream identity provider (GitHub, Google, etc.). This blocks multi-tier architectures where the user's upstream token needs to be sent or exchanged to access to backend services.
The proposed change exposes the upstream OAuth token only to the backend MCP server, while the MCP client continues to receive only the OAuthProxy access token.

This is the architecture we would like to implement:
```
MCP Client → MCP Server (validates with OAuthProxy) → Backend Services (needs user upsteam token)
```
or 
```
MCP Client → Proxy MCP (validates with OAuthProxy) → Backend MCP (needs user upsteam token)
```

## Proposed Solution

Expose the upstream token through `Context.upstream_token`:

```python
@server.tool
async def call_backend_api(ctx: Context) -> str:
    # Access the original OAuth token
    token = ctx.upstream_token
    if token:
        headers = {"Authorization": f"Bearer {token}"}
        # Forward to backend service
        ...
```

**Implementation proposal**:

1. Store upstream token in `AccessToken` when `OAuthProxy` validates tokens
2. Transfer token to request scope via `RequireAuthMiddleware` 
3. Expose through `Context.upstream_token` property

If the proposal is accepted I can make a PR.

TheITRx:
Just out of  curiosity what are the use cases of the downstream needing the upstream token? Wouldn’t you use your downstream token to OBO to another downstream.

jlowin:
The `get_access_token` method already returns the upstream token, I'm pretty sure. Are you seeing a different behavior where it returns the client-provided token?

nscendoni:
@jlowin Thanks for the tip! It works as you described it. You can resolve this ticket.

jlowin:
Great!

Issue #2447, opened 2025-11-18
AzureProvider cannot authenticate with external Microsoft APIs - PowerBI


rpgeddam:
### Description

I'm trying to create an MCP server that interacts with the PowerBI REST API using Azure AD authentication. I need the powerBI scope `https://analysis.windows.net/powerbi/api/Report.Read.All`

FastMCP documentation Forces `required_scopes` to be custom API scopes like `api://my-client-id/my.scope` 
Entra ID considers the powerBI scope and the custom API scope as belonging to different resources and will not issue a token that works for both.

```
# Attempt #1
auth_provider = AzureProvider(
    client_id=os.getenv("AZURE_CLIENT_ID"),
    client_secret=os.getenv("AZURE_CLIENT_SECRET"),
    tenant_id=os.getenv("AZURE_TENANT_ID"),
    base_url="http://localhost:8000",
    required_scopes=["custom.scope"],
    additional_authorize_scopes=["openid", "offline_access", "https://analysis.windows.net/powerbi/api/Report.Read.All"],
)

# Wont work because custom.scope and powerbi scopes are on different resources
# AADSTS28000: Provided value for the input parameter scope is not valid 
# because it contains more than one resource.
```

```
# Attempt #2
auth_provider = AzureProvider(
    client_id=os.getenv("AZURE_CLIENT_ID"),
    client_secret=os.getenv("AZURE_CLIENT_SECRET"),
    tenant_id=os.getenv("AZURE_TENANT_ID"),
    base_url="http://localhost:8000",
    required_scopes=["https://analysis.windows.net/powerbi/api/Report.Read.All"],
)

# Wont work because AzureProvider replaces the prefix on the scope
# The application asked for scope 'Report.Read.All' that doesn't exist.
```
What I expect would work:
```
  auth_provider = AzureProvider(
      client_id=os.getenv("AZURE_CLIENT_ID"),
      client_secret=os.getenv("AZURE_CLIENT_SECRET"),
      tenant_id=os.getenv("AZURE_TENANT_ID"),
      base_url="http://localhost:8000",
      required_scopes=["https://analysis.windows.net/powerbi/api/Report.Read.All"],
      additional_authorize_scopes=["openid", "offline_access"],
  )
```

 I expect his affects anyone using AzureProvider with:
  - PowerBI
  - Azure Storage
  - Azure DevOps
  - Any Microsoft API that isn't Graph or a custom API

### Example Code

```Python

```

### Version Information

```Text
[...]

PR #2918, opened 2026-01-19
Add Azure OBO dependencies, auth token injection, and documentation

jlowin:
Inspired by @pamelafox's [blog post on OBO flow for Entra-based MCP servers](https://blog.pamelafox.org/2026/01/using-on-behalf-of-flow-for-entra-based.html).

FastMCP's OAuth Proxy pattern already enables On-Behalf-Of (OBO) flows with Azure/Entra - when a user authenticates, the upstream Azure token is stored and returned via `get_access_token().token`. But actually *using* that token for OBO required boilerplate: configuring MSAL, calling `acquire_token_on_behalf_of`, handling errors.

This PR adds dependency injection support and documentation that makes token access and OBO exchanges declarative.

## Token Access (Any Auth Provider)

Extract the authenticated user's token or specific claims directly into your function parameters:

```python
from fastmcp.server.dependencies import CurrentAccessToken, TokenClaim
from fastmcp.server.auth import AccessToken

@mcp.tool()
async def my_tool(
    token: AccessToken = CurrentAccessToken,  # Full token object
    user_id: str = TokenClaim("oid"),          # Just the claim you need
):
    # No get_access_token() boilerplate
    ...
```

## Azure OBO (requires `fastmcp[azure]`)

OBO typically requires setting up MSAL's `ConfidentialClientApplication` with your credentials, calling `acquire_token_on_behalf_of` with the user's assertion token, and handling the response. We provide dependencies that wrap all of this.

**`EntraOBOToken`** handles the complete exchange - just declare what scopes you need:

```python
from fastmcp.server.auth.providers.azure import EntraOBOToken

@mcp.tool()
async def get_emails(
    graph_token: str = EntraOBOToken(["https://graph.microsoft.com/Mail.Read"]),
):
    # graph_token is ready - OBO exchange happened automatically
    async with httpx.AsyncClient() as client:
        resp = await client.get(
            "https://graph.microsoft.com/v1.0/me/messages",
            headers={"Authorization": f"Bearer {graph_token}"}
        )
```

**`MSALApp`** provides direct access to a pre-configured MSAL client for custom scenarios:

```python
from fastmcp.server.auth.providers.azure import MSALApp

@mcp.tool()
async def custom_exchange(msal: ConfidentialClientApplication = MSALApp):
    result = msal.acquire_token_on_behalf_of(
        user_assertion=get_access_token().token,
        scopes=["https://graph.microsoft.com/.default"],
    )
    # Handle result, caching, silent acquisition, etc.
```

## Other Changes

- Added `fastmcp[azure]` extra with `msal` dependency
- Split `context.mdx` into separate Context and Dependencies doc pages

Related: #2447, #2495

### DIFF
--- a/src/fastmcp/server/dependencies.py
+++ b/src/fastmcp/server/dependencies.py
@@ -1124,6 +1124,50 @@ class Progress(Dependency):  # type: ignore[misc]
         pass
 
 
+# --- Access Token dependency ---
+
+
+class _CurrentAccessToken(Dependency):  # type: ignore[misc]
+    """Async context manager for AccessToken dependency."""
+
+    async def __aenter__(self) -> AccessToken:
+        token = get_access_token()
+        if token is None:
+            raise RuntimeError(
+                "No access token found. Ensure authentication is configured "
+                "and the request is authenticated."
+            )
+        return token
+
+    async def __aexit__(self, *args: object) -> None:
+        pass
+
+
+def CurrentAccessToken() -> AccessToken:
+    """Get the current access token for the authenticated user.
+
+    This dependency provides access to the AccessToken for the current
+    authenticated request. Raises an error if no authentication is present.
+
+    Returns:
+        A dependency that resolves to the active AccessToken
+
+    Raises:
+        RuntimeError: If no authenticated user (use get_access_token() for optional)
+
+    Example:
+        ```python
+        from fastmcp.server.dependencies import CurrentAccessToken
+        from fastmcp.server.auth import AccessToken
+
+        @mcp.tool()
+        async def get_user_id(token: AccessToken = CurrentAccessToken()) -> str:
+            return token.claims.get("sub", "unknown")
+        ```
+    """
+    return cast(AccessToken, _CurrentAccessToken())
+
+
 # --- Token Claim dependency ---
 
 


docs/integrations/azure.mdx
```diff
@@ -278,3 +278,179 @@ Parameters (`jwt_signing_key` and `client_storage`) work together to ensure toke
...
+from fastmcp.server.auth.providers.azure import AzureProvider
+
+auth_provider = AzureProvider(
+    client_id="your-client-id",
+    client_secret="your-client-secret",
+    tenant_id="your-tenant-id",
+    base_url="http://localhost:8000",
+    required_scopes=["mcp-access"],  # Your API scope
+    # Include Graph scopes for OBO
+    additional_authorize_scopes=[
```
pamelafox, line 389: Is this still necessary, given admin consent? I don't believe I needed to do this in my implementation, since admin consent is given on the Entra app registration that has those scopes.

mvgadagi:
Hello @jlowin 

Thank you for the PR and the nice FastMCP package. I wanted to get your inputs on a similar use case we’re working on.

We had setup  FastMCP in proxy mode and connecting it to 2–3 MCP servers that rely on Azure AD authentication and it worked as POC where it managed for one client scenario, but we are yet to make changes for multiple clients using the FastMCP proxy
To support this, we’ve added a custom implementation for the On-Behalf-Of (OBO) flow, following the guidance here:
https://learn.microsoft.com/en-us/entra/identity-platform/v2-oauth2-on-behalf-of-flow

With this PR, do you think it would be possible to use FastMCP in proxy mode with authentication and then leverage the OBO flow for downstream MCP servers?

Also right now FastMCP support storing the auth session the redis store, do we think we could extend this to even for OBO token

One additional note: we ran into a small issue with refresh tokens when using OBO. Our understanding is that the MSAL library maintains an internal cache and automatically refreshes tokens, but in our case the token was not refreshed as expected.

Thanks a lot!

JayDoubleu:
Great PR! The dependency injection pattern for OBO is really clean.

For anyone looking for a **Managed Identity-only approach** (no client_secret at all), I wrote a guide that uses:
- `RemoteAuthProvider` + `JWTVerifier` instead of `AzureProvider`
- Federated Identity Credentials to link MI to the App Registration
- Custom `OBOTokenService` that uses MI token as client assertion

This works well for Azure Container Apps deployments where you want zero secrets.

Guide: https://jaydoubleu.dev/guides/mcp-azure-oauth2-obo

The approaches are complementary - this PR is great for scenarios where you have a client_secret, while the MI approach avoids secrets entirely.

pamelafox:
I think it'd be fantastic for MCP AzureProvider to support the Federated Identity Credentials, since then developers could have the best of both worlds. Last I looked, the underlying OAuthProvider class would need to be more flexible to allow for use of managed identity, it currently assumes everyone is using a secret-based approach.

I think that could be done in a separate PR, however, as that aspect is orthogonal to the OBO functionality, from my perspective? Or @JayDoubleu do you think it'd require changes to this?

(Disclosure: I'm from Microsoft)

pamelafox:
@mvgadagi Were you seeing a token expiry error? Are you using TokenCache() for the token_cache setting? After how much time did the token refresh issue happen? Was that with a deployed server in production? It'd be good to test this setup doesnt run into that issue. I haven't seen it yet with my similar OBO demo (https://github.com/Azure-Samples/python-mcp-demos/blob/main/servers/auth_entra_mcp.py#L110), but haven't battle tested in production.

JayDoubleu:
>I think that could be done in a separate PR, however, as that aspect is orthogonal to the OBO functionality, from my perspective? Or @JayDoubleu do you think it'd require changes to this?

You're right. AzureProvider can't be used for MI because it requires client_secret for OAuth proxy functionality. Perhaps another PR addressing it would be fine, although there might be extra rework required to make Azure happy. I'm currently rewriting my accelerator against `FastMCP==2.14.4` with this and other workarounds mentioned in #3002. Might contribute it if no one else beats me to it.

mvgadagi:
> @mvgadagi Were you seeing a token expiry error? Are you using TokenCache() for the token_cache setting? After how much time did the token refresh issue happen? Was that with a deployed server in production? It'd be good to test this setup doesnt run into that issue. I haven't seen it yet with my similar OBO demo (https://github.com/Azure-Samples/python-mcp-demos/blob/main/servers/auth_entra_mcp.py#L110), but haven't battle tested in production.

Thank you @pamelafox for the response
/ After exactly an hour we faced the issued (As we have v2.0 token and they have have expiration set to 1 hour after they issued)
/ We had used the token_cache in logic already
/ It's not in production yet, but planning in by end of Q1 after testing the whole flow

And I will go through the reference you given and try to related to our implementation.

Thanks a lot !!

JonasKs:
Hello everyone! I've taken some time to assess the AzureProvider again. 

First of all, great writeup @pamelafox! There are some details I want to address, and I hope it's okay to do it here, as I think I've uncovered a bug or two. 

From your blog:

> However, since we are configuring this server to work with arbitrary MCP clients, we don't have that option

This is not true. This is auth code flow under the hood, and the FastMCP server is responsible for constructing the redirect URI. In other words, we can send something like this:

```python
azure = AzureProvider(
    client_id=settings.MCP_SERVER_CLIENT_ID,
    client_secret=settings.MCP_SERVER_CLIENT_SECRET.get_secret_value(),
    tenant_id=settings.TENANT_ID,
    required_scopes=[f"user_impersonation"],
    additional_authorize_scopes=[
         "openid", 
         "profile", 
         "offline_access", 
         "api://<appID1>/user_impersonation",  # <-- request scope for multiple clients
         "api://<appID2>/user_impersonation"   # <-- these are NOT admin consented
    ],
)
```

and it would correctly prompt us for consent for `appID1` and `appID2`.   
However, this triggers a bug:

image

I haven't really looked into a beautiful fix here, but something like this should work:

```python
        ... code
            # Advertise full scopes including OIDC (even though we only validate non-OIDC)
            valid_scopes=parsed_required_scopes,
        )

        token_exchange_scopes = self._prefix_scopes_for_azure(
            parsed_required_scopes or []
        )
        # Only add OIDC scopes from additional_authorize_scopes, not other API scopes
        if parsed_additional_scopes:
            oidc_scopes = [s for s in parsed_additional_scopes if s in OIDC_SCOPES]
            token_exchange_scopes.extend(oidc_scopes)
        # Deduplicate while preserving order
        token_exchange_scopes = list(dict.fromkeys(token_exchange_scopes))
        self._extra_token_params = {"scope": " ".join(token_exchange_scopes)}
        logger.debug("Token exchange scopes: %s", token_exchange_scopes)
```
[on this line](https://github.com/jlowin/fastmcp/blob/main/src/fastmcp/server/auth/providers/azure.py#L212).

This basically gets the prefixed `required_scope` + OIDC scopes and set them as an `self._extra_token_params`.    
With this fix, the consent flow has been done, and the token fetched is correct.
We can then fetch tokens for multiple APIs, using this token: 
```python
@mcp.tool(name="change_state", description="Use this tool to change the state")
async def change_state(ctx: Context):
    # token = get_access_token() 
    token1 = OnBehalfOfCredential(...).get_token(scope=api://<appID1>/user_impersonation)
    token2 = OnBehalfOfCredential(...).get_token(scope=api://<appID1>/user_impersonation)
    print(f"{token1=}")
    print(f"{token2=}")
    ...
```


As for this PR — I really like this! OBO is a very common use-case for Azure, but I find it really nice for MCPs in general — no reason to double-implement authorization — just use the upstream token.  
**EDIT(Sorry, I clicked comment a bit too quick)**  
But `from azure.identity import OnBehalfOfCredential` exist. The nice thing about azure.identity (I believe opposed to msal) is that it's a higher level library which also deals with token caching, refreshing etc.   
A common pattern is to create the `obo = OnBehalfOfCredential()` on startup, and then only use `obo.get_token(scope=scope)` in your code. If the token is expired, the `get_token()` would automatically refresh it.  
So, I would highly recommend not re-implementing these, but rather re-export them.

docs/integrations/azure.mdx
JonasKs: I answered below, but admin consent isn't a requirement, nor should we make it one.

JonasKs: But — `additional_authorize_scopes` isn't just for `graph scopes`, so we should remove that comment! 😊

pamelafox:
@JonasKs In that code, you have appID1, so wouldn't that be the case of a pre-known client ID? Or is app ID referring to something else? I think I'm a bit confused.

JonasKs:
@pamelafox AppId1 and 2 are other appregs, yes. But they are not admin consented! ☺️

pamelafox:
@JonasKs Ah, yes, for known client apps, the admin consent is not needed. I do not know how often that will be the case, but it's a good idea to make that clear in documentation.

JonasKs:
What do you mean with "known"? Example of an unknown app?

JonasKs:
I've raised a PR for the bug discovered above: https://github.com/jlowin/fastmcp/pull/3013

mvgadagi:
> I've raised a PR for the bug discovered above: #3013

@JonasKs @jlowin I too confirm the issue and tested with fastmcp==3.0.0b1, adding extra scopes fails with:

Token exchange with identity provider failed: invalid_request: AADSTS28003: Provided value for the input parameter scope cannot be empty when requesting an access token using the provided authorization code. Please specify a valid scope.

```
additional_authorize_scopes=[
"api://app-registration-client-id-1/default",
"api://app-registration-client-id-2/default"
]
```

Probably the https://github.com/jlowin/fastmcp/pull/3013 the fix is not released yet

JonasKs:
Correct. Install from the main branch and try again.

jlowin:
Updated this PR to use `azure.identity.aio.OnBehalfOfCredential` instead of raw MSAL, based on @JonasKs's feedback. The `EntraOBOToken` dependency API is unchanged — the internals are just simpler now (async-native, no thread pool hack, automatic token caching/refresh). Dropped `MSALApp` since it was MSAL-specific; for advanced scenarios you can use `CurrentAccessToken()` and construct your own credential directly. Would love any thoughts.

docs/integrations/azure.mdx
pamelafox: True, yes. I think it's a bit confusing since the section just before here describes how to grant admin consent. Perhaps a note here that if you have granted admin consent to the Entra app registration, additional_authorize_scopes may not be necessary?

sandeepkunusoth:
Does OBO flow work on tools created using FastMCP.from_openapi( directly? i am getting invalid_token issue when used with openapi spec tools

PR #2946, opened 2026-01-19
Add PromptsAsTools transform

jlowin:
Some MCP clients only support tools and can't use the prompts protocol directly. This transform bridges that gap by generating tools that expose prompt functionality.

`PromptsAsTools` follows the same pattern as the recently merged `ResourcesAsTools` - it captures a provider reference at construction and generates two tools that clients can call:

```python
from fastmcp import FastMCP
from fastmcp.server.transforms import PromptsAsTools

mcp = FastMCP("Server")

@mcp.prompt
def analyze_code(code: str, language: str = "python") -> str:
    """Analyze code for potential issues."""
    return f"Analyze this {language} code:\n{code}"

mcp.add_transform(PromptsAsTools(mcp))
# Now has list_prompts and get_prompt tools
```

The `list_prompts` tool returns JSON with prompt metadata including argument information. The `get_prompt` tool renders prompts with arguments and returns the result as a messages array. When the provider is a FastMCP server, both tools route through the middleware chain just like direct prompt calls would.

### DIFF
--- a/docs/development/v3-notes/v3-features.mdx
+++ b/docs/development/v3-notes/v3-features.mdx
@@ -108,6 +108,8 @@ Transforms modify components (tools, resources, prompts) as they flow from provi
 - `ToolTransform` - modifies tool schemas (rename, description, tags, argument transforms)
 - `Enabled` - sets enabled state on components by key or tag (backs `enable()`/`disable()` API)
 - `VersionFilter` - filters components by version range (`version_gte`, `version_lt`)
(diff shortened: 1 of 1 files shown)


jlowin:
Addressed the P2 feedback. Changed `_format_prompt_result()` to use `.model_dump(mode="json", exclude_none=True)` for non-text content instead of `str()`, so embedded resources are now properly serialized as JSON-compatible dicts that tool-only clients can parse.

PR #2950, opened 2026-01-20
Rename Enabled transform to Visibility

jlowin:
The `Enabled` class name was awkward—it describes a state rather than the concept being controlled. Renaming to `Visibility` makes the API clearer since the transform controls which components are visible to clients.

```python
from fastmcp.server.transforms import Visibility

# Hide components tagged "internal"
mcp.add_transform(Visibility(False, tags={"internal"}))

# Session-level visibility control
await ctx.reset_visibility()  # was reset_components()
```

The `enable()` and `disable()` methods on servers/contexts remain unchanged—those are action verbs that still make sense.


jlowin:
@coderabbitai review

PR #2962, opened 2026-01-20
Fix FastAPI mounting examples in docs

### LINKED ISSUE #2961: FastMCP integration "quick example" is definitely incorrect
### Description

This one: https://gofastmcp.com/deployment/http#fastapi-integration

For two reasons, described elsewhere in the document:

1. The FastMCP app's lifecycle *must* be passed to the FastAPI app, as mentioned earlier for Starlette (since FastAPI is just Starlette): https://gofastmcp.com/deployment/http#mounting-in-starlette
2. Mounting the FastMCP app at /mcp doesn't put it at http://hostname/mcp but at http://hostname/mcp/mcp ... it actually works to mount it at / to get it at /mcp, but I'm not sure if that's a good idea or not

### Example Code

```Python

```

### Version Information

```Text
(latest version of docs on the website)
```

jlowin:
Fixes #2961

The FastAPI integration quick example had two bugs:

1. **Missing lifespan**: FastAPI (like Starlette) doesn't execute mounted sub-app lifespans automatically. Without passing `lifespan=mcp_app.lifespan`, the session manager won't initialize and requests will fail.

2. **Double path prefix**: `http_app()` defaults to `path="/mcp"`, so mounting at `/mcp` produced `/mcp/mcp` instead of `/mcp`.

The fix uses `http_app(path="/")` when mounting at `/mcp`, and properly passes the lifespan:

```python
mcp_app = mcp.http_app(path="/")
api = FastAPI(lifespan=mcp_app.lifespan)
api.mount("/mcp", mcp_app)
```

Also fixed the same path issue in the FastAPI integration guide's "Lifespan Management" and "Combining Lifespans" examples.

### DIFF
--- a/docs/deployment/http.mdx
+++ b/docs/deployment/http.mdx
@@ -344,13 +344,6 @@ Here's a quick example showing how to add MCP to an existing FastAPI application
 from fastapi import FastAPI
 from fastmcp import FastMCP
 
-# Your existing API
-api = FastAPI()
-
-@api.get("/api/status")
-def status():
-    return {"status": "ok"}
-
 # Create your MCP server
 mcp = FastMCP("API Tools")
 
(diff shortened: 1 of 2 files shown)


lalapeshori (approved):
reviewed

PR #2970, opened 2026-01-21
[Draft] Add enabled field to ToolTransformConfig

### LINKED ISSUE #2969: 3.0 - ToolTransformConfig lost its enabled setting
### Description

in 2.0, ToolTransformConfig had an enabled parameter, likely inherited from the shared base model, that's is gone now. We should re-add it

### Example Code

```Python

```

### Version Information

```Text
3.0.0b
```

strawgate:
ToolTransformConfig was missing the `enabled` field that exists in all similar Meta classes (ToolMeta, ResourceMeta, PromptMeta). This change adds the field back to maintain consistency across the codebase.

The `enabled` field is excluded from being passed to `TransformedTool.from_tool()` since it's not a valid parameter for that method, similar to how `arguments` is excluded.

Fixes #2969

### DIFF
--- a/tests/server/test_tool_transformation.py
+++ b/tests/server/test_tool_transformation.py
@@ -118,3 +118,33 @@ async def test_server_level_transforms_apply_to_mounted_servers():
 
     assert "renamed_sub_tool" in tool_names
     assert "sub_tool" not in tool_names
+
+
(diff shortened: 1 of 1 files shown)


strawgate:
/marvin we also need tests to verify behavior

jlowin:
Closing for #2991

PR #2971, opened 2026-01-21
docs: use --prerelease=explicit for install instructions

zzstoatzz:
Using `--prerelease=allow` when installing fastmcp can pull in prerelease versions of dependencies like httpx. httpx 1.0.dev* removes `TransportError` which breaks httpx-sse, which breaks mcp imports.

`--prerelease=explicit` only allows prereleases for packages explicitly requested with a prerelease specifier (like `fastmcp>=3.0a0`), keeping dependencies on stable versions.

```bash
# before (could install httpx==1.0.dev3)
uv add fastmcp --prerelease=allow

# after (installs httpx==0.28.1 stable)
uv add "fastmcp>=3.0a0" --prerelease=explicit
```

### DIFF
--- a/docs/getting-started/installation.mdx
+++ b/docs/getting-started/installation.mdx
@@ -12,13 +12,13 @@ FastMCP 3.0 is currently in beta. Install by specifying the version explicitly.
 </Note>
 
 ```bash
-pip install fastmcp==3.0.0b1
+uv add "fastmcp>=3.0.0b1" --prerelease=explicit
 ```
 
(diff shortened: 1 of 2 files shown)


jlowin:
`pip install "fastmcp>=3.0.0b1" --pre` is still problematic (gets the bad httpx version) -- but it turns out `pip install 'fastmcp>=3.0.0b1'` is fine! Going to close this and adopt that in #2975. Thanks!

PR #2997, opened 2026-01-23
feat: option to add upstream claims to the FastMCP proxy JWT

### LINKED ISSUE #2868: Upstream claims in FastMCP tokens
### Enhancement

Hi!

I've been involved in the [Azure Integration](https://gofastmcp.com/integrations/azure) and it still works great for us. However, the FastMCP Oauth Proxy JWT token does not preserve upstream claims. I want these available for our audit log, which is implemented in the gateway layer. 

Current token looks like this:

```json
{
  "aud":"my-aud/mcp",
  "client_id":"my-client",
  "exp":1768337534,
  "iat":1768333836,
  "iss":"my-iss",
  "jti":"CHW49fCOYHPKlsODbn8UKEQnShou6kMydeOa6NnulT8",
  "scope":"user_impersonation"
}
```

but I'd like it to contain upstream claims, looking something like this:

```json
{
  "aud":"my-aud/mcp",
  "client_id":"my-client",
  "exp":1768337534,
  "iat":1768333836,
  "iss":"my-iss",
  "jti":"CHW49fCOYHPKlsODbn8UKEQnShou6kMydeOa6NnulT8",
  "scope":"user_impersonation",
  "upstream_claims": {
    "sub": "user-subject-id",
    "oid": "user-object-id",
    "name": "Test User",
    "email": "test@example.com",
    "roles": ["Admin", "Reader"]
  }
}
```

Sometimes code is easier to understand than text, so I made Claude implement it [here](https://github.com/JonasKs/fastmcp/commit/cf3ec254de9d5a11ba367820685fff2349ababa5#diff-9a0303687382bc6bd653a3d0f28dbd3ec9c6bbac63f794d0e304bdeae6bcd3fb). I'd love to clean this up and create a PR if you accept this feature request.

JonasKs:
## Description
Implements `upstream_claims` which each auth provider can override to match that provider's need.  
I have been very conflicted in the implementation details, but ended up with a design where each provider must implement the upstream claims extraction, making it easy to customize which claims are actually valuable for each provider.  
There are a few alternatives to my implementation, which I'm very happy to implement if you like those better:
- always just forward all upstream claims (I'm not that familiar with other auth providers, so I am hesitant)
- boolean to enable/disable upstream claims in the token

Example on a token fetched from FastMCP with the `AzureProvider`:
```json
{
  "aud": "my-aud",
  "client_id": "my-clientid",
  "exp": 1769176333,
  "iat": 1769171340,
  "iss": "my-iss",
  "jti": "7OHhgdtyMR8tavlzqYcA1EmH9-nUODTw00MPFAr7oy0",
  "scope": "user_impersonation",
  "upstream_claims": {
    "name": "Jonas Krüger Svensson / Intility AS",
    "oid": "upstream-oid",
    "preferred_username": "jonas.svensson@intility.no",
    "roles": [
      "Reader"
    ]
  }
}
```

--- 

**Note on AI usage:**
I initially wrote this myself. It's been refactored with Claude Code, and then reviewed by gemini CLI with the `gemini-2.5-pro` model. Claude Code Opus4.5 wrote the tests. 

**Contributors Checklist**
- [x] My change closes #2868
  - There should be none, given that you agree that we enable this by default. If you want a configuration option, please let me know. 

**Review Checklist**

### DIFF
--- a/src/fastmcp/utilities/auth.py
+++ b/src/fastmcp/utilities/auth.py
@@ -25,7 +25,7 @@ def _decode_jwt_part(token: str, part_index: int) -> dict[str, Any]:
         raise ValueError("Invalid JWT format (expected 3 parts)")
 
     part_b64 = parts[part_index]
-    part_b64 += "=" * (4 - len(part_b64) % 4)  # Add padding
+    part_b64 += "=" * (-len(part_b64) % 4)  # Add padding
     return json.loads(base64.urlsafe_b64decode(part_b64))
 
 
--- a/tests/utilities/test_auth.py
+++ b/tests/utilities/test_auth.py
@@ -60,6 +60,20 @@ class TestDecodeJwtHeader:
         with pytest.raises(ValueError, match="Invalid JWT format"):
             decode_jwt_header("a.b.c.d")
 
+    def test_decode_header_length_divisible_by_4(self):
+        """Test decoding when base64 length is divisible by 4 (no padding needed).
+
+        This tests the edge case where len(part) % 4 == 0.
+        The padding calculation (-len % 4) correctly yields 0 in this case.
+        """
+        # Create a header that encodes to exactly 12 chars (divisible by 4)
+        header = {"x": ""}  # eyJ4IjogIiJ9 = 12 chars
+        payload = {"sub": "user"}
+        token = create_jwt(header, payload)
+
+        result = decode_jwt_header(token)
+        assert result == header
+
 
 class TestDecodeJwtPayload:
     """Tests for decode_jwt_payload utility."""
@@ -111,6 +125,20 @@ class TestDecodeJwtPayload:
             token = create_jwt({"alg": "RS256"}, payload)
             result = decode_jwt_payload(token)
             assert result == payload
+
+    def test_decode_payload_length_divisible_by_4(self):
+        """Test decoding when base64 length is divisible by 4 (no padding needed).
+
+        This tests the edge case where len(part) % 4 == 0.
+        The padding calculation (-len % 4) correctly yields 0 in this case.
+        """
+        # Create a payload that encodes to exactly 12 chars (divisible by 4)
+        header = {"alg": "RS256"}
+        payload = {"x": ""}  # eyJ4IjogIiJ9 = 12 chars
+        token = create_jwt(header, payload)
+
+        result = decode_jwt_payload(token)
+        assert result == payload
 
 
 class TestParseScopes:


src/fastmcp/server/auth/oauth_proxy/proxy.py
```diff
@@ -990,6 +995,32 @@ def _prepare_scopes_for_upstream_refresh(self, scopes: list[str]) -> list[str]:
         """
         return scopes
 
+    async def _extract_upstream_claims(
```
JonasKs, line 998: I made this async to support fetching external claims here, but I'm not sure if it's needed.

src/fastmcp/server/auth/providers/azure.py
```diff
@@ -356,3 +358,81 @@ def _prepare_scopes_for_upstream_refresh(self, scopes: list[str]) -> list[str]:
...
+        try:
+            # Azure access tokens are JWTs - decode without verification
+            # (already validated by token_verifier during token exchange)
+            parts = access_token.split(".")
+            if len(parts) != 3:
+                logger.debug("Azure token is not a valid JWT (expected 3 parts)")
+                return None
+
+            # Decode payload (urlsafe_b64decode handles padding)
+            payload = json.loads(urlsafe_b64decode(parts[1].encode()))
```
JonasKs: I couldn't find a nice method in `authlib` to just extract claims..

src/fastmcp/server/auth/oauth_proxy/proxy.py
```diff
@@ -990,6 +995,32 @@ def _prepare_scopes_for_upstream_refresh(self, scopes: list[str]) -> list[str]:
...
+            Dict of claims to embed in JWT under the "upstream_claims" key,
+            or None to not embed any upstream claims.
+
+        Example:
+            For Azure/Entra ID, you might decode the access_token JWT and
+            extract claims like sub, oid, name, preferred_username, email,
+            roles, and groups.
+        """
+        return None
+
```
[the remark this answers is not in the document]
JonasKs: It wasn't addressed in that commit, but I've addressed it in 53d55b99295bae82b303b0ced13de2495954035e

src/fastmcp/server/auth/providers/azure.py
```diff
@@ -6,8 +6,10 @@
 
 from __future__ import annotations
 
+import json
 from typing import TYPE_CHECKING, Any
 
+from authlib.common.encoding import urlsafe_b64decode
```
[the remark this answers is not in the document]
JonasKs: I wish we had error handling in Python like we do in Rust.  I feel like this is shaky (what if another error is introduced, or something else changes?), but I'm happy to implement if you'd like.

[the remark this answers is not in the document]
JonasKs: I suspect you don't want this either, @jlowin ? Pre-commit removes any such change if I try to commit.

Let me know what approach you'd like 😊

JonasKs:
I'll have to investigate the last coderabbitai feedback. I'll be back on Monday. Marking as draft for now

JonasKs:
Ok @jlowin , that last coderabbitai comment put me down a bit of a rabbit hole, but I am pretty certain in my conclusion: 

We do not need to add support for the potential "If the refresh response lacks identity fields..." argument. 

The flow is:
- Extract identity (claims) from `access_token`, not `id_token` on [exchange_authorization_code](https://github.com/jlowin/fastmcp/blob/fde9068efbb0ff1c6cdd026ef0039743eec70f6d/src/fastmcp/server/auth/oauth_proxy/proxy.py#L793)
- When FastMCP refreshes a token, the whole flow requires `access_token` - which is what is returned when refreshed. See [this line](https://github.com/jlowin/fastmcp/blob/fde9068efbb0ff1c6cdd026ef0039743eec70f6d/src/fastmcp/server/auth/oauth_proxy/proxy.py#L880) and [this line for refresh tokens](https://github.com/jlowin/fastmcp/blob/fde9068efbb0ff1c6cdd026ef0039743eec70f6d/src/fastmcp/server/auth/oauth_proxy/proxy.py#L1112).
  - This token would have the same claims as the initial flow

I am pretty impressed with coderabbit raising this as a potential issue though, and I'm happy I investigated. 

I'm ready for review.

src/fastmcp/server/auth/oauth_proxy/proxy.py
JonasKs: It is needed. Some providers (such as GitHub) would need an external lookup, as they don't use OIDC.

src/fastmcp/server/auth/providers/azure.py
jlowin: I wonder if you could use the `get_access_token()` dependency function here? It returns the upstream token when used in a request (though I am not sure if its been set at this point, would need to verify)? See e.g. https://github.com/jlowin/fastmcp/pull/2918/changes#diff-66c6267d895d1f8044ba9e7ea298a23e36213d5101b3222a8bd86b9ce4d96f9eR1048

jlowin:
@JonasKs this looks great! I left a small comment about potentially simplifying your decoding of the access token because I believe we have a helper for it. I'd also appreciate your thoughts on the approach in #2918. To be clear that's for a different outcome (OBO within the *server*) as opposed to forwarding claims to the *client* token. but maybe relevant?

src/fastmcp/server/auth/providers/azure.py
JonasKs: I'll try tomorrow when I'm back on a PC. Thanks!

JonasKs: It's not set, unfortunately. 

This somewhat ugly code also exist in [jwt.py](https://github.com/jlowin/fastmcp/blob/main/src/fastmcp/server/auth/providers/jwt.py#L237-L246), so I added a commit where this is now an util, and we re-use this functionality. 

See my new commit https://github.com/jlowin/fastmcp/pull/2997/changes/021b70d2790585931e24b7e3bb58de6a9ac2e853

JonasKs:
`prek` runs fine locally, and seems to hit a cache. I've just rebased by branch with main and the same issue happens. 

Ready for re-review @jlowin. I'll look into the other PRs now.

jlowin (approved):
Thanks!

PR #3017, opened 2026-01-28
Move tests to a reusable action and enable nightly checks

strawgate:


strawgate:
@coderabbitai no thoughts?

jlowin:
Looks great - my only question is what constitutes passing tests for the main branch. Lets say the 2am test runs and fails, but then someone merges a PR (with all tests passing). Does the "repo-level" test indicator reflect the 2am one or the most recent merge? Maybe a quick way to fix it is to run the 2am ones at 2am and also on merge? /marvin what do you think

PR #3050, opened 2026-01-31
Fix Field() handling in prompts

### LINKED ISSUE #3036: Prompt arguments with Field() defaults receive FieldInfo object instead of default value
### Description

When a prompt parameter uses `Field(default="value", ...)` and the caller omits that argument, the prompt function receives the `FieldInfo` object itself rather than extracting its `.default` value.

**Expected behavior:**
```
Perform 'summary' analysis on data.csv.
```

**Actual behavior:**
```
Perform 'annotation=NoneType required=False default='summary' description='Type of analysis.'' analysis on data.csv.
```

The `FieldInfo` object is being passed directly to the function and gets stringified via `__repr__`.

### Example Code

```Python
import asyncio
from pydantic import Field

mcp = FastMCP()

@mcp.prompt()
def analyze_data(
    data_uri: str = Field(description="The URI of the data."),
    analysis_type: str = Field(default="summary", description="Type of analysis.")
) -> str:
    return f"Perform '{analysis_type}' analysis on {data_uri}."

async def demo():
        # Call prompt WITHOUT providing analysis_type (should use default "summary")
        result = await client.get_prompt("analyze_data", {"data_uri": "data.csv"})
        print(result.messages[0].content.text)
        # Expected: "Perform 'summary' analysis on data.csv."
        # Actual:   "Perform 'annotation=NoneType required=False default='summary' description='Type of analysis.'' analysis on data.csv."

if __name__ == "__main__":
    asyncio.run(demo())
```

### Version Information

```Text
FastMCP version:                                                                2.14.4
MCP version:                                                                    1.25.0
Python version:                                                                3.12.11
Platform:                               Linux-5.15.0-139-generic-x86_64-with-glibc2.31
FastMCP root path: /home/alex/PycharmProjects/anker/.venv/lib/python3.12/site-packages
```

strawgate:
When a prompt parameter used `Field(default="value")` and the caller omitted that argument, the prompt function received the `FieldInfo` object itself rather than extracting its `.default` value. This caused the FieldInfo's `__repr__` to be stringified in the output.

Prompts now correctly extract default values from `Field(default=...)` to match the behavior of tools and resource templates.

**Changes:**
- Modified `FunctionPrompt.render()` to use `type_adapter.validate_python()`
- Added parameter filtering for security (prevents dependency override)
- Added comprehensive tests for prompts and resource templates

Fixes #3036

### DIFF
--- a/src/fastmcp/prompts/function_prompt.py
+++ b/src/fastmcp/prompts/function_prompt.py
@@ -306,7 +306,7 @@ class FunctionPrompt(Prompt):
             kwargs = {k: v for k, v in kwargs.items() if k in valid_params}
 
             # Use type adapter to validate arguments and handle Field() defaults
-            # This matches the behavior of tools in function_tool.py:258,282
+            # This matches the behavior of tools in function_tool
             type_adapter = get_cached_typeadapter(self.fn)
 
             # self.fn is wrapped by without_injected_parameters which handles


AlexanderKazakov:
I've tested this branch on my case and it works properly

PR #3062, opened 2026-02-03
Add helpers for converting FunctionTool and TransformedTool to SamplingTool

### LINKED ISSUE #3021: Add helpers for creating SamplingTool
### Enhancement

Add helpers for converting FunctionTool and Transformed tool into SamplingTool

Maybe also a helper that turns every tool (and one for converting a specific tool) on a connected client into a sampling tool too

That way you can provide existing tools on your server to the sampling API without having to create new ones

strawgate:
## Overview

This PR adds helpers to convert existing server tools (FunctionTool and TransformedTool) into SamplingTool objects, enabling servers to reuse their tools in sampling contexts without duplication.

Previously, if you had a tool defined with `@mcp.tool` and wanted to use it in `ctx.sample()`, you had to recreate the function separately. Now you can pass FunctionTool and TransformedTool instances directly to sampling methods.

## Changes

1. **Added `SamplingTool.from_callable_tool()`** - Converts FunctionTool or TransformedTool to SamplingTool
   - Special handling for TransformedTool to unwrap ToolResult objects
   - Supports name/description overrides

2. **Enhanced `prepare_tools()`** - Now accepts FunctionTool and TransformedTool directly
   - `ctx.sample()` and `ctx.sample_step()` work with these tool types
   - Backward compatible with plain functions and SamplingTool instances

## Example

```python
@mcp.tool
def search(query: str) -> str:
    """Search the web."""
    return do_search(query)

# Use tool directly in sampling
result = await ctx.sample(
    "Research Python frameworks",
    tools=[search]  # FunctionTool works directly!
)
```

Fixes #3021

### DIFF
--- a/src/fastmcp/server/sampling/sampling_tool.py
+++ b/src/fastmcp/server/sampling/sampling_tool.py
@@ -148,36 +148,31 @@ class SamplingTool(FastMCPBaseModel):
                 "Only callable tools can be converted to SamplingTools."
             )
 
-        # For TransformedTool, we need to use .run() and unwrap ToolResult
-        # because .fn might be a forwarding function that returns ToolResult
-        if isinstance(tool, TransformedTool):
+        # Both FunctionTool and TransformedTool need .run() to ensure proper
+        # result processing (serializers, output_schema, wrap-result flags)
+        async def wrapper(**kwargs: Any) -> Any:
+            result = await tool.run(kwargs)
+            # Unwrap ToolResult - extract the actual value
+            if isinstance(result, ToolResult):
+                # If there's structured_content, use that
+                if result.structured_content is not None:
+                    # Check tool's schema - this is the source of truth
+                    if tool.output_schema and tool.output_schema.get(
+                        "x-fastmcp-wrap-result"
+                    ):
+                        # Tool wraps results: {"result": value} -> value
+                        return result.structured_content.get("result")
+                    else:
+                        # No wrapping: use structured_content directly
+                        return result.structured_content
+                # Otherwise, extract from text content
+                if result.content and len(result.content) > 0:
+                    first_content = result.content[0]
+                    if isinstance(first_content, TextContent):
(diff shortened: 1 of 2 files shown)


src/fastmcp/server/sampling/run.py
```diff
@@ -334,20 +334,40 @@ def prepare_messages(
 
 
 def prepare_tools(
-    tools: Sequence[SamplingTool | Callable[..., Any]] | None,
+    tools: Sequence[SamplingTool | Callable[..., Any] | Any] | None,
```
strawgate: Sequence of any seems like a bad idea

strawgate:
/marvin do.we actually need to do inline imports? I'm not convinced there's a circular dependency issue. Also adding any to the type hint is crazy, please address

jlowin:
One suggestion on the ToolResult unwrapping logic (similar to codex):

```python
async def wrapper(**kwargs: Any) -> Any:
    result = await tool.run(kwargs)
    if isinstance(result, ToolResult):
        # Check tool's schema - this is the source of truth
        if tool.output_schema and tool.output_schema.get("x-fastmcp-wrap-result"):
            # Tool wraps results: {"result": value} -> value
            return result.structured_content["result"]
        else:
            # No wrapping: use structured_content directly
            return result.structured_content
    return result
```
This mirrors what the client does in `_parse_call_tool_result` (lines 374-378 in `client/mixins/tools.py`).

^ may need a fallback for no structured content which I think you already ahve

strawgate:
/marvin can you address jeremiah's feedback please

jlowin:
/marvin add a brief note about this to the v3-features document rc1 section (you'll see it if you pull main)

strawgate:
/marvin there's one conflict, what are you talking about. please resolve conflicts

strawgate:
/marvin can you investigate the feedback? should we be using run directly?

PR #3086, opened 2026-02-05
feat: add Static Client Registration (#3085)

### LINKED ISSUE #3085: Support Static OAuth Client Credentials (Bypass Dynamic Client Registration)
## Context

Some MCP servers do **not** support OAuth 2.0 Dynamic Client Registration (DCR). For example:
[https://api.githubcopilot.com/mcp](https://api.githubcopilot.com/mcp)

The MCP specification explicitly does **not** require DCR support:

> MCP clients and authorization servers **SHOULD** support the OAuth 2.0 Dynamic Client Registration Protocol (RFC 7591) to allow MCP clients to obtain OAuth client IDs without user interaction.

reference: https://modelcontextprotocol.io/specification/2025-11-25/basic/authorization#preregistration

So we should be able to manually register the OAuth client and obtain a `client_id` and `client_secret`.

## Summary
This PR adds support for providing a pre-registered `client_id` and `client_secret` to the `OAuth` client provider in `FastMCP`. When provided, these credentials are used to pre-seed the token storage, effectively bypassing the OAuth 2.0 Dynamic Client Registration (DCR) step.

## Motivation
The MCP specification suggests that clients and servers SHOULD support RFC 7591 (Dynamic Client Registration). However, some MCP servers (e.g., GitHub Copilot's MCP server) do not support DCR and require clients to be manually registered or use static credentials.

Currently, users wishing to connect to such servers must manually instantiate `OAuthClientInformationFull` and inject it into the `TokenStorage` before creating the client, which relies on internal implementation details.

This change exposes `client_id` and `client_secret` as first-class arguments in the `OAuth` class, streamlining the connection process for these servers.

### `src/fastmcp/client/auth/oauth.py`

Update `OAuth.__init__` to accept `client_id` and `client_secret`:

```python
class OAuth(OAuthClientProvider):
    def __init__(
        self,
        # ... existing args ...
        client_id: str | None = None,
        client_secret: str | None = None,
    ):
        # ... existing setup ...

        # 1. Create client_metadata as before
        client_metadata = OAuthClientMetadata(
            client_name=client_name,
            redirect_uris=[AnyHttpUrl(redirect_uri)],
            grant_types=["authorization_code", "refresh_token"],
            response_types=["code"],
            scope=scopes_str,
            **(additional_client_metadata or {}),
        )

        # 2. If client_id/secret are provided, pre-seed the storage
        if client_id and client_secret:
            logger.info("Using static OAuth client credentials (bypassing DCR)")
            
            # Combine metadata with credentials to create full client info
            # This ensures validity against strict schema validation
            client_info_data = client_metadata.model_dump()
            client_info_data.update({
                "client_id": client_id,
                "client_secret": client_secret,
            })
            
            static_client_info = OAuthClientInformationFull.model_validate(client_info_data)
            
            # Initialize storage adapter
            self.token_storage_adapter = TokenStorageAdapter(
                async_key_value=token_storage, server_url=mcp_url
            )
            
            # Inject into storage (synchronously run the async method)
            # This runs the storage set operation to "cache" the credentials
            import anyio
            anyio.run(
                self.token_storage_adapter.set_client_info, 
                static_client_info
            )
```

## Example Usage

```python
from fastmcp.client.auth import OAuth

# For a server that doesn't support Dynamic Client Registration
auth = OAuth(
    mcp_url="https://api.githubcopilot.com/mcp",
    client_id="Iv1...",
    client_secret="e9a...",
    client_name="My App"
)
```

## Considerations
- **Sync/Async Execution**: The `OAuth` class initializes in `__init__` (sync). To pre-seed the async storage, `anyio.run` is used. This works for standard CLI/script usage but should be verified for cases where an event loop is already running.
- **Metadata Consistency**: We ensure that the pre-seeded `OAuthClientInformationFull` object includes the correct `redirect_uris` and other metadata generated during initialization, ensuring consistency if the SDK validates this later.

martimfasantos:
## Context

Some MCP servers do **not** support OAuth 2.0 Dynamic Client Registration (DCR).

The MCP specification explicitly does **not** require DCR support:

> MCP clients and authorization servers **SHOULD** support the OAuth 2.0 Dynamic Client Registration Protocol (RFC 7591) to allow MCP clients to obtain OAuth client IDs without user interaction.

reference: https://modelcontextprotocol.io/specification/2025-11-25/basic/authorization#preregistration

So we should be able to manually register the OAuth client and obtain a `client_id` and `client_secret`.

## Summary
This PR adds support for providing a pre-registered `client_id` and `client_secret` to the `OAuth` client provider in `FastMCP`. When provided, these credentials are used to pre-seed the token storage, effectively bypassing the OAuth 2.0 Dynamic Client Registration (DCR) step.

## Motivation
The MCP specification suggests that clients and servers SHOULD support RFC 7591 (Dynamic Client Registration). However, some MCP servers (e.g., GitHub Copilot's MCP server) do not support DCR and require clients to be manually registered or use static credentials.

Currently, users wishing to connect to such servers must manually instantiate `OAuthClientInformationFull` and inject it into the `TokenStorage` before creating the client, which relies on internal implementation details.

This change exposes `client_id` and `client_secret` as first-class arguments in the `OAuth` class, streamlining the connection process for these servers.

### `src/fastmcp/client/auth/oauth.py`

Update `OAuth.__init__` to accept `client_id` and `client_secret`:

```python
class OAuth(OAuthClientProvider):
    def __init__(
        self,
        # ... existing args ...
        client_id: str | None = None,
        client_secret: str | None = None,
    ):
        # ... existing setup ...

        # 1. Create client_metadata as before
        client_metadata = OAuthClientMetadata(
            client_name=client_name,
            redirect_uris=[AnyHttpUrl(redirect_uri)],
            grant_types=["authorization_code", "refresh_token"],
            response_types=["code"],
            scope=scopes_str,
            **(additional_client_metadata or {}),
        )

        # 2. If client_id is provided, create client full information
        if self._client_id:
            # Create the full static client info directly which will avoid DCR
            self._static_client_info = OAuthClientInformationFull(
                client_id=self._client_id,
                client_secret=self._client_secret,
                **(self._additional_client_metadata or {}),
            )

...
```

## Example Usage

```python
from fastmcp.client.auth import OAuth

# For a server that doesn't support Dynamic Client Registration
auth = OAuth(
    mcp_url="https://api.githubcopilot.com/mcp",
    client_id="Iv1...",
    client_secret="e9a...",
    client_name="My App"
)

# Using the FastMCP client
async with Client(
    "https://mcp-server.example.com/mcp",
    auth=OAuth(
        client_id="my-registered-client-id",
        client_secret="my-client-secret",
    ),
) as client:
   ....

```

---

**Contributors Checklist**

- [x] My change closes #3085 

**Review Checklist**

### DIFF
--- a/src/fastmcp/client/auth/oauth.py
+++ b/src/fastmcp/client/auth/oauth.py
@@ -233,10 +233,17 @@ class OAuth(OAuthClientProvider):
             # Create the full static client info directly which will avoid DCR.
             # Spread client_metadata so redirect_uris, grant_types, response_types,
             # scope, etc. are included — servers may validate these fields.
+            metadata = client_metadata.model_dump(exclude_none=True)
+            # Default token_endpoint_auth_method based on whether a secret is
+            # provided, unless the caller already set it via additional_client_metadata.
+            if "token_endpoint_auth_method" not in metadata:
+                metadata["token_endpoint_auth_method"] = (
+                    "client_secret_post" if self._client_secret else "none"
+                )
             self._static_client_info = OAuthClientInformationFull(
                 client_id=self._client_id,
                 client_secret=self._client_secret,
-                **client_metadata.model_dump(exclude_none=True),
+                **metadata,
             )
 
         token_storage = self._token_storage or MemoryStore()
--- a/tests/client/auth/test_oauth_static_client.py
+++ b/tests/client/auth/test_oauth_static_client.py
@@ -5,7 +5,7 @@ from unittest.mock import patch
 import httpx
 import pytest
 from mcp.shared.auth import OAuthClientInformationFull
-from pydantic import AnyHttpUrl
+from pydantic import AnyUrl
 
 from fastmcp.client import Client
 from fastmcp.client.auth import OAuth
@@ -43,6 +43,7 @@ class TestStaticClientInfoConstruction:
         assert info.response_types is not None
         assert "code" in info.response_types
         assert info.scope == "read write"
+        assert info.token_endpoint_auth_method == "client_secret_post"
 
     def test_static_client_info_without_secret(self):
         """Public clients can provide client_id without client_secret."""
@@ -55,6 +56,7 @@ class TestStaticClientInfoConstruction:
         assert info is not None
         assert info.client_id == "public-client"
         assert info.client_secret is None
+        assert info.token_endpoint_auth_method == "none"
         # Metadata should still be present
         assert info.redirect_uris is not None
         assert info.grant_types is not None
@@ -177,6 +179,7 @@ class TestStaticClientE2E:
     async def test_static_client_with_dcr_disabled(self):
         """Static client_id should work when the server has DCR disabled."""
         port = find_available_port()
+        callback_port = find_available_port()
         issuer_url = f"http://127.0.0.1:{port}"
 
         provider = InMemoryOAuthProvider(
@@ -193,24 +196,26 @@ class TestStaticClientE2E:
         def greet(name: str) -> str:
             return f"Hello, {name}!"
 
-        # Pre-register a client directly in the provider
+        # Pre-register a client directly in the provider.
+        # The redirect_uri must match what the OAuth client will use.
         pre_registered = OAuthClientInformationFull(
             client_id="pre-registered-client",
             client_secret="pre-registered-secret",
-            redirect_uris=[AnyHttpUrl("http://localhost:0/callback")],
+            redirect_uris=[AnyUrl(f"http://localhost:{callback_port}/callback")],
             grant_types=["authorization_code", "refresh_token"],
             response_types=["code"],
+            token_endpoint_auth_method="client_secret_post",
             scope="read write",
         )
         await provider.register_client(pre_registered)
 
         async with run_server_async(server, port=port, transport="http") as url:
-            # Use HeadlessOAuth with static credentials
             oauth = HeadlessOAuth(
                 mcp_url=url,
                 client_id="pre-registered-client",
                 client_secret="pre-registered-secret",
                 scopes=["read", "write"],
+                callback_port=callback_port,
             )
 
             async with Client(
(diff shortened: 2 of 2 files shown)


martimfasantos:
> The feature makes sense — the implementation has a few issues though:
> 
> `anyio.run()` in **init** crashes in async contexts, which is the common case. The MCP SDK's own ClientCredentialsOAuthProvider shows the right pattern: store credentials in **init**, then override _initialize() to set context.client_info directly. There can be no async work in the constructor at all.
> 
> `_static_client_info` AttributeError — the _initialize override references it unconditionally but it's only set when credentials are provided, breaking all existing usage.
> 
> Static client info needs the full metadata — just passing client_id/client_secret without redirect_uris, grant_types, etc. will likely fail server-side validation. Since client_metadata is already constructed at that point, just spread it in.
> 
> `client_secret` should be optional — the spec says "client ID (and, if applicable, client credentials)," explicitly accommodating public client pre-registration (no secret, relying on PKCE + redirect URI validation). The current if client_id and client_secret guard silently falls through to DCR when only client_id is provided.
> 
> The fix for all four (roughly):
> 
> ```python
> # in __init__
> self._static_client_info: OAuthClientInformationFull | None = None
> if client_id is not None:
>     self._static_client_info = OAuthClientInformationFull(
>         client_id=client_id,
>         client_secret=client_secret,
>         **client_metadata.model_dump(exclude_none=True),
>     )
> 
> # override _initialize
> async def _initialize(self) -> None:
>     await super()._initialize()
>     if self._static_client_info is not None:
>         self.context.client_info = self._static_client_info
>     if self.context.current_tokens and self.context.current_tokens.expires_in:
>         self.context.update_token_expiry(self.context.current_tokens)
> ```

I’ve addressed all these points @jlowin, thanks for calling them out. 
Should we also add documentation, as was done in #2871?

jlowin (approved):
@martimfasantos thank you! I've pushed a few doc changes to the branch, this LGTM

PR #3121, opened 2026-02-09
fix get_access_token and lifespan_context returning None in background tasks - issue (#3095)

### LINKED ISSUE #3095: get_access_token returns None in background tasks (task=True)
### Enhancement

I'm experiencing an issue where the get_access_token dependency returns None when used in tools marked with task=True, even though authentication is properly configured and
working.

Setup:
- FastMCP server with JWT authentication configured
- Tool functions using task=True for background execution
- Attempting to use get_access_token dependency to access authenticated user info

Problem:
When a tool is marked with task=True, the access_token parameter injected via Depends(get_access_token) is always None, even when the request is properly authenticated. The
same dependency works fine in non-background tools.

Minimal reproducible example:

```python
from fastmcp import FastMCP
from fastmcp.server.auth import JWTVerifier, AccessToken

# Server with JWT auth
jwks_uri = "https://cognito-idp.us-east-1.amazonaws.com/us-east-1_XXXXX/.well-known/jwks.json"
auth = JWTVerifier(jwks_uri=jwks_uri)
mcp = FastMCP("TestServer", auth=auth)

# This works - access_token is populated
@mcp.tool(name="test_sync")
async def test_sync(
  access_token: AccessToken | None = Depends(get_access_token),
) -> str:
  if access_token:
      return f"Token: {access_token.token[:20]}..."
  return "No token!"  # Never reaches here when authenticated

# This doesn't work - access_token is always None
@mcp.tool(name="test_background", task=True)
async def test_background(
  access_token: AccessToken | None = Depends(get_access_token),
) -> str:
  if access_token:
      return f"Token: {access_token.token[:20]}..."
  return "No token!"  # Always returns this, even with valid auth
```

Expected behavior:
Dependencies should be resolved in the request context before the background task is spawned, so access_token should contain the authenticated user's token.

Actual behavior:
access_token is None in background tasks, making it impossible to access authenticated user information in task=True functions.

Questions:
1. Is this a known limitation of how background tasks handle dependency injection?
2. Is there a recommended pattern for accessing authentication context in background tasks?
3. I noticed LifespanContext is always empty in background tasks as well, is this expected?
4. Should dependencies be resolved before spawning the background task, or is the current behavior intentional?

Any guidance would be appreciated! Thanks for the great framework.

cristiangreco94:
## Description
 When a tool is marked with task=True, get_access_token() always returns None and lifespan_context is always empty — even when authentication is properly configured and a lifespan is set. This happens because neither the HTTP request scope nor the SDK's auth context var survive the Docket worker boundary, and request_ctx isn't set in workers.

The fix snapshots the AccessToken into Redis at submit_to_docket() time (alongside existing task metadata), restores it into a ContextVar when the background task Context is created in the Docket worker, and adds it as a fallback in get_access_token(). Expired tokens (past expires_at) return None rather than raising. For lifespan_context, the fix falls back to reading from the FastMCP server instance directly when request_context is unavailable — no Redis needed.

```python
@mcp.tool(name="test_background", task=True)
async def test_background(
    access_token: AccessToken | None = Depends(get_access_token),
) -> str:
    # access_token is now properly populated in background tasks
    if access_token:
        return f"Token: {access_token.token[:20]}..."
    return "No token!"
```

**Contributors Checklist**

- [x] My change closes #3095
- [N/A] I have performed all required documentation updates

**Review Checklist**

---

### DIFF
--- a/tests/server/tasks/test_context_background_task.py
+++ b/tests/server/tasks/test_context_background_task.py
@@ -3,7 +3,9 @@
 import pytest
 
 from fastmcp import FastMCP
+from fastmcp.server.auth import AccessToken
 from fastmcp.server.context import Context
+from fastmcp.server.dependencies import get_access_token
 from fastmcp.server.elicitation import AcceptedElicitation
 from fastmcp.server.tasks.elicitation import elicit_for_task, handle_task_input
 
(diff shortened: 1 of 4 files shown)


tests/server/tasks/test_context_background_task.py
```diff
@@ -860,3 +862,332 @@ async def ask_for_name(ctx: Context) -> str:
 
             # Verify the tool received the elicited value and returned correctly
             assert result.data == "Hello, Bob!"
+
+
+class TestAccessTokenInBackgroundTasks:
```
chrisguidry, line 867: @cristiangreco94 I've got [similar feedback](https://github.com/jlowin/fastmcp/pull/2906#issuecomment-3854664422) about this test suite as I did with @gfortaine's last PR.  He did a really phenomenal job of [de-mocking it](https://github.com/jlowin/fastmcp/pull/2906), can you give this test suite a pass too?  

The general idea here is that we have a fully functional in-memory Docket and redis (via Docket's memory broker implementation using `fakeredis`) so you shouldn't really need to mock anything here.  Just write the test using the FastMCP client and example MCP servers and tools/prompts/resources to demonstrate the end-to-end flow.  You can absolutely use mocks when you want to emulate failure conditions, etc, but these tests as they stand right now create such a closed mock world that they could easily drift from reality without failing.

tests/server/tasks/test_context_background_task.py
```diff
@@ -860,3 +862,332 @@ async def ask_for_name(ctx: Context) -> str:
...
+        mock_server._docket = mock_docket
+        server_token = _current_server.set(MagicMock(return_value=mock_server))
+        docket_token = _current_docket.set(mock_docket)
+
+        mock_session = MagicMock()
+        mock_session._fastmcp_state_prefix = "test-session-id"
+        _task_sessions["test-session-id"] = MagicMock(return_value=mock_session)
+
+        try:
+            task_info = TaskContextInfo(
```
chrisguidry, line 1046: Having a `try` in a unit test is kind of a code smell for me.  If we need to ensure some clean up across tests, let's have a pytest fixture that handles that instead so that the test can just be "flat" and more obvious.

gfortaine:
@cristiangreco94 @chrisguidry — i have a consolidated branch (`feat/distributed-task-enhancements`) that integrates the access token snapshot + lifespan fallback from this PR with the notification queue and BLPOP elicitation work from #2906.

the test suite is fully end-to-end: in-memory docket + fakeredis, no mock world, no try blocks in tests — addresses the review feedback here. also picked up a non-monotonic progress delta bug along the way.

4 commits, rebased on upstream/main. PR is up at #2906. happy to adjust if @chrisguidry or @jlowin prefer a different structure.

chrisguidry:
@gfortaine Thanks but I actually _just_ merged #2906 :D  

@cristiangreco94 if you can rebase this one on main, I can review it soon after

cristiangreco94:
I think [this PR](https://github.com/jlowin/fastmcp/pull/3138) from @gfortaine supersedes mine, so i will proceed and close it. if not, let me know and I will reopen it and work on it.

PR #3133, opened 2026-02-10
Fix unhandled exceptions in OpenAPI POST tool calls

### LINKED ISSUE #3113: FastMCP.from_openapi() tool call fails on some models like claude when tool call with POST parameters in prompt
### Description

When using VS Code Copilot Chat with an MCP server generated from an OpenAPI spec, Copilot fails with:

Error parsing JSON stream data for request id ... Internal Server Error

Image

Image

### Example Code

```Python
import httpx
from fastmcp import FastMCP

# Using OpenAPI 3.0 spec instead of Swagger 2.0
# The v3 spec is compatible with FastMCP's parser
PETSTORE_SPEC_URL = "https://petstore3.swagger.io/api/v3/openapi.json"
PETSTORE_API_BASE = "https://petstore3.swagger.io/api/v3"

def build_server() -> FastMCP:
    try:
        # Fetch the OpenAPI spec once at startup
        print(f"📥 Fetching OpenAPI spec from {PETSTORE_SPEC_URL}...")
        response = httpx.get(PETSTORE_SPEC_URL, timeout=20)
        response.raise_for_status()
        spec = response.json()
        print(f"✅ Successfully fetched OpenAPI spec (version: {spec.get('openapi', 'unknown')})")

        # Client used to call the actual API
        client = httpx.AsyncClient(base_url=PETSTORE_API_BASE, timeout=30)

        # Generate MCP server directly from OpenAPI
        print("🔧 Generating MCP server from OpenAPI spec...")
        mcp = FastMCP.from_openapi(
            openapi_spec=spec,
            client=client,
            name="Petstore MCP"
        )
        print(f"✅ Successfully generated MCP server with tools")

    except httpx.HTTPError as e:
        print(f"❌ Failed to fetch OpenAPI spec: {e}")
        raise
    except Exception as e:
        print(f"❌ Failed to build server: {e}")
        raise

    return mcp

mcp = build_server()

if __name__ == "__main__":
    # Streamable HTTP MCP endpoint will be at http://localhost:8000/mcp
    mcp.run(transport="http", host="127.0.0.1", port=8000)
```

### Version Information

```Text
FastMCP version:                                                           2.14.5
MCP version:                                                               1.26.0
Python version:                                                            3.10.9
Platform:                                              macOS-26.1-arm64-arm-64bit
FastMCP root path: /Users/sandeepkunusoth/miniconda3/lib/python3.10/site-packages
```

jlowin:
`OpenAPITool.run()` only catches `httpx.*` exceptions, so any failure during request building (parameter routing, body construction) escapes unhandled and surfaces to clients as a generic "Internal Server Error" with no diagnostic info. This particularly affects POST requests where the body construction code exercises more fragile paths.

Two changes: harden `_unflatten_arguments()` to guard against empty `content_schema` dicts and non-dict body schemas, and add a catch-all in `run()` that wraps unexpected errors with the tool's method, path, and original error — so users see something like `"Error building request for POST /items: KeyError: missing_param"` instead of a silent 500.

```python
mcp = FastMCP.from_openapi(openapi_spec=spec, client=client)
# POST tool calls that previously failed with "Internal Server Error"
# now produce actionable error messages
```

Closes #3113

### DIFF
--- a/docs/python-sdk/fastmcp-server-providers-openapi-components.mdx
+++ b/docs/python-sdk/fastmcp-server-providers-openapi-components.mdx
@@ -27,7 +27,7 @@ run(self, arguments: dict[str, Any]) -> ToolResult
 Execute the HTTP request using RequestDirector.
 
 
-### `OpenAPIResource` <sup><a href="https://github.com/jlowin/fastmcp/blob/main/src/fastmcp/server/providers/openapi/components.py#L228" target="_blank"><Icon icon="github" style="width: 14px; height: 14px;" /></a></sup>
+### `OpenAPIResource` <sup><a href="https://github.com/jlowin/fastmcp/blob/main/src/fastmcp/server/providers/openapi/components.py#L227" target="_blank"><Icon icon="github" style="width: 14px; height: 14px;" /></a></sup>
 
 
 Resource implementation for OpenAPI endpoints.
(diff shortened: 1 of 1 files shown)


jlowin:
@coderabbitai review

sandeepkunusoth:
hi @jlowin how can i test this change with 2.14. version can this be released in 2.14.6?

jlowin:
This went out in 3.0 rc1

PR #3134, opened 2026-02-10
Add validate_output option for OpenAPI tools

### LINKED ISSUE #2472: FastMCP Output Validation Bug: Incorrectly Expects Array for Object Response
### Description

# FastMCP Output Validation Bug: Incorrectly Expects Array for Object Response

## Version Information
- **FastMCP Version:** 2.13.1
- **Python Version:** 3.11

## The Problem

I'm using FastMCP to create an MCP server from an OpenAPI spec, and I'm running into a validation error when calling certain endpoints. FastMCP seems to incorrectly assume that all API responses must have a `data` array field, but some of my endpoints return object properties instead (like `user`, `account`, etc.).

## What's Happening

When I call the `get_current_user` tool, I get this error:

```
Output validation error: {} is not of type 'array'
```

But the API is actually returning the correct response! The problem is in FastMCP's validation logic, not the API.

## The Endpoint That's Broken

**Operation ID:** `get_current_user`  
**Path:** `/api/v1/account/me/`  
**Method:** `GET`

## OpenAPI Schema

The OpenAPI schema correctly defines this endpoint as returning an object with a `user` property:

```json
{
  "type": "object",
  "properties": {
    "success": {
      "type": "boolean",
      "example": true
    },
    "user": {
      "type": "object",
      "properties": {
        "id": {
          "type": "integer",
          "example": 1
        },
        "username": {
          "type": "string",
          "example": "john"
        },
        "email": {
          "type": "string",
          "example": "john@example.com"
        },
        "first_name": {
          "type": "string",
          "example": "John"
        },
        "last_name": {
          "type": "string",
          "example": "Doe"
        },
        "is_active": {
          "type": "boolean",
          "example": true
        },
        "date_joined": {
          "type": "string",
          "format": "date-time",
          "example": "2023-01-01T00:00:00Z"
        },
        "profile": {
          "type": "object",
          "properties": {
            "preferences": {
              "type": "array",
              "items": {
                "type": "string"
              }
            },
            "settings": {
              "type": "object"
            }
          }
        }
      }
    }
  }
}
```

## What the API Actually Returns

The API correctly returns this format (I've verified this with logging):

```json
{
  "success": true,
  "user": {
    "id": 1,
    "username": "user@example.com",
    "email": "user@example.com",
    "first_name": "John",
    "last_name": "Doe",
    "is_active": true,
    "date_joined": "2023-01-01T00:00:00Z",
    "profile": {
      "preferences": ["setting1", "setting2"],
      "settings": {
        "theme": "dark"
      }
    }
  }
}
```

The response matches the schema perfectly, but FastMCP still throws a validation error.

## Why This Is Confusing

I have another endpoint that works fine - `accounts_list`. The difference is:

**✅ `accounts_list` works** - returns `{"success": true, "data": [...], "total": 10}`  
**❌ `get_current_user` fails** - returns `{"success": true, "user": {...}}`

The working one has a `data` array, the broken one has a `user` object. FastMCP seems to only accept responses with a `data` array field, even though the OpenAPI schema clearly defines `user` as an object.

## What I Think Is Wrong

It looks like FastMCP's validation logic is:
1. Assuming all responses must have a `data` array field
2. Not properly reading the OpenAPI schema to see what the actual response structure should be
3. Failing validation when responses have object properties instead of array properties

## What Should Happen

FastMCP should:
- Read the OpenAPI schema correctly for each endpoint
- Validate responses based on what the schema actually says
- Accept object properties (like `user`) when the schema defines them as objects
- Only expect arrays when the schema explicitly says `type: "array"`

## How to Reproduce

1. Create a FastMCP server from an OpenAPI spec
2. Have an endpoint that returns `{"success": true, "user": {...}}` or any other object property (not `data`)
3. Call the tool that maps to that endpoint
4. See the error: `Output validation error: {} is not of type 'array'`

## Additional Notes

- The API response is correct and matches the OpenAPI schema exactly
- The error happens in FastMCP's validation, not in the actual API call
- I've added logging that shows the API returns the correct response before FastMCP's validation fails
- This will affect any endpoint that returns object properties instead of array properties

Thanks for looking into this! Let me know if you need any more details.

### Example Code

```Python

```

### Version Information

```Text
fastmcp==2.13.1
```

jlowin:
When backends don't conform to their own OpenAPI response schemas, the MCP SDK's output validation rejects the response and the tool fails with "Output validation error." Users have no way to opt out — they're stuck if the backend is wrong.

This adds `validate_output` (default `True`) to `OpenAPIProvider` and `FastMCP.from_openapi()`. When set to `False`, extracted output schemas are replaced with a permissive `{"type": "object", "additionalProperties": true}` that accepts any response. Structured JSON still flows to clients normally; only the strict schema checking is disabled. The `x-fastmcp-wrap-result` flag is preserved so non-object responses (arrays, primitives) continue to be wrapped correctly.

```python
mcp = FastMCP.from_openapi(
    openapi_spec=spec,
    client=client,
    validate_output=False,  # accept any response shape
)
```

For per-tool control, `mcp_component_fn` can be used to selectively disable validation on specific endpoints.

Closes #2472

### DIFF
--- a/docs/python-sdk/fastmcp-server-providers-openapi-components.mdx
+++ b/docs/python-sdk/fastmcp-server-providers-openapi-components.mdx
@@ -27,7 +27,7 @@ run(self, arguments: dict[str, Any]) -> ToolResult
 Execute the HTTP request using RequestDirector.
 
 
-### `OpenAPIResource` <sup><a href="https://github.com/jlowin/fastmcp/blob/main/src/fastmcp/server/providers/openapi/components.py#L227" target="_blank"><Icon icon="github" style="width: 14px; height: 14px;" /></a></sup>
+### `OpenAPIResource` <sup><a href="https://github.com/jlowin/fastmcp/blob/main/src/fastmcp/server/providers/openapi/components.py#L233" target="_blank"><Icon icon="github" style="width: 14px; height: 14px;" /></a></sup>
 
 
 Resource implementation for OpenAPI endpoints.
@@ -35,7 +35,7 @@ Resource implementation for OpenAPI endpoints.
 
 **Methods:**
 
-#### `read` <sup><a href="https://github.com/jlowin/fastmcp/blob/main/src/fastmcp/server/providers/openapi/components.py#L257" target="_blank"><Icon icon="github" style="width: 14px; height: 14px;" /></a></sup>
+#### `read` <sup><a href="https://github.com/jlowin/fastmcp/blob/main/src/fastmcp/server/providers/openapi/components.py#L263" target="_blank"><Icon icon="github" style="width: 14px; height: 14px;" /></a></sup>
 
 ```python
 read(self) -> ResourceResult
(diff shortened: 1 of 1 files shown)


jlowin:
@coderabbitai review

PR #3136, opened 2026-02-10
Relay task elicitation through standard MCP protocol

chrisguidry:
Stacked on #2906.

When a background task calls `ctx.elicit()`, the notification subscriber detects
the `input_required` notification and sends a standard `elicitation/create`
request to the client via `session.elicit()`. The client's `elicitation_handler`
fires, and the relay pushes the response to Redis for the blocked worker.

This means clients can respond to background task elicitation using the same
`elicitation_handler` they'd use for any other elicitation — no need to interact
with Redis or call `handle_task_input()` directly.

```python
mcp = FastMCP("Demo")

@mcp.tool(task=True)
async def plan_dinner(ctx: Context) -> str:
    result = await ctx.elicit(
        "What kind of dinner?",
        response_type=DinnerPrefs,
    )
    if isinstance(result, AcceptedElicitation):
        return f"Tonight: {result.data.cuisine} dinner!"
    return "Cancelled"

# Client just uses the standard elicitation_handler
async with Client(mcp, elicitation_handler=my_handler) as client:
    task = await client.call_tool("plan_dinner", {}, task=True)
    result = await task.result()
```

Also fixes `self.session` → `self._session` in `_elicit_for_task()` (the
`.session` property raises in external workers where there's no request
context), and corrects the `session` type hint in `elicit_for_task()` to
`ServerSession | None`.

Fixes the `_meta` related-task key from `modelcontextprotocol.io/related-task`
to `io.modelcontextprotocol/related-task` to match the reverse-DNS style in
the [current spec](https://modelcontextprotocol.io/specification/2025-11-25/basic/utilities/tasks).

### DIFF
--- a/docs/python-sdk/fastmcp-server-tasks-elicitation.mdx
+++ b/docs/python-sdk/fastmcp-server-tasks-elicitation.mdx
@@ -50,7 +50,29 @@ in a Docket worker context where there's no active MCP request.
 - `McpError`: If the elicitation request fails
 
 
-### `handle_task_input` <sup><a href="https://github.com/jlowin/fastmcp/blob/main/src/fastmcp/server/tasks/elicitation.py#L234" target="_blank"><Icon icon="github" style="width: 14px; height: 14px;" /></a></sup>
+### `relay_elicitation` <sup><a href="https://github.com/jlowin/fastmcp/blob/main/src/fastmcp/server/tasks/elicitation.py#L234" target="_blank"><Icon icon="github" style="width: 14px; height: 14px;" /></a></sup>
+
+```python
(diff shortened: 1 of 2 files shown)


chrisguidry:
@gfortaine I was doing some testing here and was able to connect the full loop!  How does this look to you?  Does it match what you were thinking?

gfortaine:
yes — this is the piece that was missing. having the client use the same `elicitation_handler` for foreground and background tasks is the right call. it means the distributed complexity stays invisible to tool authors.

the `self._session` fix and the reverse-dns key correction are solid catches too.

PR #3148, opened 2026-02-11
Remove deprecated FastMCP() constructor kwargs

jlowin:
FastMCP v1 conflated server identity, transport config, and global settings into a single `FastMCP()` constructor — v2 deprecated the transport params but kept them for backwards compat, resulting in a 34-parameter `__init__` with a `_deprecated_settings` indirection layer bridging old kwargs to transport methods.

This removes all 16 deprecated kwargs and replaces them with a `**kwargs` catch that raises `TypeError` with specific migration instructions. Users passing `host=` get told to use `run_http_async(host=...)` or `FASTMCP_HOST`; users passing `include_tags=` get told to use `server.enable(tags=..., only=True)`. Environment variables continue to work — transport methods now read directly from `fastmcp.settings` instead of the deleted per-instance `_deprecated_settings`.

```python
# Before: 34-param constructor with deprecated indirection
mcp = FastMCP("server", host="0.0.0.0", port=8080, include_tags={"public"})

# After: clean constructor, settings where they belong
mcp = FastMCP("server")
mcp.enable(tags={"public"}, only=True)
mcp.run_http(host="0.0.0.0", port=8080)

# Helpful error if you forget:
# TypeError: FastMCP() no longer accepts `host`. Pass `host` to `run_http_async()`, or set FASTMCP_HOST.
```

Also deletes `ExperimentalSettings` (dead code) and fixes `mcp_config.py` which was passing `include_tags`/`exclude_tags` through `create_proxy()` to the now-strict constructor.

### DIFF
--- a/docs/development/v3-notes/v3-features.mdx
+++ b/docs/development/v3-notes/v3-features.mdx
@@ -70,7 +70,7 @@ Background tasks now use a distributed Redis notification queue for reliable del
 
 Sixteen deprecated keyword arguments have been removed from `FastMCP.__init__`. Passing any of them now raises `TypeError` with a migration hint. Environment variables (e.g., `FASTMCP_HOST`) continue to work — only the constructor kwargs moved.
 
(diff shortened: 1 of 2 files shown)


jlowin:
@coderabbitai review

PR #3152, opened 2026-02-11
Support async auth checks

### LINKED ISSUE #3150: run_auth_checks should support async checkers
### Enhancement

https://github.com/jlowin/fastmcp/blob/263e0bf6e0262149269863e09a99b8de1f08bbf0/src/fastmcp/server/auth/authorization.py#L160-L163

Currently, tool auth check functions can only be synchronous based on the above. And with FastMCP 3, retrieving state from the tool context is a asynchronous call (`Context.get_state`). This makes it impossible to write a custom auth check using state from the tool context without some workaround like starting asyncio in another thread.

My use case specifically is that I store a per-user API client that caches user info into the tool context. The user's permission list is retrieved and cached by this API client as we are facing difficulties migrating the permissions into the auth token. Hence, I need access to the user's API client within the auth check, requiring `Context.get_state` which is unfortunately asynchronous, hence this problem.

On a side note, trying to use a async check function fails "open" silently since a coroutine object is truthy. At the least, the above function should error if its passed an async function to prevent that.

jlowin:
`run_auth_checks` was sync-only, which meant auth check functions couldn't use async operations like `Context.get_state` or call external services. Worse, passing an async function silently "passed" because the unawaited coroutine object is truthy — a subtle bug.

`run_auth_checks` is now `async` and uses `inspect.isawaitable()` to handle both sync and async checkers transparently. Sync and async checks can be freely mixed in a list.

```python
from fastmcp.server.auth import AuthContext

async def check_user_permissions(ctx: AuthContext) -> bool:
    if ctx.token is None:
        return False
    permissions = await fetch_permissions(ctx.token.claims["sub"])
    return "admin" in permissions

@mcp.tool(auth=check_user_permissions)
def admin_tool() -> str: ...
```

Also eliminates the `AuthCheckCallable` type alias — there's now just `AuthCheck`, defined once in `authorization.py` and used everywhere.

Closes #3150

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -96,7 +96,7 @@ When modifying MCP functionality, changes typically need to be applied across al
 
 - Uses Mintlify framework
 - Files must be in docs.json to be included
-- Never modify `docs/python-sdk/**` (auto-generated)
+- Do not manually modify `docs/python-sdk/**` — a bot automatically updates these files via commits added to PRs
 - **Core Principle:** A feature doesn't exist unless it is documented!
 
 ### Documentation Guidelines


jlowin:
@coderabbitai review

PR #3154, opened 2026-02-11
feat: Search transforms for tool discovery

jlowin:
**This may hold for 3.1.**

When a server exposes hundreds of tools, sending the full catalog to an LLM wastes tokens and hurts selection accuracy. Search transforms solve this by replacing `list_tools()` with a search interface — the LLM discovers tools on demand instead of seeing everything upfront. 

Two strategies, both zero-dependency: `RegexSearchTransform` for pattern matching and `BM25SearchTransform` for natural-language relevance ranking. Adding one collapses the entire catalog into two synthetic tools:

```python
from fastmcp import FastMCP
from fastmcp.server.transforms.search import RegexSearchTransform

mcp = FastMCP("Server")

@mcp.tool
def search_database(query: str) -> str: ...

@mcp.tool  
def send_email(to: str, subject: str, body: str) -> str: ...

# Clients now see only search_tools + call_tool
mcp.add_transform(RegexSearchTransform())
```

Search results respect the full auth pipeline — middleware, visibility transforms, session-level `disable_components`, and component auth checks all filter what's discoverable. The search tool queries `list_tools()` through the complete pipeline at search time using a contextvar bypass that only skips its own hiding behavior.

### DIFF
--- a/docs/servers/transforms/tool-search.mdx
+++ b/docs/servers/transforms/tool-search.mdx
@@ -10,28 +10,28 @@ import { VersionBadge } from '/snippets/version-badge.mdx'
 
 <VersionBadge version="3.0.0" />
 
-When a server exposes hundreds or thousands of tools, sending the full catalog to an LLM wastes tokens and degrades tool selection accuracy. Search transforms solve this by replacing the tool listing with a search interface, letting the LLM discover tools on demand.
+When a server exposes hundreds or thousands of tools, sending the full catalog to an LLM wastes tokens and degrades tool selection accuracy. Search transforms solve this by replacing the tool listing with a search interface — the LLM discovers tools on demand instead of receiving everything upfront.
+
+## How It Works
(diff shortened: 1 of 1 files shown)


whatevertogo:
Sorry, wrong link earlier — please ignore 😅
We’ve been waiting for this feature for a long time.

jlowin:
The review flagged removing `GetToolNext`, `GetResourceNext`, `GetResourceTemplateNext`, and `GetPromptNext` from `__all__` as a breaking change. That removal is intentional and correct — these are Protocol types for the transform system's internal `call_next` plumbing, not part of the public surface. Authors subclassing `Transform` don't need to import them explicitly; the base class method signatures carry the types already. If someone does want to import them directly, they're still importable from `fastmcp.server.transforms` — they're just no longer advertised as public API via `__all__`.

PR #3172, opened 2026-02-12
Fix stale request context in StatefulProxyClient handlers

### LINKED ISSUE #3169: ctx.elicit() not forwarded in FastMCPProxy + StatefulProxyClient, causing stateful tools (ctx.set_state) to fail behind gateway/proxy
### Description

**Version** : 
FastMCP 3.0.0b2

**Context:**
We use a FastMCP gateway/proxy in front of backend MCP servers as a security layer: it enforces tool allowlists, applies tool-description sanitization (prompt-injection hardening), and performs request/response auditing/redaction before forwarding calls. Within this setup, direct backend stateful elicitation with set_state works, but the same flow behind the proxy fails.

**Impact**: 
`ctx.set_state` appears broken behind gateway because execution never reaches the successful post-elicitation path; direct backend calls work, but proxied elicitation tools can hang and return timeout/cancel outcomes.

Could you please help us to understand the possible way to make StatefulProxyClient working elicitations flow along with MCP Servers managing the state using the ctx.set_state?

Thanks a lot in advance !!

### Example Code

```Python
**FastMCP as  Proxy**

stateful_client = StatefulProxyClient(target)
proxy = FastMCPProxy(client_factory=stateful_client.new_stateful, name=f"{entry.name}-proxy")

**FastMCP Server**

cached = await ctx.get_state("favorites")
result = await ctx.elicit(message="What is your favorite color?", response_type=FavoriteColorInput)
await ctx.set_state("favorites", {"color": color, "sport": sport})
```

### Version Information

```Text
FastMCP version:                                                                                    3.0.0b2
MCP version:                                                                                         1.25.0
Python version:                                                                                      3.12.9
Platform:                                                                        macOS-26.3-arm64-arm-64bit
FastMCP root path: /Users/mgadagi/workspace/.venv/lib/python3.12/site-packages
```

jlowin:
`StatefulProxyClient` reuses MCP sessions across requests so that stateful backends (like Playwright) maintain their state. But the receive-loop task that handles server-initiated messages (elicitation, sampling, roots, etc.) inherits a frozen `request_ctx` ContextVar snapshot from when the session was first created. On the second request, handlers like `default_proxy_elicitation_handler` call `get_context()` and resolve a stale `request_id` — the streamable HTTP transport routes the response to a closed stream, and the call hangs forever.

The fix wraps the default proxy handlers in `StatefulProxyClient` only (not `ProxyClient`, which creates fresh sessions per request) with a shim that restores the `request_ctx` ContextVar before forwarding. `ProxyTool.run()` stashes the current `RequestContext` before each backend call, and the shim detects staleness (same session, different request_id) to avoid corrupting concurrent sessions that share the ref via `copy.copy`.

```python
stateful_client = StatefulProxyClient(backend)
proxy = FastMCPProxy(client_factory=stateful_client.new_stateful)

# Second call now works — elicitation routes correctly
async with Client(proxy_url, elicitation_handler=handler) as client:
    result1 = await client.call_tool("ask_name", {})  # fine
    result2 = await client.call_tool("ask_name", {})  # previously hung forever
```

Note: concurrent tool calls within the same stateful session can race on the shared context ref. This is a pre-existing limitation of the stateful session model (concurrent mutations to stateful backends like Playwright are already semantically broken), not something introduced by this fix.

Closes #3169

### DIFF
--- a/docs/development/upgrade-guide.mdx
+++ b/docs/development/upgrade-guide.mdx
@@ -91,6 +91,20 @@ await ctx.set_state("key", "value")
 value = await ctx.get_state("key")
 ```
 
+#### State Values Must Be Serializable
+
+Session state values must now be JSON-serializable by default (dicts, lists, strings, numbers, etc.), since state is persisted across requests using a pluggable storage backend.
+
+If you need to store non-serializable values (e.g., passing an HTTP client from middleware to a tool), use `serializable=False`. These values are request-scoped and only available during the current tool call, resource read, or prompt render:
+
+```python
+# Middleware sets up a client for the current request
+await ctx.set_state("client", my_http_client, serializable=False)
+
+# Tool retrieves it in the same request
+client = await ctx.get_state("client")
+```
+
 #### Server Banner Environment Variable Renamed
 
 `FASTMCP_SHOW_CLI_BANNER` is now `FASTMCP_SHOW_SERVER_BANNER`.
--- a/docs/python-sdk/fastmcp-server-context.mdx
+++ b/docs/python-sdk/fastmcp-server-context.mdx
@@ -77,6 +77,9 @@ async def my_tool(x: int, ctx: Context) -> str:
     await ctx.set_state("key", "value")
     value = await ctx.get_state("key")
 
+    # Store non-serializable values for the current request only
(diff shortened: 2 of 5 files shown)


jlowin:
@coderabbitai review

PR #3185, opened 2026-02-13
Drop diskcache dependency (CVE-2025-69872)

### LINKED ISSUE #3166: Consider making diskcache optional or using safer serialization (CVE-2025-69872)
## Summary

FastMCP depends on `py-key-value-aio[disk]` which brings in `diskcache 5.6.3`. This version has a known vulnerability (CVE-2025-69872 / GHSA-w8v5-vhqr-4h9v) related to unsafe pickle deserialization.

## Vulnerability Details

- **CVE:** CVE-2025-69872
- **Package:** diskcache <= 5.6.3
- **Issue:** DiskCache uses Python's pickle for serialization, which allows arbitrary code execution if an attacker can write to the cache directory
- **Fix available:** No - diskcache hasn't released a fix (last release Aug 2023)

## Dependency Chain

```
fastmcp -> py-key-value-aio[disk] -> diskcache
```

## Request

Would you consider one of the following:

1. **Make `[disk]` optional** - Allow users to install fastmcp without disk caching if they don't need it
2. **Switch to safer serialization** - Use JSON or another safe serialization method instead of pickle
3. **Pin to a fixed version** - Once diskcache releases a fix

## Impact

This vulnerability is flagged by security scanners (pip-audit, safety, etc.) and may block deployments in security-conscious environments, even though the actual risk is low (requires write access to cache directories).

## Environment

- fastmcp version: 2.14.5
- diskcache version: 5.6.3 (transitive)

Thank you for maintaining this excellent library!

### LINKED ISSUE #3192: Bump py-key-value to 0.4.1
### Enhancement

Bump to 0.4.1 for filetreestore improvements

jlowin:
FastMCP pulls in `diskcache` transitively via `py-key-value-aio[disk]`. diskcache uses pickle serialization, which has a known vulnerability (CVE-2025-69872) — and since we only ever store JSON strings, pickle is unnecessary overhead. The dependency alone is enough to trigger pip-audit and similar scanners, blocking deployments in security-conscious environments.

This switches the default OAuth proxy storage from `DiskStore` (diskcache/SQLite/pickle) to an encrypted `FileTreeStore` (file-per-key JSON), and changes the `py-key-value-aio` extras from `[disk,keyring,memory]` to `[filetree,keyring,memory]`. diskcache is no longer in the dependency tree at all.

The storage directory (`~/.../fastmcp/oauth-proxy`) now contains plain JSON files instead of a SQLite database. Existing users will need to re-authenticate after upgrading — stored OAuth tokens from the old format won't carry over.

Closes #3166
Closes #3192

### DIFF
--- a/docs/python-sdk/fastmcp-server-auth-oauth_proxy-proxy.mdx
+++ b/docs/python-sdk/fastmcp-server-auth-oauth_proxy-proxy.mdx
@@ -26,7 +26,7 @@ production use with enterprise identity providers.
 
 ## Classes
 
-### `OAuthProxy` <sup><a href="https://github.com/jlowin/fastmcp/blob/main/src/fastmcp/server/auth/oauth_proxy/proxy.py#L119" target="_blank"><Icon icon="github" style="width: 14px; height: 14px;" /></a></sup>
+### `OAuthProxy` <sup><a href="https://github.com/jlowin/fastmcp/blob/main/src/fastmcp/server/auth/oauth_proxy/proxy.py#L118" target="_blank"><Icon icon="github" style="width: 14px; height: 14px;" /></a></sup>
 
 
 OAuth provider that presents a DCR-compliant interface while proxying to non-DCR IDPs.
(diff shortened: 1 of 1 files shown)


strawgate:
https://github.com/jlowin/fastmcp/issues/3192

We should probably document how to keep using diskstore if you need to

PR #3212, opened 2026-02-18
Cache OBO credentials on AzureProvider for token reuse

jlowin:
Every `EntraOBOToken` dependency was creating a fresh `OnBehalfOfCredential` and closing it after each tool call. The Azure SDK caches tokens *inside* the credential instance, but since we were throwing it away immediately, every tool invocation made a full OBO round-trip to Azure's token endpoint.

`AzureProvider` now holds an LRU cache of credentials keyed by user assertion hash, so repeated tool calls for the same user reuse the same credential and hit the SDK's internal cache. Token refreshes produce new assertions, which naturally get new credentials.

```python
@mcp.tool()
async def get_emails(
    graph_token: str = EntraOBOToken(["https://graph.microsoft.com/Mail.Read"])
):
    # Second call with same user token reuses the cached credential
    # and skips the OBO exchange entirely
    ...
```

### DIFF
--- a/docs/python-sdk/fastmcp-server-auth-providers-azure.mdx
+++ b/docs/python-sdk/fastmcp-server-auth-providers-azure.mdx
@@ -14,7 +14,7 @@ using the OAuth Proxy pattern for non-DCR OAuth flows.
 
 ## Functions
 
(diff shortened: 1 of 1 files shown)


jlowin:
@codex review

PR #3233, opened 2026-02-19
Fix middleware state not propagating to mounted sub-server tools

### LINKED ISSUE #3230: Mounted mcsp servers failing to receive middleware in v3.0
### Description

I think this is a breaking change between v2.x and v3.0. If I apply middleware to the main mcp instance and mount a sub mcp, then:

* in v2.x the middleware is applied
* in v3.0 the middleware is not applied

### Example Code

```Python
from fastmcp import FastMCP
from fastmcp import Context
from fastmcp.server.middleware import Middleware, MiddlewareContext


mcp = FastMCP("MyProject")
sub_mcp = FastMCP("MySubproject")
# mcp.mount(sub_mcp, prefix="mounted__")
mcp.mount(sub_mcp, namespace="mounted__")

class CustomContextMiddleware(Middleware):

    async def on_call_tool(self, context: MiddlewareContext, call_next):
        # context.fastmcp_context.set_state("test_ctx_val", "wOrLd")
        await context.fastmcp_context.set_state("test_ctx_val", "wOrLd")

mcp.add_middleware(CustomContextMiddleware())


@sub_mcp.tool()
async def hello_context(ctx: Context) -> str:
    # name = ctx.get_state("test_ctx_val")
    name = await ctx.get_state("test_ctx_val")
    return f"Hello, {name}!"

if __name__ == "__main__":
    mcp.run()

# In v2.x: Hello, wOrLd!
# In v3.0: Hello, None!
```

### Version Information

```Text
FastMCP version:                                                           3.0.0
Python version:                                                           3.13.7
Platform:                            Linux-5.14.0-1059-oem-x86_64-with-glibc2.31
FastMCP root path: /home/james/PycharmProjects/fastmcp-testing/.venv/lib/python…
```

dudujuju828:
## Description

Attempt to fix middleware state not propagating to mounted sub-server tools, as reported in #3230.

When middleware on a parent `FastMCP` instance sets state via `context.fastmcp_context.set_state()`, tools on mounted sub-servers cannot read that state — `ctx.get_state()` returns `None`. This appears to be caused by each `FastMCP` instance having its own independent `_state_store`. This PR chains state stores via a `_parent_state_store` reference set during `mount()`, and falls back to the parent store in `get_state()` when a key isn't found locally.

**Contributors Checklist**

- [x] My change closes #3230

**Review Checklist**

### DIFF
--- a/src/fastmcp/server/context.py
+++ b/src/fastmcp/server/context.py
@@ -1207,7 +1207,8 @@ class Context:
         """Get a value from the state store.
 
         Checks request-scoped state first (set with ``serializable=False``),
-        then falls back to the session-scoped state store.
+        then falls back to the session-scoped state store, then to the
+        parent state store (for mounted servers inheriting parent middleware state).
 
         Returns None if the key is not found.
         """
@@ -1215,7 +1216,14 @@ class Context:
         if prefixed_key in self._request_state:
             return self._request_state[prefixed_key]
         result = await self.fastmcp._state_store.get(key=prefixed_key)
-        return result.value if result is not None else None
+        if result is not None:
+            return result.value
+        # Fall back to parent state store (for mounted servers)
+        parent_store = self.fastmcp._parent_state_store
+        if parent_store is not None:
+            result = await parent_store.get(key=prefixed_key)
+            return result.value if result is not None else None
+        return None
 
     async def delete_state(self, key: str) -> None:
         """Delete a value from the state store.
--- a/src/fastmcp/server/server.py
+++ b/src/fastmcp/server/server.py
@@ -261,6 +261,8 @@ class FastMCP(
             pydantic_model=StateValue,
             default_collection="fastmcp_state",
         )
+        # Parent state store for mounted servers (set by mount())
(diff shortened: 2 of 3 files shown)


jlowin:
Thanks for the PR! There is a lot of complexity here that makes me hesitate. In very simple examples this has a nice feel — nested server automatically asks the parent for state on a lookup miss. But it also creates potential for error or confusion. For example, clearing state in a child server may or may not mutate a parent, which could be surprising. The Codex reviews also flag that this only chains one level deep and that remounting the same server under a different parent silently keeps the first binding.

For now I think opt-in is the right way to share state without surprises — you can pass the same store to both servers:

```python

store = MemoryStore()
mcp = FastMCP("Parent", session_state_store=store)
sub = FastMCP("Child", session_state_store=store)
mcp.mount(sub, namespace="child")
```

If this becomes a persistent user expectation we'd revisit — but to accept an automatic approach we'd need it to handle the full chain (grandchild sees root state), work correctly when a server is mounted under multiple parents, and have clear semantics for writes and deletes (shadow vs mutate).

Going to close this, but appreciate you taking a look at the problem!

PR #3248, opened 2026-02-20
Add verify_id_token option to OIDCProxy

### LINKED ISSUE #3240: `OIDCProxy`: Support id token verification for providers with opaque access tokens
## Summary

`OIDCProxy` currently verifies upstream tokens by passing the `access_token` to a `TokenVerifier`. For OIDC providers that issue opaque (non-JWT) access tokens, this means the default `JWTVerifier` can't be used, and the `id_token` — which is always a standard JWT verifiable via the provider's JWKS — goes unused.

It would be great if OIDCProxy could verify the `id_token` instead, since it already has everything needed: the JWKS URI and issuer from OIDC discovery, and the `id_token` stored in `raw_token_data`.

## Current Behavior

Token verification happens at proxy.py:1408:

```
validated = await self._token_validator.verify_token(
    upstream_token_set.access_token
)
```

Only the `access_token` is passed to the verifier. When it's opaque, the default `JWTVerifier` fails.

## What I tried

1. Custom `/userinfo` verifier — Works, but adds a network round-trip on every request:

```
class UserInfoVerifier(TokenVerifier):
        response = await client.get(userinfo_url, headers={"Authorization": f"Bearer {token}"})
        claims = response.json()
        return AccessToken(token=token, client_id=claims["sub"], scopes=[], claims=claims)

OIDCProxy(config_url=..., token_verifier=UserInfoVerifier(...))
```

2. `IntrospectionTokenVerifier` — RFC 7662 support varies across providers. Some only support introspection for specific token types (e.g., id tokens only, not access tokens), making it unusable for validating opaque access tokens. As a standalone `auth=`, it also doesn't provide the full OAuth flow that MCP clients need. And like `/userinfo`, it's still a per-request network call.

## Proposed Solution

A parameter on `OIDCProxy` to opt into id token verification:

```
OIDCProxy(
    config_url="https://provider.example/.well-known/openid-configuration",
    client_id="...",
    client_secret="...",
    base_url="https://my-server.example",
    verify_id_token=True,  # <-- new
)
```

When enabled:
1. In `load_access_token`, verify `upstream_token_set.raw_token_data["id_token"]` instead of `upstream_token_set.access_token`
2. Use the JWKS URI and issuer already available from OIDC discovery (oidc_proxy.py:419-425)
3. Populate the `AccessToken` with claims from the id token

This would be a lightweight change — no protocol changes, no new classes, just reusing the existing JWTVerifier on a different token from the same upstream response. The OIDC spec (OpenID Connect Core 1.0 §2) defines the id token as the primary authentication assertion, and `OIDCProxy` already stores it — it's just not wired up for verification today.

jlowin:
`OIDCProxy` always verifies the upstream `access_token`, but many OIDC providers (Clerk, Zoho, others) issue opaque access tokens that the default `JWTVerifier` can't validate. The existing workaround — passing a custom `token_verifier` that hits `/userinfo` or the introspection endpoint — works but adds a network round-trip on every request. The `id_token` is already sitting in `raw_token_data`, is always a standard JWT per the OIDC spec, and can be verified locally using the JWKS URI that `OIDCProxy` already discovers.

This adds a `verify_id_token` flag that tells the proxy to verify the `id_token` instead of the `access_token`:

```python
auth = OIDCProxy(
    config_url="https://provider.example/.well-known/openid-configuration",
    client_id="...",
    client_secret="...",
    base_url="https://my-server.example",
    verify_id_token=True,
)
```

The implementation adds a `_get_verification_token` hook to `OAuthProxy` so subclasses can control which token gets passed to the verifier without duplicating the token swap flow. Three important edge cases are handled:

- **Audience semantics:** In OIDC, `id_token.aud` is the OAuth `client_id`, not the API audience. When `verify_id_token=True`, the default `JWTVerifier` uses `client_id` as its audience while the `audience` parameter still flows upstream for API access.
- **Token refresh:** `raw_token_data` is now merged (not replaced) on refresh, preserving the `id_token` when providers omit it from refresh responses.
- **AccessToken identity:** After id_token verification, the returned `AccessToken` carries the upstream access_token and its scopes — the id_token is used purely as a verification gate, not as the token of record.

Closes #3240

### DIFF
--- a/docs/python-sdk/fastmcp-server-auth-oidc_proxy.mdx
+++ b/docs/python-sdk/fastmcp-server-auth-oidc_proxy.mdx
@@ -52,7 +52,7 @@ that is OIDC compliant.
 
 **Methods:**
 
-#### `get_oidc_configuration` <sup><a href="https://github.com/PrefectHQ/fastmcp/blob/main/src/fastmcp/server/auth/oidc_proxy.py#L425" target="_blank"><Icon icon="github" style="width: 14px; height: 14px;" /></a></sup>
+#### `get_oidc_configuration` <sup><a href="https://github.com/PrefectHQ/fastmcp/blob/main/src/fastmcp/server/auth/oidc_proxy.py#L432" target="_blank"><Icon icon="github" style="width: 14px; height: 14px;" /></a></sup>
 
 ```python
 get_oidc_configuration(self, config_url: AnyHttpUrl, strict: bool | None, timeout_seconds: int | None) -> OIDCConfiguration
@@ -66,7 +66,7 @@ Gets the OIDC configuration for the specified configuration URL.
 - `timeout_seconds`: HTTP request timeout in seconds
 
 
-#### `get_token_verifier` <sup><a href="https://github.com/PrefectHQ/fastmcp/blob/main/src/fastmcp/server/auth/oidc_proxy.py#L442" target="_blank"><Icon icon="github" style="width: 14px; height: 14px;" /></a></sup>
+#### `get_token_verifier` <sup><a href="https://github.com/PrefectHQ/fastmcp/blob/main/src/fastmcp/server/auth/oidc_proxy.py#L449" target="_blank"><Icon icon="github" style="width: 14px; height: 14px;" /></a></sup>
 
 ```python
 get_token_verifier(self) -> TokenVerifier


jlowin:
**Codex review triage**

Addressed in commits:
- **Refresh without id_token** — fixed via `raw_token_data` merge (old id_token preserved when refresh omits it)
- **Audience semantics** — `client_id` used as verifier audience per OIDC Core §2
- **Scope enforcement on verifier** — verifier gets no required_scopes (id_tokens don't carry scope claims); scopes still advertised/enforced at the FastMCP token level
- **Derived scope state** — `_default_scope_str`, `valid_scopes`, and CIMD `default_scope` all recomputed after restoring `required_scopes`

Dismissed:
- **Expired id_token after refresh** — If the provider doesn't return a fresh id_token on refresh and the old one expires, failing auth is *correct*. The user's identity is no longer verified. Skipping `exp` validation would be a security hole, not a fix. This is a known OIDC tradeoff.
- **Require openid scope** — Silently injecting scopes is surprising. If `verify_id_token=True` and the provider doesn't yield an id_token, `_get_verification_token` returns `None` with a clear warning on first auth attempt. That's the right signal for misconfiguration.
- **Claims divergence between id_token and access_token** — This is the intended design. The id_token provides identity claims (sub, email, etc.), the access_token is opaque and used for upstream API calls. There's no divergence, there's complementary data.

PR #3261, opened 2026-02-21
Exclude authorization header from get_http_headers()

### LINKED ISSUE #3260: OpenAPI tool forwards MCP transport auth headers to downstream APIs
## Summary

When an MCP server is created with `FastMCP.from_openapi()` and served over Streamable HTTP, the `get_http_headers()` call in `OpenAPITool.run()` forwards the incoming MCP transport's `Authorization` header to the downstream API, **overwriting** the API key configured on the httpx client.

This means the downstream API receives the MCP client's auth token instead of the intended API key, causing auth failures (503/401/403).

## Root cause

In `fastmcp/server/providers/openapi/components.py` lines 173-175:

```python
mcp_headers = get_http_headers()
if mcp_headers:
    request.headers.update(mcp_headers)
```

`get_http_headers()` extracts headers from the current MCP HTTP request context and forwards them. The `authorization` header is **not** in the `exclude_headers` set in `dependencies.py`, and `.update()` overwrites any existing headers — including the `Authorization` header that was already set from the httpx client's configured headers.

## Minimal reproduction

```python
# server.py
import json, os, httpx
from fastmcp import FastMCP

# Any OpenAPI spec will do
spec = json.loads(open("openapi.json").read())

backend = FastMCP.from_openapi(
    openapi_spec=spec,
    client=httpx.AsyncClient(
        headers={"Authorization": f"Bearer {os.environ['MY_API_KEY']}"},
    ),
    name="My API",
)

backend.run(transport="streamable-http")
```

When a client connects over Streamable HTTP with its own auth header (e.g. Claude Code, or any MCP client authenticating with the server), that auth header gets forwarded to `api.example.com`, replacing `MY_API_KEY`.

This does **not** happen when:
- The server runs over stdio (no HTTP request context, `get_http_headers()` returns `{}`)
- You call `backend.call_tool()` directly in-process without an MCP HTTP context

## Suggested fix

Add `"authorization"` to the `exclude_headers` set in `get_http_headers()`:

```python
exclude_headers = {
    "host",
    "content-length",
    "content-type",
    "connection",
    "transfer-encoding",
    "upgrade",
    "te",
    "keep-alive",
    "expect",
    "accept",
    "authorization",  # <-- add this
    # ...
}
```

Or alternatively, don't forward MCP transport headers to OpenAPI downstream requests by default.

## Environment

- fastmcp 3.0.0
- httpx 0.28.1
- Python 3.13

jlowin:
When an MCP server is served over Streamable HTTP, `get_http_headers()` was forwarding the MCP transport's `Authorization` header to downstream APIs, overwriting API keys configured on the httpx client. This adds `"authorization"` to the exclude set so transport credentials don't leak to backends.

Closes #3260

### DIFF
--- a/docs/python-sdk/fastmcp-server-dependencies.mdx
+++ b/docs/python-sdk/fastmcp-server-dependencies.mdx
@@ -169,7 +169,7 @@ By default, strips problematic headers like `content-length` that cause issues
 if forwarded to downstream clients. If `include_all` is True, all headers are returned.
 
 
-### `get_access_token` <sup><a href="https://github.com/PrefectHQ/fastmcp/blob/main/src/fastmcp/server/dependencies.py#L483" target="_blank"><Icon icon="github" style="width: 14px; height: 14px;" /></a></sup>
+### `get_access_token` <sup><a href="https://github.com/PrefectHQ/fastmcp/blob/main/src/fastmcp/server/dependencies.py#L487" target="_blank"><Icon icon="github" style="width: 14px; height: 14px;" /></a></sup>
 
 ```python
 get_access_token() -> AccessToken | None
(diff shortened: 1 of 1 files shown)


jlowin:
Close for #3262 which is more complete

PR #3262, opened 2026-02-21
fix: prevent MCP transport auth header from leaking to downstream OpenAPI APIs (#3260)

### LINKED ISSUE #3260: OpenAPI tool forwards MCP transport auth headers to downstream APIs

stakeswky:
## Problem

When an MCP server is created with `FastMCP.from_openapi()` and served over Streamable HTTP, the `get_http_headers()` call in `OpenAPITool.run()` forwards the incoming MCP transport's `Authorization` header to the downstream API, **overwriting** the API key configured on the httpx client.

This means the downstream API receives the MCP client's auth token instead of the intended API key, causing auth failures (503/401/403).

This is a **security issue** — MCP transport credentials should never leak to downstream third-party APIs.

## Root Cause

Two problems in `OpenAPITool.run()` (`components.py`):

1. `get_http_headers()` does not exclude `authorization` from forwarded headers
2. `request.headers.update(mcp_headers)` overwrites existing headers, while the client headers block above uses `if key not in request.headers` (non-overwriting) — inconsistent behavior

```python
# Client headers: non-overwriting ✓
if self._client.headers:
    for key, value in self._client.headers.items():
        if key not in request.headers:
            request.headers[key] = value

# MCP headers: overwrites everything ✗
mcp_headers = get_http_headers()
if mcp_headers:
    request.headers.update(mcp_headers)  # <-- overwrites Authorization!
```

## Fix

1. Add `authorization` to `exclude_headers` in `get_http_headers()` (`dependencies.py`) — MCP transport credentials should not be forwarded to downstream APIs by default
2. Change MCP header forwarding to use the same non-overwriting pattern as client headers — only set if key not already present

```python
# After fix: consistent non-overwriting behavior
mcp_headers = get_http_headers()
if mcp_headers:
    for key, value in mcp_headers.items():
        if key not in request.headers:
            request.headers[key] = value
```

## Impact

- **Security**: Prevents MCP client credentials from leaking to third-party APIs
- **Correctness**: httpx client's configured API key is now respected
- **Consistency**: Both client headers and MCP headers use the same non-overwriting pattern

Fixes #3260

### DIFF
--- a/src/fastmcp/server/dependencies.py
+++ b/src/fastmcp/server/dependencies.py
@@ -1048,7 +1048,7 @@ class _CurrentHeaders(Dependency):  # type: ignore[misc]
     """Async context manager for HTTP Headers dependency."""
 
     async def __aenter__(self) -> dict[str, str]:
-        return get_http_headers()
+        return get_http_headers(include={"authorization"})
 
     async def __aexit__(self, *args: object) -> None:
         pass
@@ -1057,9 +1057,10 @@ class _CurrentHeaders(Dependency):  # type: ignore[misc]
 def CurrentHeaders() -> dict[str, str]:
     """Get the current HTTP request headers.
 
-    This dependency provides access to the HTTP headers for the current request.
-    Returns an empty dictionary when no HTTP request is available, making it
-    safe to use in code that might run over any transport.
+    This dependency provides access to the HTTP headers for the current request,
+    including the authorization header. Returns an empty dictionary when no HTTP
+    request is available, making it safe to use in code that might run over any
+    transport.
 
     Returns:
         A dependency that resolves to a dictionary of header name -> value


jlowin:
@stakeswky thank you! making some tweaks to avoid breaking proxy forwarding

PR #3264, opened 2026-02-21
docs: add context-aware tool factory example

### LINKED ISSUE #1841: Example for Context-Aware Tool Factories
### Enhancement

This section on [Context Aware Tool Factories](https://gofastmcp.com/patterns/tool-transformation#context-aware-tool-factories) mentions how we can generate specialised versions of a tool for different contexts. It would be better if there were any examples to explain the same better.

machov:
Fixes #1841

Adds a concrete code example for the Context-Aware Tool Factories pattern in the tool transformation docs. The example shows how to create a factory function that generates a user-specific tool by hiding the `user_id` parameter via `Tool.from_tool()` with `ArgTransform(hide=True, default=...)`.

Updated both v2 and v3 docs.


jlowin (approved):

jlowin:
Thank you!

PR #3268, opened 2026-02-21
feat: authorization server url in Client

martimfasantos:
### Support pre-configured `authorization_server_url` in `OAuth`

Today, the `OAuth` client always sends an unauthenticated request first, receives a `401`, and only then triggers the full discovery chain (PRM → OASM → registration → token exchange) — even when the authorization server is already known.

> [!NOTE]
> In most scenarios where the client provides a `client_id` and/or `client_secret`, the authorization server is already known. These credentials are typically issued by — and bound to — a specific identity provider.

In enterprise and other controlled environments — and especially in cases where a `client_id` and/or `client_secret` is already configured — the authorization server URL is known upfront (via configuration files, environment variables, or deployment manifests). Since these credentials are issued by a specific identity provider, they implicitly determine the authorization server. In such scenarios, the initial `401` round-trip introduces unnecessary latency and overhead, without adding any functional value.

### Solution

This adds an optional `authorization_server_url` parameter to `OAuth` that lets clients go straight to OASM discovery and token acquisition before the first real request, eliminating the unnecessary 401.

```python
from fastmcp import Client
from fastmcp.client.auth import OAuth

auth = OAuth(
    client_id="my-client",
    client_secret="my-secret",
    authorization_server_url="https://auth.example.com",  # skip 401 + PRM
)

async with Client("https://mcp-server.example.com/mcp", auth=auth) as client:
    # First request is already authenticated — no 401 round-trip
    await client.ping()
```

### How it works

The MCP SDK's `OAuthClientProvider.async_auth_flow` only triggers OASM discovery and token acquisition reactively — inside the `if response.status_code == 401:` branch of the httpx auth generator. There's no hook to pre-populate the context before that branch runs.

Rather than monkeypatching the parent class, this PR adds a `_proactive_auth()` method on FastMCP's `OAuth` subclass. It runs *before* delegating to the parent's `async_auth_flow` generator:

1. Sets `context.auth_server_url` directly from the provided URL
2. Fetches OASM via the standard `build_oauth_authorization_server_metadata_discovery_urls` + `handle_auth_metadata_response` helpers from the MCP SDK — the same code paths used during normal 401-triggered discovery
3. Registers the client (DCR, CIMD, or static credentials — all three paths are supported)
4. Completes the authorization code grant (browser + callback + PKCE)
5. Exchanges for tokens via `_exchange_token_authorization_code` / `_handle_token_response`

When `async_auth_flow` then delegates to the parent, it finds valid tokens already in `context.current_tokens`, adds the `Authorization: Bearer` header, and yields the request — the server responds with 200 on the first try.

**Without** `authorization_server_url` (current behavior, unchanged):
```
POST /mcp → 401 → GET PRM → GET OASM → POST /register → GET /authorize → POST /token → POST /mcp (200)
```

**With** `authorization_server_url`:
```
GET OASM → POST /register → GET /authorize → POST /token → POST /mcp (200)
```

### Protocol compliance

This is fully compliant with the MCP spec's OAuth 2.1 authorization flow. The feature only changes *when* discovery happens (proactively vs. reactively) — it still fetches OASM from `/.well-known/oauth-authorization-server` per RFC 8414, still performs DCR per RFC 7591 when needed, and still uses Authorization Code + PKCE per OAuth 2.1. The parameter is purely opt-in; omitting it preserves the exact same 401-triggered behavior.

---

**Contributors Checklist**

- [x] My change closes #(issue number)

**Review Checklist**

---

### DIFF
--- a/docs/clients/auth/oauth.mdx
+++ b/docs/clients/auth/oauth.mdx
@@ -61,6 +61,7 @@ You don't need to pass `mcp_url` when using `OAuth` with `Client(auth=...)` — 
 - **`token_storage`** (`AsyncKeyValue`, optional): Storage backend for persisting OAuth tokens. Defaults to in-memory storage (tokens lost on restart). See [Token Storage](#token-storage) for encrypted storage options
 - **`additional_client_metadata`** (`dict[str, Any]`, optional): Extra metadata for client registration
 - **`callback_port`** (`int`, optional): Fixed port for OAuth callback server. If not specified, uses a random available port
+- **`authorization_server_url`** (`str`, optional): URL of the OAuth authorization server. When provided, skips the initial 401 round-trip and protected resource metadata discovery, going straight to fetching authorization server metadata from this URL. See [Pre-Configured Authorization Server](#pre-configured-authorization-server)
 - **`httpx_client_factory`** (`McpHttpClientFactory`, optional): Factory for creating httpx clients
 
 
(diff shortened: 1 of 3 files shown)


jlowin:
There is a lot of complexity here and despite the claim, this is not a spec-compliant change. The spec is clear that PRM discovery is mandatory, not optional

> MCP clients MUST use OAuth 2.0 Protected Resource Metadata for authorization server discovery.

And:

> MCP clients MUST support both discovery mechanisms and use the resource metadata URL from the parsed WWW-Authenticate headers when present; otherwise, they MUST fall back to constructing and requesting the well-known URIs.

If it were possible to inject this into an existing workflow as a simple keyword argument it might be worth the tradeoff for development workflows, but introducing a new auth pattern is too much complexity; we already don't like that we've had to reimplement the SDK just for retry hooks. If the spec relaxes this requirement and the SDK provides an easier approach we can revisit.

Issue #3292, opened 2026-02-25
FastMCP 3.0.2 eagerly imports heavy unused deps (~4.3s overhead for simple stdio servers)


theflysurfer:
### Description

**Summary**
`from fastmcp import FastMCP` in v3.0.2 takes **2.5–5.5 seconds** even for simple stdio tool servers that use no auth, no Redis, no task scheduling, and no runtime type checking. The overhead comes from eager imports of heavy transitive dependencies that are never needed at runtime.

This is different from #2205 (which focused on `mcp.types`). The MCP SDK has since fixed their lazy-loading (#1508). **The remaining bottleneck is FastMCP's own dependency chain.**

**Environment**
- FastMCP 3.0.2 / MCP SDK 1.26.0
- Python 3.13.9, Windows 11
- MetaMcp: HTTP multiplexer proxying 19 MCP backends via `create_proxy()`

**Impact**
We proxy 19 backends via `create_proxy()`. Each stdio backend spawns a Python process that imports FastMCP. Total cold-start cost across all backends: **100+ seconds**. This delays tool availability for clients after a restart.

### Profiling Data

Profiled with `python -X importtime` on 3 real-world servers:

| Backend | Total import | FastMCP share | Own code |
|---------|-------------|--------------|----------|
| word-mcp (115 tools, COM automation) | 2.6s | 2.4s (92%) | 0.2s |
| claude-code-admin (124 tools) | 5.0s | 4.3s (86%) | 0.7s |
| google-workspace (138 tools, Google APIs) | 9.3s | 5.6s (60%) | 1.2s |

### Breakdown: What FastMCP imports eagerly but never uses

| Module | Time | Loaded via | Used by these servers? |
|--------|------|-----------|----------------------|
| `beartype` (runtime type checking) | ~520ms | `docket` → dependency injection | No |
| `docket` + `key_value.aio` + `redis` | ~550ms | `fastmcp.server.dependencies` | No (no Redis, no task queue) |
| `authlib` + `cryptography` (OAuth/JWT) | ~1,100ms | `fastmcp.server.auth.oauth_proxy` | No (stdio servers, no OAuth) |
| `mcp.client.session` | ~1,400ms | FastMCP internal imports | No (these are servers, not clients) |
| `httpx` | ~800ms | Auth proxy, client transports | No (stdio transport only) |

**Total avoidable overhead: ~4.3 seconds** for a simple `FastMCP("my-server")` + `@mcp.tool()` server.

[...]

PR #3301, opened 2026-02-25
Replace vendored DI with uncalled-for

### LINKED ISSUE #3251: Support docket's `Shared()` dependencies in FastMCP tools
## Description

FastMCP already supports docket's `Depends()` in tool functions — `_resolve_fastmcp_dependencies` correctly initializes the `_Depends.cache` and `_Depends.stack` context vars so that per-request dependency injection works out of the box.

However, docket's `Shared()` dependencies do not work in FastMCP tools. Attempting to use a `Shared()` dependency raises:

```
LookupError: <ContextVar name='shared_resolved' at 0x...>
```

## Root Cause

`Shared()` relies on `SharedContext`, which requires three context vars to be initialized before resolution:
- `SharedContext.resolved` — a `dict` caching already-resolved shared dependencies
- `SharedContext.lock` — an `asyncio.Lock` for safe concurrent resolution
- `SharedContext.stack` — an `AsyncExitStack` for managing shared dependency lifetimes

These context vars are only set up inside docket's worker runtime (`SharedContext.__init__` expects docket and worker arguments). Since FastMCP doesn't run inside a docket worker, the context vars are never initialized.

## Suggested Fix

Initialize the `SharedContext` context vars during the FastMCP server lifespan (not per-request like `Depends`). Conceptually, `Shared` means "resolve once, reuse across all invocations" — which maps naturally to the server's lifetime.
1. At server startup, initialize the three context vars (resolved → {}, lock → asyncio.Lock(), stack → AsyncExitStack())
2. Enter the AsyncExitStack so shared dependency lifetimes are tied to the server
3. On server shutdown, exit the stack to clean up shared resources

## Minimal Reproduction

```
from contextlib import asynccontextmanager
from docket import Depends, Shared
from fastmcp import FastMCP

@asynccontextmanager
async def create_pool():
    pool = await SomePool.create()
    try:
        yield pool
    finally:
        await pool.close()

@asynccontextmanager
async def get_connection(pool=Shared(create_pool)):
    async with pool.connection() as conn:
        yield conn

@mcp.tool()
async def my_tool(query: str, conn=Depends(get_connection)) -> list:
    return await conn.fetch(query)
```

Calling `my_tool` raises `LookupError` on `shared_resolved`. Changing `Shared` to `Depends` works but creates a new pool per call, defeating the purpose.

## Motivation

`Shared` is the natural choice for expensive-to-create, reusable resources (connection pools, HTTPcclients, etc.) in tool dependencies. Since FastMCP already supports `Depends`, users may reasonably expect `Shared` to work too — especially given both libraries share the same DI model.

Happy to contribute a PR if there's alignment on the approach! Let me know what you think @chrisguidry

### LINKED ISSUE #3320: Linter does not reckognize `Depends` anymore
### Description

I am trying to use `Depends` to define a custom dependency, I am using the following import:
```
```

However, I get the following error from my linter `basedpyright`:
```
"Depends" is unknown import symbolbasedpyright[reportAttributeAccessIssue](https://docs.basedpyright.com/v1.32.1/configuration/config-files/#reportAttributeAccessIssue)
```
Image

Is there anything I missed ? How can I solve this issue ?

### Example Code

```Python
```

### Version Information

```Text
FastMCP version:                                                                                                    3.0.2
MCP version:                                                                                                       1.26.0
Python version:                                                                                                   3.12.12
Platform:                                                                                      macOS-26.2-arm64-arm-64bit
FastMCP root path: /Users/henri.chabert/dev/mistral/dashboard/eolienne-private/mcp_hub/.venv/lib/python3.12/site-packages
```

chrisguidry:
FastMCP vendored a minimal DI engine extracted from Docket (~164 lines) with try/except fallback patterns everywhere. The [`uncalled-for`](https://pypi.org/project/uncalled-for/) package is a clean, typed extraction of this same system, and since Docket will also depend on it (chrisguidry/docket#353), `uncalled_for.Dependency` becomes the single canonical base class.

This deletes the `_vendor/docket_di/` directory, replaces all the try/except import patterns with direct `uncalled_for` imports, and updates the `Dependency.execution` → `current_execution` ContextVar references to match the Docket branch. The `Progress` class now delegates to an internal impl and returns `self` from `__aenter__` (matching Docket's pattern) so that ty's generic resolution works without `type: ignore` suppressions.

```python
# before (repeated everywhere)
try:
    from docket.dependencies import Dependency
except ImportError:
    from fastmcp._vendor.docket_di import Dependency

# after
from uncalled_for import Dependency
```

Temporarily points pydocket at the `use-uncalled-for` branch so both sides can be validated together in CI. Once chrisguidry/docket#353 is merged and released, the pydocket dep goes back to a normal version pin.

Closes #3251
Makes #3292 better

(Maybe) closes #3320

### DIFF
--- a/src/fastmcp/server/context.py
+++ b/src/fastmcp/server/context.py
@@ -24,6 +24,7 @@ from mcp.types import Resource as SDKResource
 from pydantic.networks import AnyUrl
 from starlette.requests import Request
 from typing_extensions import TypeVar
+from uncalled_for import SharedContext
 
 from fastmcp.resources.resource import ResourceResult
 from fastmcp.server.elicitation import (
@@ -283,8 +284,6 @@ class Context:
         else:
             # Without docket, the lifespan won't provide a SharedContext,
             # so create one scoped to this Context for Shared() dependencies.
-            from uncalled_for import SharedContext
-
             self._shared_context = SharedContext()
             await self._shared_context.__aenter__()
 


chrisguidry:
@jlowin this draft is still pointing to a branch of docket where I'm validating that we won't have any compatibility problems with 2.x when I release it.  Will bump the version here once I'm sure.

omikader:
I believe this also closes https://github.com/PrefectHQ/fastmcp/issues/3251

chrisguidry:
I think it could!  We can do a little more here to support `Shared` in the case where you don't have tasks enabled, let me take a look at that, it should be a fairly quick add.

chrisguidry:
@omikader thanks for the nudge, this was definitely straightforward to ensure: https://github.com/PrefectHQ/fastmcp/pull/3301/commits/80526ba499b23e3b1786d1b683191860f389ec3f

chrisguidry:
I took off the `breaking change` label because I disagree with that assessment, I've gone to a lot of trouble in Docket to make sure it's still presenting the same import surface and behavior as it did in 0.17.x

PR #3306, opened 2026-02-26
OpenAPI: rewrite $ref under propertyNames and patternProperties in _replace_ref_with_defs; add regression test for dict[StrEnum, Model]

### LINKED ISSUE #3303: `propertyNames.$ref` not converted by `_replace_ref_with_defs`
### Description

When using `FastMCP.from_fastapi()` with a FastAPI app whose response model contains `dict[StrEnum, Model]`, Pydantic generates a `propertyNames` keyword with a `$ref` to the enum schema. `_replace_ref_with_defs` converts `additionalProperties.$ref` from `#/components/schemas/` to `#/$defs/`, but does not handle `propertyNames`, leaving a dangling reference.

This causes MCP clients (e.g. Cursor) to fail resolving the reference, preventing all tools from loading.

### Example Code

```python
import asyncio
from enum import StrEnum

from fastapi import FastAPI


class Category(StrEnum):
    BOOKS = "Books"
    ELECTRONICS = "Electronics"


class ItemInfo(BaseModel):
    count: int


class InventoryResponse(BaseModel):
    items: dict[Category, ItemInfo]


api = FastAPI()


@api.get("/inventory", response_model=InventoryResponse)
def get_inventory() -> InventoryResponse:
    return InventoryResponse(items={Category.BOOKS: ItemInfo(count=5)})


mcp = FastMCP.from_fastapi(app=api)


async def demo():
        for tool in tools:
            schema = tool.outputSchema or {}
            # additionalProperties.$ref is correctly #/$defs/ItemInfo
            # but propertyNames.$ref is still #/components/schemas/Category
            print(schema)


if __name__ == "__main__":
    asyncio.run(demo())
```

Output (trimmed):

```json
{
  "properties": {
    "items": {
      "additionalProperties": {"$ref": "#/$defs/ItemInfo"},
      "propertyNames": {"$ref": "#/components/schemas/Category"},
      "type": "object"
    }
  },
  "$defs": {
    "Category": {"type": "string", "enum": ["Books", "Electronics"]},
    "ItemInfo": {"properties": {"count": {"type": "integer"}}, "type": "object"}
  }
}
```

### Version Information

```
FastMCP version:                                                           3.0.1
Python version:                                                           3.12.4
Platform:                                             macOS-26.2-arm64-arm-64bit
```

Also confirmed on `main` (`schemas.py` at `bb057f2`).

manojPal23234:
## Summary

This PR fixes a gap in the OpenAPI → JSON Schema normalization path. The function `_replace_ref_with_defs` already rewrote `#/components/schemas/...` to `#/$defs/...` in common locations (e.g., properties, items, `additionalProperties`, and combinators), but it did **not** recurse into **`propertyNames`** (and **`patternProperties`**).  
As a result, schemas like `dict[StrEnum, Model]` (where keys are a `StrEnum`) produced a dangling `propertyNames.$ref` pointing to `#/components/schemas/...`, which some MCP clients failed to resolve.

This change adds explicit handling for both **`propertyNames`** and **`patternProperties`** inside the established OpenAPI normalization chokepoint.

**Fixes #3303.**

---

## What changed

- **`src/fastmcp/utilities/openapi/schemas.py`**
  - In `_replace_ref_with_defs(...)`, also recurse into:
    - `propertyNames` (mirrors the existing pattern used for `additionalProperties`)
    - `patternProperties` (recurses into each sub‑schema)

- **Tests (OpenAPI utilities)**
  - Added a focused regression test that reproduces the `dict[StrEnum, Model]` shape and asserts:
    - `additionalProperties.$ref` → `#/$defs/ItemInfo`
    - `propertyNames.$ref` → `#/$defs/Category`
    - No remaining `#/components/schemas/` in the result

  **Test file path:**  
  `tests\utilities\openapi\test_propertynames_ref_rewrite.py`

---

## Rationale & scope

- Keeps the fix **localized to the OpenAPI layer** (`_replace_ref_with_defs`) per review feedback, avoiding a second, parallel normalization path in generic JSON‑Schema utilities.
- No collision/aliasing logic is introduced; this path is not expected to receive schemas with conflicting `"$defs"` and `components.schemas`.
- Minimal, keyword‑targeted diff that closes the specific bug without altering broader behavior.

---

## Regression test (excerpt)

```python
from fastmcp.utilities.openapi.schemas import _replace_ref_with_defs

def test_replace_ref_with_defs_rewrites_propertyNames_ref():
    """
    Regression test for dict[StrEnum, Model] case:
    propertyNames.$ref and additionalProperties.$ref must both normalize to #/$defs/...
    """
    schema = {
        "type": "object",
        "propertyNames": {"$ref": "#/components/schemas/Category"},
        "additionalProperties": {"$ref": "#/components/schemas/ItemInfo"},
    }

    result = _replace_ref_with_defs(schema)

    assert result["additionalProperties"]["$ref"] == "#/$defs/ItemInfo"
    assert result["propertyNames"]["$ref"] == "#/$defs/Category"
    assert "#/components/schemas/" not in str(result)

### DIFF
--- a/src/fastmcp/utilities/openapi/schemas.py
+++ b/src/fastmcp/utilities/openapi/schemas.py
@@ -132,6 +132,8 @@ def _replace_ref_with_defs(
         if isinstance(pattern_properties, dict):
             schema["patternProperties"] = {
                 pattern: _replace_ref_with_defs(subschema)
+                if isinstance(subschema, dict)
+                else subschema
                 for pattern, subschema in pattern_properties.items()
             }
     if info.get("description", description) and not schema.get("description"):


manojPal23234 (commented):
**Deterministic migration:** replaced `dst.setdefault(k, v)` with a collision-aware policy.
- If both `$defs.Name` and `components.schemas.Name` exist and differ:
  - Preserve the old `$defs.Name` **only if** it’s referenced directly via `#/$defs/Name`, by moving it to an alias (e.g., `Name__preexisting`) and rewriting those direct refs.
  - Then overwrite `$defs[Name]` with the **component** definition so normalized component refs resolve to the intended schema.
This removes migration order dependence and avoids mixed schemas silently picking stale `$defs` entries.

manojPal23234:
> Thanks for digging into this — the issue is real and your investigation of the collision edge case shows careful thinking about the problem space.
> 
> That said, I think the fix belongs in a different layer. The function `_replace_ref_with_defs` in `src/fastmcp/utilities/openapi/schemas.py` is the established chokepoint for converting OpenAPI-style `#/components/schemas/` refs into JSON Schema `#/$defs/` refs. It already handles `properties`, `items`, `additionalProperties`, and combinators — it just doesn't recurse into `propertyNames` (or `patternProperties`, which has the same shape of problem).
> 
> This is essentially a one-line fix in `_replace_ref_with_defs` — add `propertyNames` handling the same way `additionalProperties` is handled. Adding a second, parallel normalization path in `json_schema.py` (which handles generic JSON Schema and shouldn't need to know about OpenAPI conventions) creates two different rewriting strategies that don't know about each other, which is a maintenance problem.
> 
> The collision handling (`_migrate_components_to_defs` with aliasing) isn't something we need — schemas coming through this path won't have both `$defs` and `components.schemas` with conflicting definitions. Let's keep this scoped to the actual bug.
> 
> So the ideal version of this PR would be:
> 
> 1. Add `propertyNames` and `patternProperties` handling to `_replace_ref_with_defs` (same pattern as `additionalProperties`)
> 2. Add a regression test for the `dict[StrEnum, Model]` case from #3303, in the existing test suite under `tests/utilities/openapi/`
> 
> That should be a pretty small diff. Happy to help if you have questions about the codebase structure.

Hi! Thanks for the guidance and for clarifying the right layer to fix this in.
I’ve updated the PR to:

Scope the normalization to the OpenAPI chokepoint (_replace_ref_with_defs) as suggested.
Add recursive handling for:

propertyNames (mirrors existing additionalProperties handling), and
patternProperties (recurses into each sub‑schema).


Include a regression test for the dict[StrEnum, Model] case (keys via propertyNames, values via additionalProperties) that verifies both refs are rewritten to #/$defs/... and no #/components/schemas/... remains.

Per your review, I removed the earlier generic normalization and the collision/aliasing logic from json_schema.py to avoid duplicating strategies and to keep this fix focused on the actual gap.

PR #3335, opened 2026-02-28
Add MultiAuth for composing multiple token verification sources

### LINKED ISSUE #1802: Introduce multi-tenant support for OAuthProxy
## Context

The current **OAuth Proxy** implementation allows configuring only a single upstream authorization server. This limitation makes it unsuitable for multi-tenant scenarios where requests must be routed to different upstream servers based on tenant context.

## Proposal

Introduce a `TenantResolver` interface (protocol/abstract base class) with a single method:

```python
class TenantResolver(Protocol):
    def resolve(self, request: Request) -> TenantConfiguration: ...
```

1. `resolve` receives the incoming request and returns a `TenantConfiguration` containing all required upstream details.
2. A custom `TenantResolver` implementation can be optionally provided when initializing the `OAuthProxy`.

This lets developers plug in tenant-specific logic to determine which authorization server should handle each request.

## Implementation Details

If no `TenantResolver` is provided, or if resolution fails, the proxy should fall back to the existing static upstream configuration.

This preserves backward compatibility and avoids breaking existing setups.

PS: I am willing to provide an implementation if the proposal is accepted.

### LINKED ISSUE #2499: Documentation Request: How to combine GoogleProvider OAuth with API key authentication
### Use Case

I'm building an MCP server deployed on Google Cloud Run that needs to support **two authentication methods**:

1. **Google OAuth** (via `GoogleProvider`) - for interactive clients like Claude Desktop
2. **Static API key** - for programmatic/non-interactive access (CI/CD, scripts, service-to-service calls)

This is a common pattern for production APIs where you want both user-facing OAuth and machine-to-machine API key authentication.

### What I Tried

I created a `DualAuthProvider` that extends `GoogleProvider` and overrides `verify_token()` to check for an API key first, then delegate to the parent:

```python
from mcp.server.auth.provider import AccessToken

class DualAuthProvider(GoogleProvider):
    def __init__(self, api_key: str, client_id: str, client_secret: str, base_url: str, **kwargs):
        self.api_key = api_key
        super().__init__(
            client_id=client_id,
            client_secret=client_secret,
            base_url=base_url,
            **kwargs,
        )

    async def verify_token(self, token: str) -> Optional[AccessToken]:
        # Check API key first
        if token == self.api_key:
            return AccessToken(
                token=token,
                client_id="api-key-client",
                scopes=["openid", "email", "profile"],
                expires_at=None,
            )
        
        # Delegate to GoogleProvider for OAuth tokens
        return await super().verify_token(token)
```

### What Went Wrong

The OAuth flow completes successfully (token is issued via `/token` endpoint), but when the client uses the token, it's rejected with `invalid_token`:

```
POST /token HTTP/1.1" 200 OK          # Token issued successfully
POST /mcp HTTP/1.1" 401 Unauthorized  # Token immediately rejected
INFO Auth error returned: invalid_token
```

It appears that `GoogleProvider.verify_token()` expects Google-issued tokens (validated against Google's token endpoint), not the MCP server's self-issued OAuth tokens. By extending `GoogleProvider` and relying on `super().verify_token()`, I'm breaking the MCP token verification.

### Request

Could you provide documentation or examples showing the **recommended approach** for combining `GoogleProvider` OAuth with API key authentication? Specifically:

1. **Is inheritance from `GoogleProvider` supported** for adding custom token verification logic?
2. **Should composition be used instead** (wrapping `GoogleProvider` rather than extending it)?
3. **Is middleware the right approach** for API key auth while keeping `GoogleProvider` intact?
4. **Are there plans for a built-in `DualAuthProvider`** or similar that supports this pattern?

### Additional Context

This pattern would be valuable for many production deployments where:
- Human users authenticate via OAuth (Claude Desktop, web clients)
- Automated systems authenticate via API keys (CI/CD, monitoring, internal services)

The current documentation covers each auth method individually but doesn't address combining them.

### Example Code

```Python
from mcp.server.auth.provider import AccessToken

class DualAuthProvider(GoogleProvider):
    def __init__(self, api_key: str, client_id: str, client_secret: str, base_url: str, **kwargs):
        self.api_key = api_key
        super().__init__(
            client_id=client_id,
            client_secret=client_secret,
            base_url=base_url,
            **kwargs,
        )

    async def verify_token(self, token: str) -> Optional[AccessToken]:
        # Check API key first
        if token == self.api_key:
            return AccessToken(
                token=token,
                client_id="api-key-client",
                scopes=["openid", "email", "profile"],
                expires_at=None,
            )
        
        # Delegate to GoogleProvider for OAuth tokens
        return await super().verify_token(token)
```

### Version Information

```Text
FastMCP version:                                                                2.13.0rc3.dev9+811e5166
MCP version:                                                                                     1.18.0
Python version:                                                                                 3.12.11
Platform:                                                                    macOS-26.1-arm64-arm-64bit
FastMCP root path: /Users/nickyoungblut/dev/python/arc-benchling-mcp/.venv/lib/python3.12/site-packages
```

### LINKED ISSUE #3035: M2M Authentication Failing with Entra ID OAuth
### Description
Unable to authenticate machine-to-machine (M2M) requests to our OAuth-protected resource using Entra ID (Azure AD) client credentials flow.

### Environment
- Identity Provider: Microsoft Entra ID (formerly Azure AD)
- Grant Type: Client Credentials (OAuth 2.0)
- Target: FastMCP server tools/resources
- Configuration: MCP server configuration (mcpServers)

### Current Behavior
Receiving 401 Unauthorized when attempting to access FastMCP tools with tokens obtained from Entra ID and passed via the `mcpServers` configuration.

**Root cause:** Token algorithm mismatch between Entra ID and FastMCP's internal JWT issuer. Entra ID issues tokens using the RS256 (RSA with SHA-256) asymmetric algorithm, while the FastMCP internal `jwt_issuer` uses HS256 (HMAC with SHA-256) symmetric algorithm. When FastMCP attempts to verify an Entra ID token using the HS256 verifier, it fails validation, resulting in the error: `"Auth error returned: invalid_token (status=401)"`.

This means M2M clients cannot authenticate to FastMCP tools because their RS256-signed tokens cannot be validated by FastMCP's HS256-only token verification.

### Expected Behavior
Should successfully authenticate M2M requests to FastMCP tools using valid access tokens obtained from Entra ID and configured in the `mcpServers` configuration with `Authorization: Bearer {access_token}` headers.

### Steps to Reproduce
1. Register application in Entra ID
2. Create client secret
3. Grant API permissions (application permissions, not delegated)
4. Request token using client credentials flow:
```bash
curl -X POST https://login.microsoftonline.com/{tenant-id}/oauth2/v2.0/token \
  -H "Content-Type: application/x-www-form-urlencoded" \
  -d "client_id={client-id}" \
  -d "client_secret={client-secret}" \
  -d "scope=api://{fastmcp-app-id}/.default" \
  -d "grant_type=client_credentials"
```
5. Extract the `access_token` from the response and configure it in the MCP server configuration:
```json
{
  "mcpServers": {
    "my-fastmcp-server": {
      "url": "http://localhost:8001/mcp",
      "headers": {
        "Authorization": "Bearer {access_token}"
      }
    }
  }
}
```
6. Attempt to use FastMCP tools from the configured server
7. Observe 401 Unauthorized response with error: `"Auth error returned: invalid_token (status=401)"` due to algorithm mismatch (RS256 token being validated with HS256 verifier)

### Possible Solution
Implement support for multiple JWT verification algorithms in FastMCP to allow flexible authentication:

1. **Primary:** Continue validating tokens using HS256 for FastMCP-issued tokens (current internal JWT issuer)
2. **Fallback:** Add RS256 JWT verifier that validates tokens against Entra ID's (or other OAuth providers') public keys

This would allow:
- Internal tokens (issued through FastMCP's `jwt_issuer` using HS256) to continue working as they do now
- External tokens (issued directly from Entra ID or other OAuth providers using RS256) to be validated via JWT signature verification

**Implementation approach:**
- Detect the algorithm from the JWT header (`alg` claim)
- For HS256 tokens: Use existing internal verification with shared secret
- For RS256 tokens: Fetch and cache public keys from the issuer's JWKS endpoint (e.g., `https://login.microsoftonline.com/{tenant-id}/discovery/v2.0/keys`)
- Verify standard claims (issuer, audience, expiration, etc.)
- Allow request to proceed if verification succeeds

jlowin:
Production servers often need to accept tokens from multiple authentication sources — an OAuth proxy for interactive MCP clients alongside JWT verification for machine-to-machine tokens, for example. Today there's no built-in way to compose these; you'd have to subclass `AuthProvider` and wire the chain yourself. `MultiAuth` makes this a one-liner.

`MultiAuth` wraps an optional auth server (the "primary" that owns routes and OAuth metadata) with additional `TokenVerifier` instances. On each request it tries the server first, then each verifier in order, accepting the first successful result. Routes and middleware delegate to the server; verifiers contribute only `verify_token`.

```python
from fastmcp import FastMCP
from fastmcp.server.auth import MultiAuth, OAuthProxy

auth = MultiAuth(
    server=OAuthProxy(
        issuer_url="https://login.example.com/...",
        client_id="my-app",
        client_secret="secret",
        base_url="https://my-server.com",
    ),
    verifiers=[
        JWTVerifier(
            jwks_uri="https://internal.example.com/.well-known/jwks.json",
            issuer="https://internal.example.com",
            audience="my-mcp-server",
        ),
    ],
)

```

🤖 Generated with Claude Code

https://claude.ai/code/session_01WwKYDCqjM2FqYwY5ZNVvjb

Closes #3035 
Closes #2499 
Closes #1802

### DIFF
--- a/docs/servers/auth/authentication.mdx
+++ b/docs/servers/auth/authentication.mdx
@@ -180,7 +180,7 @@ This example shows the basic structure of a custom OAuth provider. The actual im
 
 ### MultiAuth
 
-<VersionBadge version="2.14.0" />
+<VersionBadge version="3.1.0" />
 
 `MultiAuth` composes multiple authentication sources into a single `auth` provider. When a server needs to accept tokens from different issuers — for example, an OAuth proxy for interactive clients alongside JWT verification for machine-to-machine tokens — `MultiAuth` tries each source in order and accepts the first successful verification.
 
--- a/docs/servers/auth/multi-auth.mdx
+++ b/docs/servers/auth/multi-auth.mdx
@@ -7,7 +7,7 @@ icon: layer-group
 
 import { VersionBadge } from "/snippets/version-badge.mdx"
 
-<VersionBadge version="2.14.0" />
+<VersionBadge version="3.1.0" />
 
 Production servers often need to accept tokens from multiple authentication sources. An interactive application might authenticate through an OAuth proxy, while a backend service sends machine-to-machine JWT tokens directly. `MultiAuth` composes these sources into a single `auth` provider so every valid token is accepted regardless of where it was issued.
 


jlowin:
gtg

joar:
It would be nice to provide a mechanism to see which `verifier` out of `[server] + verifiers` the current `AuthenticatedUser` belongs to, I guess that can be worked around by comparing metadata in the tokens.

Something that's missing for my use case as well is the ability to have simultaneous authentication of the agent that is forwarding the request, as well as the end-user.

For example:
- I would like to deny tools/call by the agent.
- The end-user would be allowed to do anything FastMCP supports.

with this MultiAuth, I would have an MCP middleware, but then I would have to compare what I would expect to see in a JWT for the agent, with what I would expect to see in a JWT for an end-user, and deny tool calls based on that.

and also, I would only be able to authenticate either the agent or the end-user.

I have a hand-crafted application-code solution that solves the simultaneous auth of agent and end-user by separating the tokens into different headers, I also have an upstream WIP that attempts to do the same: https://github.com/PrefectHQ/fastmcp/compare/main...joar:fastmcp:multi-auth?expand=1

The requirement to list tools before end-user is authenticated is a common source of friction between agent frameworks (ADK, langchain) and the MCP standard, adding a separate `Service-Authorization` header in the agent with e.g., a Google Service Account OIDC JWT, and allowing tools/list on the MCP side given that the `Service-Authorization` JWT is valid is a solution to that problem.

jlowin:
@joar thanks for the detailed writeup! I took a look at your branch — the `MultiHeaderAuthProvider` approach is clever, and it's nice to see it working through the existing `get_middleware()` extension point without needing SDK changes.

That said, I think this is fighting against the MCP spec in a way that's hard to support at a framework level. MCP doesn't distinguish between list and execute authorization, and maintaining that split (where a client can discover tools it can't call) is something we actively try to avoid in FastMCP. Your `require_user_for_execution` check is a reasonable workaround, but it's inherently application-specific policy.

I think MultiAuth (multiple verifiers for a single identity) and what you're building (multiple simultaneous identities with different authorization policies) are different enough that they don't belong in the same abstraction. For now I'd suggest keeping your multi-header approach as application-level customization — the extension points are there for exactly this kind of thing.

Re: knowing which verifier authenticated the user — happy to consider that separately.

PR #3337, opened 2026-02-28
Add search_result_serializer hook and serialize_tools_for_output_markdown

### LINKED ISSUE #3336: CodeMode search results inflate context for tools with complex schemas
## Problem

While migrating the [OpenBB MCP server](https://github.com/OpenBB-finance/openbb-mcp) to FastMCP v3 and experimenting with `CodeMode`, I ran into a context inflation problem. `CodeMode` is designed to reduce context by hiding the full tool catalog behind a search interface — but the search results themselves turn out to be a significant source of token overhead for servers with complex tool schemas.

OpenBB tools are generated from an OpenAPI spec, so each tool carries a large, detailed JSON schema. When the LLM calls `search`, it gets back full MCP-serialized tool definitions in JSON. That format includes a lot of structural boilerplate the LLM doesn't need at that stage — it only needs to know the tool name, a description, and the parameter names and types to decide whether to use the tool and how to call it. The full JSON payload undoes a meaningful portion of the context savings that `CodeMode` is supposed to provide.

## Example

A single OpenBB tool like `equity_price_historical` produces a search result of roughly 700 tokens as JSON. The same information as compact markdown is around 80 tokens — the LLM can read the parameter names, types, and which are required just as well from the shorter format.

A [benchmark script](https://gist.github.com/MagnusS0/6b010566c98ac71e32046546d63fd393) across a representative 11-tool catalog (including one complex financial tool) measured a **65.1% token reduction** (2219 → 775 tokens) switching from JSON to markdown serialization.

## Expected behavior

There should be a way to control how search results are serialized, so servers with verbose tool schemas can reduce the token cost of search responses without losing the information the LLM actually needs.

MagnusS0:
## Description

While migrating the OpenBB MCP server to FastMCP v3 I was playing around with `CodeMode` and noticed that search results were eating a lot of the context savings it's supposed to provide. The LLM gets back full JSON tool definitions (same as `list_tools` ) which is fine for simple tools but balloons fast with anything schema-heavy.

My solution is a `search_result_serializer` hook on `BaseSearchTransform` (and `CodeMode`), so you can swap in whatever serialization makes sense for your use case. The built-in `serialize_tools_for_output_markdown` strips the JSON boilerplate and renders just what the LLM actually needs to pick and call a tool. In my simple benchmark across an 11-tool catalog with some complex OpenBB-style schemas, it cut search result tokens by ~65-70% ([benchmark script](https://gist.github.com/MagnusS0/6b010566c98ac71e32046546d63fd393)). Default behavior is unchanged, fully opt-in.

```python
from fastmcp.experimental.transforms import CodeMode
from fastmcp.server.transforms.search import serialize_tools_for_output_markdown

mcp = FastMCP("Server", transforms=[
    CodeMode(search_result_serializer=serialize_tools_for_output_markdown)
])
```

The markdown output for a tool looks like this:

```
### create_document

Create a new document in the workspace with the given metadata.

**Parameters**
- `title` (string, required)
- `content` (string, required)
- `tags` (string[], required)
- `author` (string, required)
- `published` (boolean)
- `parent_id` (string?)

**Returns**
- `value` (object)
```

Both built-in serializers (`serialize_tools_for_output_json` and `serialize_tools_for_output_markdown`) are exported from `fastmcp.server.transforms.search` and work on standalone search transforms too, not just `CodeMode`.

*Generated with [Claude Code](https://claude.ai/code)*

**Contributors Checklist**

- [x] My change closes #3336

**Review Checklist**

### DIFF
--- a/src/fastmcp/cli/cli.py
+++ b/src/fastmcp/cli/cli.py
@@ -212,6 +212,13 @@ async def inspector(
             help="Directories to watch for changes (default: current directory)",
         ),
     ] = None,
+    module: Annotated[
+        bool,
+        cyclopts.Parameter(
+            name=["--module", "-m"],
+            help="Run a Python module (python -m <module>) instead of importing a server object",
+        ),
+    ] = False,
 ) -> None:
     """Run an MCP server with the MCP Inspector for development.
 
@@ -254,7 +261,11 @@ async def inspector(
             logger.error("No configuration available")
             sys.exit(1)
         assert config is not None  # For type checker
-        await config.source.load_server()
+
+        # Skip server-object validation in module mode — the module
+        # manages its own startup and may not expose an importable server.
+        if not module:
+            await config.source.load_server()
 
         env_vars = {}
         if ui_port:
@@ -277,6 +288,10 @@ async def inspector(
 
         # Build the fastmcp run command
         fastmcp_cmd = ["fastmcp", "run", server_spec, "--no-banner"]
+
+        # Forward module mode flag
+        if module:
+            fastmcp_cmd.append("--module")
 
         # Add reload flags if enabled - the server will handle reloading
         if reload:
@@ -424,6 +439,13 @@ async def run(
             help="Run in stateless mode (no session, used internally for reload)",
         ),
     ] = False,
+    module: Annotated[
+        bool,
+        cyclopts.Parameter(
+            name=["--module", "-m"],
+            help="Run a Python module (python -m <module>) instead of importing a server object",
+        ),
+    ] = False,
 ) -> None:
     """Run an MCP server or connect to a remote one.
 
@@ -434,6 +456,7 @@ async def run(
     4. MCPConfig file: "mcp.json" - runs as a proxy server for the MCP Servers in the MCPConfig file
     5. FastMCP config: "fastmcp.json" - runs server using FastMCP configuration
     6. No argument: looks for fastmcp.json in current directory
+    7. Module mode: "-m my_module" - runs the module directly via python -m
 
     Server arguments can be passed after -- :
     fastmcp run server.py -- --config config.json --debug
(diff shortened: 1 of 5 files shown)


MagnusS0:
More general comment when it comes to token usage and search tools, maybe a follow-up PR.
But from my experience, from a context perspective, I have found the approach taken by  [mcp-cli](https://github.com/philschmid/mcp-cli) where you have a two step search:

1) candidate selection (`name` + short `description`)
2) full schema for only 1-2 finalists

A lot more token efficient and scales better when you have a lot of tools.

Could be something like:

```py
# default `summary`: name, short description, tags/category
# optional `detail="full"` for current behavior

search(query, detail="summary")

# return full input/output schema for selected tool(s)
get_schema(name | [name1, name2, ...] )

 ```

**Comment by jlowin** [comment]
Thanks @MagnusS0 this is really solid work! The serializer hook and the markdown rendering are both well thought out, and I totally agree with the direction you're going here. Your follow-up comment about two-stage search actually resolves some tension I've been feeling about how search results scale with complex tool catalogs, so thanks for surfacing that.

Here's where I think this should land architecturally: we'll split the current search tool into a progressive disclosure flow —

- **`search`** returns just names and short descriptions, super lightweight. Option to get more detail. 
- **`get_schema`** takes tool name(s) and returns abbreviated markdown (your serializer) by default, with an option to get the full JSON schema when the LLM needs it
- **`execute`** stays as-is

The LLM regulates its own token budget — scan candidates cheaply, pull schema detail only for the tools it actually cares about. Some default to let users err for more detail on simple servers (where two calls is worse than one due to latency) or less detail on complex servers (where two search calls is probably more efficient overall).

Since you already suggested this as a follow-up, I'd love to merge this as-is and then quickly follow up with a PR that implements the staged approach on top. You've already done the heavy lifting here with the serializer — the rest is reshaping the tool surface around it. This will all land together before it ships. Ideally targeting later this week or next.

**Comment by MagnusS0** [comment]
Thanks @jlowin, really glad it landed well and that you like the direction!

One more thought for the staged approach, since you're designing the tool surface now anyway, for MCPs or converted APIs with say 10–20 domains,  even a lightweight search over all candidates can get noisy. A list_categories() call as a cheap first step would let the agent scope before it does anything else:

```py
list_categories()   # ["equity", "sec", "etf", ...]
search("price", category="equity", detail="summary")   # scoped candidates
get_schema("price_history")   # full detail
```

This could map to MCP tags if tools already have them, so no extra annotation burden. For servers where tags aren't set maybe don't expose the tool at all. 

One pattern I've hit personally: sub-categories (e.g. equity → {price, fundamentals}). Probably niche, but if it's easy to expose as an opt-in, say a custom categories mapping on the transform, it would be very nice.

Just flagging before the surface is locked in, easier to bake in now than retrofit later.

jlowin:
So there is this interesting fallback (we actually started with it but it's too complicated for simple cases) where you can use the *call tool* to actually learn what the tools are. like we could actually expose the schema as a data structure and allow an LLM to process it arbitrarily. I think its a interesting extreme case. Maybe we find a way to support a spectrum of tools with some signature. I'll think on it!

PR #3349, opened 2026-03-02
Add include_unversioned option to VersionFilter

### LINKED ISSUE #3339: VersionFilter includes unversioned components with a version range
### Description

Applying `VersionFilter(version_gte="2.0")` still exposes components whose `version` is `None`, which looks inconsistent with the `VersionFilter` documentation that says only components within the version range are visible.

### Example Code

```Python
import anyio

from fastmcp import FastMCP
from fastmcp.server.transforms import VersionFilter

mcp = FastMCP("version-filter-repro")

@mcp.tool
def bare_tool() -> str:
    return "bare"

mcp.add_transform(VersionFilter(version_gte="2.0"))

async def main() -> None:
    tools = await mcp.list_tools()
    print([(tool.name, tool.version) for tool in tools])

if __name__ == "__main__":
    anyio.run(main)
```

### Version Information

```Text
FastMCP version: 3.0.2
MCP version: 1.26.0
Python version: 3.10.12
Platform: Linux-6.6.87.2-microsoft-standard-WSL2-x86_64-with-glibc2.35
FastMCP root path: /home/yang/fast_mcp_test/.venv/lib/python3.10/site-packages
```

yangbaechu:
## Description

This PR adds an `include_unversioned` option to `VersionFilter`, defaulting to `True` to preserve the current behavior. When set to `False`, unversioned components are excluded from `VersionFilter` list results by passing `match_none=self.include_unversioned` through the existing `VersionSpec.matches()` calls.

  I kept the change narrowly scoped to the list filtering behavior discussed on the issue. The PR also adds regression coverage for tools, resources, and mounted tools, updates `VersionFilter`'s docstring, and refreshes the versioning guide to document the new option and its default behavior.

I used gpt-5.3-codex to help draft and refine this change

**Contributors Checklist**

- [X] My change closes #3339
- [X] I have performed all required documentation updates

**Review Checklist**

---

### DIFF
--- a/tests/server/versioning/test_filtering.py
+++ b/tests/server/versioning/test_filtering.py
@@ -4,6 +4,7 @@
 from __future__ import annotations
 
 from fastmcp import FastMCP
+from fastmcp.server.transforms import VersionFilter
 from fastmcp.utilities.versions import (
     VersionSpec,
 )
@@ -151,7 +152,6 @@ class TestVersionFilter:
 
     async def test_include_unversioned_false_excludes_unversioned_tools(self):
         """Setting include_unversioned=False hides unversioned tools."""
-        from fastmcp.server.transforms import VersionFilter
 
         mcp = FastMCP()
 
@@ -244,7 +244,6 @@ class TestVersionFilter:
 
     async def test_include_unversioned_false_excludes_unversioned_resources(self):
         """Setting include_unversioned=False hides unversioned resources."""
-        from fastmcp.server.transforms import VersionFilter
 
         mcp = FastMCP()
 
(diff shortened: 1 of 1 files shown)


jlowin (changes requested):
Thanks for picking this up! Two things before this is ready:

The tests cover tools, resources, and mounted tools, but prompts and resource templates have no coverage for `include_unversioned=False` despite being implemented. Same pattern as the others — should be straightforward to add.

Also, imports belong at the top of the test file, not inside individual test methods. The existing tests in that file do this too, but please don't extend the pattern in the new ones.

yangbaechu:
Thanks for the review. I updated this PR by adding `include_unversioned=False` coverage for prompts and resource templates and moving imports to the top for the tests touched here. I also aligned the test scenarios so each test now defines three inputs (unversioned, in-range versioned, and out-of-range versioned) and verifies that only the in-range versioned entry remains after filtering.

PR #3385, opened 2026-03-04
Add FastMCPApp — a Provider for composable MCP applications

jlowin:
Today, `app=True` on a tool is the way to build an MCP app — it wires up the Prefab renderer and CSP, and the tool returns a component tree. That works great for standalone servers, but breaks down under composition. When a server is mounted under a namespace, `save_contact` becomes `crm_save_contact`, and the `CallTool("save_contact")` baked into the UI stops working.

`FastMCPApp` is a Provider that makes applications a first-class concept. It binds entry-point tools (the model calls these) together with backend tools (the UI calls these via CallTool), and gives backend tools global keys — UUID-suffixed stable identifiers that survive namespace transforms.

```python
from fastmcp import FastMCP, FastMCPApp

app = FastMCPApp("CRM")

@app.ui()
def contact_form():
    return Form(on_submit=CallTool(save_contact))

@app.tool()
def save_contact(name: str, email: str) -> dict:
    return {"name": name, "email": email}

# Works standalone
app.run()

# Works composed — global keys survive the namespace transform
server = FastMCP("Platform")
server.add_provider(app, namespace="crm")
```

`@app.ui()` registers model-visible entry points with auto-wired Prefab renderer and CSP. `@app.tool()` registers app-visible backend tools with global keys. The server's `call_tool` checks the global key registry as a fallback after normal resolution, so `CallTool` references reach the right tool regardless of how the server is composed. Auth still applies to global-key routed tools.

The callable resolver (`CallTool(fn)` instead of `CallTool("string")`) is wired defensively — it no-ops until Prefab ships `set_call_tool_serializer`. The FastMCP side is self-contained and independently mergeable.

### DIFF
--- a/src/fastmcp/server/app.py
+++ b/src/fastmcp/server/app.py
@@ -104,7 +104,7 @@ def _resolve_tool_ref(fn: Any) -> Any:
     2. ``__fastmcp__`` metadata (decorated but not on a FastMCPApp)
     3. ``fn.__name__`` (bare function — works for standalone servers)
     """
-    from prefab_ui.app import ResolvedTool  # ty: ignore[unresolved-import]
+    from prefab_ui.app import ResolvedTool
 
     global_key = _FN_TO_GLOBAL_KEY.get(id(fn))
     if global_key is not None:
--- a/tests/test_fastmcp_app.py
+++ b/tests/test_fastmcp_app.py
@@ -15,7 +15,7 @@ import re
 from unittest.mock import AsyncMock
 
 import pytest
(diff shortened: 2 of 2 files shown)


jlowin:
**Note:** The callable resolver wiring (`to_json(tool_resolver=...)`) requires a Prefab release with prefecthq/prefab#222. The current code detects support via signature inspection and falls back to plain `to_json()` on older versions, so there's no hard breakage. Once Prefab ships with `tool_resolver`, we should bump the `apps` extra to require that version (`prefab-ui>=0.X.0`).

PR #3394, opened 2026-03-05
fix: dereference_refs() v2 backport (fixes #3153)

### LINKED ISSUE #3153: Backport of dereference_refs() to v2 doesn't actually work
### Description

https://github.com/jlowin/fastmcp/pull/2861 does nothing - it adds derefence_refs(), but doesn't actually call it. Maybe that's a good thing considering the controversial nature of the change, but it doesn't seem intentional :-)

### Example Code

```Python

```

### Version Information

```Text
2.14.5
```

hkc5:
## Summary

This PR fixes issue #3153 where the backport of `dereference_refs()` to v2 was incomplete. The function was added but was missing critical cycle detection logic, causing circular references to fail with "ValueError: Circular reference detected" instead of gracefully falling back to root-only resolution.

## Changes

- Added `_defs_have_cycles()` function to detect circular references in $defs before attempting dereferencing
- Added cycle detection at the start of `dereference_refs()` to fall back to `resolve_root_ref()` for circular schemas
- This prevents `jsonref.replace_refs()` from creating Python object cycles that fail JSON serialization

## Testing

- All existing tests in `tests/utilities/test_json_schema.py::TestDereferenceRefs` pass
- Verified circular reference schemas now serialize correctly to JSON
- Code style checks pass (black, ruff)

## Related

- Fixes #3153
- Backports cycle detection logic from main branch

### DIFF
--- a/src/fastmcp/utilities/json_schema.py
+++ b/src/fastmcp/utilities/json_schema.py
@@ -6,6 +6,53 @@ from typing import Any
 from jsonref import JsonRefError, replace_refs
 
 
+def _defs_have_cycles(defs: dict[str, Any]) -> bool:
+    """Check whether any definitions in ``$defs`` form a reference cycle.
+
+    A cycle means a definition directly or transitively references itself
+    (e.g. Node → children → Node, or A → B → A).  ``jsonref.replace_refs``
+    silently produces Python-level object cycles for these, which Pydantic's
+    serializer rejects.
+    """
+    if not defs:
+        return False
+
+    # Build adjacency: def_name -> set of def_names it references.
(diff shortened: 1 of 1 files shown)


hkc5:
@jlowin - would appreciate your review

hkc5:
Bumping for review — all CI checks passed. This fixes the circular reference issue in dereference_refs() that was causing JSON serialization failures with schemas containing cyclic $defs. Happy to make any adjustments if needed!

hkc5:
Closing this — after reviewing the issue more carefully, I realize this needs more investigation. The fix I proposed may not address the root cause correctly. Will revisit with a better understanding of the dereference_refs flow.

PR #3404, opened 2026-03-06
fix: propagate tags from skill frontmatter to SkillResource

### LINKED ISSUE #3402: `SkillsProvider` does not add tags to `SkillResource`
### Enhancement

I wanted to implement role-based access on the skills provided by the MCP Server, using [tag-based filtering](https://gofastmcp.com/servers/visibility#filter-criteria) as I saw that the resources provider return the following object.

```
SkillResource(uri=AnyUrl('skill://name/SKILL.md'), name='name/SKILL.md', description='description', tags=set())
```

But the issue is that the tag is never set. 
I went digging in the code and found that if I add a `tags` header to my skill, the tags are registers in `frontmatter` and could theorically be added to the resource as:

```python
SkillResource(
                uri=AnyUrl(f"skill://{skill.name}/{self._main_file_name}"),
                name=f"{skill.name}/{self._main_file_name}",
                description=skill.description,
                mime_type="text/markdown",
                tags=set(skill.frontmatter.get("tags", [])),
                skill_info=skill,
                is_manifest=False,
            )
```

But it's not the case actually, do you think that could be an interesting addition ?

clawtom:
## Summary

Fixes #3402.

`SkillsProvider` (and `SkillProvider`) were constructing `SkillResource` objects without populating the `tags` field, even when `tags` were declared in the skill's YAML frontmatter. This prevented tag-based filtering from working for skill resources.

### Root cause

In `skill_provider.py`, both the `list_resources()` and `get_resource()` paths constructed `SkillResource` with an empty/default `tags` set:

```python
SkillResource(
    uri=AnyUrl(f"skill://{skill.name}/{self._main_file_name}"),
    name=f"{skill.name}/{self._main_file_name}",
    description=skill.description,
    mime_type="text/markdown",
    # tags were never passed — defaults to set()
    skill_info=skill,
    is_manifest=False,
)
```

The `tags` data is available in `skill.frontmatter` (parsed from YAML frontmatter), but was never forwarded to `SkillResource`.

### Fix

Pass `tags=set(skill.frontmatter.get("tags", []))` in both construction sites (list and get paths). The manifest resources are intentionally left without tags, as they represent directory listings rather than the skill content itself.

### Testing

To reproduce before fix:
```markdown
---
description: My skill
tags: [python, testing]
---
```
```python
resources = await provider.list_resources()
# resources[0].tags == set()  # wrong — should be {"python", "testing"}
```

After fix, `resources[0].tags == {"python", "testing"}`.

### DIFF
--- a/src/fastmcp/server/providers/skills/skill_provider.py
+++ b/src/fastmcp/server/providers/skills/skill_provider.py
@@ -290,6 +290,7 @@ class SkillProvider(Provider):
                 name=f"{skill.name}/{self._main_file_name}",
                 description=skill.description,
                 mime_type="text/markdown",
+                tags=set(skill.frontmatter.get("tags", [])),
                 skill_info=skill,
                 is_manifest=False,
             )
@@ -362,6 +363,7 @@ class SkillProvider(Provider):
                 name=f"{skill_name}/{self._main_file_name}",
                 description=skill.description,
                 mime_type="text/markdown",
+                tags=set(skill.frontmatter.get("tags", [])),
                 skill_info=skill,
                 is_manifest=False,
             )


jlowin:
Tags are not part of skills frontmatter - this requires discussion in issue before opening a PR.

PR #3408, opened 2026-03-06
Avoid stale context leakage when proxying with an already‑connected ProxyClient

jlowin:
### Motivation

- Proxy forwarding could bind backend-initiated interactions to a stale request context when a long‑lived `ProxyClient` session is reused, enabling cross‑client information leakage. This change targets the vulnerable path while keeping existing semantics for other client types.

### Description

- Special-case the client factory selection in `_create_client_factory` so a connected `ProxyClient` produces fresh disconnected copies via `client.new()` instead of reusing the same session. 
- Preserve existing reuse behavior for other connected plain `Client` instances to avoid broad regressions. 
- Add/adjust a unit test to assert that a connected `ProxyClient` target yields fresh, disconnected client copies from `_create_client_factory`.

### Testing

- Ran `uv sync` successfully. 
- Ran the full test suite with `uv run pytest -n auto`; the suite executed but the environment produced several unrelated timeouts/failures (11 failed, 4677 passed in this run) that appear orthogonal to this change. 
- Ran focused tests: the new/modified proxy test `tests/server/providers/proxy/test_proxy_client.py::TestProxyClient::test_connected_proxy_client_uses_fresh_sessions` passed, and a short selection of other tests showed two unrelated failures/timeouts under this environment. 
- `uv run prek run --all-files` failed to initialize hooks due to an external network clone error.

------
[Codex Task](https://chatgpt.com/codex/tasks/task_b_69ab2b332b28832db71093426d8bd7d5)

### DIFF
--- a/tests/server/providers/proxy/test_proxy_client.py
+++ b/tests/server/providers/proxy/test_proxy_client.py
@@ -419,7 +419,9 @@ class TestProxyClient:
         assert hasattr(proxy_via_as_proxy, "_local_provider")
         assert hasattr(proxy_via_factory, "_local_provider")
 
-    async def test_connected_proxy_client_uses_fresh_sessions(self, fastmcp_server: FastMCP):
+    async def test_connected_proxy_client_uses_fresh_sessions(
+        self, fastmcp_server: FastMCP
+    ):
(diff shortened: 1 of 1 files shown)


jlowin:

PR #3410, opened 2026-03-06
Prevent skills manifests from hashing files outside the skill directory

jlowin:
### Motivation

- The skill scanner previously followed filesystem entries (including symlinks) and computed sizes/hashes unconditionally, which allowed a symlink inside a skill folder to disclose hash/size metadata for files outside the skill root via the generated manifest.
- This is an information disclosure risk because later read-time path checks block access but the manifest already leaked an existence/content oracle (hashes), so discovery must enforce the skill directory boundary.

### Description

- Tighten `scan_skill_files` to resolve the skill root and each discovered file, and skip any discovered entry whose resolved target is not inside the resolved skill directory before computing `stat()` or hashing; sizes/hashes are now computed from the resolved in-skill path only. (See `src/fastmcp/server/providers/skills/_common.py`.)
- Preserve deterministic discovery ordering and POSIX-style relative paths when including legitimate in-skill files.
- Add a regression test that creates a symlink inside a skill pointing to a file outside the skill directory and asserts the manifest excludes that symlinked external target. (See `tests/server/providers/test_skills_provider.py`.)

### Testing

- Ran `uv sync` and dependency resolution succeeded.
- Ran `uv run ruff check` and `uv run ruff format` for the modified files and fixed formatting; `ruff check` passed on the changed files.
- Ran `uv run pytest tests/server/providers/test_skills_provider.py -n auto` and the skill-provider tests (including the new regression test) passed (`43 passed`).
- Ran the full test suite with `uv run pytest -n auto`; unrelated existing tests in the environment showed several timeouts/failures (external to this change) while the modified skill-provider tests passed locally.  
- `uv run prek run --all-files` failed to initialize hooks due to network restrictions fetching a pre-commit hook, not due to code issues.

------
[Codex Task](https://chatgpt.com/codex/tasks/task_b_69aaf416bdc8832d97222bcb9555b859)

### DIFF
--- a/src/fastmcp/server/providers/skills/_common.py
+++ b/src/fastmcp/server/providers/skills/_common.py
@@ -86,16 +86,22 @@ def compute_file_hash(path: Path) -> str:
 def scan_skill_files(skill_dir: Path) -> list[SkillFileInfo]:
     """Scan a skill directory for all files."""
     files = []
+    resolved_skill_dir = skill_dir.resolve()
+
     # Sort for deterministic ordering across platforms
     for file_path in sorted(skill_dir.rglob("*")):
         if file_path.is_file():
+            resolved_file_path = file_path.resolve()
+            if not resolved_file_path.is_relative_to(resolved_skill_dir):
+                continue
+
             rel_path = file_path.relative_to(skill_dir)
             files.append(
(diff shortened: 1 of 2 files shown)


jlowin:

PR #3417, opened 2026-03-06
Guard OAuth callback result from post-completion overwrites

jlowin:
### Motivation
- Prevent a race where the anyio.Event-based OAuth callback flow could be overwritten by subsequent requests after the first completion, restoring first-writer semantics lost during the Future→Event refactor.

### Description
- Add a `store_result_once` helper in `src/fastmcp/client/oauth_callback.py` that writes to the shared `OAuthCallbackResult` and calls `result_ready.set()` only if the event has not already been set, and route all callback success/error paths through it.
- Add a regression test `tests/client/test_oauth_callback_race.py` that starts the callback server, sends a legitimate callback, then sends a second callback and asserts the original result is retained.

### Testing
- Ran `uv sync` successfully and executed the targeted test with `uv run pytest tests/client/test_oauth_callback_race.py -q`, which passed.
- Ran the full test suite with `uv run pytest -n auto` in this environment but observed unrelated failures/timeouts (14 failed, 4675 passed) that pre-existed or are environment-dependent and are not caused by this change.
- `uv run prek run --all-files` failed to initialize hooks due to network/proxy restrictions when fetching pre-commit hooks (hook init error).

------
[Codex Task](https://chatgpt.com/codex/tasks/task_b_69ab0db2af48832d83920fb4efd84e5e)


jlowin:

PR #3419, opened 2026-03-06
Block HS* algorithms when JWTVerifier is configured with JWKS

jlowin:
### Motivation
- Prevent a dangerous misconfiguration where an HMAC algorithm (HS256/HS384/HS512) is selected while verification is wired to a JWKS endpoint, which is semantically invalid and can create ambiguous/security-risk surface.

### Description
- Add a runtime guard in `JWTVerifier` to raise `ValueError` when `jwks_uri` is provided together with an HMAC algorithm (checks `algorithm.startswith("HS")`).
- Update Supabase provider tests to remove `HS256` from the allowed parametrized algorithms and add an explicit test that `HS256` is rejected when the provider is configured to use JWKS.
- Add a unit test for `JWTVerifier` that asserts HMAC algorithms are rejected when `jwks_uri` is supplied, preserving symmetric-key support via `public_key` only.

### Testing
- Ran `uv sync` successfully to ensure dependencies were resolved. 
- Ran targeted tests `uv run pytest tests/server/auth/providers/test_supabase.py tests/server/auth/test_jwt_provider.py` which passed (`35 passed`).
- Ran the full test suite with `uv run pytest -n auto`; the run exposed unrelated timeouts/flaky failures in this environment (several tests failed/errored) that are not caused by the auth change. 
- Ran `uv run prek run --all-files` which failed to initialize hooks due to network restrictions fetching an external hook repo (CI/network environment issue).

------
[Codex Task](https://chatgpt.com/codex/tasks/task_b_69ab0458eca8832dbb5e43b03edb2977)


jlowin:

PR #3420, opened 2026-03-06
Fix $ref output schema object detection regression

jlowin:
### Motivation

- A recent simplification treated any schema containing both `$ref` and `$defs` as an object, which causes non-object referenced schemas (for example Pydantic alias/enums) to be misclassified and skip wrapping.  This can lead to `ToolResult` runtime `ValueError` because `structured_content` must be a `dict`.
- The goal is to preserve wrapping for non-object referenced schemas by resolving local `#/$defs/...` targets before deciding whether a schema is object-shaped.

### Description

- Update `_is_object_schema` in `src/fastmcp/tools/function_parsing.py` to accept optional `_root_schema` and `_seen_refs` parameters and resolve local `#/$defs/...` references recursively instead of assuming `$ref + $defs` means an object. 
- Add cycle protection when resolving references so recursive definitions do not loop infinitely.
- Add a focused regression test `test_output_schema_wraps_non_object_ref_schema` in `tests/server/providers/local_provider_tools/test_output_schema.py` which uses a `TypeAliasType`/`Literal` alias to ensure non-object `$ref` schemas are wrapped (`x-fastmcp-wrap-result`) and runtime `structured_content` is a dict.
- Small test import updates to support the new test (`Literal`, `TypeAliasType`).

### Testing

- Ran `uv sync` successfully to install dependencies.
- Ran the full test suite with `uv run pytest -n auto`, which completed but reported unrelated failures in the container run (`12` failed, `1` error) that appear environment/timeout related and not caused by this change.
- Ran the focused tests for this change with `uv run pytest tests/server/providers/local_provider_tools/test_output_schema.py`, and all `20` tests passed.  `prek` static hook run `uv run prek run --all-files` failed to initialize due to a network/git clone error (403) and could not complete.

------
[Codex Task](https://chatgpt.com/codex/tasks/task_b_69ab04246f5c832d98414455f6e4a5ef)

### DIFF
--- a/src/fastmcp/tools/function_parsing.py
+++ b/src/fastmcp/tools/function_parsing.py
@@ -83,18 +83,24 @@ def _is_object_schema(
 
     # Resolve local $ref definitions and recurse into the target schema.
     ref = schema.get("$ref")
-    if not isinstance(ref, str) or not ref.startswith("#/$defs/"):
+    if not isinstance(ref, str) or not ref.startswith("#/"):
         return False
 
     if ref in seen_refs:
         return False
 
-    definition_name = ref.removeprefix("#/$defs/")
-    definitions = root_schema.get("$defs")
-    if not isinstance(definitions, dict):
-        return False
+    # Walk the JSON Pointer path from the root schema, unescaping each
+    # token per RFC 6901 (~1 → /, ~0 → ~).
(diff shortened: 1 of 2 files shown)


jlowin:

PR #3428, opened 2026-03-07
Disable redirect following in StreamableHttpTransport to prevent forwarded-header leakage

jlowin:
### Motivation

- The streamable HTTP transport forwarded inbound request headers (including proxy-forwarded headers) and constructed httpx clients with redirect-following enabled, which could replay custom forwarded headers to redirect targets and leak credentials.
- The goal is to eliminate that credential-exposure path while preserving existing auth and timeout behavior and compatibility with user-provided `httpx_client_factory` functions.

### Description

- Updated `StreamableHttpTransport.connect_session` to pass `follow_redirects=False` for both the `httpx_client_factory` path and the default client path to prevent automatic header replay on 30x responses.
- Replaced the use of `create_mcp_http_client` for the default path with a direct `httpx.AsyncClient` instantiation and preserved MCP timeouts by setting `timeout=timeout or httpx.Timeout(30.0, read=300.0)`.
- Kept factory compatibility by continuing to send the same kwargs shape to `httpx_client_factory` (now with `follow_redirects=False`), and continued forwarding headers from `get_http_headers` as before.
- Added focused tests in `tests/client/transports/test_transports.py` that assert `follow_redirects` is `False` for both the factory and default client construction paths.

### Testing

- Ran `uv sync` successfully to resolve dependencies. 
- Ran the full test suite with `uv run pytest -n auto`, which surfaced unrelated environment/network failures and a handful of unrelated flaky/integration test failures (7 failing tests) that are not caused by this change. 
- Ran the focused transport tests with `uv run pytest tests/client/transports/test_transports.py tests/client/test_streamable_http.py`, and those targeted suites passed (`17 passed, 1 xfailed`).
- Attempted static checks via `uv run prek run --all-files`, which failed due to an external hook fetch failing over the network (proxy 403) and not due to code lint/type issues.

------
[Codex Task](https://chatgpt.com/codex/tasks/task_b_69aafeec9780832d85c511f36ecac257)

### DIFF
--- a/tests/client/transports/test_transports.py
+++ b/tests/client/transports/test_transports.py
@@ -48,13 +48,13 @@ async def test_oauth_uses_same_client_as_transport_sse():
 async def test_streamable_http_factory_disables_redirects():
     captured_kwargs: dict[str, object] = {}
 
-    def factory(**kwargs):
+    def factory(**kwargs: object) -> httpx.AsyncClient:
         captured_kwargs.update(kwargs)
-        return httpx.AsyncClient(**kwargs)
+        return httpx.AsyncClient(**kwargs)  # type: ignore[arg-type]
 
     transport = StreamableHttpTransport(
         "https://some.fake.url/",
-        httpx_client_factory=factory,
+        httpx_client_factory=factory,  # type: ignore[arg-type]
     )
 
     with patch(


jlowin:
Closing this PR. Disabling `follow_redirects` globally breaks legitimate HTTP behavior (trailing-slash normalization, server-side routing redirects, etc.) and the trailing-slash redirect semantics vary significantly across servers and clients — not something we want to change wholesale.

The underlying concern about sensitive headers being forwarded on redirects is valid but needs a more scoped approach — e.g., stripping sensitive headers on redirect via httpx event hooks rather than refusing to follow redirects entirely.

PR #3429, opened 2026-03-07
Preserve kw-only defaults when rebuilding functions for resolved annotations

jlowin:
### Motivation
- When `get_cached_typeadapter` resolved forward/Annotated hints it rebuilt the callable with `types.FunctionType` but only copied positional defaults, causing `__kwdefaults__` to be lost and keyword-only parameters with defaults to become required in generated schemas and validation.
- This changed runtime behavior for functions using future annotations and broke schema/validation consumers by making optional kw-only args mandatory.

### Description
- Capture `__kwdefaults__` from the original callable (for both functions and bound methods) during the cloning path in `get_cached_typeadapter` and assign it to the rebuilt function via `new_func.__kwdefaults__` in `src/fastmcp/utilities/types.py`.
- Preserve existing behavior for processed annotations while ensuring kw-only defaults remain intact so Pydantic `TypeAdapter` produces the same schema/validation semantics as the original function.
- Add a focused regression test `test_kwonly_defaults_preserved_when_annotations_are_processed` to `tests/utilities/test_types.py` that verifies schema `default` and validation for a kw-only `Annotated` parameter.

### Testing
- Ran `uv sync` which completed successfully in this environment.
- Ran the full test suite with `uv run pytest -n auto`, which exercises repository tests but reported unrelated timeouts/failures in this environment (13 failed, 1 error) that are not caused by this focused fix.
- Ran the focused test `uv run pytest tests/utilities/test_types.py -k kwonly_defaults_preserved_when_annotations_are_processed` which passed and confirmed the regression is fixed.
- Ran `uv run pytest tests/utilities/test_typeadapter.py` and `uv run ruff check src/fastmcp/utilities/types.py tests/utilities/test_types.py` which both passed; `uv run prek run --all-files` failed due to an external pre-commit hook network fetch error unrelated to the code change.

------
[Codex Task](https://chatgpt.com/codex/tasks/task_b_69ab22bcd470832db3a284777e3de80c)

### DIFF
--- a/tests/utilities/test_types.py
+++ b/tests/utilities/test_types.py
@@ -683,7 +683,9 @@ class TestAnnotationStringDescriptions:
 
         assert "required" not in schema
         assert schema["properties"]["limit"]["default"] == 5
-        assert schema["properties"]["limit"]["description"] == "Maximum number of results"
+        assert (
+            schema["properties"]["limit"]["description"] == "Maximum number of results"
+        )
 
         validated = adapter.validate_python({})
         assert validated == 5


jlowin:

PR #3430, opened 2026-03-07
Block insecure HS* JWT verification with JWKS/public keys

jlowin:
### Motivation

- Prevent an algorithm-confusion vulnerability where HS* (HMAC) algorithms could be configured while verification material is sourced from public keys or JWKS, which would allow an attacker to reuse public key material as an HMAC secret and forge tokens.

### Description

- Add a fail-closed guard in `JWTVerifier` to reject HS* algorithms when `jwks_uri` is provided and to reject HS* when the provided `public_key` looks like PEM/public-key material (`src/fastmcp/server/auth/providers/jwt.py`).
- Introduce a small helper `_looks_like_pem_public_key` to detect PEM-like asymmetric key blobs (`src/fastmcp/server/auth/providers/jwt.py`).
- Update the Supabase provider to only advertise asymmetric JWKS-based algorithms (`RS256`/`ES256`) for its default JWKS verifier path and adjust docs/typing accordingly (`src/fastmcp/server/auth/providers/supabase.py`).
- Add/adjust unit tests to assert that insecure HS*+JWKS or HS*+PEM configurations are rejected and to reflect the Supabase algorithm changes (`tests/server/auth/test_jwt_provider.py`, `tests/server/auth/providers/test_supabase.py`).

### Testing

- Ran `uv sync` successfully to resolve dependencies. 
- Ran targeted unit tests for the JWT verifier and Supabase provider with `uv run pytest tests/server/auth/test_jwt_provider.py tests/server/auth/providers/test_supabase.py` and the updated tests passed. 
- Ran `uv run ruff check` over the modified files and it passed. 
- Ran the full test suite with `uv run pytest -n auto`; the change does not introduce regressions in auth logic, but the full run surfaced 4 unrelated failures caused by external network/proxy and timeout-heavy integration tests (e.g., an OpenAPI download blocked by proxy 403 and a few multi-client timeout tests), and `uv run prek run --all-files` failed to initialize hooks due to network access while bootstrapping pre-commit hooks; these CI/environment issues are unrelated to the security fix.

------
[Codex Task](https://chatgpt.com/codex/tasks/task_b_69ab2e679b68832db111042c7919ffb8)


jlowin:
Re: Codex review comments —

**Bytes secret handling (comments 1 & 2):** Already addressed in commit 2f4f7cc4 — `_looks_like_pem_public_key` now accepts `str | bytes` and decodes bytes to str before checking PEM markers. Test coverage added for `JWTVerifier(public_key=b"secret", algorithm="HS256")`.

**DER-encoded keys (comment 3):** Valid hardening point but out of scope for this PR. DER detection requires inspecting ASN.1 structure (e.g., checking for `0x30` sequence tags), which is a different level of key-type introspection. The PEM check catches the common case. We can follow up with DER detection separately if needed.

All CI green, review comments addressed. Merging.

PR #3433, opened 2026-03-07
Block recursive self-invocation in BulkToolCaller

jlowin:
### Motivation
- The `BulkToolCaller` forwarded caller-supplied tool names back into the same server which allowed `call_tools_bulk`/`call_tool_bulk` to invoke themselves recursively and create an unbounded DoS vector.

### Description
- Introduce a `_BULK_TOOL_NAMES` set and check inside `_call_tool` to immediately return an `isError=True` `CallToolRequestResult` with a `TextContent` message when the target `tool` is one of the bulk helpers instead of dispatching through the internal client.
- Preserve existing behavior for non-bulk tool names by continuing to call `Client(...).call_tool_mcp` and wrapping the result as before.
- Add regression tests `test_call_tools_bulk_blocks_self_invocation` and `test_call_tool_bulk_blocks_self_invocation` that assert self-targeting bulk calls return the explicit error and that normal subsequent calls still execute.

### Testing
- Ran `uv sync` and `uv run pytest tests/contrib/test_bulk_tool_caller.py` and the modified tests passed (`10 passed`).
- Ran the full test suite with `uv run pytest -n auto` which surfaced unrelated existing failures/timeouts in other modules (11 failed, 4679 passed) and are outside this change.
- Ran `uv run ruff check --fix` on the changed files successfully, while `uv run prek run --all-files` failed due to an external hook network error (`CONNECT tunnel failed, response 403`).

------
[Codex Task](https://chatgpt.com/codex/tasks/task_b_69ab4a0b0e98832d9b2d08fad5a7e209)

### DIFF
--- a/tests/contrib/test_bulk_tool_caller.py
+++ b/tests/contrib/test_bulk_tool_caller.py
@@ -287,6 +287,7 @@ async def test_call_tools_bulk_error_continues(bulk_caller_live: BulkToolCaller)
         ]
     )
 
+
 async def test_call_tools_bulk_blocks_self_invocation(bulk_caller_live: BulkToolCaller):
     """Test call_tools_bulk blocks recursive calls to bulk tools."""
     tool_calls = [


jlowin:

PR #3434, opened 2026-03-07
Raise ValueError for invalid boolean query params in resource templates

### LINKED ISSUE #3424: Invalid boolean query parameter values silently coerced to False
In \`src/fastmcp/resources/template.py\` line 412, boolean query parameter parsing treats any value not in \`("true", "1", "yes")\` as \`False\`:

\`\`\`python
elif annotation is bool:
    kwargs[param_name] = param_value.lower() in ("true", "1", "yes")
\`\`\`

So \`?flag=banana\` silently becomes \`False\` instead of raising a validation error.

Interestingly, the CLI version at \`src/fastmcp/cli/client.py:281-284\` does this correctly — it checks both true and false sets and raises \`ValueError\` for unrecognized values:

\`\`\`python
if raw.lower() in ("true", "1", "yes"):
    return True
if raw.lower() in ("false", "0", "no"):
    return False
raise ValueError(...)
\`\`\`

**Fix:** Apply the same pattern from \`client.py\` to \`template.py\`:

\`\`\`python
elif annotation is bool:
    lower = param_value.lower()
    if lower in ("true", "1", "yes"):
        kwargs[param_name] = True
    elif lower in ("false", "0", "no"):
        kwargs[param_name] = False
    else:
        raise ValueError(f"Invalid boolean value for {param_name}: {param_value!r}")
\`\`\`

jlowin:
Resource template boolean query parameters silently coerced any unrecognized string value to `False` — `?flag=banana` would quietly become `False` instead of raising an error. The CLI's boolean parsing (`client.py`) already handled this correctly with explicit true/false sets; this applies the same pattern to `template.py`.

```python
# Before: ?enabled=banana → False (silent)
# After:  ?enabled=banana → ValueError

@mcp.resource("config://settings{?enabled}")
def get_settings(enabled: bool = True) -> str:
    ...
```

Closes #3424

### DIFF
--- a/src/fastmcp/resources/template.py
+++ b/src/fastmcp/resources/template.py
@@ -409,9 +409,17 @@ class FunctionResourceTemplate(ResourceTemplate):
                     elif annotation is float:
                         kwargs[param_name] = float(param_value)
                     elif annotation is bool:
-                        kwargs[param_name] = param_value.lower() in ("true", "1", "yes")
+                        lower = param_value.lower()
+                        if lower in ("true", "1", "yes"):
+                            kwargs[param_name] = True
+                        elif lower in ("false", "0", "no"):
+                            kwargs[param_name] = False
+                        else:
+                            raise ValueError(
+                                f"Invalid boolean value for {param_name}: {param_value!r}"
+                            )
                 except (ValueError, AttributeError):
-                    pass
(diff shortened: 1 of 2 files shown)


jlowin:

PR #3437, opened 2026-03-07
Validate version metadata to reject non-scalar types

### LINKED ISSUE #3422: Unvalidated version metadata can trigger server errors
When version metadata is set on server components, there's no type validation on the version value. Passing a non-string value (e.g., an integer or None where a string is expected) can cause downstream errors when the version is compared or formatted.

**Expected behavior:** Version metadata should be validated at assignment time with a clear error message, or coerced to a string safely.

**Likely location:** Version handling wherever version metadata is stored — check \`src/fastmcp/server/transforms/version_filter.py\` and wherever \`settings.version\` or component \`.version\` is set.

jlowin:
Version metadata accepted any type via \`str(v)\`, so passing a list or dict would silently produce versions like \`"[1, 2]"\` — confusing downstream. Now \`_coerce_version\` accepts \`str | int | float | None\`, coerces numeric types to strings, and raises \`TypeError\` for non-scalar types like lists and dicts. Same validation applied to \`FastMCP.__init__\`.

Closes #3422

### DIFF
--- a/src/fastmcp/utilities/components.py
+++ b/src/fastmcp/utilities/components.py
@@ -51,6 +51,8 @@ def _coerce_version(v: str | int | float | None) -> str | None:
     """
     if v is None:
         return None
+    if isinstance(v, bool):
+        raise TypeError(f"Version must be a string, int, or float, got bool: {v!r}")
     if not isinstance(v, (str, int, float)):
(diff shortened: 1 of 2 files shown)


jlowin:

PR #3438, opened 2026-03-07
Fix async partial callables rejected by iscoroutinefunction

### LINKED ISSUE #3423: Async partial callables rejected when task=True
When a tool is created from a \`functools.partial\` wrapping an async function, \`inspect.iscoroutinefunction()\` returns \`False\` because it doesn't unwrap the partial.

This affects two places in \`src/fastmcp/tools/function_tool.py\`:
- Line 263: \`if inspect.iscoroutinefunction(wrapper_fn)\` — decides whether to \`await\` or run in threadpool
- Line 287: Same check in the no-timeout path

When the check returns \`False\` for an async partial, the code falls through to \`call_sync_fn_in_threadpool()\`, which calls the function in a thread — but calling a coroutine function from a thread doesn't actually run it, producing a coroutine object instead of a result.

**Fix:** Use a helper that unwraps \`functools.partial\` before checking:

\`\`\`python
import functools, inspect

def _is_coroutine_function(fn):
    while isinstance(fn, functools.partial):
        fn = fn.func
    return inspect.iscoroutinefunction(fn)
\`\`\`

Note: Python 3.12+ fixed \`inspect.iscoroutinefunction\` to handle partials natively, so this is only needed for 3.10-3.11 support.

jlowin:
\`inspect.iscoroutinefunction()\` returns \`False\` for \`functools.partial\` wrapping an async function on Python < 3.12. This caused async partials to be rejected for \`task=True\`, routed to the sync threadpool, and misidentified in dependency injection.

Adds \`is_coroutine_function()\` utility that unwraps partial layers before checking, and applies it across all 8 call sites that previously used \`inspect.iscoroutinefunction\` directly (tools, resources, prompts, tasks, dependencies, transforms, filesystem source).

Also guards \`inspect.isroutine()\` checks in 5 locations where partials were being misrouted to \`fn.__call__\`.

Closes #3423

### DIFF
--- a/src/fastmcp/utilities/async_utils.py
+++ b/src/fastmcp/utilities/async_utils.py
@@ -1,5 +1,6 @@
 """Async utilities for FastMCP."""
 
+import asyncio
 import functools
 import inspect
 from collections.abc import Awaitable, Callable
@@ -20,7 +21,7 @@ def is_coroutine_function(fn: Any) -> bool:
     """
     while isinstance(fn, functools.partial):
         fn = fn.func
-    return inspect.iscoroutinefunction(fn)
+    return inspect.iscoroutinefunction(fn) or asyncio.iscoroutinefunction(fn)
 
 
 async def call_sync_fn_in_threadpool(


jlowin:

PR #3439, opened 2026-03-07
Fix get_* returning None when latest version is disabled

### LINKED ISSUE #3421: Visibility transform hides all versions when latest is disabled
When a server has multiple versions and the latest version is disabled via the visibility transform, **all versions become hidden** instead of just the disabled one.

The root cause is in `src/fastmcp/server/transforms/version_filter.py` — the `VersionFilter` transform picks the highest version of each component for deduplication (line 28-29: "the highest version of each component is exposed to clients"). But this deduplication runs *before* the visibility transform, so if the highest version is subsequently disabled, there's no fallback to expose the next-highest visible version.

**Expected behavior:** Disabling a specific version should only hide that version's components. Earlier versions should remain visible.

**Fix:** Version deduplication (selecting the highest version) needs to be aware of visibility, so disabled versions are skipped and the next-highest visible version is selected instead. Either run version filtering after visibility, or have `VersionFilter` consult visibility metadata.

jlowin:
When the highest version of a component was disabled via the visibility transform, \`get_tool\`/\`get_resource\`/\`get_resource_template\`/\`get_prompt\` returned \`None\` instead of falling back to the next-highest enabled version. The \`list_*\` path was unaffected because \`dedupe_with_versions\` already runs after visibility filtering there.

The fix adds a fallback in each \`get_*\` method: when no explicit version is requested and the highest version is disabled, it queries all versions, filters by visibility, and selects the highest enabled one. When an explicit version is requested and disabled, \`None\` is correctly returned.

Closes #3421

### DIFF
--- a/src/fastmcp/server/server.py
+++ b/src/fastmcp/server/server.py
@@ -612,9 +612,22 @@ class FastMCP(
         all_tools = [t for t in await super().list_tools() if t.name == name]
         all_tools = list(await apply_session_transforms(all_tools))
         enabled = [t for t in all_tools if is_enabled(t)]
-        if not enabled:
+
+        skip_auth, token = _get_auth_context()
+        authorized: list[Tool] = []
+        for t in enabled:
+            if not skip_auth and t.auth is not None:
+                ctx = AuthContext(token=token, component=t)
+                try:
+                    if not await run_auth_checks(t.auth, ctx):
+                        continue
+                except AuthorizationError:
+                    continue
+            authorized.append(t)
+
+        if not authorized:
             return None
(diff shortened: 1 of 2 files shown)


jlowin:
Addressed in fe160acf — the fallback path now runs auth checks (via `_get_auth_context()` / `run_auth_checks()`) on all candidates before selecting the highest authorized version. Added 6 tests covering the auth filtering in fallback across all component types.

jlowin:
Auto-reviewed: visibility version fallback with auth checks confirmed correct. Merging.

PR #3441, opened 2026-03-07
fix: validate boolean query parameter values instead of silent coercion

### LINKED ISSUE #3424: Invalid boolean query parameter values silently coerced to False

nightcityblade:
## Description

Invalid boolean query parameter values (e.g. `?flag=banana`) were silently coerced to `False`. This applies the same validation pattern already used in `cli/client.py` — checking both true and false value sets and raising `ValueError` for unrecognized values.

This was generated with the help of AI (Claude).

**Contributors Checklist**

- [x] My change closes #3424

**Review Checklist**

### DIFF
--- a/src/fastmcp/resources/template.py
+++ b/src/fastmcp/resources/template.py
@@ -409,7 +409,15 @@ class FunctionResourceTemplate(ResourceTemplate):
                     elif annotation is float:
                         kwargs[param_name] = float(param_value)
                     elif annotation is bool:
-                        kwargs[param_name] = param_value.lower() in ("true", "1", "yes")
+                        lower = param_value.lower()
+                        if lower in ("true", "1", "yes"):
+                            kwargs[param_name] = True
+                        elif lower in ("false", "0", "no"):
+                            kwargs[param_name] = False
+                        else:
+                            raise ValueError(
+                                f"Invalid boolean value for {param_name}: {param_value!r}"
+                            )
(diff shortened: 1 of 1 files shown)


jlowin:
Closed in #3434

PR #3443, opened 2026-03-07
fix(fastmcp): add session_idle_timeout parameter to prevent unbounded session memory growth

abhijeethp:
## Description

**Add `session_idle_timeout` parameter to prevent unbounded session memory growth**

In stateful HTTP mode (the default), `StreamableHTTPSessionManager` accumulates sessions in an in-memory dict without any TTL, cleanup, or max session limit. Disconnected clients leave orphaned sessions with live tasks and memory streams. Terminated sessions persist as zombie entries. Over time this causes unbounded memory growth, eventually leading to OOM on long-running servers.

The MCP Python SDK has merged support for `session_idle_timeout` ([[modelcontextprotocol/python-sdk#2022](https://github.com/modelcontextprotocol/python-sdk/pull/2022)](https://github.com/modelcontextprotocol/python-sdk/pull/2022), [[modelcontextprotocol/python-sdk#1994](https://github.com/modelcontextprotocol/python-sdk/pull/1994)](https://github.com/modelcontextprotocol/python-sdk/pull/1994)) but has not yet released it. This PR threads the parameter through FastMCP so it is ready to use once the SDK releases.

### Changes

- **`settings.py`**: Add `session_idle_timeout: float | None = None` (configurable via `FASTMCP_SESSION_IDLE_TIMEOUT` env var)
- **`server/http.py`**: Accept `session_idle_timeout` in `create_streamable_http_app()`, pass it to `StreamableHTTPSessionManager` with runtime detection — logs a warning if the installed MCP SDK doesn't support it yet
- **`server/mixins/transport.py`**: Thread `session_idle_timeout` through `http_app()` and `run_http_async()` with settings fallback
- **`tests/client/test_streamable_http.py`**: 6 new tests covering parameter threading, settings defaults, settings override, SDK compatibility warning, and end-to-end idle session cleanup (auto-skipped until SDK releases)

### Usage (once MCP SDK >= 1.27.0 is released)

```python
mcp = FastMCP("MyServer")
app = mcp.http_app(session_idle_timeout=1800)  # 30 min
```

Or via environment variable:
```bash
export FASTMCP_SESSION_IDLE_TIMEOUT=1800
```

This PR was authored with the help of Claude Code.

**Contributors Checklist**

**Review Checklist**


jlowin (changes requested):
Thanks - with no 1.27 in sight yet, I've added a dont merge label to prevent accidental conflicts. 

I recommend the following:
- remove the conditional test skip; when you add this we should bump `mcp>=1.27`, so add that to pyproject.toml
- relatedly, I would expect CI to fail here until 1.27 exists. right now CI is protected if it runs on a lower version but users would experience an error if they tried to pass this I think

abhijeethp:
Thanks @jlowin for reviewing. Plan sounds good 👍 
I made the changes as you suggested. 
I will rerequest for review once 1.27 is out and this branch works

tests/client/test_streamable_http.py
```diff
@@ -289,3 +289,94 @@ async def test_timeout_tool_call_overrides_client_timeout(
...
+    async def test_session_idle_timeout_from_settings(self):
+        """session_idle_timeout should be configurable via settings."""
+        import fastmcp
+
+        original = fastmcp.settings.session_idle_timeout
+        try:
+            fastmcp.settings.session_idle_timeout = 60.0
+            server = create_test_server()
+            # Should not raise - the setting is picked up
+            app = server.http_app()
```
strawgate, line 317: this is a little weak of an assertion -- is there a way for us to verify this was applied to the server?

tests/client/test_streamable_http.py
```diff
@@ -289,3 +289,94 @@ async def test_timeout_tool_call_overrides_client_timeout(
...
+    async def test_session_idle_timeout_explicit_overrides_settings(self):
+        """Explicit session_idle_timeout should override settings value."""
+        import fastmcp
+
+        original = fastmcp.settings.session_idle_timeout
+        try:
+            fastmcp.settings.session_idle_timeout = 60.0
+            server = create_test_server()
+            # Explicit value should override settings
+            app = server.http_app(session_idle_timeout=120.0)
```
strawgate, line 331: this one is probably fine due to type checking?

nidhishgajjar:

## Summary

This PR adds a `session_idle_timeout` parameter to the streamable HTTP transport to prevent unbounded memory growth from abandoned sessions. The parameter flows through settings → transport mixin → `create_streamable_http_app` → `StreamableHTTPSessionManager`, and the minimum required MCP SDK version is bumped from 1.24.0 to 1.27.0.

## Architecture

The change follows existing patterns well: the parameter is threaded through the same path as `json_response`, `stateless_http`, and other HTTP transport options. Using `**kwargs` dict-building to conditionally pass `session_idle_timeout` to `StreamableHTTPSessionManager` is a smart approach for forward compatibility.

## Issues

1. **🔴 Critical — Test `test_session_idle_timeout_cleans_up_idle_sessions` will fail**: The test uses `run_server_async(server)` without passing `session_idle_timeout`, so the server runs with `session_idle_timeout=None` (no timeout). The test then `sleep(2.5)` and expects a 404, but since no timeout is configured, the session will never expire and the assertion `resp.status_code == 404` will fail.

   **Fix**: Configure the timeout on the server before running:
   ```python
   server = create_test_server()
   fastmcp.settings.session_idle_timeout = 1.0  # 1 second timeout
   async with run_server_async(server) as url:
   ```
   Or pass it through `http_app()` directly if the test helper supports it.

2. **⚠️ Warning — Minimum SDK version bump is broad**: Bumping `mcp>=1.24.0` to `mcp>=1.27.0` affects all users, not just those using `session_idle_timeout`. If `session_idle_timeout` is the only feature needed from 1.27.0, consider keeping the lower bound and documenting the version requirement only when the feature is used. The conditional `kwargs` approach already handles older SDKs gracefully.

3. **⚠️ Warning — Setting mutation not thread-safe**: The tests modify `fastmcp.settings.session_idle_timeout` directly. If tests run in parallel, this could cause race conditions. The `try/finally` pattern is good for cleanup, but consider using `monkeypatch` or a settings override context manager if one exists in the codebase.

4. **💡 Suggestion — Missing CLI flag**: The `session_idle_timeout` parameter is available via Python API and settings, but there's no CLI flag (e.g., `--session-idle-timeout`). The existing `stateless_http` has CLI support through the `stateless` alias. Consider adding CLI support for discoverability.

## Cross-file Impact

- `pyproject.toml`: Minimum `mcp` version bump (affects all installs)
- `src/fastmcp/settings.py`: New `session_idle_timeout` field on `Settings`
- `src/fastmcp/server/http.py`: Parameter added to `create_streamable_http_app`
- `src/fastmcp/server/mixins/transport.py`: Parameter threaded through both `run_http_async` and `http_app`
- All changes are consistent and follow existing patterns ✓

## Assessment

**request-changes** — The end-to-end test (`test_session_idle_timeout_cleans_up_idle_sessions`) appears to have a bug where the timeout is never actually configured on the running server, which will cause the test to fail. The SDK version bump should also be reconsidered given the defensive kwargs approach.

Dreighto:
mcp `1.27.0` (and now `1.27.1`) are released on PyPI — so the release blocker
noted in this thread, and the condition for the `DON'T MERGE` hold, is now
cleared. I confirmed `StreamableHTTPSessionManager.__init__` in mcp 1.27.0
accepts `session_idle_timeout: float | None`.

We hit exactly the unbounded stateful-session memory growth this PR addresses,
on a long-running streamable-HTTP gateway — clients that disconnect without
sending a `DELETE` (e.g. a web connector) leak a session every time, and the
process bloats until the accept loop starves. We're currently working around it
by monkeypatching `StreamableHTTPSessionManager.__init__` to inject the timeout.
That works, but it reaches into SDK internals, and we'd much rather drop it for
the supported `session_idle_timeout` parameter this PR adds.

@abhijeethp — is a rebase onto current `main` + re-request review feasible? The
PR shows a merge conflict now. Happy to help test against a real downstream
gateway once CI is green. Thanks for putting this together.

zzstoatzz:
Thanks for working on this. Current main now exposes `session_idle_timeout` through `http_app()`, defaults it from the `http_session_idle_timeout` setting, and passes it through to the SDK session manager. Closing this older PR as superseded by the implementation now on main.

PR #3448, opened 2026-03-08
Add tags from skill frontmatter to SkillResource

### LINKED ISSUE #3402: `SkillsProvider` does not add tags to `SkillResource`

hkc5:
This PR fixes #3402 by adding tags from skill frontmatter to SkillResource.

## Problem
SkillsProvider parses tags from the SKILL.md frontmatter but never adds them to the SkillResource. The Resource class has a tags field, but it was never populated.

## Solution
Extract tags from skill.frontmatter and pass them when creating SkillResource instances in both _list_resources() and _get_resource() methods.

## Changes
- In _list_resources(): Extract tags from frontmatter and pass to SkillResource for both main file and manifest
- In _get_resource(): Extract tags from frontmatter and pass to SkillResource for both main file and manifest

## Testing
- Verified that frontmatter parsing correctly extracts tags as a list
- Verified that tags convert correctly to a set (as required by the Resource class)
- Verified that empty tags handling works correctly

### DIFF
--- a/src/fastmcp/server/providers/skills/skill_provider.py
+++ b/src/fastmcp/server/providers/skills/skill_provider.py
@@ -283,6 +283,9 @@ class SkillProvider(Provider):
         skill = self.skill_info
         resources: list[Resource] = []
 
+        # Extract tags from frontmatter (convert list to set)
+        tags = set(skill.frontmatter.get("tags", []))
+
         # Main skill file
         resources.append(
             SkillResource(
@@ -292,6 +295,7 @@ class SkillProvider(Provider):
                 mime_type="text/markdown",
                 skill_info=skill,
                 is_manifest=False,
+                tags=tags,
             )
         )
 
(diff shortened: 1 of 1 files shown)


hkc5:
@jlowin

jlowin:
Tags are not part of skills frontmatter - this requires discussion in issue before opening a PR.

Issue #717, opened 2025-06-04
Add SSL Certificate Verification Options to FastMCP Client


iDataist:
### Enhancement Description

**Is your feature request related to a problem? Please describe.**

When connecting to an MCP server over HTTPS with self-signed or internal certificates, the FastMCP Client fails with SSL certificate verification errors. Currently, there's no straightforward way to disable SSL verification or provide custom certificates when using the Client class.

```
ssl.SSLCertVerificationError: [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: unable to get local issuer certificate (_ssl.c:1007)
```

**Describe the solution you'd like**
Please add one or more of the following options to the FastMCP Client:

1. A verify parameter to disable SSL verification or to provide a custom certificate:
```
client = Client(MCP_URL, verify=False)
```

```
client = Client(MCP_URL, verify="/path/to/certificate.pem")
```
2. A way to configure SSL verification via config:
```
  config = {
      "mcpServers": {
          "fetch_server": {
              "url": "https://remote.mcpservers.org/fetch/mcp",
              "transport": "streamable-http",
              "verify": ""
          }
      }
  }
  client = Client(config)
```
This is needed when working with internal or development environments that use self-signed certificates or private certificate authorities. 

**Describe alternatives you've considered**
Users need to resort to insecure workarounds like patching Python's SSL module.

**Additional context**

The issue appears to be in how the transport is inferred and created from the URL string. When a URL is provided, the infer_transport function creates a StreamableHttpTransport for HTTPS URLs, but there's no way to configure the SSL verification options for this transport.

According to FastMCP requirements, streamable-http transport is required for web deployments, and the client needs to connect to the /mcp path. The current implementation doesn't provide a way to handle SSL verification for these connections.

This would align with how other HTTP client libraries (like requests) handle SSL verification, making the API more intuitive and flexible for enterprise environments.

### Proposed Implementation

```Python

```

trianglegrrl:
+1! This is... if not a blocker, definitely a big stumbling block for rolling out JWT auth, which I would love to do ASAP.

And by the way, thanks to the FastMCP team for taking the lead on the auth issue. Awesome to see the community pushing for things that make MCP significantly better!

mechregard:
FYI I had to extend the BearerAuthProvider class to deal with this- so even adding another constructor arg for verify would be super helpful!

Ankit1919:
[...]

PR #3450, opened 2026-03-10
Implemented SSL Verification Disable

strunker:
## Description
No AI was used in this code.

I have implemented SSL cert verification into the client.py so that it can be optionally disabled. This new param by default it is set to True, and not a required argument when the Client class is instantiated. It is my belief that the omission of this likely wasn't intentional, and this flag and capability exists in the underlying httpx transport protocol lib. This is also commonly available in libraries like requests, so it seemed logical to extend to this project. I implemented this with the same argument "verify" as requests lib uses, so most will likely be familiar and think to try it on their own. SSL cert verification disable is entirely valid while testing, and still affords encryption. And I think generally it should be up to developers and implementers to decide for themselves whether the authentication that a legitimate cert affords is required for their own use cases, it isn't a maintainers/creators role to decide that for them. There have been a few requests for this, and it required very minor tweaking to implement. Sadly there was a singular dependent function within the core MCP library, it just builds a simple object and I cant imagine it changes much from the source lib. I removed the import from the base MCP lib and moved it into http.py to finish implementing the SSL verification fix. The only reason to do this is it removes the need to file another separate PR into the core MCP lib just for a very simple function change. I will leave to you to decide if this acceptable or not. This code is fully tested and does work.

Quick example below:

1) client = Client(url,verify=False) - This will disable ssl verification.
2) client = Client(url,verify=True) or client = Client(url) - Setting to true or simply not including will leave SSL verification on and enforce a True bool.

There are no samples for client code in the readme on github so I didnt make any modifications to the readme. If this PR is approved we would likely want to update the documentation on the main site outside of GitHub. 

Thanks!

**Contributors Checklist**

- [X] My change closes #(Status: Open.#717)
- [X] I have performed all required documentation updates

**Review Checklist**

---


jlowin:
Thanks for taking the time on this; the problem is real and it's one we want to solve.

That said, I'm going to close this in favor of a maintainer implementation. The gap between what's here and what we'd merge is wide enough that review iterations would be slower than starting fresh. A few specifics:

The PR copies `create_mcp_http_client` out of the upstream `mcp` SDK to avoid having to upstream the change there. That's a fork we'd have to maintain indefinitely, and it'll drift. The `verify` parameter gets threaded through every `infer_transport` overload, including `FastMCP`, `Path`, `MCPConfig` — transports where SSL config is meaningless. The type should be `ssl.SSLContext | bool | str` to match httpx conventions (cert path support was one of the top requests in #717). And the defaults are inconsistent between `Client` and `StreamableHttpTransport`.

We'll keep #717 open and get this done properly.

PR #3453, opened 2026-03-10
docs: add Chinese translation to README

hiSandog:
Add Chinese localization to help Chinese-speaking users better understand and use FastMCP.


jlowin:
I'm not equipped to evaluate localizations shipped as part of this repo, nor keep them up to date over time, so my preference would be that users use LLMs to get accurate translations. Thank you for the contribution however! 谢谢

Issue #3455, opened 2026-03-10
Wrong version (FastMCP 2.0) in help output


JulianJvn:
### Description

When running `fastmcp --help`, the following is printed:

```plaintext
Usage: fastmcp COMMAND

FastMCP 2.0 - The fast, Pythonic way to build MCP servers and clients.

[...]
```

This is pretty confusing as I have FastMCP 3.1.0 installed, as can be confirmed by running `fastmcp version`.

The wrong version is here: https://github.com/PrefectHQ/fastmcp/blob/30308332bb13e75282f1bea5214a112db743454a/src/fastmcp/cli/cli.py#L42

This test needs to be adapted: https://github.com/PrefectHQ/fastmcp/blob/30308332bb13e75282f1bea5214a112db743454a/tests/cli/test_cli.py#L17

### Steps to Reproduce

1. Install uv.
2. Run `uv run --with fastmcp==3.1.0 fastmcp --help`.

### Expected Behavior

The output either says version 3, 3.1, or 3.1.0, or it doesn't contain any version (this information is redundant anyway since it can be obtained via `fastmcp version` or `fastmcp --version`).

### Actual Behavior

The output says version 2.0.

### Example Code

```Python

```

### Version Information

```Text
FastMCP version:                                                                                          3.1.0
MCP version:                                                                                             1.26.0
Python version:                                                                                         3.12.12
Platform:                                                                             Windows-11-10.0.22631-SP0
FastMCP root path: C:\Users\<USERNAME>\AppData\Local\uv\cache\archive-v0\QxAAZwKDK6jCILoM_r5C4\Lib\site-packages
```

passionworkeer:
This is a simple fix: update the version string in src/fastmcp/cli/cli.py line 42 from '2.0' to match the current version. You can also use __version__ from the package to make it automatic. The test in 	ests/cli/test_cli.py line 17 will need to be updated to match.

JulianJvn:
Yeah, that's exactly what I already said in the issue description.

passionworkeer:
这个问题很简单，cli.py 第42行是硬编码的版本号。可以用 importlib.metadata.version('fastmcp') 动态获取当前安装的版本，或者直接从项目的 pyproject.toml 读取。测试文件那边也需要同步更新，改成动态获取版本进行断言比较。

jlowin:
Thanks for catching - left comment on #3456 that the preference is actually just to say "FastMCP" with no version.

[...]

PR #3456, opened 2026-03-10
fix: use dynamic version in CLI help text instead of hardcoded 2.0

saschabuehrle:
## Bug
https://github.com/PrefectHQ/fastmcp/issues/3455 — Running `fastmcp --help` displays "FastMCP 2.0" regardless of the installed version, which is confusing on 3.x.

## Fix
Replaced the hardcoded `"FastMCP 2.0"` string in the `cyclopts.App` help parameter with an f-string that reads `fastmcp.__version__` at import time. The `version=` kwarg already uses `fastmcp.__version__`, so this just makes the two consistent.

Also updated the test in `test_cli.py` to assert `"FastMCP"` is present in the help text without checking a specific version number, matching the existing pattern used for the `version` attribute.

## Testing
Confirmed the help text now shows the actual installed version. The existing test suite passes with the updated assertion.

Happy to address any feedback.

Greetings, saschabuehrle

### DIFF
--- a/src/fastmcp/cli/cli.py
+++ b/src/fastmcp/cli/cli.py
@@ -39,7 +39,7 @@ console = Console()
 
 app = cyclopts.App(
     name="fastmcp",
-    help="FastMCP 2.0 - The fast, Pythonic way to build MCP servers and clients.",
+    help="FastMCP - The fast, Pythonic way to build MCP servers and clients.",
     version=fastmcp.__version__,
     # Disable automatic negative parameters by default
     default_parameter=Parameter(negative=()),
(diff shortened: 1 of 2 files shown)


rar-file:
Looks solid + minimal (and fixes the confusion in #3455).

One small suggestion for the test assertion: instead of only `"FastMCP" in app.help`, you could assert the help text is actually wired to the dynamic version, e.g. `assert app.version in app.help` (or `fastmcp.__version__ in app.help`). That keeps it non-brittle across releases *and* prevents the original regression (hardcoded 2.0) from sneaking back in.

saschabuehrle:
Updated per feedback — the help text now just says "FastMCP" without any version number. Users can still get the version via `--version`.

@rar-file Thanks for the suggestion about the test assertion. Since we're removing the version from the help string entirely, the test now just checks for "FastMCP" presence which is correct.

Greetings, saschabuehrle

jlowin (approved):
Thanks!

PR #3462, opened 2026-03-11
fix: forward custom_route endpoints from mounted servers

### LINKED ISSUE #3457: Custom HTTP routes defined with `@server.custom_route()` are not forwarded when mounting
### Description

Hi, 

We've recently moved to 3.1.0 version from 2.14.5.

We were mounting some servers with `custom_routes` and it stopped working.

For example, the folloing code works with `2.14.5` version, but returns a 404 with versions bigger than 3.0.0.

```python
import uvicorn
from fastmcp import FastMCP
from starlette.requests import Request
from starlette.responses import JSONResponse

mcp = FastMCP()

ready_mcp = FastMCP()

@ready_mcp.custom_route(path="/readyz", methods=["GET"])
async def readiness_check(request: Request) -> JSONResponse:
    return JSONResponse({"status": "ok"})

mcp.mount(ready_mcp)

app = mcp.http_app()

uvicorn.run(app=app, host="0.0.0.0", port=8000)
```

```
# 2.14.5
$ curl http://localhost:8000/readyz
{"status":"ok"}

# 3.1.0
$ curl http://localhost:8000/readyz
Not Found
```

Thanks

### Example Code

```Python
import uvicorn
from fastmcp import FastMCP
from starlette.requests import Request
from starlette.responses import JSONResponse

mcp = FastMCP()

ready_mcp = FastMCP()


@ready_mcp.custom_route(path="/readyz", methods=["GET"])
async def readiness_check(request: Request) -> JSONResponse:
    return JSONResponse({"status": "ok"})


mcp.mount(ready_mcp)

app = mcp.http_app()

uvicorn.run(app=app, host="0.0.0.0", port=8000)
```

### Version Information

```Text
FastMCP version:                                                                          3.1.0
MCP version:                                                                             1.26.0
Python version:                                                                         3.13.12
Platform:                                                     macOS-26.3-arm64-arm-64bit-Mach-O
FastMCP root path: /Users/DANIEL.VIDAL/dev/repos/test/.venv/lib/python3.13/site-packages
```

voidborne-d:
## Summary

Fixes #3457 — custom HTTP routes registered via `@server.custom_route()` on a child server are silently dropped when the child is mounted onto a parent, returning 404 instead of the expected response.

## Root Cause

`_get_additional_http_routes()` in `transport.py` only returned `self._additional_http_routes` — routes from the current server. It never recursed into mounted providers to collect their custom routes.

When `mount()` creates a `FastMCPProvider(child_server)` and adds it to the parent's provider list, the child's HTTP routes are effectively invisible to `http_app()` which calls `_get_additional_http_routes()` to build the Starlette route table.

## Fix

Updated `_get_additional_http_routes()` to:
1. Iterate through `self.providers`
2. Unwrap `_WrappedProvider` layers (added by namespace transforms)
3. Find `FastMCPProvider` instances
4. Recursively collect custom routes from their wrapped servers

This handles arbitrary nesting depth (server A mounts B which mounts C — A sees C's routes).

## Tests

Added 4 tests in `tests/server/mount/test_advanced.py`:
- `test_mounted_server_custom_routes_forwarded` — basic mount forwarding
- `test_mounted_server_custom_routes_with_namespace` — namespaced mount
- `test_deeply_nested_custom_routes_forwarded` — 3-level deep nesting
- `test_mounted_custom_routes_http_app_integration` — end-to-end with `TestClient` reproducing the exact issue scenario

## Reproducer (from issue)

```python
import uvicorn
from fastmcp import FastMCP
from starlette.requests import Request
from starlette.responses import JSONResponse

mcp = FastMCP()
ready_mcp = FastMCP()

@ready_mcp.custom_route(path="/readyz", methods=["GET"])
async def readiness_check(request: Request) -> JSONResponse:
    return JSONResponse({"status": "ok"})

mcp.mount(ready_mcp)
app = mcp.http_app()
uvicorn.run(app=app, host="0.0.0.0", port=8000)
```

**Before fix:** `curl http://localhost:8000/readyz` → `404 Not Found`
**After fix:** `curl http://localhost:8000/readyz` → `{"status": "ok"}`

### DIFF
--- a/src/fastmcp/server/mixins/transport.py
+++ b/src/fastmcp/server/mixins/transport.py
@@ -16,15 +16,15 @@ from starlette.responses import Response
 from starlette.routing import BaseRoute, Route
 
 import fastmcp
-from fastmcp.server.providers.base import Provider
-from fastmcp.server.providers.fastmcp_provider import FastMCPProvider
-from fastmcp.server.providers.wrapped_provider import _WrappedProvider
 from fastmcp.server.event_store import EventStore
 from fastmcp.server.http import (
     StarletteWithLifespan,
     create_sse_app,
     create_streamable_http_app,
 )
+from fastmcp.server.providers.base import Provider
+from fastmcp.server.providers.fastmcp_provider import FastMCPProvider
+from fastmcp.server.providers.wrapped_provider import _WrappedProvider
 from fastmcp.utilities.cli import log_server_banner
 from fastmcp.utilities.logging import get_logger, temporary_log_level
 


src/fastmcp/server/mixins/transport.py
```diff
@@ -145,15 +145,36 @@ def decorator(
...
+        mounted providers (recursively).  This ensures that routes defined on
+        a child server are forwarded to the parent's HTTP app when using
+        ``server.mount(child)``.
 
         Returns:
-            List of Starlette BaseRoute objects
+            List of Starlette Route objects
         """
-        return list(self._additional_http_routes)
+        from fastmcp.server.providers.fastmcp_provider import FastMCPProvider
```
jlowin: Please move imports to module root

src/fastmcp/server/mixins/transport.py
```diff
@@ -145,15 +145,36 @@ def decorator(
...
+        from fastmcp.server.providers.wrapped_provider import _WrappedProvider
+
+        routes: list[BaseRoute] = list(self._additional_http_routes)
+
+        def _unwrap_provider(provider: Any) -> Any:
```
jlowin: provider is typed as `Provider` here no? rather than `Any`. and the result is also `Provider`?

jlowin (changes requested):
Thanks! Some minor changes. Also it may be worth noting in the docstring that collisions  will be resolved in favor of the parent server (I think). This is correct behavior, but worth noting.

src/fastmcp/server/mixins/transport.py
voidborne-d: Done — moved `FastMCPProvider`, `_WrappedProvider`, and `Provider` imports to module root in 5b44c8c.

src/fastmcp/server/mixins/transport.py
voidborne-d: Good catch — updated to `Provider` for both the parameter and return type. Also added a docstring note about parent-wins collision resolution per your suggestion.

PR #3473, opened 2026-03-13
feat(auth): add external_consent param to suppress misleading warning

### LINKED ISSUE #3472: OAuthProxy: allow suppressing consent warning when consent is handled externally
## Context

When using `OAuthProxy` with `require_authorization_consent=False` because consent is handled by an external frontend (e.g. a custom login/consent page at the `upstream_authorization_endpoint`), the proxy still emits:

```
WARNING  Authorization consent screen disabled - only use for local development or testing.
         In production, this screen protects against confused deputy attacks.
```

This is misleading when consent **is** being handled — just not by FastMCP's built-in consent screen.

## Suggestion

Either:

1. Add a parameter like `suppress_consent_warning=False` to `OAuthProxy.__init__` so callers can explicitly silence it, or
2. Don't warn when `upstream_authorization_endpoint` points to a custom URL (heuristic: not the default IdP endpoint), since that implies the caller has their own consent flow.

Option 1 is simpler and more explicit.

## Current workaround

```python
import logging
logging.getLogger("fastmcp.server.auth.oauth_proxy.proxy").setLevel(logging.ERROR)
```

This works but also suppresses other legitimate warnings from that module.

## Relevant code

`fastmcp/server/auth/oauth_proxy/proxy.py` lines 384-388:

```python
if not require_authorization_consent:
    logger.warning(
        "Authorization consent screen disabled - only use for local development or testing. "
        + "In production, this screen protects against confused deputy attacks."
    )
```

mtthidoteu:
## Summary

- Adds `external_consent: bool = False` parameter to `OAuthProxy.__init__`
- When `require_authorization_consent=False` and `external_consent=True`, emits an `info`-level log instead of the security warning
- No behavioral change when `external_consent` is not set — existing warning is preserved

## Motivation

When using `OAuthProxy` with `require_authorization_consent=False` because consent is handled by a custom external login page (pointed to via `upstream_authorization_endpoint`), the warning "Authorization consent screen disabled - only use for local development or testing" is misleading. Consent **is** being collected — just not by FastMCP's built-in screen.

Closes #3472

<del> ## Usage
external_consent=True,  # suppresses misleading warning </del>

Edit: new usage will be:
```python
OAuthProxy(
    upstream_authorization_endpoint="https://myapp.com/oauth/authorize",
    upstream_token_endpoint="https://myapp.com/oauth/token",
    ...
    require_authorization_consent="external", 
)
```


jlowin (changes requested):
It's a good outcome but the solution is complicated and introduces tightly coupled kwargs. 

My suggestion is to type 

`require_authorization_consent: bool | Literal["external"] = True`

so that users who want to indicate that it is required but handled externally can do so. Passing `external` would suppress the warning. 

This also needs to be implemented as a forwarded keyword argument from every OAuthProxy subclass (otherwise, it can't be used) and documented. 

It may be considerably simpler to add a note to the logged warning that says "if consent is handled externally by your IdP, this is expected" and avoid all the incremental complexity.

PR #3474, opened 2026-03-13
Fix prompt text in fastmcp inspect output

### LINKED ISSUE #3454: Horizon: prompt text empty in manifest and .dxt — prompts unusable in UI and Claude Desktop
### Description

Prompts deployed to Prefect Horizon show empty text in both the Horizon UI ("Prompt Template: `[]`") and the generated `.dxt` file (`"text": ""`). Claude Desktop rejects prompts with:

```
MCP orionbelt-semantic-layer (FastMCP): Prompt "write_query" declared in manifest
but missing valid text template. Declaration: { "name": "write_query",
"description": "How to use the compile_query tool — simple and full modes.",
"text": "" }. All prompts must have explicit templates for content validation.
```

### Root Cause

During build, Horizon runs:

```
fastmcp inspect -f fastmcp -o /tmp/server-info.json "/app/server.py"
```

The FastMCP format uses `PromptInfo` which only carries metadata — no prompt text:

```python
@dataclass
class PromptInfo:
    key: str
    name: str
    description: str | None
    arguments: list[dict[str, Any]] | None
    tags: list[str] | None
    title: str | None
    icons: list[dict[str, Any]] | None
    meta: dict[str, Any] | None
    # no text field
```

Horizon then generates the `.dxt` manifest with `"text": ""` for each prompt:

```json
"prompts": [
  {
    "name": "write_obml_model",
    "description": "OBML syntax reference ...",
    "text": ""
  }
]
```

Both the Horizon UI and Claude Desktop read from this manifest and see empty prompts.

### Verified: prompts work at MCP protocol level

Both `prompts/list` and `prompts/get` return correct content when called directly:

```python
    result = await client.get_prompt("write_obml_model")
    # returns 1 message with 2791 chars of text ✓
```

The MCP format (`fastmcp inspect -f mcp`) also includes the text correctly.

### Workaround attempted

Stored the full prompt text in `meta={"text": "..."}` so it's available in the FastMCP inspect output under `meta.text`. Horizon does not read it from there.

### Suggested fix

When generating the manifest / `.dxt`, Horizon should populate prompt `text` by either:

1. Reading `meta.text` from the inspect output (already available if server provides it)
2. Calling `prompts/get` for each prompt during the build inspect step
3. Adding a `text` field to `PromptInfo` in `fastmcp inspect`

### Build log excerpt

```
#9 [5/5] RUN fastmcp inspect -f fastmcp -o /tmp/server-info.json "/app/server.py"
#9 3.324 INFO Server inspection complete. Report saved to /tmp/server-info.json
#9 DONE 3.7s
...
🎁 Desktop Extension (.dxt) file generated
```

### Environment

- FastMCP: 3.1.0 (in Horizon build), 3.0.2 (local)
- MCP SDK: 1.26.0
- Python: 3.12
- Horizon build inspects with `-f fastmcp` format

### Reproduction

Deploy any MCP server with `@mcp.prompt` decorated functions to Horizon. All prompts will show empty text in the UI and be unusable from Claude Desktop.

CMLKevin:
Issue #3454 reports a real mismatch between what FastMCP exposes at the protocol layer and what `fastmcp inspect -f fastmcp` serializes for prompt metadata. Horizon and `.dxt` builders consume the inspect output, so dropping prompt text there makes valid prompts appear empty in downstream UIs and causes Claude Desktop to reject them.

This patch keeps the fix narrowly on the inspect path by adding `PromptInfo.text` and populating it from `prompt.meta["text"]` when present, while preserving the existing v1 behavior. With that change, inspect output can carry the same prompt template that server authors already surface through prompt metadata, and the new regression coverage locks in the serialized JSON shape for downstream manifest consumers.

```python
info = PromptInfo(
    key="write_query",
    name="write_query",
    description="How to use the compile_query tool",
    text="SELECT * FROM table WHERE ...",
)
```

Closes #3454.

🤖 Generated with OpenClaw OSS Maintainer.

### DIFF
--- a/tests/utilities/test_inspect.py
+++ b/tests/utilities/test_inspect.py
@@ -174,8 +174,18 @@ class TestGetFastMCPInfo:
         prompt_names = [prompt.name for prompt in info.prompts]
         assert "analyze_data" in prompt_names
         assert "custom_prompt" in prompt_names
-        assert next(prompt for prompt in info.prompts if prompt.name == "analyze_data").text == "Analyze: {data}"
-        assert next(prompt for prompt in info.prompts if prompt.name == "custom_prompt").text is None
+        assert (
+            next(
+                prompt for prompt in info.prompts if prompt.name == "analyze_data"
+            ).text
+            == "Analyze: {data}"
+        )
+        assert (
+            next(
+                prompt for prompt in info.prompts if prompt.name == "custom_prompt"
+            ).text
+            is None
+        )
 
     async def test_format_fastmcp_info_includes_prompt_text_from_meta(self):
         """Prompt inspect output should surface prompt template text from meta."""


jlowin:
Parent issue is closed

PR #3488, opened 2026-03-14
fix(transforms): prevent multiple ResourcesAsTools from overwriting

### LINKED ISSUE #3391: Tools with multiple ResourcesAsTools are overwritten
### Description

I have structured the code with one provider per feature, and my idea was to be able to add them all as tools so that when the corresponding LLM cannot use resources or templates, it uses the tools.

The problem I see is that if I put FastMCP several transforms with each provider, the last one remains (I imagine this is the one that overwrites the list_resources method).

```
❯ Execute tool list-resouces in test-bug mcp

● test-bug - list_resources (MCP)
  ⎿  {
       "result": "[\n  {\n    \"uri\": \"data://resource2\",\n    \"name\": \"resource2\",\n    \"de
     scription\": null,\n    \"mime_type\": \"text/plain\"\n  }\n]"
     }

● The list_resources tool returned one resource:

  - URI: data://resource2
  - Name: resource2
  - MIME type: text/plain
  - Description: none
```

Should this work like this? Or is it an error in how it is implemented?

### Example Code

```Python
import asyncio

import uvicorn
from fastmcp import FastMCP
from fastmcp.server.providers import LocalProvider
from fastmcp.server.transforms import ResourcesAsTools

provider1 = LocalProvider()
@provider1.resource("data://resource1")
async def resource1() -> str:
    return "Hello from resource1!"

provider2 = LocalProvider()
@provider2.resource("data://resource2")
async def resource2() -> str:
    return "Hello from resource2!"

mcp = FastMCP(
    version="1",
    providers=[provider1, provider2],
    transforms=[
        ResourcesAsTools(provider1),
        ResourcesAsTools(provider2),
    ],
    strict_input_validation=True,
)

async def main() -> None:  # pragma: no cover
    app = mcp.http_app(path="/")

    server: uvicorn.Server = uvicorn.Server(
        config=uvicorn.Config(
            app=app,
            host="127.0.0.1",
            port=8000,
            server_header=False,
        )
    )

    try:
        await server.serve()
        await server.shutdown()


if __name__ == "__main__":
    asyncio.run(main())
```

### Version Information

```Text
FastMCP version:                                                                                   3.1.0
MCP version:                                                                                      1.26.0
Python version:                                                                                   3.14.3
Platform:                                                      Linux-6.12.75-1-lts-x86_64-with-glibc2.43
FastMCP root path: /home/jcmencia/Projects/QA/automated-tests-app-sqa/.venv/lib/python3.14/site-packages
```

rar-file:
Fixes #3391.

When multiple `ResourcesAsTools` transforms are applied, the generated `list_resources` / `read_resource` tools could be overridden (and/or duplicated) depending on transform ordering.

This change:
- avoids adding duplicate tools if they already exist in the tool list
- snapshots the provider methods at tool creation time so later transforms don’t change the behavior of the generated tools
- adds a regression test for adding ResourcesAsTools twice

Tests: `pytest -k resources_as_tools`

### DIFF
--- a/docs/apps/development.mdx
+++ b/docs/apps/development.mdx
@@ -1,0 +1,54 @@
+---
+title: Development
+sidebarTitle: Development
+description: Preview and test your app tools locally without a full MCP host.
+icon: flask
+---
+
+import { VersionBadge } from '/snippets/version-badge.mdx'
+
+<VersionBadge version="3.2.0" />
+
+`fastmcp dev apps` launches a browser-based preview for your app tools. It starts your MCP server and a local dev UI side by side — you pick a tool, fill in its arguments, and see the rendered result in a new tab. No MCP host client needed.
+
+This works with both [Prefab apps](/apps/prefab) and [custom HTML apps](/apps/low-level).
+
+## Quick Start
+
+```bash
+pip install "fastmcp[apps]"
+fastmcp dev apps server.py
+```
+
+The dev UI opens at `http://localhost:8080`. Your MCP server runs on port 8000 with auto-reload enabled by default — save a file and the server restarts automatically.
+
+## How It Works
+
+The dev server does three things:
+
+The **picker page** connects to your MCP server, finds all tools with UI metadata, and renders a form for each one. The forms are auto-generated from the tool's input schema — text fields, dropdowns, checkboxes, all wired up.
+
+When you submit a form, the dev server **calls your tool** via the MCP protocol and opens the result in a new tab. The result page loads the tool's UI resource (the Prefab renderer or your custom HTML) inside an AppBridge — the same protocol that real MCP hosts use.
+
+A **reverse proxy** on `/mcp` forwards requests from the browser to your MCP server, avoiding CORS issues that would otherwise block the iframe-based renderer from talking to a different port.
+
+## Options
+
+```bash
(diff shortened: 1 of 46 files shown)


jlowin:
Thanks for the contribution! I think this is working around a usage pattern that the framework already handles.

The intended way to expose all resources as tools is a single transform on the server itself, since the server already aggregates resources across all its providers:

```python
provider1 = LocalProvider()
provider2 = LocalProvider()

mcp = FastMCP(providers=[provider1, provider2])
mcp.add_transform(ResourcesAsTools(mcp))
# list_resources now returns resources from both providers
```

If you need per-provider disambiguation (separate tool names per source), that's what `Namespace` and mounting are for:

```python
server1 = FastMCP("Provider1")
server2 = FastMCP("Provider2")

mcp = FastMCP("Main")
mcp.mount(server1, namespace="p1")
mcp.mount(server2, namespace="p2")
mcp.add_transform(ResourcesAsTools(mcp))
# Resources appear with namespaced URIs like p1/... and p2/...
```

I also don't think this PR fully addresses the issue as reported. It deduplicates the `list_tools` output (first transform's tools win), but doesn't change the `get_tool` middleware chain, where the last transform still intercepts execution. So the listed tool and the executed tool would come from different providers.

Going to close this, but I appreciate you digging into it!

PR #3562, opened 2026-03-19
feat: add block-no-verify PreToolUse hook to .claude/settings.json

### LINKED ISSUE #3561: feat: add block-no-verify PreToolUse hook to .claude/settings.json
## Add `block-no-verify` to prevent agents from bypassing git hooks

**Problem**

`.claude/settings.json` currently configures settings but does not include a `PreToolUse` hook to prevent AI agents from running `git commit` or `git push` with the hook-bypass flag. When an agent uses this flag, it silently skips pre-commit, commit-msg, and pre-push hooks — defeating linting, formatting, and test-before-commit checks.

**Solution**

Add [`block-no-verify`](https://github.com/tupe12334/block-no-verify) as a `PreToolUse` Bash hook. I'm happy to open a PR if the maintainers are interested.

tupe12334:
## Summary

Adds `block-no-verify@1.1.2` as a `PreToolUse` Bash hook alongside the existing SessionStart hook. When agents use the hook-bypass flag, it blocks the git command. Closes #3561

---
_Disclosure: I am the author and maintainer of `block-no-verify`._

### DIFF
--- a/.claude/settings.json
+++ b/.claude/settings.json
@@ -10,6 +10,17 @@
           }
         ]
       }
+    ],
+    "PreToolUse": [
+      {
+        "matcher": "Bash",
+        "hooks": [
+          {
+            "type": "command",
+            "command": "npx block-no-verify@1.1.2"
+          }
+        ]
+      }
     ]
   }
 }


jlowin:
Thanks for the PR.

That said, I don't think we need to enforce this mechanically. CI already runs pre-commit checks before anything merges, so `--no-verify` locally just means the author gets a failing CI run instead of a local failure — annoying, but not a correctness risk. The hooks are a convenience to catch things earlier, not a gate we need to guarantee runs.

Adding `npx` execution on every Bash tool call introduces a third-party dependency into every contributor's development loop, which is more surface area than the problem warrants. Going to close this one out, but thanks for thinking about it.

PR #3563, opened 2026-03-19
Proxy remote server instructions via create_proxy

### LINKED ISSUE #3559: ProxyProvider should proxy original MCP's server instructions to the FastMCPProxy's instructions
### Enhancement

Couldn't find any previous issue about it so bring this up. I am aware `client.initialize_result.instructions` contains the server instructions, and that ProxyProvider uses a fastmcp Client to work. As such, I think the server instructions should be proxied too for FastMCPProxy. Currently instructions is always None unless you add your own.

jlowin:
`create_proxy()` now automatically fetches the remote server's instructions at startup and exposes them to connecting clients. This was a clear oversight — proxies forwarded tools, resources, prompts, sampling, roots, and elicitation, but not instructions.

The fix keeps a clean separation of concerns: `ProxyProvider` exposes a `fetch_remote_instructions()` capability, and `create_proxy()` wires it into the server's lifespan. `FastMCPProxy` used directly doesn't auto-fetch — it's the lower-level escape hatch where users handle this themselves.

```python

# Instructions automatically inherited from the backend
proxy = create_proxy("http://example.com/mcp")

# Explicit instructions take precedence
proxy = create_proxy("http://example.com/mcp", instructions="Custom instructions")
```

Closes #3559

### DIFF
--- a/docs/servers/providers/proxy.mdx
+++ b/docs/servers/providers/proxy.mdx
@@ -129,6 +129,7 @@ Proxies automatically forward MCP protocol features:
 
 | Feature | Description |
 |---------|-------------|
+| Instructions | Server instructions from the backend |
 | Roots | Filesystem root access requests |
 | Sampling | LLM completion requests |
 | Elicitation | User input requests |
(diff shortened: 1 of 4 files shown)


jlowin:
Re: the auth concern from Codex — we explored using middleware (`on_initialize`) to patch instructions at request time, but the SDK sends the `InitializeResult` to the client *before* middleware sees the return value (see `MiddlewareServerSession` in `low_level.py` — `capturing_respond` writes to the stream immediately). So middleware can observe the initialize response but can't mutate what the client receives.

The lifespan approach — setting `server.instructions` before any client connects — is the only way to make this work given the current SDK architecture. For auth-protected backends where the anonymous startup fetch fails, the exception is swallowed and the proxy starts without instructions, same as before this PR. Users can pass `instructions=` explicitly in that case.

Issue #2327, opened 2025-10-31
Is there a session context?


Maxi91f:
### Enhancement

I was used to store session-long variables in the lifespan context. Now that the lifespan context is actually a _life_ span context, I have to make kind of a hack to store different values for a session.

The context that can be set with `ctx.set_state` only lives for one request AFAIK.

Am I missing something? Or do we need a new feature?

strawgate:
Part of the problem is that "session" is different across each connection method.

Are you using SSE?

Maxi91f:
> Part of the problem is that "session" is different across each connection method.

Fair point. I was thinking about something like adding a `set_session_state` in `MiddlewareServerSession`.

> Are you using SSE?

I'm using http.

strawgate:
Storing in the lifespan context was working for http? For some reason I thought every http request was its own context before

Maxi91f:
Yeah, it was kind of a hack, but it worked. And since every session had its own different lifespan, we stored info there.

jlowin:
I think trying to be opinionated on this runs some risk of bumping up against protocol so what's coming to mind to me is that we ensure that the `mcp-session-id` is set in context which will let users adopt it as a key for whatever purpose they have. But it should be noted that for STDIO, the session is de facto equivalent to the lifetime of the server (and session id is never set) and for stateless http, the session id is never set

bill4689:
Is there some way to manage the session start and end in the session lifespan, just like the Fastmcp app lifespan?
I need to initialize something in the first connection and release it when the session is closed.

Maxi91f:
>I think trying to be opinionated on this runs some risk of bumping up against protocol so what's coming to mind to me is that we ensure that the mcp-session-id is set in context which will let users adopt it as a key for whatever purpose they have. 

Yeah, I tried to implement an `on_disconnect` and found that there's not a message for disconnection in the protocol, and it start's to get messy really quick. Now I know why implementing the `on_initialize` took so much.

Right now I get the id in `on_message` where we have all the ids avilable, and store it in a singleton dict. The tricky part is cleaning that dict when the user disconnect, with a weak ref, or other tricks.

>But it should be noted that for STDIO, the session is de facto equivalent to the lifetime of the server (and session id is never set) and for stateless http, the session id is never set

Yeah, the documentation already has some notes about this, and makes sense

[...]

Issue #1894, opened 2025-09-22
Enhance session start/end hook


rcxwhiz:
### Enhancement

At the moment the `FastMCP` `lifespan` parameter does not contain any info about the client connection that is beginning/ending (specifically the MCP session ID). This would be useful for remote MCP servers where you want to take some action when a client connects/disconnects. As far as I understand these are the best workarounds right now:

- Use `access_token()` in the lifespan to get the identity of an authenticated user (does not differentiate between sessions)
- Attach some HTTP middleware to the application listening for connection attempts, harvesting MCP session ID from the response
- Attach some MCP middleware with the `on_message()` hook to track the first MCP message sent for a session

I believe it would be helpful to add access to the MCP session ID in the lifespan handler.

rcxwhiz:
Just now seeing that this may be covered by #1546

Maxi91f:
PR #2359 might partially solve the issue, at least for connection

Emeralden:
Hi! Could a maintainer assign me to this? I have a PR ready (#4232) with both on_connect and on_disconnect hooks, and 6 passing tests. Previous PRs (#2360, #3568) are closed so ready to go

jlowin:
The MCP spec is about to become stateless which will make this issue moot. To the extent these semantics survive, we can reexamine them, but I think it's likely this will become a virtual concern.

Emeralden:
Streamable HTTP can go stateless. But on_connect and on_disconnect will still be useful for STDIO and SSE transports, right? Can we keep hooks even if they become no-ops for stateless HTTP?

PR #3568, opened 2026-03-20
feat: add on_disconnect middleware lifecycle hook

### LINKED ISSUE #2360: Add middleware event on_disconnect
### Enhancement

Recently a new method was added to middleware `on_initialize`. The next logical step is adding `on_disconnection`.

With this method the user can do something if a client disconnects, such as liberate a resource, or track usage. This will probably could be a good workaround for https://github.com/jlowin/fastmcp/issues/2327 and https://github.com/jlowin/fastmcp/issues/1894 and comments like [this one](https://github.com/jlowin/fastmcp/issues/2327#issuecomment-3479293306).

I [started to code something](https://github.com/jlowin/fastmcp/compare/main...Maxi91f:fastmcp:add-disconnection) to implement this and it is really tricky. It's easy to trigger a middleware when a session ends if we [implement the dunder](https://github.com/Maxi91f/fastmcp/blob/add-disconnection/src/fastmcp/server/low_level.py#L60) `__aexit__` in `MiddlewareServerSession`. But there are some problems:

- MCP standard doesn't define a message for disconnection, so [there's not that message](https://github.com/modelcontextprotocol/python-sdk/blob/main/src/mcp/types.py#L345) in MCP SDK library. The message when calling the context needs to be None, or some custom message, neither of those are great ideas.
- There's no context. Such as `on_initialize`, a new context can be created, but this is even more abstract, because no request has been made, and the session instance exists, but is about to be destroyed. We can create a new context just as is done on `on_initialize`, but that context is basically empty, and not useful at all. Why trigger a `on_disconnect` if you are not telling me who was? If https://github.com/jlowin/fastmcp/pull/2359 or something similar gets merged, or someone calls `context.session_id` to create the session_id, we can have at least one `session_id` in session. But we need to add that session in the context.

So, in short, this can be really useful, but for now it's tricky to do because it's not a MCP standard, it's not a request, it needs to be done in `low_level` touching `MiddlewareServerSession`, but it's not impossible.

I'm opening this issue because I spent some time trying to solve this, and wanted to share my findings, and I think there's not other issue exactly for this.

If there's something I'm missing, I'm available for more testing and PoC and even to work on a PR if needed.

Note: I'm biased thinking mostly of HTTP, maybe there's something I'm missing for STDIO.

mtthidoteu:
## Problem

There is no public API to run code when a client disconnects. The only workaround is accessing `session._exit_stack` directly, which is fragile across upgrades.

Since `on_initialize` exists as a lifecycle hook, `on_disconnect` is the natural counterpart.

## Use case

MCP server that auto-generates session summaries on disconnect — needs access to session state set during earlier tool calls.

## Changes

- `MiddlewareServerSession.__aexit__` fires the middleware chain before session teardown
- `Middleware.on_disconnect()` hook with `MiddlewareContext[None]`
- `Middleware.on_lifecycle()` generic hook for lifecycle events
- 8 tests covering all scenarios

Closes #2360
Related: #1894, #2327

### DIFF
--- a/tests/server/middleware/test_disconnect_middleware.py
+++ b/tests/server/middleware/test_disconnect_middleware.py
@@ -2,8 +2,6 @@
 
 from typing import Any
 
-import pytest
-
 from fastmcp import Client, FastMCP
 from fastmcp.server.context import Context
 from fastmcp.server.middleware import CallNext, Middleware, MiddlewareContext


jlowin:
Given that there are about to be significant changes to the SDK, I'm not enthusiastic about adding new low-level hooks/modifications at this time (it is complicated to understand whether we will be able to support them in a non-breaking way after the SDK v2 is released). Given that session disconnects are not required and major MCP changes are coming to how sessions and transports work, I think this is cementing a pattern that we may want to revisit when the protocol makes it easier. Is there a clear use case that is currently impossible despite those frictions? The issue this PR closes is already closed.

mtthidoteu:
> Given that there are about to be significant changes to the SDK, I'm not enthusiastic about adding new low-level hooks/modifications at this time (it is complicated to understand whether we will be able to support them in a non-breaking way after the SDK v2 is released). Given that session disconnects are not required and major MCP changes are coming to how sessions and transports work, I think this is cementing a pattern that we may want to revisit when the protocol makes it easier. Is there a clear use case that is currently impossible despite those frictions? The issue this PR closes is already closed.

The use-case is to be able to run something on session ending, in my case it is summarising the session and doing something with it.

jlowin:
Well, of course the use case of a PR is always the thing the PR does — what I'm asking is whether this is a generalizable framework concern, given that "disconnect" isn't defined in MCP and most clients don't formally close their sessions.

This hook fires when the transport closes, not when "the client is done." Those are only the same thing for stdio, where process lifecycle and session lifecycle are identical. For stateless streamable HTTP — increasingly the standard for production — this would fire after every request, which is semantically wrong. And in a horizontally scaled deployment with persistent sessions, the transport might close on a different node than the one that handled the tool calls, or never close at all.

I'm not comfortable adding a lifecycle abstraction that promises something the protocol can't deliver. At this moment in particular, with SDK v2 coming and sessions being redesigned by the core committee, I think the bar for a low-level modification is very high. If MCP adds session lifecycle semantics in the future, we can support this properly then!

PR #3580, opened 2026-03-22
Add `encoding` parameter to `FileResource`

### LINKED ISSUE #3579: FileResource.read() raises UnicodeDecodeError on Windows for UTF-8 files with non-ASCII characters
### Enhancement

`FileResource.read()` doesn't specify an encoding when calling `read_text()`, so it falls back to the system locale. On Windows (where the default is often `cp1252`), this causes a `UnicodeDecodeError` when reading UTF-8 files that contain non-ASCII characters like smart quotes or em dashes.

**Use case:** Registering a UTF-8 markdown file as a `FileResource` with `mime_type="text/markdown"` on a Windows machine with a `cp1252` locale.

**Expected:** The file is read successfully.

**Actual:**
```
UnicodeDecodeError: 'charmap' codec can't decode byte 0x9d in position 73832:
character maps to <undefined>
```

The byte `0x9d` is part of the UTF-8 sequence for U+201D (RIGHT DOUBLE QUOTATION MARK). The file is valid UTF-8, but the system codec can't decode it.

shulkx:
Closes #3579 

`FileResource.read()` calls `read_text()` without specifying an encoding, which defaults to the system locale. On Windows with a `cp1252` locale, this raises `UnicodeDecodeError` for UTF-8 files containing non-ASCII characters like smart quotes or em dashes — a common scenario when serving markdown content.

This adds an optional `encoding` field to `FileResource` that gets passed through to `read_text()`. The default is `None` (preserving system behavior), so this is a non-breaking change. Users who need cross-platform reliability can now specify the encoding explicitly:

```python
resource = FileResource(
    uri=f"file://{path.as_posix()}",
    path=path,
    mime_type="text/markdown",
    encoding="utf-8",
)
```


jlowin:
Thanks @shulkx. Does it make sense to change the default from system to `utf-8`? Then this would move from escape hatch to avoiding a footgun. But perhaps that introduces other unexpected issues, so having your new keyword permits users to help themselves if/when they run into trouble.

shulkx:
> Thanks @shulkx. Does it make sense to change the default from system to `utf-8`? Then this would move from escape hatch to avoiding a footgun. But perhaps that introduces other unexpected issues, so having your new keyword permits users to help themselves if/when they run into trouble.

Good point @jlowin -- defaulting to "utf-8" makes more sense here. The risk of breaking users who intentionally read system-locale-encoded files via FileResource is minimal in an MCP context where content is overwhelmingly UTF-8. Updated.

jlowin (approved):
Looks good — defaulting to utf-8 is the right call here. Thanks for the quick turnaround on the default change.

PR #3585, opened 2026-03-22
Replace UUID global keys with (app_name, tool_name) registry

jlowin:
The old FastMCPApp routing used UUID-suffixed global keys (like `save_contact-a1b2c3d4`), three module-level registries, and a multi-fallback resolver that mapped callables and strings through several paths. It worked, but the machinery was hard to follow and the keys were process-local.

This replaces all of it with a single `_APP_TOOLS` dict keyed by `(app_name, tool_name)`. The server reads `_meta.fastmcp.app` from the MCP request to scope the lookup. The Prefab renderer (prefab-ui ≥0.13.1) reads `_meta.fastmcp.app` from the `@app.ui()` structured content and echoes it back on every `callServerTool` call.

```python
app = FastMCPApp("Contacts")

@app.tool()
def save(name: str) -> str:
    # registered as _APP_TOOLS[("Contacts", "save")]
    return name

@app.ui()
def manager() -> PrefabApp:
    # structured content includes _meta.fastmcp.app = "Contacts"
    # CallTool("save") just serializes as "save"
    with Column() as view:
        Button("Save", on_click=CallTool("save"))
    return PrefabApp(view=view)
```

Two apps with the same tool name are disambiguated by app name, not UUID. Keys are deterministic and survive across processes.

### DIFF
--- a/src/fastmcp/server/app.py
+++ b/src/fastmcp/server/app.py
@@ -326,7 +326,10 @@ class FastMCPApp(Provider):
                     visibility=["model"],
                 )
 
-            meta: dict[str, Any] = {"ui": app_config_to_meta_dict(app_config)}
+            meta: dict[str, Any] = {
+                "ui": app_config_to_meta_dict(app_config),
+                "fastmcp": {"app": self.name},
+            }
 
             tool_obj = Tool.from_function(
                 fn,
--- a/src/fastmcp/server/mixins/mcp_operations.py
+++ b/src/fastmcp/server/mixins/mcp_operations.py
@@ -220,11 +220,12 @@ class MCPOperationsMixin:
             app_name: str | None = None
             try:
                 ctx = server._mcp_server.request_context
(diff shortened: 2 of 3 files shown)


jlowin:
We're aware this PR's process-level registry (`_APP_TOOLS`) doesn't survive horizontally scaled environments where each tool call may land in a different process (e.g., Prefect Horizon's lambda handlers). This is a simplification over the previous UUID-based approach — fewer registries, deterministic keys, readable routing — but it's not the final architecture.

The next step is making app-visible tools survive transforms in the provider chain itself, so tool routing works from the provider tree at call time without any process-level state. That's a separate PR.

Re: Codex comments — the `unwrapResult` concern is handled by `_meta.fastmcp.wrap_result` on the tool result (already set by FastMCP, already read by the renderer at line 191 in actions.ts). The backend-tool Prefab view concern is valid for multi-step flows but out of scope for this simplification.

PR #3589, opened 2026-03-22
Fix loopback redirect URI port matching per RFC 8252 §7.3

### LINKED ISSUE #3588: CIMD redirect_uri validation rejects loopback URIs with dynamic ports (RFC 8252 §7.3)
### What happened?

  CIMD clients that declare `redirect_uris` without a port (e.g. `http://localhost/callback`) fail validation when the actual
  redirect uses a dynamic port, which is standard behavior for MCP clients.
```
  {"error":"invalid_request","error_description":"Redirect URI 'http://localhost:51353/callback' does not match CIMD
  redirect_uris."}
```

  This affects Claude Code, whose CIMD document declares:
  ```json
  "redirect_uris": ["http://localhost/callback", "http://127.0.0.1/callback"]
```

### Example Code

```Python
from fastmcp.server.auth.redirect_validation import matches_allowed_pattern

  # CIMD pattern without port vs actual redirect with dynamic port
  assert matches_allowed_pattern(
      "http://localhost:51353/callback",
      "http://localhost/callback",
  )  # AssertionError
```

  _match_port normalizes the missing pattern port to 80, so it never matches the dynamic port. Per RFC 8252 §7.3, loopback
  redirect URIs use ephemeral ports and the authorization server MUST allow any port.

### Version Information

```Text
FastMCP version:                                                                                         3.1.1
MCP version:                                                                                            1.26.0
Python version:                                                                                        3.13.12
Platform:                                                                  macOS-26.3.1-arm64-arm-64bit-Mach-O
FastMCP root path: /Users/rushabh/hiro__worktrees/radoshi-mcp-debugging/mcp/.venv/lib/python3.13/site-packages
```

radoshi:
Claude wrote the PR and I reviewed and guided it for some minor fixups.

---

## Summary

CIMD clients that declare `redirect_uris` without a port (e.g. `http://localhost/callback`) fail validation when the actual redirect uses a dynamic ephemeral port — standard behavior for MCP clients like Claude Code. Per RFC 8252 §7.3, the authorization server MUST allow any port for loopback redirect URIs.

The fix skips port matching in `matches_allowed_pattern` when the pattern host is a loopback address (`localhost`, `127.0.0.1`, `::1`) and no port is explicitly specified. Scheme, host, and path validation are unaffected.

```python
from fastmcp.server.auth.redirect_validation import matches_allowed_pattern

# Before: False — pattern port defaults to 80, rejects dynamic port
# After: True — loopback pattern without port accepts any port
matches_allowed_pattern(
    "http://localhost:51353/callback",
    "http://localhost/callback",
)
```

Also updated `CIMDFetcher.validate_redirect_uri` to delegate to `matches_allowed_pattern` instead of raw `fnmatch`, so both validation paths get consistent behavior.

Closes #3588

### DIFF
--- a/src/fastmcp/server/auth/cimd.py
+++ b/src/fastmcp/server/auth/cimd.py
@@ -16,7 +16,6 @@ This module provides:
 
 from __future__ import annotations
 
-import fnmatch
 import json
 import time
 from collections.abc import Mapping
@@ -28,6 +27,7 @@ from urllib.parse import urlparse
 
 from pydantic import AnyHttpUrl, BaseModel, Field, field_validator
 
+from fastmcp.server.auth.redirect_validation import matches_allowed_pattern
 from fastmcp.server.auth.ssrf import (
     SSRFError,
     SSRFFetchError,
@@ -422,6 +422,9 @@ class CIMDFetcher:
     def validate_redirect_uri(self, doc: CIMDDocument, redirect_uri: str) -> bool:
         """Validate that a redirect_uri is allowed by the CIMD document.
 
+        Uses component-level matching (scheme, host, port, path) which correctly
+        handles RFC 8252 §7.3 loopback port flexibility and wildcard patterns.
+
         Args:
             doc: The CIMD document
             redirect_uri: The redirect URI to validate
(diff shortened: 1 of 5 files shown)


daveboxaq:
Thanks all! I was going to take a stab at this, but I'm so stoked that this has been merged already! Any idea when we can expect another release? 🚀

PR #3604, opened 2026-03-24
Fix: Always return CallToolResult when structured_content exists (#3596)

### LINKED ISSUE #3596: Structured data no longer works in FastMCP 3.1
### What happened?

I just upgraded from FastMCP 2.14.5 to 3.1.1.  Many of my unit tests now fail because the `data` element of `CallToolResult` is `None`.  For example:

```
E       assert False
E        +  where False = isinstance(None, list)
E        +    where None = CallToolResult(content=[TextContent(type='text', text='[...some text data in JSON format...]')], meta=None, data=None, is_error=False).data
```
I then tried a downgrade back to 2.14.5, and it works.  This is my test code in this case:

```python
async def test_get_cml_labs(main_mcp_client: Client[FastMCPTransport], created_lab: tuple[UUID4Type, LabRequest]):
    result = await main_mcp_client.call_tool("get_cml_labs", {})
    # outsource(result.data, ".json")

    assert isinstance(result.data, list)
    assert len(result.data) > 0
    for lab in result.data:
        if isinstance(lab, dict):
            lab = Lab(**lab)
        assert isinstance(lab, Lab)
```

And `main_mcp_client` is defined as a fixture to be:

```python
@pytest.fixture()
async def main_mcp_client():
    """
    Main MCP client fixture for testing.
    Works with both mock and live modes.
    """

    from cml_mcp.server import server_mcp

    async with Client(transport=server_mcp) as mcp_client:
        yield mcp_client
```

So, it's using the direct transport.

### Example Code

```Python
See code above.
```

### Version Information

```Text
FastMCP version:                                                               3.1.1
MCP version:                                                                  1.26.0
Python version:                                                              3.13.12
Platform:                                        macOS-26.3.1-arm64-arm-64bit-Mach-O
FastMCP root path: /Users/jclarke/src/git/cml-mcp/.venv/lib/python3.13/site-packages
```

hnshah:
Fixes #3596

## Problem
`CallToolResult.data` returns `None` when structured data is present.

## Root Cause
`to_mcp_result()` returned tuple instead of `CallToolResult` when `meta=None` but `structured_content` exists.

## Solution
Return `CallToolResult` when either `meta` or `structured_content` is present (OR condition).

## MRE
```python
result = await client.call_tool("tool_name", {})
assert result.data is not None  # Now passes
```

Fixes both:
- Original bug (#3596 - structured_content without meta)  
- Regression caught by CI (meta without structured_content)

Co-authored-by: marvin-context-protocol[bot]

### DIFF
--- a/src/fastmcp/tools/base.py
+++ b/src/fastmcp/tools/base.py
@@ -125,15 +125,15 @@ class ToolResult(BaseModel):
     ) -> (
         list[ContentBlock] | tuple[list[ContentBlock], dict[str, Any]] | CallToolResult
     ):
-        # Always return CallToolResult when structured_content exists to maintain
-        # consistent .data field in client (fixes #3596)
-        if self.structured_content is not None:
+        # Return CallToolResult when either meta or structured_content exists
+        # to maintain consistent client behavior (fixes #3596)
+        if self.meta is not None or self.structured_content is not None:
             return CallToolResult(
                 structuredContent=self.structured_content,
                 content=self.content,
                 _meta=self.meta,  # type: ignore[call-arg]  # _meta is Pydantic alias for meta field
             )
-        # Only return bare content list when no structured content
+        # Only return bare content when both are None
         return self.content
 
 


hnshah:
Thanks @marvin-context-protocol! You're absolutely right - I broke the meta-only case.

Fixed in latest commit (c8cdef71):
- Changed to OR condition: `if self.meta is not None or self.structured_content is not None`
- Now preserves meta when structured_content is None  
- Still fixes original #3596 bug

The test should pass now. Apologies for the oversight!

jlowin:
See the note on #3596 please; this issue needs an MRE before it has a PR

jlowin:
Closing, see #3596

PR #3631, opened 2026-03-26
fix: HTTP request headers not accessible in background task workers

pandego:
## Summary
- Fixes PrefectHQ/fastmcp#3545

## Validation
- Local validation attempted in execute worker


jlowin:
@pandego have you discussed this approach with @chrisguidry ?

src/fastmcp/server/dependencies.py
```diff
@@ -1159,6 +1244,71 @@ async def get_auth_type(headers: dict = CurrentHeaders()) -> str:
     return cast(dict[str, str], _CurrentHeaders())
 
 
+def _wrap_task_execution(fn: Callable[..., Any]) -> Callable[..., Any]:
```
chrisguidry: Okay, I like the front half of this in `submit_to_docket` where we're stashing the headers to Redis, but I'm not sure about this "back half" where we've wrapped every task function with `_wrap_task_execution`.  This is a great use case for [making additional Docket dependencies](https://docket.lol/en/latest/dependency-injection/), or just extending the ones we have.  What I'm thinking here is that we continue registering the original function (without this wrapper) and instead update the `CurrentHeaders`, `CurrentRequest`, etc, dependencies to do this part of the job (reconstructing state from redis) internally.  I think this should end up being clearer to read and more extensible.

In fact, `submit_to_docket` could even inspect which dependencies are used and only stash what they need in redis.  Like if you're only asking for `CurrentHeaders` but not the `AccessToken` we can just stash the headers in redis.

Implementing them as dependencies also gives you a very natural way to clean up redis once the background task finishes.

How does that work for you @pandego?

pandego: Yep, that makes sense. I reworked this to keep the submission-side snapshotting in `submit_to_docket`, but move the reconstruction responsibility back into the dependency layer instead of wrapping every registered task callable.

The updated branch now registers the original tool/resource/prompt functions with Docket again, and relies on `CurrentRequest`, `CurrentHeaders`, and `CurrentAccessToken` to restore task state on demand. Focused validation passed locally:

- `uv run pytest tests/server/tasks/test_context_background_task.py -q`
- `uv run prek run --files src/fastmcp/server/dependencies.py src/fastmcp/server/tasks/handlers.py src/fastmcp/tools/function_tool.py src/fastmcp/tools/base.py src/fastmcp/resources/function_resource.py src/fastmcp/resources/template.py src/fastmcp/resources/base.py src/fastmcp/prompts/function_prompt.py src/fastmcp/prompts/base.py tests/server/tasks/test_context_background_task.py`

If you want, I can also follow up with the dependency-inspection optimization so `submit_to_docket` only snapshots the request bits a given task actually uses, but I kept this pass focused on the design change you called out.

jlowin:
@pandego @chrisguidry note this included a minor breaking change that `get_http_request()` fails to restore headers (I merged too optimistically before waiting for CI, but its failing on other branches now) -- fixed in #3681

PR #3632, opened 2026-03-26
fix(server): preserve mounted tool task metadata

pandego:
Mounted tools were losing `execution.taskSupport` metadata when exposed through a parent server, even though the child tool advertised task support correctly. That made `tools/list` disagree with actual mounted task behavior and hid SEP-1686 capability from clients.

This preserves the child tool's execution metadata in the mounted provider wrapper and adds a regression test covering the mounted `tools/list` path. For example, a child `@tool(task=True)` now still advertises `execution.taskSupport == "optional"` after `parent.mount(child)`.

### DIFF
--- a/src/fastmcp/tools/base.py
+++ b/src/fastmcp/tools/base.py
@@ -204,10 +204,12 @@ class Tool(FastMCPComponent):
             ),  # ty:ignore[unknown-argument]
         )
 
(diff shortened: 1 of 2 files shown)


pandego:
Thanks - good call. I moved the task execution metadata handling into `Tool.to_mcp_tool()`, removed the subclass-specific overrides, kept the mounted-tool regression coverage, and rebased on `main`.

PR #3644, opened 2026-03-27
fix: resolve CurrentFastMCP() and ctx.fastmcp to child server in mounted background tasks (#3571)

### LINKED ISSUE #3571: Mounted background tasks bind CurrentFastMCP() and Context.fastmcp to the parent server instead of the mounted child
## What happened?

When a task-enabled tool is defined on a child FastMCP server and that server is mounted into a parent server, background task execution (`task=True`) resolves the server context to the parent server instead of the child server.

This affects both:

- `CurrentFastMCP()`
- `ctx.fastmcp` on an injected `Context`

Foreground execution appears fine; this only reproduces in background task mode.

## Example code

```python
import asyncio

from fastmcp import Client, Context, FastMCP
from fastmcp.dependencies import CurrentFastMCP

child = FastMCP("child")

@child.tool(name="whoami", task=True)
async def whoami(server=CurrentFastMCP()) -> str:
    return f"server name: {server.name}"

@child.tool(name="whoami_ctx", task=True)
async def whoami_ctx(ctx: Context) -> str:
    return f"context server: {ctx.fastmcp.name}"

parent = FastMCP("parent")
parent.mount(child, namespace="child")

async def demo():
    async with Client(parent) as client:
        result1 = await (await client.call_tool("child_whoami", {}, task=True)).result()
        result2 = await (await client.call_tool("child_whoami_ctx", {}, task=True)).result()

        print(result1.content[0].text)
        print(result2.content[0].text)

if __name__ == "__main__":
    asyncio.run(demo())
```

## Expected behavior

Mounted child tasks should preserve the child server as the active FastMCP server during background execution.

Expected output:

- `server name: child`
- `context server: child`

## Actual behavior

The background task resolves both dependencies against the parent server instead.

Observed output:

- `server name: parent`
- `context server: parent`

## Version information

```text
FastMCP: 3.1.0
Python: 3.13
OS: linux
```

syhstanley:
When a background task is submitted from a mounted child server,
`CurrentFastMCP()` and `ctx.fastmcp` inside the task resolve to the
parent server instead of the child.
https://github.com/jlowin/fastmcp/issues/3571

Root cause: `_current_server` ContextVar is set to the parent at Docket
worker startup (`server/mixins/lifespan.py:51`). Foreground calls override
this per-request through `child_server.call_tool()`, but background task
workers skip that step.

Fix: at `submit_to_docket()` time — which runs inside the child's
`call_tool()` where `ctx.fastmcp` is already the child — store a
`task_id → weakref` mapping. `_CurrentFastMCP.__aenter__()` and
`_CurrentContext.__aenter__()` now check this map first before falling
back to `_current_server`.

Regression tests added in `tests/server/tasks/test_task_mount.py`
covering both `CurrentFastMCP()` and `ctx.fastmcp` resolution.

Closes #3571

### DIFF
--- a/src/fastmcp/server/dependencies.py
+++ b/src/fastmcp/server/dependencies.py
@@ -174,6 +174,26 @@ def get_task_session(session_id: str) -> ServerSession | None:
 _current_server: ContextVar[weakref.ref[FastMCP] | None] = ContextVar(
     "server", default=None
 )
+
+# --- Background task server map ---
+
+
+_task_server_map: dict[str, weakref.ref[Any]] = {}
+
+
+def register_task_server(task_id: str, server: Any) -> None:
+    """Register the owning server for a background task.
+
+    Called at task-submission time so that background workers can resolve
+    CurrentFastMCP() / ctx.fastmcp to the child server for mounted tasks (#3571).
+
+    Args:
+        task_id: The server-generated task UUID
+        server: The FastMCP server that owns the component being executed
+    """
+    _task_server_map[task_id] = weakref.ref(server)
+
+
 _current_docket: ContextVar[Docket | None] = ContextVar("docket", default=None)
 _current_worker: ContextVar[Worker | None] = ContextVar("worker", default=None)
 _task_access_token: ContextVar[AccessToken | None] = ContextVar(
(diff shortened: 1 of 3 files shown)


jlowin:
Thanks for taking a look at this. We ended up going a different direction — this fix needed some design iteration around where to intercept (the wrapping approach doesn't work because Docket resolves dependencies before calling the function, so the fix has to live in get_server() itself). We're shipping our own implementation that centralizes the logic there. In the future, for this type of sensitive machinery, please work on design in the issue to ensure you are treating the cause, not the symptom!

PR #3645, opened 2026-03-27
feat: implement resource subscriptions (resources/subscribe / resources/unsubscribe)

### LINKED ISSUE #3641: MCP conformance: resource subscriptions (subscribe/unsubscribe)
The MCP conformance test suite expects `resources/subscribe` and `resources/unsubscribe` support. FastMCP doesn't implement resource subscriptions, so both checks are listed as expected failures in `tests/conformance/expected-failures.yml`.

When implemented, remove `resources-subscribe` and `resources-unsubscribe` from the expected failures file and verify the conformance suite passes.

syhstanley:
FastMCP didn't implement `resources/subscribe` or `resources/unsubscribe`, leaving both as expected failures in the conformance suite. This PR adds full support for both endpoints, removing them from `expected-failures.yml` so the conformance runner exercises them for real.

Subscriptions are tracked in a `ResourceSubscriptionRegistry` — a weakref-based singleton that maps resource URIs to subscriber sessions. The low-level server registers handlers for both MCP methods and cleans up a session's subscriptions on disconnect. Capabilities are advertised as `subscribe: true`. To push change notifications to subscribers, call `ctx.notify_resource_updated(uri)` from any resource or tool handler:

```python
@mcp.tool
async def update_data(ctx: Context) -> str:
    # ... mutate the resource ...
    await ctx.notify_resource_updated("myapp://data/live")
    return "updated"
```

Closes #3641

### DIFF
--- a/src/fastmcp/server/context.py
+++ b/src/fastmcp/server/context.py
@@ -788,6 +788,38 @@ class Context:
         """
         await self.session.send_notification(mcp.types.ServerNotification(notification))
 
+    async def notify_resource_updated(self, uri: str) -> None:
+        """Notify all subscribed clients that a resource has been updated.
+
+        Sends ``notifications/resources/updated`` to every session that has
+        previously called ``resources/subscribe`` for *uri*.  Sessions that
+        have since disconnected are silently skipped.
+
+        Args:
+            uri: The resource URI that has changed.
+
+        Example:
+            ```python
+            @server.tool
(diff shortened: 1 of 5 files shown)


jlowin:
There are some significant issues with this implementation (most notably a global registry instead of per-server) that suggest it hasnt been designed in a framework context. Please make a design proposal rather than implementing quickly.

PR #3660, opened 2026-03-27
[Do not merge] Fix FileSystemProvider reload race condition, redundant reloads, and missing warning

### LINKED ISSUE #3625: bug: several correctness issues in FileSystemProvider import machinery
`FileSystemProvider` has several bugs in how it imports discovered modules, plus a few smaller behavioral gaps.

### 1. `sys.path` entries are never removed

In non-package mode, the parent directory of each discovered file is inserted into `sys.path[0]` and never removed. In package mode, the package root's parent is inserted. For a typical layout with `tools/`, `resources/`, and `prompts/` subdirectories, three entries accumulate per provider instance. They persist for the lifetime of the process and affect all imports everywhere.

**Fix:** wrap the addition in a `try/finally` and remove it after `exec_module` returns. This is safe because `exec_module` resolves all top-level imports synchronously; by the time it returns, every `import` statement in the file is already a live object reference in `module.__globals__`. The path entry is not needed at request time.

```python
added = parent_dir not in sys.path
if added:
    sys.path.insert(0, parent_dir)
try:
    spec.loader.exec_module(module)
finally:
    if added:
        sys.path.remove(parent_dir)
```

### 2. Non-package files registered under their bare stem, shadowing stdlib

In non-package mode, the module is registered in `sys.modules` using `file_path.stem` as the key — so `mcp/json.py` is registered as `sys.modules["json"]`, silently replacing the stdlib module for the entire process. Any code that subsequently does `import json` gets the provider file.

```
mcp/
└── json.py   # contains @tool functions

# After FileSystemProvider("mcp"):
import json
json.dumps({})   # AttributeError: module 'json' has no attribute 'dumps'
```

Common risky names: `os`, `json`, `http`, `re`, `time`, `types`, `logging`, `asyncio`, `requests`, `httpx`, `pydantic`.

**Fix:** use a private, path-based key that can't collide with real packages:

```python
import hashlib
key = f"_fastmcp_{file_path.stem}_{hashlib.sha1(str(file_path).encode()).hexdigest()[:12]}"
sys.modules[key] = module
```

### 3. `_find_package_root` walks above the provider root

`_find_package_root` walks up the directory tree until it finds a directory without `__init__.py`. If the provider root is nested inside a larger Python package (ancestor directories have `__init__.py`), it escapes past the provider boundary. For a provider rooted at `myproject/mcp/`, it returns `myproject/` as the package root and adds `myproject/`'s parent to `sys.path`. The generated module name is then `myproject.mcp.tools` instead of `mcp.tools`.

**Fix:** stop the walk at the provider root:

```python
def _find_package_root(file_path: Path, stop_at: Path | None = None) -> Path | None:
    current = file_path.parent
    package_root = None
    while current != current.parent:
        if stop_at is not None and current == stop_at.parent:
            break
        if _is_package_dir(current):
            package_root = current
            current = current.parent
        else:
            break
    return package_root
```

### 4. Lock released before components are read — race condition

In reload mode, `_list_tools` (and all other `_list_*`/`_get_*` overrides) calls `_ensure_loaded()` then `super()._list_tools()`. `_ensure_loaded` acquires the lock, reloads (which clears and rebuilds `_components`), then **releases the lock** before returning. `super()._list_tools()` reads `_components` without holding the lock.

A concurrent request can therefore acquire the lock, call `_components.clear()` in a threadpool thread, and race with the first request's read. Result: `list_tools` returns an empty list.

**Fix:** swap the components dict atomically instead of clear-and-rebuild, so readers always see either the old or new complete dict:

```python
def _load_components(self) -> None:
    result = discover_and_import(self._root)
    new_components: dict[str, FastMCPComponent] = {}
    # ... populate new_components ...
    self._components = new_components  # atomic reference swap under CPython's GIL
```

### 5. Concurrent requests each trigger a full reload

The inner condition inside `_ensure_loaded`'s lock — `if self._reload or not self._loaded` — is always `True` when `reload=True`. The lock serializes requests but doesn't deduplicate them: every caller that acquires the lock reloads, even if the previous holder just finished. Ten concurrent requests trigger ten sequential full directory scans.

**Fix:** track a reload generation counter. Callers that were waiting for the lock check whether the generation changed while they waited; if it did, the reload already happened:

```python
async with self._reload_lock:
    if not self._loaded or (self._reload and self._reload_generation == generation_before):
        await asyncio.to_thread(self._load_components)
        self._reload_generation += 1
```

## Smaller gaps

**No warning for non-existent root.** `FileSystemProvider(Path("./mcp"))` with a typo silently creates an empty provider with zero components. A `logger.warning` in `_load_components` when `not self._root.exists()` would surface this immediately.

strawgate:
The reload path in `FileSystemProvider` had two correctness issues and a usability gap. The component dict was cleared and rebuilt in place, so a concurrent reader could see an empty dict mid-reload. Every concurrent request also triggered its own full directory scan because the lock serialized but didn’t deduplicate. And a typo in the root path silently produced zero components.

Now `_load_components` builds into a fresh dict and swaps the reference atomically, a generation counter lets waiters skip redundant reloads, and a non-existent root logs a warning.

```python
# Atomic swap — readers never see an empty dict
new_components = {}
self._components = new_components
# ... register into new_components ...

# Generation counter — deduplicate concurrent reloads
generation_before = self._reload_generation
async with self._reload_lock:
    if self._reload_generation == generation_before:
        self._load_components()
        self._reload_generation += 1
```

Closes #3625 (bugs 4, 5, and missing warning)


strawgate:
Superseded by #3938 which fixes the race condition correctly by holding the reload lock across both reload and read, instead of the atomic-swap approach which still had a partial-dict window.

PR #3665, opened 2026-03-27
[DO NOT MERGE] Fix ErrorHandlingMiddleware classifying by __cause__ type

### LINKED ISSUE #3663: ErrorHandlingMiddleware._transform_error classifies by __cause__ type instead of the raised exception type
`_transform_error` maps exceptions to MCP error codes, but when an exception has a `__cause__` (from `raise X from Y`), it classifies based on the *cause* rather than the exception that was actually raised:

```python
# error_handling.py:92
error_type = type(error.__cause__) if error.__cause__ else type(error)
```

This inverts the semantics — the outer exception is the one the developer deliberately raised, and should determine the error code.

```python
from fastmcp import FastMCP

mcp = FastMCP(middleware=[ErrorHandlingMiddleware(transform_errors=True)])

@mcp.tool()
def lookup(key: str) -> str:
    try:
        return some_dict[key]
    except KeyError as e:
        raise ValueError(f"Invalid key: {key}") from e
```

If `lookup` is called with a bad key, the developer explicitly raises `ValueError` (invalid params, `-32602`), but the middleware sees `KeyError` via `__cause__` and returns "Not found" (`-32001`) instead.

The fix is to classify on `type(error)` — or if the intent was to unwrap, check `__cause__` only as a fallback, not as the primary.

strawgate:
`ErrorHandlingMiddleware._transform_error` inspects `error.__cause__` to determine the error type when `transform_errors` is enabled. This means that when a developer deliberately raises one exception chained from another — e.g., `raise ValueError("bad input") from KeyError("missing_key")` — the middleware classifies the error based on the `KeyError` (the cause) rather than the `ValueError` (the exception that was actually raised). The outer exception is the one the developer chose to represent the error, and it should determine the MCP error code.

The fix removes the `__cause__` lookup so `_transform_error` always classifies by the type of the exception itself:

```python
# Before: a chained ValueError gets classified as KeyError
raise ValueError("bad input") from KeyError("k")
# _transform_error saw KeyError → mapped to "resource not found" (-32002)

# After: the raised exception determines the error code
raise ValueError("bad input") from KeyError("k")
# _transform_error sees ValueError → mapped to "invalid params" (-32602)
```

Fixes #3663

🤖 Generated with Claude Code

### DIFF
--- a/src/fastmcp/server/middleware/error_handling.py
+++ b/src/fastmcp/server/middleware/error_handling.py
@@ -89,7 +89,7 @@ class ErrorHandlingMiddleware(Middleware):
             return error
 
         # Map common exceptions to appropriate MCP error codes
-        error_type = type(error.__cause__) if error.__cause__ else type(error)
+        error_type = type(error)
 
         if error_type in (ValueError, TypeError):
             return McpError(


strawgate:
Closing — the fix is wrong. See #3663 for details.

Issue #3676, opened 2026-03-28
Support Clerk Auth Provider


mostafa6765:
### Enhancement
Clerk is a popular enterprise auth provider.
Website: https://clerk.com/

We should support it!

PR #3677, opened 2026-03-28
Add Clerk OAuth provider

mostafa6765:
## Description

Add a Clerk OAuth provider to the FastMCP server auth stack, mirroring the existing Google style providers.

## Contribution type

- [x] Enhancement

## Checklist

- [x] This PR addresses an existing issue #3676 


mostafa6765:
Hey @jlowin — could you take a look at this?

The failing test appears to be a pre-existing issue on main, unrelated to the Clerk OAuth changes in this PR.

The marvin-context-protocol bot analysis highlights that get_http_request() does not restore headers when called directly inside a background task (without using CurrentRequest() / CurrentHeaders()), which causes the failure.

mostafa6765:
@jlowin  Thanks for the thoughtful review — really appreciate the detailed feedback.

On the **OIDCProxy vs OAuthProxy** point:
You’re absolutely right that Clerk is an OIDC provider and that, in theory, using `OIDCProxy` would simplify things by leveraging JWKS discovery and local JWT verification.

That said, our current approach is `intentionally` aligned with the existing Google provider pattern, where we **fetch user info during token verification**. The main reason is that, in practice, we almost always need enriched user attributes (especially `email`, `email_verified`, etc.) immediately as part of the auth context.

With `OIDCProxy`, while we’d get validated JWT claims locally, we’d still need to make a **follow-up request to `/userinfo`** to retrieve those fields reliably. So the trade-off becomes:

* **OIDCProxy** → 0 calls for verification + 1 call for user info
* **Current approach** → 1–2 calls but returns fully enriched claims in a single step

We chose the latter to keep the auth pipeline simpler for downstream consumers and consistent with existing providers.

---

On **audience validation**:
Agreed — this was a gap. The latest update enforces strict audience checking when `client_id` is configured, ensuring tokens are only accepted if issued for the correct application.

---

On **scope validation fallback**:
Also agreed — the previous behavior was too permissive. This is now updated to **fail closed**:

* If `required_scopes` is configured but scopes cannot be verified → reject the token
* Scope checks are only skipped when not configured at all

---

### Summary

* We’re aware that `OIDCProxy` could reduce verification overhead and simplify parts of the implementation.
* However, since **user profile data is required in most flows**, we’d still need an additional request — so we opted for a design that returns **fully enriched claims upfront**, consistent with the Google provider.
* The key security concerns (audience + scope validation) are now addressed with strict checks.

mostafa6765:
@jlowin please take a look — I’ve fixed the oauth/token_info issue using Basic Auth.

mostafa6765:
@jlowin Thanks — updated.

* Switched to introspection-first (security checks before userinfo)
* Enforced strict `active` handling (missing ⇒ reject per RFC 7662)
* Made userinfo after introspection (enrichment only)
* Adjusted tests accordingly

PR #3736, opened 2026-04-01
fix: FastAPI TestClient compatibility and lifespan re-initialization

### LINKED ISSUE #2375: FastMCP breaks FastAPI Testclient
### Description

Mounting an FastMCP `http_app` leads to errors when testing the FastAPI endpoints using `fastapi.testclient.TestClient`.

Consider the following minimal app in `example_api/app.py`:
```python
from contextlib import asynccontextmanager
from fastapi import FastAPI

from fastmcp import FastMCP
    
@asynccontextmanager
async def lifespan(app: FastAPI):
    """Define FastAPI lifespan."""
    print('Starting FastAPI app')
    try:
        yield
    finally:
        print('Stopping FastAPI app')

mcp = FastMCP('tools')
mcp_app = mcp.http_app(path='/mcp', transport='http')

@mcp.tool
def add_two_numbers(first: float, second: float) -> float:
    """Add two numbers."""
    return first + second

@asynccontextmanager
async def combined_lifespan(app: FastAPI):
    """Run both lifespans."""
    async with lifespan(app):
        async with mcp_app.lifespan(app):
            yield

app = FastAPI(lifespan=combined_lifespan)

@app.get('/health')
def health():
    """Return health status."""
    return {'status': 'ok'}

app.mount('/analytics', mcp_app)
```

and a test like this:
```python
from fastapi.testclient import TestClient
from example_app.app import app

def test_health():
    with TestClient(app) as client:
        response = client.get('/health')
        assert response.status_code == 200

def test_health2():
    with TestClient(app) as client:
        response = client.get('/health')
        assert response.status_code == 200
```
Running the tests with `pytest test_app.py` leads to the following error:

`RuntimeError: StreamableHTTPSessionManager .run() can only be called once per instance. Create a new instance if you need to run again.`

The bug has already been found in this issue: https://github.com/jlowin/fastmcp/issues/946 and in a linked blog a solution was proposed. Changing the app to following fixes the error:

```python
from contextlib import asynccontextmanager
from fastapi import FastAPI

from fastmcp import FastMCP

class CustomFastMCP():
    def __init__(self, mcp):
        self.mcp = mcp
    
    def create_app(self):
        self.app = self.mcp.http_app(path='/mcp', transport='http')
        self.generator = self.app.lifespan(self.app)
        self.generator_started = False
        return self.app
    
    async def start_session(self):
        if self.generator_started:
            return
        await self.generator.__aenter__()
        self.generator_started = True
    
@asynccontextmanager
async def lifespan(app: FastAPI):
    """Define FastAPI lifespan."""
    print('Starting FastAPI app')
    try:
        yield
    finally:
        print('Stopping FastAPI app')

mcp = FastMCP('tools')
custom = CustomFastMCP(mcp)

@mcp.tool
def add_two_numbers(first: float, second: float) -> float:
    """Add two numbers."""
    return first + second

@asynccontextmanager
async def combined_lifespan(app: FastAPI):
    """Run both lifespans."""
    async with lifespan(app):
        await custom.start_session() 
        yield

app = FastAPI(lifespan=combined_lifespan)

@app.get('/health')
def health():
    """Return health status."""
    return {'status': 'ok'}

app.mount('/analytics', custom.create_app())
```

### Example Code

```Python

```

### Version Information

```Text
FastMCP version:                                                                          2.13.0.2
MCP version:                                                                                1.20.0
Python version:                                                                            3.10.18
Platform:                                                                Windows-10-10.0.19044-SP0
```

kvdhanush06:
## Description

Closes #2375

`FastMCP` raised a `RuntimeError` when using FastAPI's `TestClient` across multiple test runs. The root cause was `StreamableHTTPSessionManager` enforcing a strict single-execution constraint on its lifespan. Since `TestClient` triggers the full application lifespan on every context entry, each subsequent `with TestClient(app)` block attempted to re-enter an already-consumed manager, causing the failure.

**Root Cause**

`StreamableHTTPSessionManager` is designed to be run once. Re-entering its lifespan raises a `RuntimeError`. Because `TestClient` spins up and tears down the ASGI lifespan on every `with` block, any test suite with more than one `TestClient` context would fail on the second invocation.

**Fix**

Introduced `ReusableSessionManagerWrapper` in `src/fastmcp/server/http.py`. The wrapper accepts a factory callable instead of a manager instance directly. Each time the lifespan is entered, it lazily constructs a fresh `StreamableHTTPSessionManager` via the factory, ensuring every test run — and every lifespan cycle in general — operates on a correctly initialized instance.

No changes to the public API or existing behavior for single-run scenarios.

**Verification**

Added a regression suite in `tests/server/test_fastapi_testclient_compat.py` covering:
- Multiple sequential `TestClient` context entries
- Correct teardown and re-initialization between runs
```python
# Previously raised RuntimeError on the second block — now works correctly
with TestClient(app) as client:
    client.get("/health")

with TestClient(app) as client:
    client.get("/health")
```

## Contribution type

## Checklist

### DIFF
--- a/src/fastmcp/server/http.py
+++ b/src/fastmcp/server/http.py
@@ -32,11 +32,15 @@ logger = get_logger(__name__)
 class StreamableHTTPASGIApp:
     """ASGI application wrapper for Streamable HTTP server transport."""
 
-    def __init__(self, session_manager):
+    def __init__(self, session_manager: StreamableHTTPSessionManager | None):
         self.session_manager = session_manager
 
     async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:
         try:
+            if self.session_manager is None:
+                raise RuntimeError(
+                    "Task group is not initialized. Make sure to use run()."
+                )
             await self.session_manager.handle_request(scope, receive, send)
         except RuntimeError as e:
             if str(e) == "Task group is not initialized. Make sure to use run().":
@@ -262,37 +266,6 @@ def create_sse_app(
     return app
 
 
-class ReusableSessionManagerWrapper:
-    """Wrapper for StreamableHTTPSessionManager that supports multiple runs.
-
-    FastAPI's TestClient can trigger an app's lifespan multiple times, but
-    the underlying StreamableHTTPSessionManager can only be run once.
-    This wrapper recreates the manager for each run.
-    """
-
-    def __init__(self, factory: Callable[[], StreamableHTTPSessionManager]):
-        self._factory = factory
-        self._current_manager: StreamableHTTPSessionManager | None = None
-
-    @asynccontextmanager
-    async def run(self) -> AsyncGenerator[None, None]:
-        # Create a new manager instance for this run
-        self._current_manager = self._factory()
-        try:
-            async with self._current_manager.run():
-                yield
-        finally:
-            # Clear the manager after run, even if an exception occurred
-            self._current_manager = None
-
-    async def handle_request(self, scope: Scope, receive: Receive, send: Send) -> None:
-        if self._current_manager is None:
-            # Re-raise the error that StreamableHTTPSessionManager would raise
-            # if it wasn't running.
-            raise RuntimeError("Task group is not initialized. Make sure to use run().")
-        await self._current_manager.handle_request(scope, receive, send)
-
-
 def create_streamable_http_app(
     server: FastMCP[LifespanResultT],
     streamable_http_path: str,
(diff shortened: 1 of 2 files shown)


kvdhanush06:
@jlowin Thanks for the feedback. I've simplified the fix by moving the `StreamableHTTPSessionManager `creation directly into the lifespan, removing the wrapper and factory entirely. The regression tests have also been updated to the simplified status as requested. Ready for another look!

PR #3750, opened 2026-04-03
fix: propagate upstream_claims in load_access_token

### LINKED ISSUE #3723: load_access_token discards upstream_claims embedded by _extract_upstream_claims
## Summary

`_extract_upstream_claims` (added in #2997) allows subclasses to embed custom claims in the FastMCP JWT under the `upstream_claims` key. These claims are correctly signed into the JWT at issuance. However, `load_access_token` doesn't surface them during validation — it swaps the JWT for the upstream provider token and returns that, losing the `upstream_claims`.

## Expected Behavior

Claims returned by `_extract_upstream_claims` should be accessible on the `AccessToken` returned by `load_access_token`, since they were signed into the JWT and verified during validation.

## Actual Behavior

`load_access_token`:
1. Verifies the FastMCP JWT signature (payload includes `upstream_claims`) ✅
2. Extracts `jti` from the payload
3. Looks up the upstream token via JTI mapping
4. Validates the upstream token → returns `AccessToken` with upstream provider claims only
5. `upstream_claims` from step 1 is discarded ❌

## Reproduction

1. Subclass `OIDCProxy` and override `_extract_upstream_claims` to return custom claims
2. Authenticate via OAuth — the FastMCP JWT correctly contains `upstream_claims`
3. Make an MCP tool call — `load_access_token` validates the JWT but the returned `AccessToken.claims` only has the upstream provider's claims, not the `upstream_claims` from the FastMCP JWT

## Workaround

Override `load_access_token` to re-decode the already-verified JWT and merge `upstream_claims` into the `AccessToken`:

```python
async def load_access_token(self, token: str) -> AccessToken | None:
    access_token = await super().load_access_token(token)
    if access_token is None:
        return None
    payload = pyjwt.decode(token, options={"verify_signature": False})
    upstream_claims = payload.get("upstream_claims")
    if upstream_claims:
        access_token.claims["upstream_claims"] = upstream_claims
    return access_token
```

## Suggested Fix

In `OAuthProxy.load_access_token`, after verifying the JWT and looking up the upstream token, merge `upstream_claims` from the verified JWT payload into the returned `AccessToken.claims`. This makes `_extract_upstream_claims` useful for both external JWT consumers (API gateways) and internal FastMCP tool dependencies.

kvdhanush06:
## Description

Closes #3723

FastMCP's `OAuthProxy` discarded custom claims embedded in the JWT under the `upstream_claims` key during the `load_access_token` process. While these claims were correctly signed into the JWT at issuance (via `_extract_upstream_claims`), they were lost during the token swap where the FastMCP JWT is exchanged for the upstream provider's token.

**Root Cause**

The `load_access_token` method verified the FastMCP JWT to retrieve the `jti` for mapping lookup but failed to preserve the verified payload. It then validated the upstream token and returned a fresh `AccessToken` object that only contained the provider's claims, losing the contextually relevant `upstream_claims` that the server had previously verified and signed.

**Fix**

Updated `load_access_token` in `src/fastmcp/server/auth/oauth_proxy/proxy.py` to:
1. Extract `upstream_claims` from the verified FastMCP JWT payload.
2. Merge these claims into the returned `AccessToken.claims` dictionary.

This ensures that any custom data intended to survive the token swap (e.g. cross-provider user IDs or internal metadata) is accessible to the downstream MCP server.

**Verification**

Added a regression test suite `TestUpstreamClaimsPropagation` in `tests/server/auth/test_oidc_proxy_token.py` covering:
- Correct extraction of `upstream_claims` from the FastMCP JWT.
- Successful merging of these claims into the final `AccessToken` after the upstream token swap.
```python
# The regression test ensures the final AccessToken includes the propagated claims
result = await proxy.load_access_token(fastmcp_jwt)
assert "upstream_claims" in result.claims
assert result.claims["upstream_claims"] == {"sub": "idp-user-123"}
```

## Contribution type

## Checklist

### DIFF
--- a/tests/server/auth/test_oidc_proxy_token.py
+++ b/tests/server/auth/test_oidc_proxy_token.py
@@ -1,8 +1,4 @@
-"""Tests for OIDC Proxy token management and propagation.
-
-These tests cover the OIDCProxy's ability to issue, verify, and swap tokens
-between FastMCP and upstream identity providers.
-"""
+"""Tests for OIDC Proxy verify_id_token functionality."""
 
 from unittest.mock import patch
 


kvdhanush06:
@jlowin I've refactored the tests as requested. They've been moved to `test_tokens.py` and simplified using the `TestTransparentUpstreamRefresh` pattern with real in-memory stores, which eliminates the OIDC boilerplate.

I've also:
1. Implemented the `_setup_session_with_claims` helper and the two concise test cases.
2. Removed the redundant `cast(AccessToken, result)` calls and `@pytest.mark.asyncio` decorators.
3. Verified that the propagation correctly uses `model_copy(deep=True)` to prevent cached object mutation.

Tests and static checks (prek) are passing. Ready for another look!

PR #3754, opened 2026-04-03
fix: remove CSP from tool metadata, keep on resource only

### LINKED ISSUE #3735: CSP should be set for an App's resource instead of tool
### Enhancement

Hi, as a follow-up to https://github.com/PrefectHQ/fastmcp/issues/3623, in `_maybe_apply_prefab_ui` the `meta.get("ui")` does return the specified CSP now for the **tool**'s metadata, but looking at the [MCP Apps spec](https://github.com/modelcontextprotocol/ext-apps/blob/main/specification/2026-01-26/apps.mdx#resource-discovery), the `McpUiToolMeta` should only have `resourceUri` and `visibility` fields, not `csp`. 

`csp` only exists in `UIResourceMeta` (See [UI Resource format](https://github.com/modelcontextprotocol/ext-apps/blob/main/specification/2026-01-26/apps.mdx#ui-resource-format)), so we should be setting it in `_ensure_prefab_renderer` for the UI `TextResource` we return, and not be returning any `csp` for the tool.

Let me know if that makes sense, thanks.

jlowin:
The MCP Apps spec defines `McpUiToolMeta` as `{resourceUri, visibility}` only — CSP belongs on `UIResourceMeta`, not the tool. We were putting CSP on both the tool and the renderer resource; the resource was already correct, so this just strips the redundant CSP from the tool's `_meta.ui`.

Two call sites: `_expand_prefab_ui_meta` (auto-wired prefab tools) and `FastMCPApp.ui()` (explicit app tools). Both now build `AppConfig` with only `resource_uri`/`visibility`, matching the spec's tool interface.

Note this is technically a breaking change though unused and undocumented. 

Closes #3735

### DIFF
--- a/tests/test_fastmcp_app.py
+++ b/tests/test_fastmcp_app.py
@@ -246,7 +246,8 @@ class TestAppUI:
         assert meta is not None
         assert meta["ui"]["resourceUri"] == "ui://prefab/renderer.html"
 
-    async def test_ui_has_csp(self):
+    async def test_ui_tool_has_no_csp(self):
+        """CSP belongs on the UI resource, not the tool (per MCP Apps spec)."""
         app = FastMCPApp("test")
 
         @app.ui()
@@ -256,8 +257,7 @@ class TestAppUI:
         tools = await app._list_tools()
         meta = tools[0].meta
         assert meta is not None
-        csp = meta["ui"].get("csp")
-        assert csp is not None
+        assert "csp" not in meta["ui"]
 
     async def test_ui_with_title_and_description(self):
         app = FastMCPApp("test")


Mars9934:
Hi @jlowin , not sure if this change is working quite right. In MCP Inspector, I still see the `csp` being set for the **Tool**:
image

And more importantly, still not seeing the customized `frameDomains` for the **Resource**:
image
I think it's cause `_ensure_prefab_renderer` isn't getting `tool.meta` so it only has default CSP from `get_renderer_csp`.

For reference, my tool declaration is:
```
@mcp.tool(app=PrefabAppConfig(csp=ResourceCSP(frame_domains=["https://example1234.com"])))
```

PR #3761, opened 2026-04-04
feat: auto-generate tool catalog in instructions

terry-li-hm:
## Summary

Adds an opt-in `auto_catalog` parameter to `FastMCP.__init__` that auto-appends a concise tool summary to the `instructions` field in the `initialize` response.

- **New parameter:** `FastMCP(auto_catalog=True)` -- defaults to `False` (fully backwards-compatible)
- **Catalog format:** `\n\nAvailable tools:\n  tool_name: First sentence of description.`
- **Budget-aware:** Truncates with `...` to stay within 2KB (the cap Claude Code enforces on instructions)
- **Sorted:** Tools listed alphabetically for stable, predictable output
- **Provider-aware:** Collects tools from all `LocalProvider` instances, including those added via `providers=[]`

## Motivation

The MCP spec's `instructions` field in the `initialize` response is designed to help LLMs understand what a server offers. Claude Code surfaces it in the system prompt. However:

1. **Server authors rarely write useful instructions** -- most leave it as `None` or a generic string
2. **FastMCP treats it as a static string** -- it doesn't reflect what tools are actually registered
3. **When tools are deferred/lazy-loaded**, the LLM only sees tool names without descriptions, making tools undiscoverable until explicitly listed

With `auto_catalog=True`, a server like:

```python
mcp = FastMCP("my-server", instructions="Data analysis tools", auto_catalog=True)

@mcp.tool(description="Run a SQL query against the database. Supports SELECT only.")
def query(sql: str) -> str: ...

@mcp.tool(description="List all available tables.")
def list_tables() -> str: ...
```

produces this `instructions` in the initialize response:

```
Data analysis tools

Available tools:
  list_tables: List all available tables.
  query: Run a SQL query against the database.
```

## Implementation

- `FastMCP.__init__` accepts `auto_catalog: bool = False`
- `FastMCP._collect_tools_sync()` walks the provider tree synchronously (no async context needed)
- `FastMCP._build_tool_catalog()` builds the catalog string with budget-aware truncation
- `LowLevelServer.create_initialization_options()` appends the catalog to `instructions` at server start
- 9 new tests covering: default off, appending, no base instructions, no tools, truncation, no-description tools, provider tools, alphabetical sort, first-sentence extraction

## Test plan

- [x] All 9 new tests pass (`tests/server/test_auto_catalog.py`)
- [x] All 26 existing server tests pass (`tests/server/test_server.py`)
- [x] All 10 client initialization tests pass (`tests/client/client/test_initialize.py`)
- [ ] Review: verify catalog renders correctly with mounted sub-servers

### DIFF
--- a/src/fastmcp/server/low_level.py
+++ b/src/fastmcp/server/low_level.py
@@ -182,6 +182,15 @@ class LowLevelServer(_Server[LifespanResultT, RequestT]):
         # ensure we use the FastMCP notification options
         if notification_options is None:
             notification_options = self.notification_options
+
+        # Auto-catalog: append a concise tool summary to instructions so LLMs
+        # can discover tools from the initialize response alone.
+        if self.fastmcp._auto_catalog:
+            catalog = self.fastmcp._build_tool_catalog()
+            if catalog is not None:
+                original = self.instructions or ""
+                self.instructions = original + catalog
+
         return super().create_initialization_options(
             notification_options=notification_options,
             experimental_capabilities=experimental_capabilities,
--- a/src/fastmcp/server/server.py
+++ b/src/fastmcp/server/server.py
@@ -274,6 +274,7 @@ class FastMCP(
         tools: Sequence[Tool | Callable[..., Any]] | None = None,
         on_duplicate: DuplicateBehavior | None = None,
         mask_error_details: bool | None = None,
+        auto_catalog: bool = False,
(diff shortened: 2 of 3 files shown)


jlowin:
I do not think this is good at a framework level, and it conflicts with dynamic providers in any case. Users for whom it is helpful have all the tools they need to implement it for their servers.

PR #3768, opened 2026-04-06
Fix OpenAPI 3.0 nullable fields in tool input schemas

### LINKED ISSUE #3744: OpenAPI 3.0: nullable fields in input parameter schemas not converted to JSON Schema
### What happened?

When using `FastMCP.from_openapi()` with an OpenAPI 3.0 spec, `nullable: true` on input parameter fields is passed through verbatim into the MCP tool's `inputSchema`. This is invalid JSON Schema.. nullable is an OpenAPI 3.0 extension, not a JSON Schema keyword.

Output schemas (responses) are correctly converted via `convert_openapi_schema_to_json_schema()`, but input schemas (`flat_param_schema`) are not.

For example, `{"type": "string", "nullable": true}` should become `{"type": ["string", "null"]}` in the tool's input schema, but it doesn't.

This causes downstream validation errors in MCP clients (e.g. LangChain) when tools accept nullable parameters.

### Example Code

```Python
"""
Repro: nullable fields in input schemas are not converted from OpenAPI 3.0 to JSON Schema.

Output schemas are correctly converted (nullable: true -> type: ["string", "null"]),
but input schemas (flat_param_schema) are not.

Tested on fastmcp 2.14.2 and 3.2.0.
"""

import asyncio
import json

from fastmcp import FastMCP
from fastmcp.server.openapi import MCPType, RouteMap

OPENAPI_SPEC = {
    "openapi": "3.0.3",
    "info": {"title": "Test API", "version": "1.0.0"},
    "servers": [{"url": "http://localhost:9999"}],
    "paths": {
        "/items": {
            "post": {
                "operationId": "createItem",
                "summary": "Create an item",
                "tags": ["test"],
                "requestBody": {
                    "required": True,
                    "content": {
                        "application/json": {
                            "schema": {
                                "type": "object",
                                "properties": {
                                    "name": {"type": "string"},
                                    "description": {
                                        "type": "string",
                                        "nullable": True,
                                    },
                                },
                                "required": ["name"],
                            }
                        }
                    },
                },
                "responses": {
                    "200": {
                        "description": "OK",
                        "content": {
                            "application/json": {
                                "schema": {
                                    "type": "object",
                                    "properties": {
                                        "description": {
                                            "type": "string",
                                            "nullable": True,
                                        },
                                    },
                                }
                            }
                        },
                    }
                },
            }
        }
    },
}


async def main():
    mcp = FastMCP.from_openapi(
        openapi_spec=OPENAPI_SPEC,
        client=None,
        name="Test",
        route_maps=[RouteMap(mcp_type=MCPType.TOOL, mcp_tags={"openapi"})],
    )

    tools = await mcp.get_tools()
    tool = list(tools.values())[0]

    input_schema = tool.parameters
    desc = input_schema["properties"]["description"]

    print("Input schema 'description' field:")
    print(json.dumps(desc, indent=2))

    # Bug: nullable leaks into input schema as-is
    assert "nullable" not in desc, (
        f"nullable leaked into input schema: {desc}\n"
        f'Expected: {{"type": ["string", "null"]}}'
    )


asyncio.run(main())
```

### Version Information

```Text
fastmcp: 2.14.2
mcp: 1.26.0
Python: 3.11.13
OS: macOS-26.4-arm64

Also tested on fastmcp 3.2.0, same behavior
```

kvdhanush06:
## Description

This PR fixes a bug where `nullable: true` from OpenAPI 3.0 specifications was leaking into the input schemas of MCP tools. In JSON Schema, `nullable` is not a valid keyword and must be converted to a type union (e.g., `type: ["type", "null"]`) to ensure compatibility with downstream MCP clients like LangChain.

The fix involves:
- Improving the central `convert_openapi_schema_to_json_schema` utility to handle `$defs` and `$definitions` recursively.
- Applying this utility to the final combined input schema in `_combine_schemas_and_map_params`.

Closes #3744

## Contribution type

## Checklist

### DIFF
--- a/src/fastmcp/utilities/openapi/json_schema_converter.py
+++ b/src/fastmcp/utilities/openapi/json_schema_converter.py
@@ -60,7 +60,6 @@ def convert_openapi_schema_to_json_schema(
         remove_read_only: Whether to remove readOnly properties
         remove_write_only: Whether to remove writeOnly properties
         convert_one_of_to_any_of: Whether to convert oneOf to anyOf
-        convert_one_of_to_any_of: Whether to convert oneOf to anyOf
 
     Returns:
         JSON Schema-compatible dictionary
@@ -219,7 +218,7 @@ def _needs_recursive_processing(
             if field_type is dict and isinstance(schema[field_name], dict):
                 if field_name in ("properties", "$defs", "$definitions"):
                     # Check if any schema in the map needs conversion
-                    for _name, sub_schema in schema[field_name].items():
+                    for sub_schema in schema[field_name].values():
                         if isinstance(sub_schema, dict):
                             nested_needs_conversion = (
                                 any(
--- a/tests/utilities/openapi/test_nullable_fields.py
+++ b/tests/utilities/openapi/test_nullable_fields.py
@@ -1,8 +1,12 @@
 """Tests for nullable field handling in OpenAPI schemas."""
 
+import httpx
 import pytest
 from jsonschema import ValidationError, validate
 
+from fastmcp import FastMCP
+from fastmcp.client import Client
+from fastmcp.server.providers.openapi import OpenAPIProvider
 from fastmcp.utilities.openapi.json_schema_converter import (
     convert_openapi_schema_to_json_schema,
 )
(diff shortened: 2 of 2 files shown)


jlowin (changes requested):
Thanks for the fix. The diagnosis is correct: input schemas from `_combine_schemas_and_map_params` need to go through `convert_openapi_schema_to_json_schema`, and extending the recursion to cover `$defs`/`$definitions` is the right call.

A few things to address:

**Tests.** The checklist says tests were added but there are none in the diff. This needs a regression test: an OpenAPI 3.0 spec with `nullable: true` on an input parameter, asserting the tool's input schema produces `type: ["string", "null"]` instead of `nullable: true`. A second case covering a nullable field inside `$defs` would be good too.

**Local import.** The import in `schemas.py` needs to move to module level. Project convention is all imports at the top of the file, no local imports inside functions.

**Memoization/cycle detection.** The `_memo` and `_visited` machinery adds a lot of surface area. OpenAPI 3.0 specs don't produce circular `$defs`; that's a JSON Schema draft 2019+ concern, and by that point `nullable` doesn't exist anyway. Please drop the `_memo` parameter from `convert_openapi_schema_to_json_schema` and the `_visited` parameter from `_needs_recursive_processing` unless you can point to a real OpenAPI 3.0 spec that creates circular references. The `$defs`/`$definitions` recursion itself should stay.

Happy to re-review once those are addressed.

kvdhanush06:
@jlowin Thanks for the feedback. I've addressed the comments by simplifying the schema conversion logic and following the project's import conventions.

I've also:
1. Removed the _memo and _visited machinery for cycle detection and memoization, as circular references are not a concern for OpenAPI 3.0 in this context.
2. Added regression tests in test_nullable_fields.py that specifically cover nullable: true on input parameters and within $defs.
3. Moved the convert_openapi_schema_to_json_schema import in schemas.py to the module level.
4. Verified that the conversion correctly produces JSON Schema type unions (e.g., type: ["string", "null"]) for downstream MCP compatibility.

Tests and static checks are passing. Ready for another look!

jlowin (approved):
Thanks for the follow-up — the memoization removal, module-level import, and `$defs` tests all look good.

I pushed a small cleanup commit to your branch:

- Removed the duplicated docstring line for `convert_one_of_to_any_of`
- Reverted `.items()` back to `.values()` in `_needs_recursive_processing` since the key isn't used
- Added two integration tests that exercise the full pipeline (OpenAPI spec → `OpenAPIProvider` → `tool.inputSchema`) to confirm `nullable: true` doesn't leak through on input parameters or `$defs` — this is the regression path the unit tests alone wouldn't catch

LGTM with those changes.

Issue #3760, opened 2026-04-04
Recursive outputSchema from native Pydantic tool crashes ListToolsResult serialization


xingyug:
## Description

Recursive Pydantic models used as native tool output types cause `ListToolsResult.model_dump()` to crash with `ValueError: Circular reference detected (id repeated)` during `tools/list` serialization. This makes **all tools unavailable** — not just the one with the recursive model.

## Minimal Reproduction

```python
from fastmcp import FastMCP
from pydantic import BaseModel

mcp = FastMCP("repro")

class TreeNode(BaseModel):
    name: str
    children: list["TreeNode"] = []

@mcp.tool()
def get_tree() -> TreeNode:
    return TreeNode(name="root")

if __name__ == "__main__":
    mcp.run(transport="sse")
```

Connect any MCP client and send `tools/list`. The server crashes during response serialization.

## Environment

- fastmcp: 2.14.6
- mcp: 1.27.0
- pydantic: 2.12.5
- Python: 3.12

## Expected Behavior

Recursive `$ref` in the generated output schema should be preserved as-is or safely handled. `tools/list` should return the tool definitions without crashing.

## Actual Behavior

```
ValueError: Circular reference detected (id repeated)

  File ".../mcp/shared/session.py", line 346, in _send_response
    result=response.model_dump(by_alias=True, mode="json", exclude_none=True)
  File ".../pydantic/main.py", line 464, in model_dump
    return self.__pydantic_serializer__.to_python(...)
```

## Root Cause Analysis

1. Pydantic generates a valid recursive JSON Schema with `$ref: "#/$defs/TreeNode"` for the `children` field
2. FastMCP's `compress_schema(dereference=True)` calls `jsonref.replace_refs(proxies=False)` on the output schema
3. `jsonref` resolves the recursive `$ref` into an actual Python dict circular reference (verified: `id(children_items_at_depth_2) == id(schema_at_depth_1)`)
4. When `ListToolsResult.model_dump()` serializes the tool definitions, Pydantic hits the circular dict and raises `ValueError`

The crash happens in the schema processing path, not during tool execution. All tools become unavailable because `ListToolsResult` serialization fails for the entire response.

## Relationship to Existing Issues

[...]

PR #3774, opened 2026-04-06
fix: remove dereference=True while calling compress_schema

Jazzcort:
## Description
This addresses the issues found in #3760. Currently, if a self-referencing Pydantic data type is buried in a nested field, calling dereference_refs() introduces a circular reference to the final schema. This results in a serialization error that crashes the server during tools/list requests.

I checked v3.2.0 and noticed that dereference=True is no longer being passed to compress_schema(), which essentially reverts #3170. We should implement this same behavior for the 2.x versions. While the community will eventually migrate to 3.x, I think we still need to pull this fix in now because the bug causes a complete server crash whenever tools/list is called.

## Contribution type

## Checklist

- [ ] If I used an LLM, it followed the repo's contributing conventions (not generic output)

### DIFF
--- a/src/fastmcp/tools/tool.py
+++ b/src/fastmcp/tools/tool.py
@@ -500,7 +500,6 @@ class ParsedFunction:
             input_schema,
             prune_params=prune_params,
             prune_titles=True,
-            dereference=True,
         )
 
         output_schema = None
@@ -561,7 +560,8 @@ class ParsedFunction:
                     output_schema = base_schema
 
                 output_schema = compress_schema(
-                    output_schema, prune_titles=True, dereference=True
+                    output_schema,
+                    prune_titles=True,
                 )
 
                 # Resolve root-level $ref to meet MCP spec requirement for type: object
--- a/tests/tools/test_tool.py
+++ b/tests/tools/test_tool.py
@@ -194,15 +194,18 @@ class TestToolFromFunction:
                 "tags": set(),
                 "enabled": True,
                 "parameters": {
-                    "properties": {
-                        "user": {
+                    "$defs": {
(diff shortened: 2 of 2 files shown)


Jazzcort:
Hey @jlowin, let me know what do you think about this. Do you think this is worth the fix? I can start to fix the test cases if you agree with this fix. Otherwise, you can just close this PR. 😁

jlowin:
Thanks for digging into this! The bug is real — recursive models nested in containers do break serialization.

The history here is that dereferencing was added specifically because many MCP clients (Claude, Copilot, etc.) don't resolve `$ref`/`$defs` in schemas — there are a bunch of issues documenting how contentious this was. So removing it outright fixes the recursive case but breaks the more common case.

In 3.x we solved this properly: the `dereference=True` keyword was removed from `compress_schema()` exactly where you indicate, but dereferencing was promoted to its own middleware with a public API flag — `FastMCP(dereference_schemas=False)` lets users opt out. The middleware also handles circular schemas gracefully instead of crashing.

Backporting that middleware to 2.x would be a fairly involved change. Since 3.x already addresses this (and a lot more), I'd recommend upgrading if you can — it's a much better experience overall. Happy to help if you run into any migration questions.

Jazzcort:
@jlowin Thank you for the detailed walkthrough of this bug! We currently have fastmcp pinned to 2.x just to streamline the process for our current release. However, we are definitely planning to upgrade to 3.x in a future release—the new features and improvements will really help simplify our implementation.

Thanks for maintaining such an awesome project! We'll be sure to contribute back if we encounter any issues or think of potential improvements while using it.

PR #3811, opened 2026-04-10
MCP Events: Auto-set source field from tool context

elijahr:
## Summary
- Adds `_tool_name` parameter to `Context.__init__`, threaded from `call_tool` dispatch
- `Context.emit_event` auto-sets `source` to `"tool/<tool_name>"` when called from a tool and no explicit source is provided
- Explicit source parameter always takes precedence over auto-source

## Test plan
- [ ] 5 new tests covering tool_name property, auto-source, explicit override, non-tool context
- [ ] All 95 event tests pass
- [ ] `uv run pytest tests/server/test_events.py -x`


elijahr:
Opened against wrong repo. This should be on axiomantic/fastmcp.

PR #3814, opened 2026-04-10
Enhance description parameter documentation

### LINKED ISSUE #2652: Dynamic Tool Descriptions
### Enhancement

This may already be possible, but I don't see any references in the docs.  

I would like to be able to dynamically change the tool descriptions based on several runtime factors.   

**Examples:**

- Providing user specific details (EG a `refresh_data` tool providing a timestamp of when the data was last refreshed)
- Removing mentions of other tools if they are not enabled for the configuration / user
- A `query_data` tool providing table references that the current user has access to

With the doc string descriptions I don't think this is possible.  Maybe it can be done with the annotations?  

If it is possible, it would maybe be nice to add some documentation around it, as it should be a common use case.

jordimassaguerpla:
## Description

Added information about dynamically setting the description in MCP.

Closes #2652 

## Contribution type

- [X] Documentation improvement

## Checklist

- [X] This PR addresses an existing issue (or fixes a self-evident bug)
- [X] I have read [CONTRIBUTING.md](../CONTRIBUTING.md)
- [ ] I have added tests that cover my changes
- [ ] If I used an LLM, it followed the repo's contributing conventions (not generic output)

### DIFF
--- a/docs/servers/tools.mdx
+++ b/docs/servers/tools.mdx
@@ -70,7 +70,7 @@ def search_products_implementation(query: str, category: str | None = None) -> l
 </ParamField>
 
 <ParamField body="description" type="str | None">
-  Provides the description exposed via MCP. If set, the function's docstring is ignored for this purpose
+  Provides the description exposed via MCP. If set, the function's docstring is ignored for this purpose. You can use this to dynamically set the description.
 </ParamField>
 
 <ParamField body="tags" type="set[str] | None">


jlowin:
Thanks for picking this up. I think the original issue was effectively answered in the thread — construction-time strings already work dynamically, and per-call dynamism is a custom Provider pattern. A one-line note on the parameter doesn't really add that context, so I'd rather leave it.

PR #3822, opened 2026-04-10
Bump pydocket>=0.19.0, drop fakeredis pin

chrisguidry:
pydocket 0.19.0 fixes the fakeredis 2.35.0 `FakeConnection` rename internally (https://github.com/chrisguidry/docket/pull/382), so we no longer need to carry the `fakeredis[lua]<2.35.0` ceiling in the `tasks` extra. This simplifies the dependency to just `pydocket>=0.19.0` and lets fakeredis float as a transitive.

Also bumps the `_MIN_DOCKET_VERSION` floor to match.

Follows up on #3804.

🤖 Generated with Claude Code


jlowin (approved):
👍

PR #3826, opened 2026-04-10
Add real-world schema crash test (232K schemas from APIs.guru)

strawgate:
## Summary

Adds an integration test that runs `json_schema_to_type` against every schema definition in the [APIs.guru openapi-directory](https://github.com/APIs-guru/openapi-directory) — 232,092 schemas from 4,120 real-world OpenAPI specs.

The test snapshots crash counts as regression baselines so future changes can't silently increase the crash rate. As we fix crash patterns, we ratchet the baselines down.

```
Schemas tested:  232,092
TypeErrors:      2,342  (e.x. datetime serialization in defaults)
SchemaErrors:      273  (e.x. invalid regexes in specs — not our code)
Timeouts:            0
Other errors:        0
Total crashes:   2,615  (1.13%)
```

**Skipped by default** — requires a local clone of the openapi-directory repo. Run with:

```bash
git clone --depth 1 https://github.com/APIs-guru/openapi-directory /tmp/openapi-directory
uv run pytest tests/utilities/json_schema_type/test_real_world_schemas.py -m integration -s
```

Takes ~13 minutes on a MacBook Pro. Each schema gets a 5-second SIGALRM timeout to catch infinite loops.

### DIFF
--- a/.github/workflows/run-schema-crash-test.yml
+++ b/.github/workflows/run-schema-crash-test.yml
@@ -1,0 +1,55 @@
+name: Schema Crash Test
+
+on:
+  push:
+    branches: ["main"]
+    paths:
+      - "src/fastmcp/utilities/json_schema_type.py"
+      - "src/fastmcp/utilities/json_schema.py"
+      - "src/fastmcp/utilities/openapi/**"
+      - "src/fastmcp/server/providers/openapi/**"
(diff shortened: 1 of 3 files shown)


jlowin (approved):
Well-structured crash regression suite — good isolation from the main test run, and the YAML type normalization fix is a nice bonus.

PR #3835, opened 2026-04-11
fix: accept Azure v1 tokens in AzureProvider and AzureJWTVerifier

albertogallensf:
## Problem

When `AzureProvider` is configured to request tokens scoped to a first-party Microsoft resource (e.g. Dynamics 365 Business Central with scope `https://api.businesscentral.dynamics.com/.default`), Azure emits **v1-format tokens**. These tokens differ from v2 tokens in three ways that cause validation to fail:

1. **Issuer** is `https://sts.windows.net/{tid}/` instead of `https://login.microsoftonline.com/{tid}/v2.0`
2. **Audience** is `api://{client_id}` instead of the bare `{client_id}` — and when `identifier_uri` is a custom URI (not the default `api://{client_id}`), the `api://{client_id}` format is not in the accepted audience list
3. **Scopes**: the `.default` scope does not appear in the `scp` claim, so if no other non-OIDC scopes are configured, validation raises `ValueError` before even checking the token

Microsoft documents this behaviour: [Access token formats](https://learn.microsoft.com/en-us/entra/identity-platform/access-tokens#token-formats).

### 1. Issuer v1 support

Pass both v2 and v1 issuer URLs as a list to `JWTVerifier` (which already supports `issuer: str | list[str]`):

\`\`\`python
issuer: list[str] = [
    f"https://{base_authority}/{tenant_id}/v2.0",
    f"https://sts.windows.net/{tenant_id}/",
]
\`\`\`

### 2. Audience v1 fallback

Explicitly include `api://{client_id}` alongside `client_id` and `identifier_uri`:

\`\`\`python
audience: list[str] = [client_id, f"api://{client_id}"]
if self.identifier_uri not in audience:
    audience.append(self.identifier_uri)
\`\`\`

This extends the existing fix from #3797 to also cover the case where `identifier_uri` is a custom URI different from `api://{client_id}`.

### 3. Scope validation relaxation

When all requested scopes are OIDC scopes (or the list is empty after filtering), skip scope enforcement instead of raising `ValueError`. The token is still validated by issuer and audience. This is common when the only scope requested from Azure is `<resource>/.default`.

All three fixes applied to both `AzureProvider` and `AzureJWTVerifier`.

## How we discovered this

We are building an MCP server that connects Claude.ai to Dynamics 365 Business Central via FastMCP's `AzureProvider`. The app registration uses `identifier_uri="https://api.businesscentral.dynamics.com"` and requests scope `https://api.businesscentral.dynamics.com/.default`. Azure returns v1-format tokens that FastMCP rejected on all three counts.

## Changes

- `src/fastmcp/server/auth/providers/azure.py` — issuer as list, audience with `api://` fallback, scope validation relaxation (both classes)
- `tests/server/auth/providers/test_azure.py` — updated issuer/audience assertions
- `tests/server/auth/providers/test_azure_scopes.py` — updated issuer assertions, scope tests now verify enforcement is skipped, added `test_validates_token_with_v1_issuer`

## Test plan

- [x] All 77 Azure provider tests pass (76 existing + 1 new v1 issuer test)
- [x] Multi-tenant configurations (`organizations`, `consumers`, `common`) unaffected — issuer remains `None`
- [x] OBO integration tests unaffected
- [x] Scope enforcement still works for non-OIDC scopes

### DIFF
--- a/src/fastmcp/server/auth/providers/azure.py
+++ b/src/fastmcp/server/auth/providers/azure.py
@@ -197,8 +197,14 @@ class AzureProvider(OAuthProxy):
         self.identifier_uri = identifier_uri or f"api://{client_id}"
         self.additional_authorize_scopes: list[str] = parsed_additional_scopes
 
-        # Always validate tokens against the app's API client ID using JWT
-        issuer = f"https://{base_authority}/{tenant_id}/v2.0"
+        # Always validate tokens against the app's API client ID using JWT.
+        # Accept both v2 and v1 issuers: Azure emits v1-format tokens (issuer
+        # https://sts.windows.net/{tid}/) when custom API scopes target a
+        # first-party Microsoft resource such as Dynamics 365 Business Central.
+        issuer: list[str] = [
+            f"https://{base_authority}/{tenant_id}/v2.0",
+            f"https://sts.windows.net/{tenant_id}/",
+        ]
         jwks_uri = f"https://{base_authority}/{tenant_id}/discovery/v2.0/keys"
 
         # Azure access tokens only include custom API scopes in the `scp` claim,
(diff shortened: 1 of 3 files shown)


jlowin:
Thanks for the detailed write-up and for tracking down exactly where each validation fails.

We explored v1 token support during the original Azure provider work (#1891) and decided against it — @JonasKs, who maintains `fastapi-azure-auth`, [weighed in strongly](https://github.com/PrefectHQ/fastmcp/pull/1891#issuecomment-3349998547) that v1 support introduces conditional branching that invites bugs, and that his own codebase got much simpler after dropping it. That matches our experience — this PR needs three separate workarounds across three different validation axes, and that's just for one resource.

I think the underlying issue here is architectural rather than a gap in FastMCP. Your `identifier_uri` is set to `https://api.businesscentral.dynamics.com` — that's Business Central's resource URI, not your own app's. Because BC is a first-party Microsoft resource, you don't control its `accessTokenAcceptedVersion`, which is why Azure emits v1 tokens. If your MCP server has its own app registration with its own identifier URI and scopes, it validates tokens scoped to *itself* (v2, under your control), and then uses OBO to obtain a separate BC-scoped token. That way the v1/v2 question doesn't touch FastMCP at all.

JonasKs:
@jlowin ‘s answer is 100% correct! You’re not supposed to imitate another app, which you are when the `audience` is not for an appreg you have created.   
So the decision to not support deprecated v1 tokens saved you from having a “security oopsie”.

PR #3852, opened 2026-04-11
Promote 7 ty rules from ignore to warn

strawgate:
Enables stricter type checking by promoting 7 ty rules from their default `ignore` level to `warn` (which becomes `error` via the existing `error-on-warning = true` setting).

**Rules promoted:**
| Rule | Violations | What it catches |
|------|-----------|-----------------|
| `division-by-zero` | 0 | Literal division by zero |
| `possibly-missing-attribute` | 0 | Attribute access on types that may not have it |
| `possibly-missing-import` | 0 | Names that look like they should be imported |
| `possibly-unresolved-reference` | 9 | Variables used when possibly not defined |
| `unsupported-dynamic-base` | 0 | Classes with dynamic base classes |
| `unsupported-operator` | 0 | Operators on incompatible types |
| `unused-ignore-comment` | 0 | Stale `ty: ignore` comments |

**Fixes (9 total):**
- `openapi/schemas.py`: Initialize `body_schema` before conditional block (5 hits)
- `test_input_validation.py`: Initialize `error_msg` before try/except (2 hits)
- `test_real_world_schemas.py`: Initialize `old_handler` before conditional (1 hit)
- `mcp_config.py`: Suppress walrus-operator false positive with `ty: ignore` (1 hit)

Following the pattern from [scipy-stubs](https://github.com/scipy/scipy-stubs) which promotes all of these to `error`.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -162,6 +162,15 @@ python-version = "3.10"
 # Some code uses `# ty: ignore[invalid-argument-type]` for this limitation.
 # TODO: Remove these ignores once ty supports union narrowing
 
+# Promote rules from default-ignore to warn (becomes error via error-on-warning)
+division-by-zero = "warn"
+possibly-missing-attribute = "warn"
+possibly-missing-import = "warn"
+possibly-unresolved-reference = "warn"
+unsupported-dynamic-base = "warn"
+unsupported-operator = "warn"
+unused-ignore-comment = "warn"
+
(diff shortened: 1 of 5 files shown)


jlowin (approved):
Promoting 7 ty rules from ignore to warn. The 9 accompanying fixes are legitimate — one was a latent NameError waiting to happen.

PR #3857, opened 2026-04-11
Raise on unhandled content types in sampling handler dispatch chains

strawgate:
The Anthropic and OpenAI sampling handlers have isinstance chains that dispatch on MCP content types (`TextContent`, `ImageContent`, `AudioContent`, `ToolUseContent`, `ToolResultContent`) but silently drop unhandled variants like `EmbeddedResource` and `ResourceLink`.

This adds explicit `else: raise ValueError` to the two list-content dispatch loops that were missing catch-all branches, matching the behavior of:
- The Gemini handler (already raises on all unhandled types)
- The single-content dispatch paths in both Anthropic and OpenAI (already raise)

**Why raise instead of warn-and-skip:** These handlers convert MCP messages to vendor API formats for LLM calls. A partial conversion produces a plausible-but-wrong response — the LLM confidently answers based on incomplete input, and the user has no reason to doubt it. A clear error is less surprising than a subtly wrong answer.

```python
# Before: EmbeddedResource silently dropped
for item in content:
    if isinstance(item, ToolUseContent): ...
    elif isinstance(item, TextContent): ...
    elif isinstance(item, ImageContent): ...
    elif isinstance(item, AudioContent): ...
    elif isinstance(item, ToolResultContent): ...
    # EmbeddedResource falls through silently

# After: explicit guard
    else:
        raise ValueError(
            f"Unsupported content type for Anthropic: {type(item).__name__}"
        )
```

### DIFF
--- a/src/fastmcp/client/sampling/handlers/anthropic.py
+++ b/src/fastmcp/client/sampling/handlers/anthropic.py
@@ -235,6 +235,10 @@ class AnthropicSamplingHandler:
                                 is_error=item.isError if item.isError else False,
                             )
                         )
+                    else:
+                        raise ValueError(
+                            f"Unsupported content type for Anthropic: {type(item).__name__}"
+                        )
 
                 if content_blocks:
                     anthropic_messages.append(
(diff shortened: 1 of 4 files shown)


jlowin (approved):
Exhaustiveness is the right posture for message conversion — silent-drop was the bug. Matches the Gemini handler and single-content paths.

PR #3873, opened 2026-04-12
ci: speed up schema crash test (CSafeLoader + xdist-safe aggregation)

jlowin:
Follow-up to [#3826](https://github.com/PrefectHQ/fastmcp/pull/3826).

The real-world schema crash test was running ~700 provider-level pytest cases serially, taking 25-40 minutes on PRs touching schema-to-type code. Three changes speed it up and make parallel execution correct:

1. **`CSafeLoader` instead of `yaml.safe_load`.** Profiling showed ~80% of wall-clock was pure-Python YAML parsing — `safe_load()` doesn't pick up libyaml automatically. Switching to the C loader with the same safety semantics gives ~8x speedup per file.

2. **`-n auto` in the pytest command.** GitHub's standard ubuntu runners have 2 cores, so this gives ~2x on top of the YAML win.

3. **Disk-backed result persistence + `pytest_sessionfinish` aggregation.** The previous aggregate test (`test_z_aggregate_crash_rate`) read a module-level `_results` dict, which broke under xdist because each worker is a separate process. Now each per-provider test writes its counts to `SCHEMA_CRASH_RESULTS_DIR`, and a `pytest_sessionfinish` hook in `conftest.py` reads them all on the xdist master after workers finish. Baseline violations set `session.exitstatus` to fail the run.

Net: ~3-5 minutes expected in CI (down from ~25-40). One incidental observation for @strawgate — on my local run the `TypeErrors` count dropped from the previous baseline of 388 to 0. The JSON round-trip should produce identical input regardless of YAML loader, so it's worth a second look to confirm nothing silently changed.

### DIFF
--- a/.github/workflows/run-schema-crash-test.yml
+++ b/.github/workflows/run-schema-crash-test.yml
@@ -52,4 +52,4 @@ jobs:
         env:
           RUN_REAL_WORLD_SCHEMA_TEST: "1"
           OPENAPI_DIRECTORY_PATH: /tmp/openapi-directory
-        run: uv run pytest tests/utilities/json_schema_type/test_real_world_schemas.py -m integration -v -s --timeout-method=thread
+        run: uv run pytest tests/utilities/json_schema_type/test_real_world_schemas.py -m integration -v -n auto --timeout-method=thread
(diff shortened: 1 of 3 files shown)


strawgate:
I think I had two prs that fixed bugs in this area so depending on merge order the type issues may be fixed and we can update the baseline 

Will look in a bit

PR #3880, opened 2026-04-12
fix: resolve list[dict] return type producing Root() instead of dicts

### LINKED ISSUE #3867: list[dict] return type annotation becomes Root() in testing
### What happened?

When a tool is annotated with the return type `list[dict]` the response becomes a list of `Root()` rather than the actual response, at least in a test environment:

For example, run the example code below with
```
uv run --with pytest-asyncio --with fastmcp==3.2.3 pytest --asyncio-mode=auto
```
gives
```
    async def test_weather_operations():
        # Pass server directly - no deployment needed
            result = await client.call_tool("get_temperatures")
>           assert result.data == [{"city": "NYC", "temp": 72}, {"city": "LA", "temp": 85}]
E           AssertionError: assert [Root(), Root()] == [{'city': 'NY..., 'temp': 85}]
E             
E             At index 0 diff: Root() != {'city': 'NYC', 'temp': 72}
E             Use -v to get more diff
```

### Example Code

```Python

server = FastMCP("WeatherServer")


@server.tool
def get_temperatures() -> list[dict]:
    """Get current temperatures for all cities"""
    return [
        {"city": "NYC", "temp": 72},
        {"city": "LA", "temp": 85},
    ]


async def test_weather_operations():
        result = await client.call_tool("get_temperatures")
        assert result.data == [{"city": "NYC", "temp": 72}, {"city": "LA", "temp": 85}]
```

### Version Information

```Text
FastMCP version:                                                             3.2.3
MCP version:                                                                1.27.0
Python version:                                                            3.12.13
Platform:                                             macOS-26.3.1-arm64-arm-64bit
FastMCP root path: /Users/rparini/fastmcp_debug/.venv/lib/python3.12/site-packages
```

KeWang0622:
## Description

Fixes #3867

When a tool returns `list[dict]`, the client deserializes each dict as a `Root()` dataclass with no fields instead of preserving the original dict data:

```python
@server.tool
def get_temperatures() -> list[dict]:
    return [{"city": "NYC", "temp": 72}, {"city": "LA", "temp": 85}]

# Before fix:  result.data == [Root(), Root()]
# After fix:   result.data == [{"city": "NYC", "temp": 72}, {"city": "LA", "temp": 85}]
```

## Root cause

`_get_from_type_handler`'s `"object"` branch always fell through to `_create_dataclass` for schemas without `properties` (like `{"type": "object", "additionalProperties": true}` — the JSON Schema for bare `dict`). This created an empty dataclass named `Root`.

The top-level `json_schema_to_type` already handled this correctly — returning `dict[str, Any]` — but that logic was not reachable from `_schema_to_type`, which is used for nested schemas (e.g., array items).

## Fix

Extract `_object_schema_to_type` to unify the four object-schema cases so both top-level and nested paths produce the correct type:

1. No `properties` + `additionalProperties` — `dict[str, T]`
2. No `properties` + no `additionalProperties` — `dict[str, Any]`
3. `properties` + `additionalProperties is True` — Pydantic BaseModel
4. `properties` only — dataclass

## Contribution type

- [ ] Enhancement

## Checklist

- [x] This PR addresses an existing issue (#3867)
- [x] Tests included: unit tests (schema roundtrip) + integration test (client call_tool)
- [x] All 157 json_schema_type tests pass
- [x] All 2548 server tests pass
- [x] All 294 tool tests pass
- [x] ruff passes clean

### DIFF
--- a/src/fastmcp/utilities/json_schema_type.py
+++ b/src/fastmcp/utilities/json_schema_type.py
@@ -350,23 +350,48 @@ def _return_Any() -> Any:
     return Any
 
 
+def _object_schema_to_type(
+    schema: Mapping[str, Any], schemas: Mapping[str, Any]
+) -> type:
+    """Convert an object schema to the appropriate Python type.
+
+    Handles three cases that mirror the top-level ``json_schema_to_type`` logic:
+    1. No ``properties`` with ``additionalProperties`` — return ``dict[str, T]``
+    2. No ``properties`` and no ``additionalProperties`` — return ``dict[str, Any]``
+    3. Has ``properties`` and ``additionalProperties is True`` — Pydantic model
+    4. Has ``properties`` — dataclass
+    """
+    has_properties = bool(schema.get("properties"))
+    additional_props = schema.get("additionalProperties")
+
+    if not has_properties and additional_props:
+        if additional_props is True:
+            return dict[str, Any]
+        value_type = _schema_to_type(additional_props, schemas)
+        return cast(type[Any], dict[str, value_type])  # type: ignore[valid-type]  # ty:ignore[invalid-type-form]
+
+    if not has_properties and not additional_props:
+        return dict[str, Any]
+
+    if has_properties and additional_props is True:
+        return _create_pydantic_model(schema, schema.get("title"), schemas)
+
+    return _create_dataclass(schema, schema.get("title"), schemas)
+
+
 def _get_from_type_handler(
     schema: Mapping[str, Any], schemas: Mapping[str, Any]
 ) -> Callable[..., Any]:
     """Get the appropriate type handler for the schema."""
 
-    type_handlers: dict[str, Callable[..., Any]] = {  # TODO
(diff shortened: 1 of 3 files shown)


jlowin (approved):
Thanks for the clean diagnosis and tests on this one — the root cause is exactly right, and the `list[dict[str, X]]` parametrization plus the end-to-end client test give me a lot of confidence in the fix.

One small note for anyone reading this later: the PR description says the helper unifies top-level and nested paths, but in practice the new `_object_schema_to_type` is only wired into `_get_from_type_handler`. The top-level `json_schema_to_type` (around lines 180-198) and the `anyOf` dict-in-union branch (around lines 362-374) still carry their own copies of the four-case logic. That's fine — those sites aren't broken, and this is the narrowest fix for the reported bug — but it means the divergence risk that caused #3867 in the first place still exists at two other sites. I'll open a follow-up to route all three through the new helper.

Merging. Thanks again!

PR #3885, opened 2026-04-13
Harden forced client disconnect cleanup

vonbai:
This fixes forced `Client.close()` cleanup when caller cancellation arrives while the client is waiting for the background session task to unwind. The forced disconnect path now uses the same `anyio.CancelScope(shield=True)` + `move_on_after()` pattern used during connection-startup cancellation, then clears `session_task` before releasing the session lock so a concurrent reconnect cannot reuse a dying runner.

The regression test pauses session teardown, cancels a force-close mid-wait, immediately attempts a reconnect, and asserts the reconnect gets a fresh session task and can call `list_tools()`.

Validation: `uv run pytest tests/client/client/test_client.py::test_force_close_cancelled_wait_starts_fresh_session -q`, focused neighboring cancellation tests, `uv run pytest tests/client/client/test_client.py -q`, `uv run pytest tests/test_mcp_config.py::test_multi_client_force_close -q`, `uv run pytest -n auto tests/test_mcp_config.py`, `uv run pytest -n auto`, and `uv run prek run --all-files`.

Warning note: the full `tests/test_mcp_config.py` run under xdist still emits one `PytestUnraisableExceptionWarning` from `BaseSubprocessTransport.__del__` in `test_multi_client_lifespan`. I checked an `upstream/main` baseline in a separate worktree: the same subprocess/event-loop-closed warning class is already present there, and the baseline still fails `test_multi_client_force_close` with the original cancellation bug. The focused force-close test now passes without that warning.

### DIFF
--- a/src/fastmcp/client/client.py
+++ b/src/fastmcp/client/client.py
@@ -642,10 +642,20 @@ class Client(
             # stop the active session
             if self._session_state.session_task is None:
                 return
+            session_task = self._session_state.session_task
             self._session_state.stop_event.set()
             # wait for session to finish to ensure state has been reset
-            await self._session_state.session_task
-            self._session_state.session_task = None
+            try:
+                if force:
+                    await asyncio.shield(session_task)
+                else:
+                    await session_task
+            except asyncio.CancelledError:
+                if not force or not session_task.cancelled():
+                    raise
+            finally:
+                if session_task.done():
+                    self._session_state.session_task = None
 
     async def _session_runner(self):
         """
--- a/src/fastmcp/client/client.py
+++ b/src/fastmcp/client/client.py
@@ -647,15 +647,14 @@ class Client(
             # wait for session to finish to ensure state has been reset
             try:
                 if force:
-                    await asyncio.shield(session_task)
+                    with anyio.CancelScope(shield=True):
+                        with anyio.move_on_after(self._disconnect_timeout):
+                            with suppress(asyncio.CancelledError):
+                                await session_task
                 else:
                     await session_task
-            except asyncio.CancelledError:
-                if not force or not session_task.cancelled():
-                    raise
             finally:
-                if session_task.done():
-                    self._session_state.session_task = None
+                self._session_state.session_task = None
 
     async def _session_runner(self):
         """
--- a/tests/client/client/test_client.py
+++ b/tests/client/client/test_client.py
@@ -439,6 +439,32 @@ class _DelayedConnectTransport(ClientTransport):
         await self._inner.close()
 
 
+class _DelayedDisconnectTransport(ClientTransport):
+    def __init__(
+        self,
+        inner: ClientTransport,
+        disconnect_started: anyio.Event,
+        allow_disconnect: anyio.Event,
+    ) -> None:
+        self._inner = inner
+        self._disconnect_started = disconnect_started
+        self._allow_disconnect = allow_disconnect
+
+    @contextlib.asynccontextmanager
+    async def connect_session(
+        self, **session_kwargs: Any
+    ) -> AsyncIterator[ClientSession]:
+        async with self._inner.connect_session(**session_kwargs) as session:
+            try:
+                yield session
+            finally:
+                self._disconnect_started.set()
+                await self._allow_disconnect.wait()
+
+    async def close(self) -> None:
+        await self._inner.close()
+
+
 async def test_client_nested_context_manager(fastmcp_server):
     """Test that the client connects and disconnects once in nested context manager."""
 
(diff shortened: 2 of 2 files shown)


jlowin (changes requested):
Thanks @vonbai — the diagnosis here is right. The unshielded `await self._session_state.session_task` does propagate caller-cancellation into the session task itself, cutting short `_session_runner`'s ability to process `stop_event`, unwind its `AsyncExitStack`, and close the transport cleanly. That's worth fixing.

On the asyncio-vs-anyio question: `session_task` being an `asyncio.Task` is intentional — there's an existing comment at `client.py:561-563` explaining that it has to outlive individual context manager scopes, which anyio's structured concurrency won't allow. So `asyncio.shield` + `asyncio.CancelledError` on this specific task is the right family of primitives, not an oversight.

## The blocker: Codex's P1 race is real

The shield prevents `session_task` from being cancelled when the caller is cancelled mid-await. Combined with catching `CancelledError` for the force path, that opens a window where:

1. Caller is cancelled during `await asyncio.shield(session_task)`
2. `CancelledError` is caught and swallowed (force + task not cancelled)
3. `finally:` runs — `session_task.done()` is False (still unwinding in the background), so the field isn't cleared
4. `_disconnect` returns, lock releases
5. Concurrent `_connect()` acquires the lock, sees `session_task is not None and not session_task.done()`, treats the dying session as live, increments `nesting_counter`, and doesn't start a fresh runner
6. Original session task finishes, cleanup runs, transport resets
7. The "reused" session is stale — subsequent calls fail

The pre-fix behavior avoided this because an unshielded cancelled await would cancel `session_task` too, making `.done()` True before the lock released. The shield breaks that invariant.

## The cleaner fix: match the existing pattern in this same function

There's already a working shielded-wait-with-timeout pattern a few lines up at `client.py:570-579` for the connection-startup cancellation case:

```python
with anyio.CancelScope(shield=True):
    with anyio.move_on_after(3):
        try:
            await session_task
            pass
```

Unlike `asyncio.shield`, `anyio.CancelScope(shield=True)` actually *waits* for the inner await to complete rather than re-raising cancellation immediately. By the time you release the lock, `session_task.done()` is True and the concurrent-reconnect race is closed. It also keeps this function to a single cancellation pattern instead of two.

Applied to the disconnect path:

```python
session_task = self._session_state.session_task
self._session_state.stop_event.set()
try:
    if force:
        with anyio.CancelScope(shield=True):
            with anyio.move_on_after(self._disconnect_timeout):
                try:
                    await session_task
                    pass
    else:
        await session_task
finally:
    self._session_state.session_task = None
```

Please also add a regression test covering the race directly: spawn a force-close, cancel it mid-await, then immediately attempt a reconnect, and assert the new session actually works (i.e., `_connect` started a fresh runner rather than reusing the dying one). The existing `test_multi_client_force_close` doesn't exercise the cancellation path.

## Other notes

- Your Validation section mentions the focused test "still emitted the existing subprocess/pytest-timeout warning shape as a warning, not a failure." Could you characterize that warning? If it's pre-existing noise unrelated to this change, say so explicitly; if it's new or related, the fix should address it.
- CI isn't running yet (only label triage is green). I'll approve workflows once the above is addressed.

vonbai:
Thanks, this was exactly the missing edge case.

I pushed `48fc3b56`, which changes the forced disconnect path to use the existing `anyio.CancelScope(shield=True)` + `move_on_after()` pattern and clears `session_task` before releasing the session lock. I also added a direct regression test that pauses session teardown, cancels `client.close()` mid-wait, immediately reconnects, and asserts the reconnect gets a fresh session task and can call `list_tools()`.

On the warning: I rechecked it in a separate `upstream/main` worktree. The focused force-close test passes without the warning on the updated branch, while the full `tests/test_mcp_config.py` xdist run still has a `BaseSubprocessTransport.__del__` / event-loop-closed warning in `test_multi_client_lifespan`. The same warning class is present on upstream main, where `test_multi_client_force_close` still fails with the original cancellation bug, so I’ve left that subprocess cleanup warning out of this PR’s scope.

Validation run on the updated branch:

- `uv run pytest tests/client/client/test_client.py::test_force_close_cancelled_wait_starts_fresh_session -q`
- neighboring client cancellation tests
- `uv run pytest tests/client/client/test_client.py -q`
- `uv run pytest tests/test_mcp_config.py::test_multi_client_force_close -q`
- `uv run pytest -n auto tests/test_mcp_config.py`
- `uv run pytest -n auto` (`5400 passed, 3 skipped, 14 xfailed, 1 pre-existing warning`)

jlowin (approved):
Thanks!

PR #3889, opened 2026-04-13
OTEL: Fix attribute compliance with MCP semantic conventions

strawgate:
**Step 1 of 3** in the OTEL instrumentation series. This PR fixes attribute compliance and improves the telemetry helpers. Steps 2 (#3890) and 3 (#3892) build on this.

## Attribute compliance

Removes `rpc.system`, `rpc.service`, `rpc.method` from all spans — MCP is not a traditional RPC system and the emerging MCP OTEL conventions use `mcp.method.name` instead. Adds `gen_ai.tool.name` on `tools/call` spans and `gen_ai.prompt.name` on `prompts/get` spans per the OpenTelemetry GenAI semantic conventions.

Removes resource URI from `resources/read` span names — per the MCP semantic conventions, resource URIs SHOULD NOT be included in span names by default since they can be unbounded (especially with resource templates like `users://{id}/profile`). The URI remains available in the `mcp.resource.uri` attribute. Tool names and prompt names remain in span names since they are low-cardinality.

## Telemetry helper improvements

- `is_recording()` guards on attribute-setting (avoids work on non-recording spans)
- `error.type` attribute with `__qualname__` on error spans (OTEL error convention)
- ToolError `isinstance` check → sets `"tool_error"` as error type
- Exception message in span status description for debugging
- `tool_name` / `prompt_name` params on `server_span` and `client_span`

## Client call_tool enrichment

Reflects `result.isError` on the client span so callers see `ERROR` status even when the MCP protocol call itself succeeded (the tool failed, not the transport).

**Breaking change**: Removes `rpc.*` span attributes. Consumers querying for `rpc.system == "mcp"` will need to switch to `mcp.method.name`.


src/fastmcp/server/server.py
```diff
@@ -1506,7 +1511,12 @@ async def render_prompt(
             # Core logic: find and render prompt (providers queried in parallel)
             # Use get_prompt to apply transforms and filter disabled
             with server_span(
-                f"prompts/get {name}", "prompts/get", self.name, "prompt", name
+                f"prompts/get {name}",
```
strawgate, line 1514: i think we want to keep the prompt/tool names in the span we just need to make sure it's low cardinality and not including URIs

jlowin (changes requested):
The semconv alignment is the right direction and the tests look solid. Two things before merge:

1. **Hoist `ToolError` imports to module level.** `from fastmcp.exceptions import ToolError as _ToolError` appears inside exception handlers at `src/fastmcp/server/telemetry.py:99`, `src/fastmcp/server/telemetry.py:136`, and `src/fastmcp/client/telemetry.py:50`. No import cycle exists — these should be module-level per project convention.

2. **Migration note for `rpc.*` removal.** Dropping `rpc.system`/`rpc.service`/`rpc.method` will silently break any user dashboards or alerts keyed on those attributes. A short note in the PR body (or a callout in the observability docs) pointing users at the new `mcp.method.name` / `gen_ai.tool.name` / `gen_ai.prompt.name` attributes would soften the break.

jlowin (approved):
Both change-requests addressed: ToolError imports hoisted to module level (server/telemetry.py:10, client/telemetry.py:8), migration note added as a Warning callout in docs/servers/telemetry.mdx pointing users at `mcp.method.name` / `gen_ai.*`. CI green across the matrix.

PR #3896, opened 2026-04-13
fix: handle circular JSON Pointer $ref in dereference_refs

### LINKED ISSUE #3893: DereferenceRefsMiddleware crashes with RecursionError on JSON Pointer style $ref (non-$defs)
### What happened?

DereferenceRefsMiddleware crashes with a RecursionError when schemas contain circular $ref using JSON Pointer style (e.g. #/properties/nodes/items) instead of $defs-based style (e.g.  #/$defs/Node).

This is the format emitted by .NET/C# MCP servers using System.Text.Json's schema generator. There is no $defs section, and $ref pointers reference locations directly within the schema  tree.

C#'s System.Text.Json schema generator uses JSON Pointer $ref in two cases:
- Recursive types (e.g. a Node with Children: IReadOnlyList<Node>) → circular $ref
- Shared types (e.g. the same complex type appearing in multiple properties) → non-circular $ref

As a workaround, we disable the built-in middleware (dereference_schemas=False).

### Example Code

```Python
from fastmcp.utilities.json_schema import dereference_refs

schema = {
    "type": "object",
    "properties": {
        "nodes": {
            "type": "array",
            "items": {
                "type": "object",
                "properties": {
                    "value": {"type": "string"},
                    "children": {
                        "type": "array",
                        "items": {"$ref": "#/properties/nodes/items"},
                    },
                },
            },
        },
    },
}

dereference_refs(schema)  # RecursionError
```

### Version Information

```Text
- fastmcp 3.2.0
- jsonref 1.1.0
- Python 3.13
- Upstream MCP server: C# ModelContextProtocol.AspNetCore 1.2.0
```

lawrence3699:
Closes #3893.

`dereference_refs` crashes with `RecursionError` when schemas contain circular `$ref` using JSON Pointer paths (e.g. `#/properties/nodes/items`) instead of `$defs`-based paths. This is the format emitted by .NET/C# MCP servers using `System.Text.Json`. The existing `_defs_have_cycles` check only detects cycles within `$defs`, so pointer-style cycles reach `jsonref.replace_refs` uncaught.

```python
dereference_refs({
    "type": "object",
    "properties": {
        "nodes": {
            "type": "array",
            "items": {
                "type": "object",
                "properties": {
                    "value": {"type": "string"},
                    "children": {
                        "type": "array",
                        "items": {"$ref": "#/properties/nodes/items"},
                    },
                },
            },
        },
    },
})
# Before: RecursionError
# After:  returns schema with circular $ref preserved (same fallback as $defs cycles)
```

The fix catches `RecursionError` alongside `JsonRefError` and falls back to `resolve_root_ref`, matching the existing behavior for `$defs`-based circular schemas. A regression test verifies the crash is gone and the circular `$ref` is preserved in the output.

🤖 Generated with Claude Code


jlowin:
Thanks @lawrence3699 — clean diagnosis and nicely scoped fix. Root cause is right: `_defs_have_cycles` only walks `$defs`-based references, so a JSON Pointer cycle like `#/properties/nodes/items` bypasses the proactive check and reaches `jsonref.replace_refs`, which recurses until Python's stack limit raises `RecursionError` rather than `JsonRefError`. Adding `RecursionError` to the existing exception tuple and reusing the same `resolve_root_ref` fallback matches the behavior path for `$defs`-style cycles — surgical and consistent.

One small ask before merging — Copilot's first inline suggestion is worth taking: the regression test currently asserts `"$ref" in children_items`, but doesn't verify the pointer value is preserved verbatim. Strengthening it to:

```python
assert children_items["$ref"] == "#/properties/nodes/items"
```

catches any future regression where we might accidentally rewrite the pointer during fallback. One-line change, strictly better.

Optional (feel free to skip) — the linked issue notes that .NET/`System.Text.Json` also emits JSON Pointer `$ref` for *shared non-circular* types. A second test locking in that non-circular pointer `$ref` still dereferences correctly through `replace_refs` would guard against future regressions on that path. Not a blocker, just a nice-to-have.

Copilot's second comment (about the existing "resolving only root-level $ref" wording) is a fair observation but it's about a pre-existing comment, not something this PR introduced — out of scope here.

Once the test assertion is tightened I'll get this merged.

jlowin (changes requested):
See my comment above — one small ask: strengthen the regression test assertion to check the pointer value is preserved verbatim (`assert children_items["$ref"] == "#/properties/nodes/items"`). Optional second test for non-circular JSON Pointer `$ref` would be nice but not blocking.

The core fix is correct and merge-ready once the assertion is tightened.

nidhishgajjar:

## Summary

This PR fixes a `RecursionError` crash in `dereference_refs` when schemas contain circular `$ref` using JSON Pointer paths (e.g., `#/properties/nodes/items`) instead of `$defs`-based references. This pattern is common in schemas generated by .NET's `System.Text.Json`. The fix adds `RecursionError` to the existing `except JsonRefError` handler.

## Architecture

The change is minimal and well-targeted. The existing `_defs_have_cycles` check only handles `$defs`-based cycles, but JSON Pointer-based circular refs bypass that detection and cause `jsonref.replace_refs` to recurse infinitely. Catching `RecursionError` as a fallback is the right approach since it covers all edge cases without requiring exhaustive cycle detection.

## Issues

1. **💡 Suggestion — `RecursionError` catch scope**: Catching `RecursionError` is pragmatic but slightly broad — it could also catch unrelated recursion issues in `replace_refs` or `_merge_ref_siblings`. In practice, this is fine because the fallback (`resolve_root_ref`) is safe regardless. The comment explaining the `.NET/System.Text.Json` context is very helpful for future maintainers.

2. **💡 Suggestion — Consider increasing recursion limit**: An alternative approach would be to set `sys.setrecursionlimit` around the `replace_refs` call, but the current approach (catching and falling back) is simpler and safer. No change needed.

## Cross-file Impact

None — the change is contained to `dereference_refs` in `json_schema.py` and its test file.

## Assessment

**approve** — Clean, minimal fix for a real crash bug. The test is excellent: it creates a realistic circular JSON Pointer schema and verifies both that the function doesn't crash AND that the circular `$ref` is preserved (not incorrectly inlined). Good work including the `.NET context in the comment.

jlowin (approved):

PR #3714, opened 2026-03-30
client: fix set_roots updates live session at runtime

dgenio:
## Description

Fixes #326.

Calling `Client.set_roots(...)` during an active session previously only updated pending session kwargs, so runtime roots changes were not reflected until reconnect. This PR updates the live session callback when connected and adds a regression test that verifies runtime root updates are immediately visible.

Closes #326

## Contribution type

- [x] Bug fix (simple, well-scoped fix for a clearly broken behavior)
- [ ] Documentation improvement
- [ ] Enhancement (maintainers typically implement enhancements — see [CONTRIBUTING.md](../CONTRIBUTING.md))

## Checklist

- [x] This PR addresses an existing issue (or fixes a self-evident bug)
- [x] I have read [CONTRIBUTING.md](../CONTRIBUTING.md)
- [x] I have added tests that cover my changes
- [ ] I have run `uv run prek run --all-files` and all checks pass
- [x] I have self-reviewed my changes
- [x] If I used an LLM, it followed the repo's contributing conventions (not generic output)

## Testing

- `uv run pytest tests/client/test_roots.py -q` -> `5 passed`
- `uv run prek run --all-files` was run previously; local environment was missing `loq`, so full pass should be confirmed in CI

PR #3906, opened 2026-04-13
fix: set_sampling_callback and set_elicitation_callback update live session immediately

### LINKED ISSUE #3715: set_sampling_callback and set_elicitation_callback don't update live session
## Description

`set_sampling_callback` and `set_elicitation_callback` on `Client` only update `_session_kwargs` without touching the live `ClientSession`. This means changes made while connected won't take effect until reconnect — the same bug that `set_roots` had before #3714.

## Context

Identified during review of #3714 (fix for #326). The `set_roots` fix was scoped to roots only; these sibling methods need the same treatment.

## Proposed approach

Apply the same pattern used in `set_roots`: after updating `_session_kwargs`, check `is_connected()` and mutate the live session's private callback attribute (`_sampling_callback` / `_elicitation_callback`). Since these methods are not called from `__init__`, they don't need the `hasattr(self, '_session_state')` guard.

Note: `_sampling_callback` and `_elicitation_callback` are private attributes on the upstream MCP SDK's `ClientSession` — same fragility caveat as `_list_roots_callback`.

## Files

- `src/fastmcp/client/client.py` — `set_sampling_callback`, `set_elicitation_callback`
- Tests needed for both methods

mrishav:
`set_sampling_callback` and `set_elicitation_callback` only updated `_session_kwargs`, so changes had no effect on an already-connected client — the old callback stayed active until reconnect.

Fix: after updating `_session_kwargs`, check `is_connected()` and directly mutate `_sampling_callback` / `_elicitation_callback` on the live `ClientSession`.

Closes #3715


jlowin:
Thanks for picking this up, @mrishav.

This is structurally the same fix as #3714 (roots callback), which I closed last week because I'm not comfortable reaching into private MCP SDK attributes (`_list_roots_callback`, and here `_sampling_callback`/`_elicitation_callback`) to work around the lack of a public callback-update API. Same coupling risk, same concern about silent breakage on SDK upgrades — especially with MCP SDK v2 on the horizon, which will likely change session internals significantly.

Going to close this one for the same reason. If the SDK adds a public way to update callbacks on a live session, the fix becomes a one-liner on our side and we can revisit both.

Broader note for future contributions: when an issue references a prior PR ([#3715](https://github.com/PrefectHQ/fastmcp/issues/3715) cites #3714 as the origin), please read that PR and its closing comments before opening a follow-up. It saves a round-trip when the original was closed for an architectural reason that still applies.

mrishav:
> Thanks for picking this up, @mrishav.
> 
> This is structurally the same fix as #3714 (roots callback), which I closed last week because I'm not comfortable reaching into private MCP SDK attributes (`_list_roots_callback`, and here `_sampling_callback`/`_elicitation_callback`) to work around the lack of a public callback-update API. Same coupling risk, same concern about silent breakage on SDK upgrades — especially with MCP SDK v2 on the horizon, which will likely change session internals significantly.
> 
> Going to close this one for the same reason. If the SDK adds a public way to update callbacks on a live session, the fix becomes a one-liner on our side and we can revisit both.
> 
> Broader note for future contributions: when an issue references a prior PR ([#3715](https://github.com/PrefectHQ/fastmcp/issues/3715) cites #3714 as the origin), please read that PR and its closing comments before opening a follow-up. It saves a round-trip when the original was closed for an architectural reason that still applies.

Thanks for the explanation and for pointing out #3714. I’ll check related PRs first going forward.

Makes sense on the SDK coupling concern, I’ll keep an eye out for a public callback-update API in MCP SDK v2.

PR #3929, opened 2026-04-14
fix: allow hyphens in resource template parameter names

### LINKED ISSUE #3921: Resource template registration fails when param names contain hyphens
URI templates with hyphenated parameter names fail at registration time.

```python
@mcp.resource("data://{user-id}/profile")
def get_profile(user_id: str) -> str:
    return f"profile for {user_id}"
# ValueError: URI template must contain at least one parameter
```

The regex in `template.py:505` uses `\w+` which matches `[a-zA-Z0-9_]` — hyphens are excluded:

```python
path_params = set(re.findall(r"{(\w+)(?:\*)?}", uri_template))
```

So `{user-id}` is not recognized as a parameter, and the template appears to have zero parameters.

RFC 6570 allows hyphens in variable names. The fix is `[\w-]+` or `[\w.-]+` in the regex.

strawgate:
URI templates with hyphenated parameter names like `{user-id}` failed at registration because the param-extraction regex used `\w+` which excludes hyphens. This also affected `build_regex()` which created invalid Python named groups from hyphenated names.

The fix widens param-extraction regexes to `[\w-]+`, converts hyphens to underscores for regex group names and function parameter matching, and ensures `expand_uri_template` / `match_uri_template` handle the mapping in both directions.

```python
@mcp.resource("data://{user-id}/profile")
def get_profile(user_id: str) -> str:
    return f"profile for {user_id}"
```

Fixes #3921

🤖 Generated with Claude Code

### DIFF
--- a/src/fastmcp/resources/template.py
+++ b/src/fastmcp/resources/template.py
@@ -110,7 +110,10 @@ def match_uri_template(uri: str, uri_template: str) -> dict[str, str] | None:
             if name in parsed_query:
                 # Take first value if multiple provided.
                 # Normalize hyphens to underscores to match Python param names.
-                params[name.replace("-", "_")] = parsed_query[name][0]
+                # Don't overwrite path params that were already extracted.
+                key = name.replace("-", "_")
+                if key not in params:
+                    params[key] = parsed_query[name][0]
 
     return params
 
--- a/src/fastmcp/resources/template.py
+++ b/src/fastmcp/resources/template.py
@@ -52,8 +52,11 @@ def build_regex(template: str) -> re.Pattern[str] | None:
     - `{var*}` - wildcard path parameter (captures multiple segments)
     - `{?var1,var2}` - query parameters (ignored in path matching)
 
+    Hyphens in parameter names are normalized to underscores in regex group
+    names so that matched groups are valid Python identifiers.
+
     Returns None if the template produces an invalid regex (e.g. parameter
-    names with hyphens, leading digits, or duplicates from a remote server).
+    names with leading digits or duplicates from a remote server).
     """
     # Remove query parameter syntax for path matching
     template_without_query = re.sub(r"\{\?[^}]+\}", "", template)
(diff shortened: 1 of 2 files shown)


jlowin (changes requested):
The normalization approach is defensible on ergonomic grounds, but there's a blocking correctness issue and a couple of things to tighten up before merge.

### Blocking: collision detection

Normalizing `-` ↔ `_` means `{user-id}` and `{user_id}` in the same template both resolve to the same key, and the second value silently overwrites the first. Symptoms:

- `src/fastmcp/resources/template.py:106-113` — the query-extraction loop silently drops the colliding value.
- `src/fastmcp/resources/template.py:553-556` — `from_function` dedupes into a set *before* `build_regex` sees the duplicate named groups, so a template that registers cleanly can silently fail at match time.
- `src/fastmcp/resources/template.py:128-143` — `expand_uri_template` can synthesize query values from unrelated path params, changing resource identity on roundtrip.

The fix is small: in `from_function`, before the `all_uri_params` union, check whether two *raw* param names normalize to the same key and raise `ValueError` with a clear message. That closes all three symptoms at the source.

Please also add a test for the collision case — none of the new tests cover it.

### PR body: RFC 6570 claim

The PR body states RFC 6570 permits raw hyphens in varnames. It doesn't — the grammar only allows ALPHA / DIGIT / `_` / `.` / pct-encoded. Happy to take the change on ergonomic grounds, but the RFC justification should be removed so it doesn't propagate.

### Deleted `TestMalformedURITemplates` cases

These were locking in the old "hyphens are malformed" contract. Flipping the contract is fine, but worth a sentence in the PR body noting the deliberate behavior change for anyone grepping the history later.

### CI

`ruff format` was failing per the triage bot. Please confirm it's clean on the latest commit.

jlowin (approved):
Collision guard landed at template.py:559-569 with deterministic sorted error message, plus runtime guard at :111-115 (path wins over query). Tests `test_from_function_rejects_hyphen_underscore_collision` and `test_query_param_does_not_clobber_path_param` cover both. RFC claim removed. Approving — the `minimize` CI failure is unrelated to this change.

PR #3932, opened 2026-04-14
fix: OpenAPI request director sends multipart and form-urlencoded as JSON

strawgate:
The OpenAPI request director's `build()` method detects the declared content type from the spec but only handles JSON variants. Three bugs result:

1. **multipart/form-data sent as JSON** -- When an OpenAPI spec declares `multipart/form-data` (e.g. file uploads), the body falls through to the JSON branch and is sent as `application/json`. Upstream APIs reject it. Fixed by using httpx's `files=` parameter which sends proper multipart encoding.

2. **application/x-www-form-urlencoded sent as JSON** -- Same issue for form-encoded bodies. Fixed by using httpx's `data=` parameter which sends as `application/x-www-form-urlencoded`.

3. **Cookie parameters silently dropped** -- `_unflatten_arguments()` handles path, query, header, and body locations but has no cookie case. Cookie params hit the "unknown location" warning and are discarded. Fixed by adding `cookie_params` throughout the unflatten pipeline and passing them to `httpx.Request(cookies=...)`.

```python
# Before: multipart body silently sent as application/json
director.build(multipart_route, {"file": "data"})
# Content-Type: application/json  <-- wrong

# After: proper multipart encoding
director.build(multipart_route, {"file": "data"})
# Content-Type: multipart/form-data; boundary=...  <-- correct
```

🤖 Generated with Claude Code

### DIFF
--- a/src/fastmcp/utilities/openapi/director.py
+++ b/src/fastmcp/utilities/openapi/director.py
@@ -75,14 +75,19 @@ class RequestDirector:
         json_body: dict[str, Any] | list[Any] | None = None
         content: str | bytes | None = None
 
-        # Step 5: Determine the declared content type from the OpenAPI spec
+        # Step 5: Determine the declared content type from the OpenAPI spec.
+        # Strip parameters (e.g. "; charset=utf-8") for dispatch matching.
         declared_content_type: str | None = None
         if route.request_body and route.request_body.content_schema:
-            declared_content_type = next(iter(route.request_body.content_schema))
+            raw_ct = next(iter(route.request_body.content_schema))
+            declared_content_type = raw_ct.split(";")[0].strip().lower()
 
-        # httpx requires cookie values to be strings
+        # httpx requires cookie values to be strings; use OpenAPI-style
+        # serialization (e.g. true/false for booleans, not True/False)
         cookies = (
-            {k: str(v) for k, v in cookie_params.items()} if cookie_params else None
+            {k: _query_scalar_to_str(v) for k, v in cookie_params.items()}
+            if cookie_params
+            else None
         )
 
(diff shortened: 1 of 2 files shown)


jlowin (changes requested):
The three bugs being fixed here are all real and the fixes target the right code paths. Two follow-ups to address before merge:

1. **Content-Type parameter loss at `src/fastmcp/utilities/openapi/director.py:97`.** The outgoing `Content-Type` header is now set to the stripped/lowercased dispatch key, so a spec-declared `application/json-patch+json; charset=utf-8` loses its parameters on the wire. Preserve the raw Content-Type for the outgoing header; only normalize the copy used for dispatch matching.

2. **Bare `bytes` in multipart at `src/fastmcp/utilities/openapi/director.py:108`.** Non-tuple multipart values flow through `_query_scalar_to_str`, so raw `bytes` get stringified to their Python repr (`"b'...'"`). For `type: string, format: binary` fields users currently have to wrap in a tuple — which isn't documented and is surprising. Either passthrough `bytes` / file-like values directly to `httpx`, or document the tuple requirement explicitly. Please add a bare-bytes test alongside the existing multipart coverage at `tests/utilities/openapi/test_director.py:1153+`.

The rest of the PR is ready — thorough tests, good reuse of `_query_scalar_to_str`, and the refactor to a single `httpx.Request` construction point is clean.

jlowin (approved):
Both items fixed cleanly: `raw_content_type` is preserved for the outgoing header at director.py:137 (dispatch matching uses the normalized copy), and `bytes`/`IOBase` now bypass `_query_scalar_to_str` at director.py:110-115 and are handed to httpx as `(None, v)`. New test asserts charset params survive on the wire. CI green.

PR #3938, opened 2026-04-14
fix: FileSystemProvider reload race condition

### LINKED ISSUE #3625: bug: several correctness issues in FileSystemProvider import machinery

strawgate:
Supersedes #3660, which had the same race condition it was trying to fix (the atomic swap exposed a partially-populated dict during the registration loop).

The fix is simpler: hold the reload lock across both the reload AND the subsequent read. Since every reader method (`_list_tools`, `_get_tool`, etc.) already called `_ensure_loaded()` before reading, we just widen the lock scope to cover both steps. No reader can observe a mid-rebuild `_components` dict because the lock serializes everything.

```python
async def _with_reload(self, coro_fn, *args):
    async with self._reload_lock:
        await self._maybe_reload()
        return await coro_fn(*args)  # read happens inside the lock
```

This only affects `reload=True` mode, which is explicitly a dev feature. The serialization cost is irrelevant for development.

Also warns when the root directory doesn't exist (previously produced zero components silently).

Closes #3625

### DIFF
--- a/src/fastmcp/server/providers/filesystem.py
+++ b/src/fastmcp/server/providers/filesystem.py
@@ -97,6 +97,8 @@ class FileSystemProvider(LocalProvider):
         self._warned_files: dict[Path, float] = {}
         # Lock for serializing reload operations (created lazily)
         self._reload_lock: asyncio.Lock | None = None
+        # Generation counter to deduplicate concurrent reloads
+        self._reload_generation: int = 0
 
         # Always load once at init to catch errors early
         self._load_components()
@@ -157,17 +159,16 @@ class FileSystemProvider(LocalProvider):
         else:
             logger.debug("Ignoring unknown component type: %r", type(component))
 
-    async def _maybe_reload(self) -> None:
-        """Reload components if needed. Caller must hold ``_reload_lock``."""
-        if self._reload or not self._loaded:
-            await asyncio.to_thread(self._load_components)
-
     async def _with_reload(self, coro_fn: Callable[..., Any], *args: Any) -> Any:
         """Acquire the reload lock, reload if needed, then run *coro_fn*.
 
         Holding the lock across both the reload and the read prevents
         concurrent readers from seeing a partially-rebuilt ``_components``
         dict (the ``clear()`` + re-register window).
+
+        A generation counter deduplicates concurrent reload requests:
+        if another caller already reloaded while we waited for the lock,
+        we skip the redundant reload.
         """
         if not self._reload and self._loaded:
             return await coro_fn(*args)
@@ -176,8 +177,14 @@ class FileSystemProvider(LocalProvider):
         if self._reload_lock is None:
             self._reload_lock = asyncio.Lock()
 
+        generation_before = self._reload_generation
+
         async with self._reload_lock:
-            await self._maybe_reload()
+            if not self._loaded or (
+                self._reload and self._reload_generation == generation_before
+            ):
+                await asyncio.to_thread(self._load_components)
+                self._reload_generation += 1
             return await coro_fn(*args)
 
     # Override provider methods to support reload mode
--- a/src/fastmcp/server/providers/filesystem.py
+++ b/src/fastmcp/server/providers/filesystem.py
@@ -28,8 +28,9 @@ Example:
 from __future__ import annotations
 
 import asyncio
-from collections.abc import Sequence
+from collections.abc import Callable, Sequence
 from pathlib import Path
+from typing import Any
 
 from fastmcp.prompts.base import Prompt
 from fastmcp.resources.base import Resource
@@ -102,9 +103,11 @@ class FileSystemProvider(LocalProvider):
 
     def _load_components(self) -> None:
         """Discover and register all components from the filesystem."""
-        # Clear existing components if reloading
         if self._loaded:
             self._components.clear()
+
+        if not self._root.exists():
+            logger.warning("FileSystemProvider root does not exist: %s", self._root)
 
         result = discover_and_import(self._root)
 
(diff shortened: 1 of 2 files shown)


jlowin (changes requested):
Nice fix — serializing reads during reload is the honest approach, and it's refreshing to see the `_with_reload` wrapper collapse 9 near-identical overrides to one-liners. One thing to tighten before merge:

### Blocking: module-level import

`tests/fs/test_provider.py:296` imports `asyncio` inside the test function. The file already imports `asyncio` at module scope, and the project convention is "ALL imports at module level." Please drop the local import.

### Optional follow-ups (non-blocking)

- `src/fastmcp/server/providers/filesystem.py:162` — `coro_fn: Callable[..., Any]` could be `Callable[..., Awaitable[Any]]` for a more precise type. `Any` elsewhere limits the real impact.
- No test for the new non-existent-root warning path. Trivial to add and would round out coverage.

Once the local import is hoisted, this is good to go.

nidhishgajjar:

## Summary

This PR fixes a race condition in `FileSystemProvider` where concurrent readers could observe an empty `_components` dict during reload. The old `_ensure_loaded()` released the reload lock before the read, allowing another task to acquire the lock and `clear()` the dict between the reload and read steps. The fix replaces it with `_with_reload()`, which holds the lock across both the reload and the read. A generation counter deduplicates concurrent reload requests.

## Architecture

The refactoring from "ensure loaded, then read separately" to "wrap the read in a locked reload" is the right pattern. It eliminates the vulnerable window where `_components.clear()` runs while another task is reading. The generation counter is a clean optimization: if task B waits for the lock while task A reloads, B skips the redundant filesystem scan.

## Issues

1. **⚠️ Warning — `_with_reload` runs the read under the lock**: Holding the lock across the read means all concurrent `_list_tools`, `_get_tool`, etc. calls are serialized, not just the reloads. For a filesystem-backed provider with reload=True, this is likely fine since reload mode is typically used in development. But for high-concurrency production scenarios, this could become a bottleneck. The alternative (copy-on-write with atomic dict swap) is more complex but scales better. This is an acceptable tradeoff for now.

2. **💡 Suggestion — Type signature of `coro_fn`**: `Callable[..., Any]` loses the type safety that the individual provider methods had. Since `_with_reload` is private and only called in 8 well-defined places, this is pragmatic. No change needed, but worth noting.

3. **💡 Suggestion — Root existence warning on every reload**: The new `logger.warning` in `_load_components` fires on every reload if the root doesn't exist. In reload mode, this means a warning every time any method is called. Consider logging once (similar to `_warned_files`) or downgrading to `debug` after the first occurrence.

4. **✅ Good — Generation counter deduplication**: The `_reload_generation` pattern correctly handles the thundering herd problem where N concurrent callers all trigger reload. Only the first actually scans the filesystem.

## Cross-file Impact

- `src/fastmcp/server/providers/filesystem.py`: Core refactor of reload mechanism
- `tests/fs/test_provider.py`: Tests updated to use public API (`_list_tools`) instead of `_ensure_loaded`, plus new concurrency test
- No external API changes — `_ensure_loaded` was private (`_` prefix) ✓

## Assessment

**approve** — Clean fix for a real concurrency bug. The `_with_reload` pattern is simpler and safer than the old two-step approach. The generation counter is a smart optimization. The test directly validates the race condition is fixed.

jlowin (approved):
Withdrawing my earlier changes-requested — the blocking concern about a local `asyncio` import was a misread; it's at module level (line 3).

The fix is correct and clean. Holding the lock across reload+read via `_with_reload` eliminates the window where `clear()` could run between `_ensure_loaded()` and the actual read. The generation counter is a nice dedup for concurrent callers without losing reload-on-every-call semantics for sequential access. The concurrency test directly exercises the race.

PR #3946, opened 2026-04-16
Fix raise_on_error handling for tool tasks

### LINKED ISSUE #3908: raise_on_error not forwarded during task graceful degradation
## Summary

When calling `client.call_tool(name, task=True, raise_on_error=True)` and the server doesn't support tasks (graceful degradation), `raise_on_error` is silently dropped. Errors are swallowed instead of raised.

## Reproduction

```python

server = FastMCP('test')

@server.tool
def failing_tool() -> str:
    """Always fails"""
    raise ValueError('intentional error')

    # Non-task path: correctly raises ToolError
    await client.call_tool('failing_tool', raise_on_error=True)  # raises ✅

    # Task path with graceful degradation: should raise but doesn't
    task = await client.call_tool('failing_tool', task=True, raise_on_error=True)
    result = await task.result()
    print(result.is_error)  # True — error was silently swallowed ❌
```

## Root Cause

`call_tool()` (tools.py:276) calls `_call_tool_as_task()` without forwarding `raise_on_error`:

```python
if task:
    return await self._call_tool_as_task(
        name, arguments, task_id, ttl, meta=request_meta or None
    )
    # ← raise_on_error not passed
```

`_call_tool_as_task()` (tools.py:291) doesn't accept `raise_on_error` at all. In the graceful degradation path (line 351), it calls `_parse_call_tool_result` without it, which defaults to `raise_on_error=False`:

```python
parsed_result = await self._parse_call_tool_result(name, raw_result)
# ← defaults to raise_on_error=False
```

## Fix

1. Add `raise_on_error: bool = True` parameter to `_call_tool_as_task()`
2. Pass it through from `call_tool()` at line 276
3. Forward it to `_parse_call_tool_result()` at line 351

gnanirahulnutakki:
Closes #3908

This threads `raise_on_error` through the tool-task path so `task=True` preserves the same error semantics as direct tool calls.

It also fixes the accepted-task path by teaching `ToolTask.result()` to honor the caller's `raise_on_error` choice instead of hardcoding `True`, which keeps the background and graceful-degradation branches consistent.

Validation:
- `.venv/bin/pytest tests/client/tasks/test_client_tool_tasks.py tests/client/client/test_error_handling.py -q`

### DIFF
--- a/src/fastmcp/client/mixins/tools.py
+++ b/src/fastmcp/client/mixins/tools.py
@@ -288,7 +288,12 @@ class ClientToolsMixin:
 
         if task:
             return await self._call_tool_as_task(
-                name, arguments, task_id, ttl, meta=request_meta or None
+                name,
+                arguments,
+                task_id,
+                ttl,
+                raise_on_error=raise_on_error,
+                meta=request_meta or None,
             )
 
         result = await self.call_tool_mcp(
@@ -308,6 +313,7 @@ class ClientToolsMixin:
         arguments: dict[str, Any] | None = None,
         task_id: str | None = None,
         ttl: int = 60000,
+        raise_on_error: bool = True,
         meta: dict[str, Any] | None = None,
     ) -> ToolTask:
         """Call a tool for background execution (SEP-1686).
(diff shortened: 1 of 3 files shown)


jlowin (approved):
The fix correctly threads `raise_on_error` through the task path without touching any task lifecycle mechanics (creation, tracking, cancellation, notifications). The design is right for the future/promise pattern — errors surface at `.result()` time, not at task submission time, keeping both the graceful-degradation and background-task paths consistent.

Minor: the error message construction in `ToolTask.result()` for the immediate path duplicates the logic in `_parse_call_tool_result`. Necessary here since the immediate result is already parsed, but if this pattern grows a shared helper would be cleaner. Not blocking.

PR #3952, opened 2026-04-17
docs: standardize ToolAnnotations examples

### LINKED ISSUE #3944: Documentation mixes annotations={} options (dict/ToolAnnotations)
### Enhancement

The documentation uses both approaches:

```
@mcp.tool(annotations={"readOnlyHint": True})
def get_user(user_id: str) -> dict:
    """Retrieve user information by ID."""
    return {"id": user_id, "name": "Alice"}

@mcp.tool(
    annotations=ToolAnnotations(
        readOnlyHint=True,
        idempotentHint=True,  # Same result for repeated calls
        openWorldHint=False   # Only internal data
    )
)
def search_products(query: str) -> list[dict]:
    """Search the product catalog."""
    return [{"id": 1, "name": "Widget", "price": 29.99}]
```

Is that intentional? I would think the ToolAnnotations was the better approach for more type-safety, but I'm confused by the mixing of them in one example.

gnanirahulnutakki:
## Summary
- clarify in the tools docs that `annotations` accepts either a plain dict or `ToolAnnotations`
- standardize the user-facing tool annotation examples on `ToolAnnotations` for consistency
- mirror the same example style in the ChatGPT integration docs and the v2 docs set

Closes #3944.

## Validation

### DIFF
--- a/docs/integrations/chatgpt.mdx
+++ b/docs/integrations/chatgpt.mdx
@@ -92,10 +92,12 @@ The connector must be explicitly enabled in each chat session through Developer 
 
 ### Skip Confirmations
 
-Use `annotations={"readOnlyHint": True}` to skip confirmation prompts for read-only tools:
+Use `annotations=ToolAnnotations(readOnlyHint=True)` to skip confirmation prompts for read-only tools:
 
 ```python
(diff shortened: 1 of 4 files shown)


jlowin (approved):
Thanks!

PR #3975, opened 2026-04-19
docs: clarify Azure OAuth proxy blockers beyond DCR

### LINKED ISSUE #3942: Docs feedback: Azure proxy wording overemphasizes DCR and underexplains direct Claude/Azure incompatibility
### Enhancement

The current wording made me assume that the main blocker with Azure was lack of DCR, and that manual client registration would be a valid workaround.

Because of that, I spent time manually creating the Azure app registration and configuring Claude with the Azure metadata, client ID, and client secret. That made it seem like I was on the right path.

The actual blocker turned out to be later in the flow: Claude was calling Azure’s authorization endpoint with parameters Azure did not accept, such as resource. So in practice, the main issue was direct client-to-Azure protocol incompatibility, not just lack of DCR.

I think the docs should make that clearer.

itsmejay80:
The Azure integration docs open by attributing the need for the OAuth Proxy to Azure's lack of Dynamic Client Registration. Reading that, a user can reasonably conclude that manual App registration is a sufficient workaround  it isn't. Azure AD v2.0 also rejects the `resource` parameter that MCP clients include on authorization requests per RFC 8707, so a directly-configured MCP client fails at the `/authorize` call even with a valid, manually-registered app. The [`AzureProvider`](src/fastmcp/server/auth/providers/azure.py) already strips `resource` on upstream calls precisely for this reason, but the docs didn't mention it.

This updates the opening paragraph on both [`docs/integrations/azure.mdx`](docs/integrations/azure.mdx) and the v2 mirror to name both incompatibilities and explain what the proxy does about each — so readers don't waste time on a workaround that can't actually work.

Closes #3942

### DIFF
--- a/docs/integrations/azure.mdx
+++ b/docs/integrations/azure.mdx
@@ -9,9 +9,9 @@ import { VersionBadge } from "/snippets/version-badge.mdx"
 
 <VersionBadge version="2.13.0" />
 
-This guide shows you how to secure your FastMCP server using **Azure OAuth** (Microsoft Entra ID). Azure is incompatible with direct MCP client authentication in two ways: it doesn't support Dynamic Client Registration, and Azure AD v2.0 rejects the `resource` parameter that MCP clients send on authorization requests (RFC 8707). Manually creating an Azure App registration and pointing an MCP client at Azure directly is *not* a sufficient workaround — the authorization request itself will fail.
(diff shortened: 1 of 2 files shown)


jlowin:
Closing this. The current docs already state "this integration uses the OAuth Proxy" — that isn't ambiguous. The `resource` parameter behavior is an implementation detail the proxy handles transparently, and surfacing it in the introductory paragraph adds noise without helping anyone who's actually following the integration as documented.

The underlying issue (#3942) describes a scenario where the user bypassed the FastMCP proxy entirely and pointed Claude directly at Azure — we can't document our way around that.

PR #3985, opened 2026-04-20
fix: handle JSON Pointer style circular refs in dereference_refs

### LINKED ISSUE #3893: DereferenceRefsMiddleware crashes with RecursionError on JSON Pointer style $ref (non-$defs)

Alm0stSurely:
## Problem

C# / .NET MCP servers emit schemas with circular `$ref` using JSON Pointer style (e.g. `#/properties/nodes/items`) rather than `$defs`-based style. `jsonref.replace_refs` turns these into self-referential Python dicts that Pydantic rejects with "Circular reference detected". This crashes `DereferenceRefsMiddleware` when serving tools from .NET MCP servers.

Fixes #3893

## Analysis

The existing `_defs_have_cycles` guard only detects cycles inside `$defs`. JSON Pointer circular refs bypass this check because they point directly into the schema tree. After `replace_refs`, the resulting dict contains a Python-level circular reference (the same dict appears inside itself via `id()` identity), which makes the schema unserializable.

## Solution

Add `_has_circular_refs` to detect post-dereference Python-level cycles by tracking object ids during traversal. When a cycle is found, fall back to `resolve_root_ref` which preserves the original `$ref` instead of producing an unserializable structure.

## Benchmarks

No performance impact: the cycle check is O(n) over the already-dereferenced schema and only runs when `replace_refs` succeeds. For non-circular schemas, the overhead is a single linear scan. All 33 existing `test_json_schema.py` tests pass, plus a new test covering the JSON Pointer circular ref case.

## Notes

- The fallback behavior matches the existing circular `$defs` handling: preserve the original schema rather than crash.
- This is a minimal, surgical fix that doesn't change the API or behavior for non-circular schemas.

### DIFF
--- a/src/fastmcp/utilities/json_schema.py
+++ b/src/fastmcp/utilities/json_schema.py
@@ -1,6 +1,5 @@
 from __future__ import annotations
 
-import copy
 from collections import defaultdict
 from typing import Any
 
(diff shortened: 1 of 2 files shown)


jlowin:
Thanks for the fix — the diagnosis is right and the approach works. We're going to close this in favor of [#3896](https://github.com/PrefectHQ/fastmcp/pull/3896), which addresses the same `RecursionError` with a more surgical change: adding `RecursionError` to the existing `except JsonRefError` clause in `dereference_refs`. That keeps the fix minimal and consistent with how `$defs`-style cycles are already handled.

The `_has_circular_refs` post-dereference scan is a valid technique, but the two-pass approach (dereference then scan) adds complexity that the simpler `except (JsonRefError, RecursionError)` avoids. #3896 also preserves `_strip_discriminator` and `_strip_remote_refs`, which handle important OpenAPI and SSRF edge cases.

Appreciate the effort here!

Issue #3451, opened 2026-03-10
Interoperating with OpenTelemetry instrumentation in the MCP SDK


aabmass:
### Enhancement

Hello from OpenTelemetry python 👋  First off, thank you for instrumenting with OTel directly, it's very exciting to see!

I wanted to give an FYI that I'm working on adding instrumentation to the MCP SDK directly https://github.com/modelcontextprotocol/python-sdk/issues/421 ([PR](https://github.com/modelcontextprotocol/python-sdk/pull/2132) if you're curious). I imagine that will have some impact on what's been built in this repo, and at worst there will be doubled spans. It would be relatively easy to make the layers interoperate, and I'm happy to brainstorm on this.

I've also sent a PR for https://github.com/modelcontextprotocol/python-sdk/issues/1969 which fixes contextvar propagation (which OTel uses) from the MCP Session layer to the transport layer (e.g. opentelemetry-instrumentation-httpx).

jlowin:
Thanks @aabmass! Copying @chrisguidry for visibility, as he's the OTEL guru on the FastMCP side

chrisguidry:
Thanks for the heads up, @aabmass!  I think we can definitely unify things once the SDK's spans are in place.  We tried to stick as close to the [OTel semantic conventions for MCP](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/gen-ai/mcp.md), so I'm betting that once you have spans in place, we can just remove ours and add back any fastmcp-specific attributes to them where we need.  Theoretically, our unit test suite should remain unchanged 🤞

strawgate:
Happy to help look at this too

ringerc:
I just hit this issue with `httpx` not seeing trace context. Quite confusing.

Context propagation should still work at application layer if `_meta` is decoded by the request receiver, but two parallel traces may be generated (one for transport, one for application) which is very confusing.

### Workaround

Patch `StreamableHTTPTransport.post_writer` to restore trace context from `_meta` before spawning request handler tasks:

1. Extract `traceparent`/`tracestate` from `params._meta` when processing each message
2. Wrap the task function to attach the extracted OTel context before execution
3. The httpx instrumentation then sees the correct parent and injects matching HTTP headers

```python
# Extract trace context from _meta
traceparent = params.get("_meta", {}).get("traceparent")
if traceparent:
    carrier = {"traceparent": traceparent}
    trace_ctx = extract(carrier)

# Wrap handler to run with restored trace context
[...]

PR #3997, opened 2026-04-21
OTEL: fix end-to-end MCP trace interop

### LINKED ISSUE #3993: Native telemetry breaks third-party OpenTelemetry instrumentations
FastMCP's native OpenTelemetry telemetry interferes with third-party instrumentations that need to control the span hierarchy. We encountered this while building `opentelemetry-instrumentation-fastmcp`, which follows the [MCP semantic conventions](https://opentelemetry.io/docs/specs/semconv/gen-ai/mcp/).

**The problem has three parts:**

**1. `start_as_current_span` overrides external context.** FastMCP's `client_span` and `server_span` use `tracer.start_as_current_span(...)`, which unconditionally sets FastMCP's spans as the active trace context. When a third-party instrumentation has already established a parent span (e.g., an `invoke_agent` or `execute_tool` span), FastMCP's span replaces it as the active context. Downstream `propagate.inject()` calls then propagate FastMCP's span ID instead of the instrumentation's, breaking the client → server trace link. The result: client and server spans end up on different traces.

**2. No standard flag to disable native telemetry.** When users bring their own instrumentation — either via zero-code (`opentelemetry-instrument`) or manually — there's no way to turn off FastMCP's built-in spans. This leads to duplicate, conflicting span hierarchies where both FastMCP and the external instrumentation create overlapping spans for the same operations.

**3. Native spans don't follow MCP semantic conventions.** The [OTel GenAI MCP semantic conventions](https://opentelemetry.io/docs/specs/semconv/gen-ai/mcp/) define standard attributes (`gen_ai.operation.name`, `mcp.method.name`, `network.transport`, etc.) and a specific span structure for MCP operations. FastMCP's native spans use custom naming and attributes that don't align with these conventions, making them less useful for observability tooling that expects the standard schema.

**Minimal repro for issue 1:**

```python
from opentelemetry import trace, context, propagate

tracer = trace.get_tracer("my-instrumentation")

# External instrumentation creates and activates a parent span
parent = tracer.start_span("execute_tool get_weather")
token = context.attach(trace.set_span_in_context(parent))

# FastMCP's client_span replaces the active context
with client_span("tools/call", ...):  # uses start_as_current_span internally
    carrier = {}
    propagate.inject(carrier)
    # carrier now has FastMCP's span ID, not the instrumentation's parent
    # → server extracts a different parent → traces break

context.detach(token)
```

**Expected behavior:** A way to disable native span creation so external instrumentations can own the span hierarchy while FastMCP's context propagation (`inject_trace_context`/`extract_trace_context`) continues to work.

### LINKED ISSUE #3998: OpenTelemetry: FastMCP client does not propagate trace context on initialize and list methods
## Summary

FastMCP's client still behaves like an opaque MCP client on some protocol requests. `initialize`, `tools/list`, `resources/list`, `resources/templates/list`, and `prompts/list` create local client spans but do not inject trace context into MCP `_meta`.

## Use case

Wrap a FastMCP client in an outer OpenTelemetry span, then connect and call `initialize()` or any of the list methods. The client emits a local span, but the server side does not receive `traceparent` / `baggage` through MCP `_meta`, so the handshake or discovery request is disconnected from the caller's trace.

## Expected vs actual

Expected: these client requests should propagate `_meta.traceparent` and `_meta.baggage` the same way `tools/call`, `resources/read`, and `prompts/get` already do.

Actual: the client calls the MCP SDK methods directly, so those requests cross the protocol boundary without MCP trace context.

Related: #3887 tracks broader method coverage. This issue is specifically about propagation on methods the FastMCP client already wraps or exposes today.

strawgate:
FastMCP's OpenTelemetry support worked best when FastMCP owned the entire MCP exchange. In mixed setups, native FastMCP spans could compete with outer instrumentation for ownership of the MCP trace, server spans could prefer ambient transport context over the propagated MCP parent, and the FastMCP client still sent `initialize` and discovery requests like an opaque client. The result was traces with the right raw events but the wrong hierarchy at the MCP boundary.

This makes FastMCP compose cleanly as both client and server. FastMCP now has a supported `propagation_only` mode plus `suppress_fastmcp_telemetry()` for cases where another instrumentation layer should own MCP spans, server spans parent from propagated MCP context while linking ambient transport spans, baggage stays current during execution, and the FastMCP client now propagates `_meta` on `initialize`, `tools/list`, `resources/list`, `resources/templates/list`, and `prompts/list`. The new interop tests cover mixed instrumentation, opaque third-party MCP clients, and the missing client handshake/discovery paths so this behavior is exercised end to end instead of inferred.

```python
from fastmcp import Client
from fastmcp.telemetry import suppress_fastmcp_telemetry

with tracer.start_as_current_span("execute_tool weather"):
    async with Client(server, auto_initialize=False) as client:
        with suppress_fastmcp_telemetry():
            await client.initialize()
            await client.list_tools()
            result = await client.call_tool("weather", {"city": "Chicago"})
```

Refs #3451. Closes #3993. Closes #3998.

🤖 Generated with Codex

### DIFF
--- a/docs/more/settings.mdx
+++ b/docs/more/settings.mdx
@@ -28,6 +28,12 @@ You can change which `.env` file is loaded by setting the `FASTMCP_ENV_FILE` env
 | `FASTMCP_ENABLE_RICH_TRACEBACKS` | `bool` | `true` | Use rich tracebacks for errors. |
 | `FASTMCP_DEPRECATION_WARNINGS` | `bool` | `true` | Show deprecation warnings. |
 
+## Telemetry
+
+| Environment Variable | Type | Default | Description |
+|---|---|---|---|
+| `FASTMCP_TELEMETRY_MODE` | `Literal["native", "propagation_only"]` | `native` | Controls FastMCP's native OpenTelemetry span creation. `native` emits FastMCP MCP spans and propagates trace context. `propagation_only` keeps `_meta` trace propagation but suppresses FastMCP's own spans so another instrumentation layer can own the MCP span hierarchy. |
+
 ## Transport & HTTP
 
 These control how the server listens when running with an HTTP transport.
--- a/docs/servers/telemetry.mdx
+++ b/docs/servers/telemetry.mdx
@@ -17,6 +17,8 @@ FastMCP uses the OpenTelemetry API for instrumentation. This means:
 - **Bring your own SDK** - You control collection, export, and sampling
 - **Works with any OTEL backend** - Jaeger, Zipkin, Datadog, New Relic, etc.
 
+FastMCP also propagates OpenTelemetry context through MCP `params._meta`, including `traceparent`, `tracestate`, and `baggage` when present.
+
 ## Enabling Telemetry
 
 The easiest way to export traces is using `opentelemetry-instrument`, which configures the SDK automatically:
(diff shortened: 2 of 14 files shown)


strawgate:
Closing in favor of PR #4046 which has a clean single-commit history.

PR #4000, opened 2026-04-21
feat(settings): accept FASTMCP__FIELD env vars as aliases of FASTMCP_FIELD

mtthidoteu:
## Problem

Many projects separate namespace prefix from field name with a **double** underscore in their env vars (`DATABASE__PASSWORD`, `COGNITO__CLIENT_SECRET`) — it makes namespaces visually distinct from `WORD_WORD` field names.

FastMCP doesn't follow that convention. With `env_prefix=\"FASTMCP_\"` + `env_nested_delimiter=\"__\"`, pydantic-settings parses `FASTMCP__HOME` as field `_home` (doesn't exist) and silently drops it. Only `FASTMCP_HOME` (single underscore between prefix and field) is visible to `Settings`.

For projects that want to keep a single convention across their whole stack, this means writing a bootstrap shim that mutates `os.environ` before fastmcp is imported:

```python
# Every project wiring fastmcp ends up with this somewhere
for k, v in list(os.environ.items()):
    if k.startswith(\"FASTMCP__\"):
        os.environ.setdefault(\"FASTMCP_\" + k[len(\"FASTMCP__\"):], v)
```

This PR makes that shim unnecessary.

## Change

Override `settings_customise_sources` to inject `FASTMCP_<FIELD>` aliases for any `FASTMCP__<FIELD>` entries already loaded by the env and dotenv sources. The helper mutates the existing source instances in place — substituting subclasses instead would lose init-time overrides like `Settings(_env_file=...)` that pydantic-settings bakes into the default source instances.

Backward-compatible:

- Every existing `FASTMCP_<FIELD>` env var still resolves exactly as before.
- `FASTMCP_<SUBMODEL>__<FIELD>` nested vars (e.g. `FASTMCP_DOCKET__NAME`) still resolve exactly as before — the alias only translates the namespace boundary, not the nested delimiter.
- When both `FASTMCP_HOME` and `FASTMCP__HOME` are set, the canonical single-underscore form wins (via \`setdefault\`).
- Only previously-ignored `FASTMCP__<FIELD>` env vars gain new meaning.

## Design alternatives considered

- **\`AliasChoices\` / \`validation_alias\` per field** — would duplicate the alias declaration across ~30 fields and wouldn't help future fields added to \`Settings\` or \`DocketSettings\`.
- **Subclass \`EnvSettingsSource\` and \`DotEnvSettingsSource\`** — initially tried, but substituting new source instances in \`settings_customise_sources\` loses the init-time \`_env_file\` override that pydantic-settings bakes into the default \`dotenv_settings\` it passes in. Mutating in place avoids that gotcha.
- **Runtime rename (\`home\` → nested submodel)** — breaking for every current user.

## Tests

New \`tests/test_settings_double_underscore_alias.py\`:

- \`FASTMCP_HOME\` canonical (regression)
- \`FASTMCP__HOME\` alias resolves equivalently
- Canonical wins over alias when both set
- Nested \`FASTMCP_DOCKET__NAME\` (regression)
- Nested via alias prefix \`FASTMCP__DOCKET__NAME\` resolves equivalently
- Non-FASTMCP env vars are not touched
- \`.env\` file values are covered too (via \`_env_file\` override)

All 7 pass. Existing \`tests/cli/\` + \`tests/deprecated/test_settings.py\` (482 tests) and any test matching \`-k \"settings or env\"\` (143 tests) still pass with no changes. \`uv run --frozen ty check src/fastmcp/settings.py\` is clean.

## Context

We use this shim in two downstream projects today and would rather delete it once a fastmcp release carries this fix. Happy to iterate on the approach if you'd prefer a different hook point.

### DIFF
--- a/src/fastmcp/settings.py
+++ b/src/fastmcp/settings.py
@@ -10,8 +10,10 @@ from platformdirs import user_data_dir
 from pydantic import Field, field_validator
 from pydantic_settings import (
     BaseSettings,
+    PydanticBaseSettingsSource,
     SettingsConfigDict,
 )
+from pydantic_settings.sources.providers.env import EnvSettingsSource
 
 from fastmcp.utilities.logging import get_logger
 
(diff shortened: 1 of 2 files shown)


mtthidoteu:
Closing — on reflection, this asks fastmcp to carry complexity for a convention mismatch that's ours to own. We're switching to the single-underscore form (like we do for AWS_REGION, OPENAI_API_KEY, etc.) and dropping the shim downstream. Thanks for your time.

PR #4014, opened 2026-04-22
chore: use SHA-256 for dynamic provider module name suffix

beejak:
When avoiding `sys.modules` collisions for filesystem-discovered providers, derive the private module suffix with SHA-256 instead of SHA-1 (weaker per crypto hygiene / static scanners).


jlowin:
This is purely to compute an identifier; there's no need for cryptographic security concerns. If we could do it faster we would.

PR #4019, opened 2026-04-22
fix(json_schema_to_type): add allOf and oneOf handling

Will-hxw:
## Summary
- Add allOf and oneOf handling to json_schema_to_type() function: oneOf returns Union type, allOf attempts to merge object types into a single dataclass

## Why
Issue #3839: json_schema_to_type() silently returns Any for both allOf and oneOf schemas, causing type information loss.

## Validation
- [x] Syntax check passed
- [x] Commit pushed to fork

PR #4020, opened 2026-04-22
fix(json_schema): address P1/P2 review comments on PR #4019

Will-hxw:
## Summary
Addressed two review comments from chatgpt-codex-connector[bot] on PR #4019:

- **P1 (DataclassField.MISSING → MISSING)**: The code imported `Field as DataclassField` but then referenced `DataclassField.MISSING`, which doesn't exist. Fixed to use `dataclasses.MISSING` directly (aliased as `DC_MISSING`).

- **P2 (preserve null in oneOf/allOf)**: The `oneOf` and `allOf` branches previously dropped `type(None)` and never reintroduced it, so `{"oneOf": [{"type": "string"}, {"type": "null"}]}` was converted to `str` instead of `str | None`. Fixed to check for null and preserve it in the union, matching the pattern already used by `anyOf`.

## Changes
- `_merge_object_types()`: Changed `DataclassField.MISSING` to `DC_MISSING` (aliased from `dataclasses.MISSING`)
- `_schema_to_type()`: Added `has_null` tracking for both `oneOf` and `allOf` branches, using `| None` syntax for single-type-with-null and including `type(None)` in multi-type unions

## Related
Supersedes / fixes review comments on PR #4019

### DIFF
--- a/src/fastmcp/utilities/json_schema_type.py
+++ b/src/fastmcp/utilities/json_schema_type.py
@@ -380,6 +380,56 @@ def _return_Any() -> Any:
     return Any
 
 
+def _merge_object_types(types: list[type]) -> type | None:
+    """Try to merge multiple object types into one.
+
+    For allOf schemas where all sub-schemas are objects/dataclasses,
+    we can create a merged dataclass with combined properties.
+    Returns None if types can't be meaningfully merged.
+    """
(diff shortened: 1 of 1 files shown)


Will-hxw:
This supersedes PR #4019. Contains the original allOf/oneOf fix plus P1/P2 review responses. The original PR has been noted for closure.

jlowin:
Closing as discussed in #4019

PR #4036, opened 2026-04-24
Add log_level parameter to FastMCP errors

### LINKED ISSUE #4035: Server-side exception logging for expected control-flow errors creates noise in error monitoring
## Problem

When using FastMCP errors as control flow to guide LLM behavior, every error triggers `logger.exception()` on the server side, creating noise in production error monitoring systems (Sentry, DataDog, etc.).

Example workflow where this happens:
```python
@mcp.tool
def check_model_status(model_id: str) -> dict:
    if not model_registry.is_deployed(model_id):
        # This is expected behavior to guide the LLM, not a bug
        # But it logs a full stack trace server-side
        raise ToolError(f"Model {model_id} not deployed. Call deploy_model first.")
    return model_registry.get_status(model_id)
```

In multi-step LLM workflows, errors often represent normal operation (prerequisite checks, validation, state machine transitions) rather than bugs. The LLM needs to see these error messages to know what to do next, but server operators don't need stack traces for expected behavior.

Currently, there's no way to send an error to the LLM client without also logging it server-side as an exception.

daniel-tsiang:
## Description

Closes #4035

Adds `log_level` parameter to `FastMCPError` and all subclasses (`ToolError`, `ResourceError`, `PromptError`, etc.) to control the server-side log level for expected control-flow errors.

When building multi-step LLM workflows, errors often represent normal control flow rather than bugs. For example, a tool might raise `ToolError("Model not deployed, call deploy_model first")` to guide the LLM's next action. These expected errors create noise in production error monitoring systems (Sentry, DataDog) that typically capture ERROR-level logs. Rather than suppressing logs entirely (losing all visibility), this lets server operators downgrade the log level so the error is still recorded but doesn't trigger alerts.

### Usage

```python
@mcp.tool
def check_model_status(model_id: str) -> dict:
  if not model_registry.is_deployed(model_id):
      raise ToolError(
          f"Model {model_id} not deployed. Call deploy_model first.",
          log_level=logging.WARNING  # Log at WARNING instead of ERROR
      )
  return model_registry.get_status(model_id)
```

The error still reaches the LLM client with the full message and is still logged server-side — just at the specified level instead of ERROR.

### Implementation

- Added `log_level: int = logging.ERROR` parameter to `FastMCPError.__init__`
- Updated catch blocks in `server.py` to use `logger.log(e.log_level, ..., exc_info=True)` instead of `logger.exception()`
- Updated `sampling/run.py` for sampling tool errors
- Fixed pre-existing bug where `function_prompt.py` double-wrapped `PromptError`

## Contribution type

- [x] Enhancement (maintainers typically implement enhancements — see [CONTRIBUTING.md](../CONTRIBUTING.md))

## Checklist

### DIFF
--- a/src/fastmcp/exceptions.py
+++ b/src/fastmcp/exceptions.py
@@ -1,5 +1,7 @@
 """Custom exceptions for FastMCP."""
 
+import logging
+
 from mcp import McpError  # noqa: F401
 
 
@@ -15,9 +17,9 @@ class FastMCPDeprecationWarning(DeprecationWarning):
 class FastMCPError(Exception):
     """Base error for FastMCP."""
 
-    def __init__(self, *args: object, suppress_log: bool = False) -> None:
+    def __init__(self, *args: object, log_level: int = logging.ERROR) -> None:
         super().__init__(*args)
-        self.suppress_log = suppress_log
+        self.log_level = log_level
 
 
 class ValidationError(FastMCPError):
--- a/src/fastmcp/exceptions.py
+++ b/src/fastmcp/exceptions.py
@@ -15,6 +15,10 @@ class FastMCPDeprecationWarning(DeprecationWarning):
 class FastMCPError(Exception):
     """Base error for FastMCP."""
 
+    def __init__(self, *args: object, suppress_log: bool = False) -> None:
+        super().__init__(*args)
+        self.suppress_log = suppress_log
+
 
 class ValidationError(FastMCPError):
     """Error in validating parameters or return values."""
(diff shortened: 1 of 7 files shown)


jlowin:
I don't think suppressing the log is the right remedy here? Perhaps allowing a custom log level override? Otherwise the server has no report that the error took place.

daniel-tsiang:
> I don't think suppressing the log is the right remedy here? Perhaps allowing a custom log level override? Otherwise the server has no report that the error took place.

@jlowin Thanks Jeremiah, great suggestion. I've amended the PR to allow a custom log level override instead. Please can you take another look?

jlowin (approved):
Thanks!

PR #4062, opened 2026-04-26
docs: add Japanese README

eltociear:
Add Japanese translation of the README to improve accessibility for Japanese-speaking developers.

Japan has a growing MCP developer community, and a localized README helps with adoption. The translation covers the key sections: introduction, installation, three pillars (Server/Client/Agent), and quick start examples.

### DIFF
--- a/README.ja.md
+++ b/README.ja.md
@@ -1,0 +1,84 @@
+<div align="center">
+
+# FastMCP 🚀
+
+<strong>速く動いて、ものを作ろう。</strong>
+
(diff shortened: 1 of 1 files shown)


jlowin:
Thank you for the contribution, but we do not maintain localizations here as we are not equipped to keep them up to date at this time, nor evaluate them on an ongoing basis. Users are encouraged to use AI tooling to always get the most up-to-date information in their preferred language.

Issue #2647, opened 2025-12-18
[Client] Correlation for elicitation and sampling requests to tool calls


fschoenberger:
### Enhancement

I am building a multi-user application using FastMCP's *client* where multiple tool calls are processed concurrently via asyncio. In this environment, the server often needs to trigger sampling or elicitation requests back to the client. Think of the application as something like MCP inspector. 

**The Problem**

Currently, when the global `sampling_handler` or `elicitation_handler` is triggered, the `RequestContext` does not provide metadata linking that handler execution back to the original tool call.

For example:
- User A and User B both trigger tool calls, which execute concurrently.
- The server needs to ask for a "sampling" approval.
- Without a unique identifier (like a `tool_call_id` or `stream_id`), my application cannot reliably route the approval request back to the specific user/future that initiated the chain.

It's not a solution to just use different MCP sessions per user, because the exact same problem happens if a user invokes two tools at the same time (for example, if I want to display in the application which sampling/ elicitation request originated from where). 

**Proposed Solution**

Enhance the RequestContext to include an optional field (for example, `originating_call_id` or `tool_call_id`).

From a [spec perspective](https://modelcontextprotocol.io/specification/2025-11-25/basic/transports#multiple-connections), this should be fine:

> 1. The client MAY remain connected to multiple SSE streams simultaneously.
> 2. The server MUST send each of its JSON-RPC messages on only one of the connected streams; that is, it MUST NOT broadcast the same message across multiple streams.

It should thus be clear from the stream which tool call the request belongs to.

jlowin:
I'm surprised this isn't handled automatically by the SDK, which I thought adds (or exposes) the related request ID on each of these calls. MCP sessions are per-client and sometimes per-request, so there should be no possibility of sending an elicitation request to the wrong client simply because the session isn't available. Can you say more about your situation? It sounds like maybe you are attempting to use a shared session for multiple clients?

fschoenberger:
Thanks for getting back to me!

I'm building an API that streams MCP tool execution events (sampling requests, elicitations and results) back to clients via SSE (the use case is similar to MCP inspector). The core problem is that I cannot reliably correlate futures created during tool execution back to their originating tool invocation stream.

The current architecture has an SSE endpoint that invokes tools and registers event handlers to capture futures:

```python
@router.post("/{server_id}/tools/{tool_name}/invoke")
async def invoke_tool(server_id: str, tool_name: str, request: InvokeToolRequest):
[...]

PR #4068, opened 2026-04-26
Fix sampling from background tasks

cuyua9:
## Summary
- use the captured `origin_request_id` when server-side sampling is invoked from a background task
- keep foreground sampling behavior unchanged because `origin_request_id` resolves to the active request id there
- add an end-to-end background task regression test that calls `ctx.sample()` without an active `request_context`

## Why
`Context.request_id` raises when code runs in a background task because there is no active MCP request context in the worker. `Context.sample()` still used `context.request_id` when sending the server-initiated sampling request, so a task that called `ctx.sample()` could fail before the client sampling handler was reached.

Background task contexts already capture `origin_request_id` when the task is submitted. Passing that value to `create_message()` preserves request routing for foreground calls and lets background sampling use the stored originating request id instead of touching the unavailable request context.

Related to #2647, but this is a narrow runtime fix rather than the full client-side correlation feature.

## Testing
- `uv run pytest tests/server/tasks/test_context_background_task.py -q`
- `uv run ruff check src/fastmcp/server/sampling/run.py tests/server/tasks/test_context_background_task.py`

### DIFF
--- a/src/fastmcp/server/sampling/run.py
+++ b/src/fastmcp/server/sampling/run.py
@@ -532,7 +532,7 @@ async def sample_step_impl(
             model_preferences=_parse_model_preferences(model_preferences),
             tools=sdk_tools,
             tool_choice=effective_tool_choice,
-            related_request_id=context.request_id,
+            related_request_id=context.origin_request_id,
         )
 
     # Check if this is a tool use response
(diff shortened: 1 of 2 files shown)


jlowin (approved):
Thanks!

PR #4071, opened 2026-04-26
feat(openapi): support per-call HTTP headers via http_headers / _meta

### LINKED ISSUE #4025: OpenAPI: per-call auth headers on call_tool() to enable shared FastMCP instances across tenants
## Problem

`FastMCP.from_openapi()` builds a FastMCP instance bound to one `httpx.AsyncClient`. In a multi-tenant proxy scenario — where many callers share the same OpenAPI server but each has a different `Authorization` token — there is no way to pass per-call auth headers into `call_tool()`. The only option today is to construct a separate `FastMCP` per caller.

```python
# Today: one FastMCP per caller because auth is baked into the client
for caller_token in caller_tokens:
    client = httpx.AsyncClient(
        headers={"Authorization": f"Bearer {caller_token}"},
    )
    mcp = FastMCP.from_openapi(spec, client=client)
    result = await mcp.call_tool("search", {"q": "alice"})
    await client.aclose()

# Desired: one FastMCP shared across callers, auth injected at call time
mcp = FastMCP.from_openapi(spec, client=base_client)  # built once, no auth
result = await mcp.call_tool(
    "search",
    {"q": "alice"},
    http_headers={"Authorization": "Bearer <per-caller-token>"},  # injected here
)
```

## Expected vs actual

`call_tool()` accepts arguments for the tool and nothing else. There is no way to supply headers that are scoped to a single call without rebuilding the FastMCP instance.

`OpenAPITool.run()` does call `get_http_headers()` to pick up ambient headers from the ContextVar, but:

1. `authorization` is excluded from that default set (correctly, per #3262 — I'm not asking to revert that).
2. In a non-Starlette call path (e.g. calling `call_tool()` directly in Python, outside any HTTP transport), the ContextVar is empty anyway.

## Design sketch

The simplest fit with the framework would be a `http_headers` kwarg threaded from `call_tool()` down to `OpenAPITool.run()`, applied with the same non-overwriting precedence the client-headers block already uses:

```
call_tool("search", args, http_headers={"Authorization": "Bearer ..."})
  └─► OpenAPITool.run(arguments, http_headers={...})
        └─► applied on top of client.headers, does not overwrite
```

Callers that don't need per-call headers pass nothing; the default path is unchanged. The security fix in #3262 is unaffected — transport auth still doesn't leak to downstream APIs by default.

Happy to discuss API shape or implementation details if this direction makes sense.

MukundaKatta:
## Why

Closes #4025.

`FastMCP.from_openapi()` binds auth to the `httpx.AsyncClient` you pass in. In a multi-tenant proxy where one FastMCP instance is shared across callers (each with their own bearer token), there is no way to inject per-caller auth on `call_tool()` — you have to construct a new `FastMCP` per caller and tear it down after the call.

`OpenAPITool.run()` already pulls headers from the inbound MCP transport via `get_http_headers()`, but that path is empty when there is no MCP HTTP transport — e.g. when `call_tool()` is invoked directly from Python inside another service. The reporter's case has no Starlette ContextVar, so no headers reach the upstream API. (See the issue's follow-up comment for the comparison with #346 / #575.)

## What

Two ways to inject headers for one call only — both converge on a ContextVar that the OpenAPI components read at request-build time:

```python
# Python API
result = await mcp.call_tool(
    "search",
    {"q": "alice"},
    http_headers={"Authorization": f"Bearer {caller_token}"},
)

# Wire API (inside a CallToolRequest)
_meta = {"fastmcp": {"http_headers": {"Authorization": "Bearer ..."}}}
```

- New ContextVar `_current_call_http_headers` in `src/fastmcp/server/dependencies.py`, plus the public helper `get_call_http_headers()`.
- `FastMCP.call_tool()` gains an `http_headers=` kwarg that sets the ContextVar around the call and resets it on the way out (so headers do not leak across calls).
- `_call_tool_mcp` extracts `_meta.fastmcp.http_headers` (mirroring how it already extracts `_meta.fastmcp.version`) and forwards to `call_tool(http_headers=...)`. Malformed payloads are silently dropped.
- `OpenAPITool.run()` and `OpenAPIResource.read()` apply per-call headers *over* the client's defaults and ambient transport headers — per-tenant `Authorization` wins, which is the whole point of the issue.
- Per-call headers are used to build the upstream HTTP request only. They never appear in the `ToolResult` returned to the LLM. There is a regression test for that.

The default path (no `http_headers`, no `_meta.fastmcp.http_headers`) is unchanged. The security stance from #3262 (transport `Authorization` excluded from `get_http_headers()` by default) is unaffected — this PR is a separate, opt-in injection point.

Documented in `docs/integrations/openapi.mdx` under Authentication.

## Tested

New file `tests/server/providers/openapi/test_per_call_headers.py`:

- `TestDefaultBehaviorPreserved` — no per-call headers => existing behavior.
- `TestPerCallHeadersKwarg` — `http_headers=` overrides client defaults; multiple headers; helper readable from inside the call; **does not leak into the result**; **does not leak across calls**.
- `TestPerCallHeadersViaMetaProtocol` — `_meta.fastmcp.http_headers` reaches the upstream; absent meta falls back to client default; malformed payload is ignored.
- `TestContextVarHygiene` — empty mapping outside any call; helper returns a defensive copy.

### DIFF
--- a/docs/integrations/openapi.mdx
+++ b/docs/integrations/openapi.mdx
@@ -68,6 +68,24 @@ mcp = FastMCP.from_openapi(
 )
 ```
 
+#### Per-call headers (multi-tenant)
+
+When one FastMCP instance is shared across callers — each with their own bearer token — auth can be injected at call time instead of being baked into the client. Pass `http_headers=` on `call_tool()` and the values are forwarded to the upstream HTTP request for that one call. Per-call values override anything set on the client.
+
+```python
+# One server, no auth on the client.
+mcp = FastMCP.from_openapi(openapi_spec=spec, client=httpx.AsyncClient(base_url=...))
+
+# Per-caller token, applied for this call only.
+result = await mcp.call_tool(
+    "search",
+    {"q": "alice"},
+    http_headers={"Authorization": f"Bearer {caller_token}"},
+)
+```
+
+The same headers can be supplied over the wire by setting `_meta.fastmcp.http_headers` on the tool call request. Per-call headers are scoped to a single call and never appear in the response sent back to the LLM.
+
 ## Route Mapping
 
 By default, FastMCP converts **every endpoint** in your OpenAPI specification into an MCP **Tool**. This provides a simple, predictable starting point that ensures all your API's functionality is immediately available to the vast majority of LLM clients which only support MCP tools.
--- a/src/fastmcp/server/dependencies.py
+++ b/src/fastmcp/server/dependencies.py
@@ -63,6 +63,7 @@ __all__ = [
     "TaskContextSnapshot",
     "TokenClaim",
     "get_access_token",
+    "get_call_http_headers",
     "get_context",
     "get_http_headers",
     "get_http_request",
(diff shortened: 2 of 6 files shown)


jlowin:
This seems overcomplicated and overloads meta in a way that is highly unlikely to be used in practice. We don't want to leak the internal transport into the client interface.

Issue #4055, opened 2026-04-26
ToolResult isError lost and dedupe_with_versions() shares mutable meta


strawgate:
# ToolResult and deduplication bugs

## Summary

Three bugs in tool result handling and deduplication:

### Bug 1: `ToolResult.isError` lost through `to_mcp_result()`

`src/fastmcp/tools/tool.py:129-134` — `ToolResult.to_mcp_result()` creates a `CallToolResult` with `structuredContent` and `content` but never sets `isError`. The `isError` field on `ToolResult` has no corresponding mapping to `CallToolResult.isError`, so error state is lost in the MCP protocol response.

### Bug 2: `dedupe_with_versions()` shares mutable `meta` dict

`src/fastmcp/utilities/versions.py:321` — `model_copy()` without `deep=True` shares nested mutable objects (dicts, lists) in the `meta` field between the original and deduplicated component. Mutating the deduplicated component's meta also mutates the original.

### Bug 3: `decorator_metadata_not_preserved` in direct `from_function()` calls

When `FunctionTool.from_function()` is called on a function decorated with `@tool(name="x", version="y", ...)`, the `__fastmcp__` metadata is not preserved — a new `ToolMeta` with defaults is created. (Note: the server registration path via `_ensure_tool()` handles this correctly.)

## Impact

- Error state from tools is silently lost — callers can't distinguish errors from success
- Mutation of deduplicated component metadata corrupts the original
- Direct `from_function()` API callers lose decorator metadata

## Fix Suggestion

1. Map `ToolResult.isError` to `CallToolResult.isError` in `to_mcp_result()`
2. Use `model_copy(deep=True)` in `dedupe_with_versions()`
3. Extract `__fastmcp__` metadata in `from_function()` when available

jlowin:
On **Bug 2** (`dedupe_with_versions` sharing nested `meta`): I don't think we should fix this. `meta` is supplied at construction and treated as read-only after — nothing in the framework mutates it in place, so the aliasing only bites if you reach into a built component's meta, which isn't a supported pattern. Deep-copying meta on every dedup is standing cost to defend against that, and `model_copy(deep=True)` doesn't even work here (Pydantic applies `update` after the copy). If someone needs a variant, `Tool.from_tool(..., meta=...)` is the path. Tracking/declining the split-out in #4202.

(For the record: Bug 3 was handled by #4072; Bug 1 — `isError` in `to_mcp_result` — still looks open.)

[...]

PR #4072, opened 2026-04-27
fix: preserve @tool metadata in from_function

lawrence3699:
Direct `Tool.from_function()` and `FunctionTool.from_function()` calls currently rebuild tool metadata from defaults even when the function was decorated with standalone `@tool(...)`. That drops the decorator-provided name, version, title, description, tags, and meta for direct factory users, even though the server registration path already respects the same metadata.

This restores the existing decorator metadata lookup in the direct factory path, but only when no explicit `from_function()` parameters were supplied, so explicit factory arguments keep their current behavior.

```python
@tool(name="custom-greet", version="v1")
def greet(name: str) -> str:
    return f"Hello, {name}!"

Tool.from_function(greet).name
# "custom-greet"
```

Refs #4055 (decorator metadata bug).

## Contribution type

## Checklist

Generated with Codex.

### DIFF
--- a/src/fastmcp/tools/function_tool.py
+++ b/src/fastmcp/tools/function_tool.py
@@ -24,7 +24,7 @@ from pydantic import Field
 from pydantic.json_schema import SkipJsonSchema
 
 import fastmcp
-from fastmcp.decorators import resolve_task_config
+from fastmcp.decorators import get_fastmcp_meta, resolve_task_config
 from fastmcp.exceptions import FastMCPDeprecationWarning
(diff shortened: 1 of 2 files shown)


jlowin (approved):
Tight, correct fix. Mirrors the existing _ensure_tool pattern and the parameterized test directly exercises the @tool decorator path. Note for future: the all-or-nothing semantics mean any explicit kwarg replaces the entire decorator (a user passing only `name=` will silently lose decorator `tags`/`version`/etc.) — defensible and consistent with the mutual-exclusion API design here, but worth a docstring note someday.

PR #4076, opened 2026-04-28
fix(openapi): keep blank values in parse_qs (refs #4056)

MukundaKatta:
## Problem

`HeadlessOAuth.callback_handler` parses OAuth callback query strings with `parse_qs(parsed.query)` without `keep_blank_values=True`.

That drops explicitly empty values, so `?code=abc&state=` is treated the same as a missing `state`, and `?error=invalid_request&error_description=` falls back to `Unknown error` instead of preserving the provider's empty description.

## MRE

```python
from urllib.parse import parse_qs

assert parse_qs("code=abc&state=").get("state") is None
assert parse_qs("code=abc&state=", keep_blank_values=True)["state"] == [""]
```

## Fix

Add `keep_blank_values=True` to the remaining `parse_qs` call in `src/fastmcp/utilities/tests.py`.

## Tests

Added focused coverage for:

- blank `state` round-tripping as `""`
- missing `state` still returning `None`
- blank `error_description` staying blank instead of using the fallback

Refs #4056. This PR is intentionally scoped to the remaining `HeadlessOAuth.callback_handler` call site; #4069 handled the template parser path.

### DIFF
--- a/src/fastmcp/utilities/tests.py
+++ b/src/fastmcp/utilities/tests.py
@@ -254,7 +254,11 @@ class HeadlessOAuth(OAuth):
         if response.status_code == 302:
             redirect_url = response.headers["location"]
             parsed = urlparse(redirect_url)
-            query_params = parse_qs(parsed.query)
+            # keep_blank_values=True so explicitly-empty params (e.g. ?state=)
+            # survive parsing instead of being silently dropped. Real OAuth
+            # callbacks can include empty `state` or `error_description`,
+            # and downstream code distinguishes "" from missing.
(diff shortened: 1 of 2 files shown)


jlowin (approved):
Complementary fix to #4069 — different parse_qs site (HeadlessOAuth.callback_handler), correct one-line fix. I pushed a ruff-format tweak to your branch to clear the static_analysis failure. Future nit: title prefix should be `fix(auth)` not `fix(openapi)` — this code is in the auth callback, not OpenAPI.

PR #4077, opened 2026-04-28
fix(tools): preserve isError flag and deep-copy meta in dedupe (refs #4055)

### LINKED ISSUE #4055: ToolResult isError lost and dedupe_with_versions() shares mutable meta

MukundaKatta:
## Description

Two bugs from #4055 in tool result handling and component deduplication.

`ToolResult.to_mcp_result()` built a `CallToolResult` with `content`, `structuredContent`, and `_meta` but never set `isError`, so a tool returning an error result had the flag silently dropped before it reached the wire — clients could no longer tell an error apart from a success. `ToolResult` now carries an `is_error` field that maps to `CallToolResult.isError`. The shape returned for the no-meta, no-error case is unchanged.

`dedupe_with_versions()` used `model_copy()` without `deep=True`, so the deduplicated component shared nested mutables (dicts and lists inside `meta`) with the original input components. A caller that touched the result's `meta` would silently mutate the originals (and vice versa). Switched to `model_copy(deep=True)`.

```python
from fastmcp.tools.base import ToolResult
from mcp.types import TextContent

result = ToolResult(content=[TextContent(type="text", text="boom")], is_error=True)
mcp_result = result.to_mcp_result()
mcp_result.isError  # True   (was: dropped)
```

Bug 3 from the same issue (decorator metadata in `from_function`) is being addressed separately in #4072 and is not touched here.

Closes #4055

## Contribution type

## Checklist

### DIFF
--- a/src/fastmcp/tools/base.py
+++ b/src/fastmcp/tools/base.py
@@ -76,12 +76,20 @@ class ToolResult(BaseModel):
     meta: dict[str, Any] | None = Field(
         default=None, description="Runtime metadata about the tool execution"
     )
+    is_error: bool = Field(
+        default=False,
+        description=(
+            "Whether the tool call ended in an error. Maps to "
+            "CallToolResult.isError on the wire."
+        ),
+    )
 
     def __init__(
         self,
         content: list[ContentBlock] | Any | None = None,
         structured_content: dict[str, Any] | Any | None = None,
         meta: dict[str, Any] | None = None,
+        is_error: bool = False,
     ):
         if content is None and structured_content is None:
             raise ValueError("Either content or structured_content must be provided")
@@ -118,7 +126,10 @@ class ToolResult(BaseModel):
                 )
 
         super().__init__(
-            content=converted_content, structured_content=structured_content, meta=meta
+            content=converted_content,
+            structured_content=structured_content,
+            meta=meta,
+            is_error=is_error,
         )
 
     def to_mcp_result(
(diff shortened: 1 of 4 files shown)


jlowin (changes requested):
Two changes here, and I want to take just one.

**Drop the dedupe deep-copy at `src/fastmcp/utilities/versions.py:332`.** The existing `**meta` / `**fastmcp` spread already produces fresh outer and inner dicts. The only residual aliasing is for *other* nested dicts inside `meta`, and there's no known caller that mutates them. `model_copy(deep=True)` deep-copies the whole component (tags, icons, annotations, everything) on every dedupe — that's real overhead for a hypothetical bug. If we ever observe a real mutation issue, the right fix is targeted at the call site, not a blanket deep-copy here.

**Keep the `is_error` field on `ToolResult`** — that's a real API gap and worth landing. But reframe it: `ToolResult` never had `is_error`, so this is an enhancement (new feature), not a bug fix or "preservation." Update the title and changelog framing accordingly.

Also: imports inside test bodies at `tests/tools/tool/test_tool.py:620,632` need to move to module level.

Want this as one PR (just the `is_error` change) rather than the current bundle.

jlowin:
Closing as stale

Issue #4054, opened 2026-04-26
Multiple auth handling bugs


strawgate:
# Multiple auth handling bugs

## Summary

Four unrelated auth bugs:

### Bug 1: Auth denial returns "resource not found" instead of auth error

`src/fastmcp/server/middleware/authorization.py` — When `get_resource()` returns `None` because component-level auth denied access, the middleware raises `AuthorizationError("resource not found")`. This is misleading — the resource exists, auth just denied it.

### Bug 2: OIDC discovery fetched on every call

`src/fastmcp/server/auth/oidc_proxy.py:159` — `httpx.get()` is called on every `get_oidc_configuration()` invocation with no caching. OIDC discovery documents are static and should be cached.

### Bug 3: `refresh_expires_in=0` (Keycloak "never") treated as 1 year

`src/fastmcp/server/auth/oauth_proxy/proxy.py:1067-1069`:
```python
if "refresh_expires_in" in idp_tokens and int(idp_tokens["refresh_expires_in"]):
```
Keycloak uses `refresh_expires_in=0` to mean "no expiry" (never expires), but `int(0)` is falsy, causing the else branch to take a 1-year fallback instead.

### Bug 4: `StatefulProxyClient` context `__aenter__` not called

`src/fastmcp/server/providers/proxy.py:965,971` — `_restore_request_context` calls `_current_context.set(Context(fastmcp))` without calling `__aenter__()`. When `__aexit__` later runs, it does nothing because `_tokens` was never populated, leaving stale context.

## Impact

- Users see confusing error messages when auth denies access
- Unnecessary HTTP to OIDC provider on every authentication
- Keycloak offline tokens (infinite refresh) incorrectly expire after 1 year
- Proxy context cleanup fails, potentially leaking context state

## Fix Suggestion

1. Change error message to indicate authorization failure
2. Add caching to `get_oidc_configuration()` with TTL
3. Check `refresh_expires_in == 0` explicitly before the falsy check
4. Call `__aenter__` before setting context, or use a different pattern

rschick:
Bug 2 should also be addressed. We're seeing real impact due to throttling by Auth0.

jmealo:
@rschick Perhaps opening one issue per bug would yield better results?

strawgate:
> Bug 2 should also be addressed. We're seeing real impact due to throttling by Auth0.

[...]

PR #4078, opened 2026-04-28
fix(client): forward __aenter__ to wrapped client in StatefulProxyClient (refs #4054)

MukundaKatta:
## Why

Issue #4054 catalogues four unrelated auth-handling bugs. The `StatefulProxyClient` `Context` lifecycle bug (Bug 4) is the last one outstanding:

- Bug 1 (auth-denial error message): not addressed yet.
- Bug 2 (OIDC discovery caching): not addressed yet.
- Bug 3 (`refresh_expires_in=0` handling): tracked in #4060.
- Bug 4 (this PR): `_restore_request_context` skipped `Context.__aenter__`.

This PR closes out Bug 4 only.

For visibility, two adjacent issues from the same audit batch are being addressed in #4076 (parse_qs blank values, refs #4056) and #4077 (`ToolResult.isError` + dedupe meta deep-copy, refs #4055).

## What

`src/fastmcp/server/providers/proxy.py` — `_restore_request_context` previously did:

```python
_current_context.set(Context(fastmcp))
```

That bypasses `Context.__aenter__`, so:

- `Context._tokens` stays empty, which means the matching `Context.__aexit__` is a no-op.
- `_current_server` is never set, so dependency-injection helpers that rely on it see stale values.
- `SharedContext` is never created when running without a Docket lifespan, so `Shared()` dependencies inside the handler get a stale shared store.

The fresh `Context` therefore looks half-initialised inside the proxy handlers triggered from `StatefulProxyClient`'s receive loop, and its tokens leak as the wrapper returns.

The fix:

1. Make `_restore_request_context` `async` and `await Context(fastmcp).__aenter__()` instead of touching `_current_context` directly. The function returns the entered `Context | None` so the caller can pair the enter with an exit.
2. Update `_make_restoring_handler` to wrap the handler in `try/finally` and call `await ctx.__aexit__(None, None, None)` after the handler runs, so the `_current_context` token (and any `_current_server` / `_shared_context` / docket tokens) are released cleanly.

The `request_ctx` restoration logic and the staleness-detection branch (only override when same session, different request_id) are unchanged — only the `_current_context` side now goes through the proper enter/exit.

## Tests

`tests/server/providers/proxy/test_stateful_proxy_client.py` — new `TestRestoreRequestContext` class covering:

- `test_restore_enters_context_and_sets_current_server` — the returned `Context` has populated `_tokens`, `_current_context.get()` is the fresh `Context`, `_current_server` resolves back to the same `FastMCP`, and `request_ctx` was restored to the stashed value. After `__aexit__`, `_current_context` no longer points at the entered `Context`.
- `test_make_restoring_handler_awaits_aenter_and_aexit` — the handler invoked through the wrapper sees an active `Context` with non-empty `_tokens`, and after the wrapper returns `_current_context` is back to `None` (no token leak).
- `test_restore_returns_none_when_rc_ref_empty` — pins down the `rc_ref[0] is None` shortcut.

The existing `TestStatefulProxyClient` suite (concurrent log routing, stateful state, multi-proxy isolation, elicitation-over-HTTP) should continue to pass since the request_ctx and routing behaviour is unchanged.

## Notes

- Diff is intentionally minimal: one production change to `proxy.py`, one new test class.
- `_restore_request_context` becoming `async` is internal — it's only called from `_make_restoring_handler` (also internal).
- Refs #4054. The other three bugs in that issue are out of scope for this PR.

### DIFF
--- a/src/fastmcp/server/providers/proxy.py
+++ b/src/fastmcp/server/providers/proxy.py
@@ -932,9 +932,9 @@ async def default_proxy_progress_handler(
     await ctx.report_progress(progress, total, message)
 
 
-def _restore_request_context(
+async def _restore_request_context(
     rc_ref: list[Any],
-) -> None:
+) -> Context | None:
     """Set the ``request_ctx`` and ``_current_context`` ContextVars from stashed values.
 
     Called at the start of proxy handler invocations in
@@ -948,12 +948,15 @@ def _restore_request_context(
     ContextVar-dependent and would resolve stale values in the receive
     loop.  Instead we construct a fresh ``Context`` here after restoring
     ``request_ctx``, so its property accesses read the correct values.
-    """
-    from fastmcp.server.context import Context, _current_context
 
+    The returned ``Context`` (when one is entered) must have ``__aexit__``
+    called on it by the caller to release the ``_current_context`` /
+    ``_current_server`` tokens it pushed; ``_make_restoring_handler``
+    handles that via ``try/finally``.
+    """
     stashed = rc_ref[0]
     if stashed is None:
-        return
+        return None
 
     rc, fastmcp_ref = stashed
     try:
@@ -962,13 +965,18 @@ def _restore_request_context(
         request_ctx.set(rc)
         fastmcp = fastmcp_ref()
         if fastmcp is not None:
-            _current_context.set(Context(fastmcp))
-        return
+            # Use ``__aenter__`` (not a raw ``_current_context.set``) so
+            # ``Context._tokens``, ``_current_server`` and ``SharedContext``
+            # are all populated; otherwise the matching ``__aexit__`` would
+            # be a no-op and tokens would leak.
+            return await Context(fastmcp).__aenter__()
+        return None
     if current_rc.session is rc.session and current_rc.request_id != rc.request_id:
         request_ctx.set(rc)
         fastmcp = fastmcp_ref()
         if fastmcp is not None:
-            _current_context.set(Context(fastmcp))
+            return await Context(fastmcp).__aenter__()
+    return None
 
 
 def _make_restoring_handler(handler: Callable, rc_ref: list[Any]) -> Callable:
@@ -980,8 +988,12 @@ def _make_restoring_handler(handler: Callable, rc_ref: list[Any]) -> Callable:
     """
 
     async def wrapper(*args: Any, **kwargs: Any) -> Any:
-        _restore_request_context(rc_ref)
-        return await handler(*args, **kwargs)
+        ctx = await _restore_request_context(rc_ref)
+        try:
+            return await handler(*args, **kwargs)
+        finally:
+            if ctx is not None:
+                await ctx.__aexit__(None, None, None)
 
     return wrapper
 
--- a/tests/server/providers/proxy/test_stateful_proxy_client.py
+++ b/tests/server/providers/proxy/test_stateful_proxy_client.py
@@ -199,3 +199,114 @@ class TestStatefulProxyClient:
                 # one that would hang without the fix.
                 result2 = await client.call_tool("ask_name", {})
                 assert result2.data == "Hello, Alice!"
+
+
+class TestRestoreRequestContext:
+    """Regression tests for the Context lifecycle in `_restore_request_context`.
+
+    Previously `_restore_request_context` called `_current_context.set(Context(fastmcp))`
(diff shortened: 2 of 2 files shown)


jlowin (changes requested):
Three things before merge:

1. **Use a context manager for Context lifecycle.** The current shape (`ctx = await _restore_request_context(...)` + `try/finally: await ctx.__aexit__(None, None, None)`) is hand-rolling lifecycle management that the language already has syntax for. It also loses exception info: when the handler raises, `__aexit__` gets `(None, None, None)` instead of the actual exception tuple. Two ways to fix. Make `_restore_request_context` an `@asynccontextmanager` and have the wrapper do `async with _restore_request_context(rc_ref): return await handler(...)`, or take an `AsyncExitStack` as a parameter and push the Context onto it (matches how `_lifespan_manager` handles conditional entries). Either is fine.

2. **Retitle.** The title says "forward `__aenter__` to wrapped client in `StatefulProxyClient`" but the change is in `_restore_request_context` / `_make_restoring_handler`. Nothing to do with `StatefulProxyClient.__aenter__`. Something like `fix(proxy): enter Context properly in _restore_request_context`.

3. **Hoist test imports to module level.** Every method in `TestRestoreRequestContext` has its imports inside the function body. Move them to the top of `tests/server/providers/proxy/test_stateful_proxy_client.py`.

strawgate:
This PR targets Bug 4 from #4054, and that gap is real: `_restore_request_context` never sets `_current_server`, so any user-supplied sampling/elicitation/roots callback that calls `get_server()` or uses a DI dependency raises `"No FastMCP server instance in context"`. That's worth fixing and stays tracked on #4054.

The reason we're closing this rather than iterating on it: the bug is that the path *doesn't* call `__aexit__` — but the fix isn't to *call* `__aexit__`. It's that this path shouldn't be opening a context-manager scope in the first place.

The `StatefulProxyClient` receive loop is a single long-lived task. `_restore_request_context` isn't opening a scope around a unit of work — it's *repairing* stale ContextVars in place on a task that keeps running and handles the next message, where the next repair just overwrites. A context manager models `enter → bounded work → exit`; there is no natural "exit" here. Bolting one on (here, or via the `@asynccontextmanager` refactor suggested in review) invents a per-handler scope around something deliberately unscoped, and that invented scope is what generates the follow-on problems:

- **Exception info loss**: `finally: await ctx.__aexit__(None, None, None)` drops the real exception. `Context.__aexit__` forwards the exc tuple to `SharedContext.__aexit__` → its `AsyncExitStack`, so error-sensitive `Shared()` CMs see success on a failed handler.
- **Cross-session `_request_state` bleed**: `Context.__aenter__` does `self._request_state = parent._request_state` from the current contextvar. In the shared-ref / `copy.copy` concurrent case this function's own docstring warns about, that "parent" is another session's `Context`, so the entered Context grafts a foreign session's request state.
- **Per-notification churn**: enter/exit now run on every restored callback, including each forwarded `log`/`progress`, allocating+entering+exiting a `SharedContext` (or churning docket/worker tokens) per message.

All three exist *because* a scope was opened that then has to be unwound — they aren't independent bugs to patch.

Decomposing what `__aenter__` does on this path: `_current_context.set`, `_current_server.set`, and the docket/worker sets are pure ContextVar writes that are correct to do set-only on a long-lived task — ContextVars don't leak, they're overwritten, and not `reset()`-ing is the point (we never want the stale prior value back). `_request_state` inheritance must be *skipped*, not performed. The only side-effect that genuinely wants a scoped lifetime is `SharedContext`, and a proxy receive-loop callback using a `Shared()` dependency is exotic, never worked before this PR, and isn't a regression to leave unsupported.

So the minimal, principled fix keeps the original set-only model and adds the one missing line:

```python
request_ctx.set(rc)
fastmcp = fastmcp_ref()
if fastmcp is not None:
    _current_context.set(Context(fastmcp))
    _current_server.set(weakref.ref(fastmcp))   # the actual Bug 4 gap
```

No `__aenter__`/`__aexit__`, so the exc-info issue, the cross-session bleed, and the per-call churn don't exist — not because they're handled, but because there's no scope to get wrong. If `Shared()` in receive-loop callbacks ever becomes a real requirement, the right move is to scope *only* a `SharedContext` per handler call, not to CM-wrap the whole `Context`.

This is also a good illustration of why CONTRIBUTING.md asks for the planned approach to be discussed on the issue before a PR — the lifecycle direction here would have been worth aligning on up front, before the implementation and review cycles. Closing this in favor of the set-only fix tracked on #4054.

Issue #4084, opened 2026-04-29
Windows startup crash due to UnicodeDecodeError when loading skills


jmarenin:
## Description

`SkillProvider._load_skill()` calls `Path.read_text()` without specifying an encoding parameter. On Windows, this defaults to the system locale encoding (typically cp1252), which fails when SKILL.md files contain UTF-8 characters like emoji (✅, ❌, →).

This causes the MCP server to exit immediately with exit code 1 on Windows when loading skills that contain UTF-8 characters.

## Error

```
UnicodeDecodeError: 'charmap' codec can't decode byte 0x9d in position XXX: character maps to <undefined>
```

## Root Cause

In `fastmcp/server/providers/skills/skill_provider.py`, approximately line 13:

```python
content = main_file.read_text()  # ❌ No encoding specified
```

Without an explicit encoding parameter, Python uses the system's default encoding, which is cp1252 on Windows. This cannot decode UTF-8 emoji and special characters.

### Example Code

```Python
from pathlib import Path
from tempfile import TemporaryDirectory
from fastmcp.server.providers.skills import SkillsDirectoryProvider

# Create a temporary skill directory with UTF-8 content
with TemporaryDirectory() as tmpdir:
    skill_dir = Path(tmpdir) / "test-skill"
    skill_dir.mkdir()
    
    # Create SKILL.md with UTF-8 emoji characters
    skill_file = skill_dir / "SKILL.md"
    skill_file.write_text(
        """---
name: test-skill
description: Test skill with UTF-8 characters
---

# Test Skill

- ✅ Success indicator
- ❌ Failure indicator  
- → Arrow symbol
""",
        encoding="utf-8"
    )
    
    # This will fail on Windows with UnicodeDecodeError
    provider = SkillsDirectoryProvider(roots=skill_dir.parent)
    print("Skills loaded successfully")
```

### Version Information

```Text
FastMCP version:                                                         3.2.4
MCP version:                                                            1.27.0
Python version:                                                         3.12.0
```

doneman536:
In upcoming versions of python utf-8 is supported by default. Till then we can have this.

jlowin:
Closed in #4092

PR #4092, opened 2026-05-03
#4084 [Issues] Windows startup crash due to UnicodeDecodeError when l…

doneman536:
#4084 

Added some encoding as fall back.

### DIFF
--- a/src/fastmcp/server/providers/skills/skill_provider.py
+++ b/src/fastmcp/server/providers/skills/skill_provider.py
@@ -237,17 +237,8 @@ class SkillProvider(Provider):
                 f"Main skill file not found: {main_file}. "
                 f"Expected {self._main_file_name} in {self._skill_path}"
             )
-        
-        possible_encodings = ["utf-8", "utf-16", "cp1252", "latin-1"]
-        content = None
-        for encoding in possible_encodings:
-            try:
-                content = main_file.read_text(encoding=encoding)
-                break 
-            except UnicodeDecodeError:
-                continue
-        if content is None:
-            raise UnicodeError(f"Skill file encoding not supported for {main_file}")
+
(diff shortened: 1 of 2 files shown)


jlowin (changes requested):
Thanks for tracking this down — the symptom is real (Windows defaults to cp1252 for `Path.read_text()`), but the fix is the wrong shape.

**The right fix is one line:** `main_file.read_text(encoding="utf-8")`. SKILL.md is markdown and the Anthropic skills convention is UTF-8; explicitly declaring the encoding fixes the cp1252 default cross-platform.

The current PR adds a try-each-encoding loop through utf-8 / cp1252 / latin-1. That's risky because `latin-1` decodes any byte sequence without raising — so if utf-8 ever fails for a real reason (corrupted file, BOM-prefixed UTF-16, etc.), the loop silently mojibakes instead of surfacing the problem. Encoding-sniffing without `chardet` isn't reliable, and the fallback masks bugs.

Two asks:

1. Replace the loop at `src/fastmcp/server/providers/skills/skill_provider.py:240-249` with a single `read_text(encoding="utf-8")`.
2. Add a regression test using the MRE from #4084 — write a SKILL.md with non-ASCII content (emoji is fine) using `encoding="utf-8"`, load it via `SkillsDirectoryProvider`, assert success. This works cross-platform and doesn't need a Windows runner since the fix is about explicit encoding, not OS detection.

Also: there's some trailing whitespace in the diff that prek will flag.

doneman536:
@jlowin  You're right, based on the requested changes:
1. Changed to encoding="utf-8" rather than multiple fallbacks.
2. Added a regression test using the MRE from #4084.
Thank you

jlowin (approved):
Thanks!

PR #4095, opened 2026-05-04
Make Docket reentrant; mounted servers enter their own lifespan

### LINKED ISSUE #4049: Mounted server lifespan bypasses _lifespan_manager()
# Mounted server lifespan management broken

## Summary

When a FastMCP server is mounted inside another server via `FastMCPProvider`, its lifespan is entered via `self.server._lifespan(self.server)` instead of `self.server._lifespan_manager()`. This bypasses Docket infrastructure and causes two problems:

1. **Provider lifespans are never entered** — mounted servers with database connections, caches, or other resources initialized in ` lifespan` won't have those resources available.

2. **Lifespan is re-entered per client session** — `_lifespan_result_set` is never set to `True` (because `_lifespan_manager()` is bypassed), so the lifespan generator is re-entered on every new client connection.

## Root Cause

`FastMCPProvider.lifespan()` at `src/fastmcp/server/providers/fastmcp_provider.py:707`:
```python
async with self.server._lifespan(self.server):
    yield
```

Should use `_lifespan_manager()` to properly manage the lifecycle:
```python
async with self.server._lifespan_manager():
    yield
```

## Impact

- Any mounted server relying on ` lifespan` setup will have broken initialization
- Repeated lifespan entries cause resource leaks (e.g., connection pools created multiple times)
- Background tasks started in ` lifespan` may be started multiple times

## Related

- Issue #311: "Proxied servers have even weird-er lifespans" (open, same root cause)

jlowin:
A mounted server's lifespan was entered but its yielded value was thrown away, so `ctx.lifespan_context` on a mounted-child tool returned the parent's lifespan dict (or `{}` for nested grandchildren whose lifespans never ran at all). `FastMCPProvider.lifespan()` rolled its own minimal entry path that bypassed `_lifespan_manager` — necessarily, because `_lifespan_manager` bundled Docket startup that should only happen once per process.

This makes `_lifespan_manager` itself reentrant. A new `_lifespan_root_active` ContextVar lets `_docket_lifespan` detect that a parent in the same runtime tree already owns Docket and SharedContext, and become a no-op for nested entries. Mounted children now go through the full path — same lock, ref-counting, result cache, sub-provider iteration as a root server — and `FastMCPProvider.lifespan()` collapses to three lines of delegation.

```python
@asynccontextmanager
async def child_lifespan(_mcp: FastMCP) -> AsyncIterator[dict]:
    yield {"db": db_connection}

child = FastMCP("Child", lifespan=child_lifespan)

@child.tool
def query(ctx: Context) -> str:
    db = ctx.lifespan_context["db"]  # now sees child's lifespan, not parent's
    return db.query("SELECT 1")

parent = FastMCP("Parent")
parent.mount(child, "child")
```

Nested mounts work the same way: the recursive sub-provider iteration in `_lifespan_manager` walks the tree, each level entering its own user lifespan exactly once.

Closes #4049.

### DIFF
--- a/src/fastmcp/server/context.py
+++ b/src/fastmcp/server/context.py
@@ -354,13 +354,15 @@ class Context:
     def lifespan_context(self) -> dict[str, Any]:
         """Access the server's lifespan context.
 
-        Returns the context dict yielded by the server's lifespan function.
-        Returns an empty dict if no lifespan was configured or if the MCP
-        session is not yet established.
+        Returns the context dict yielded by *this* server's lifespan function.
+        For a mounted child this is the child's own lifespan, not the parent's
+        — the MCP session always belongs to the parent, so reading from the
+        request context would return the parent's. We read directly from the
+        server's cached lifespan result instead, which is set by the
+        per-server ``_lifespan_manager`` regardless of mount position
+        (closes #4049).
 
-        In background tasks (Docket workers), where request_context is not
-        available, falls back to reading from the FastMCP server's lifespan
-        result directly.
+        Returns an empty dict if no lifespan was configured.
 
         Example:
         ```python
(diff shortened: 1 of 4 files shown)


jlowin:
Good catch from Codex on the sibling-server case. The flag belonged in `FastMCPProvider.lifespan` (the mount mechanism) rather than `_docket_lifespan` (which fires for any server, including independent siblings). Pushed in 8fa2bde4 along with a regression test that enters two unrelated servers via `AsyncExitStack` and asserts each establishes its own server-context.

jlowin:
Right again — the flag was leaking past the inner `_lifespan_manager` entry. The flag is only needed during `_docket_lifespan.__aenter__` (when the mounted child decides whether to start its own Docket); once entry is done, no one reads it. Resetting the token immediately after the inner `_lifespan_manager` finishes setting up keeps the signal scoped to entry only.

Pushed in fa7ad006 plus a regression test (`test_unrelated_server_after_parent_with_mount`) that mounts a child under a parent, then enters an unrelated server later in the same task and asserts `_current_server` resolves correctly.

PR #4101, opened 2026-05-05
fix(tool_transform): hoist $defs to schema root when ArgTransform introduces them

### LINKED ISSUE #4093: `ArgTransform(type=...)` places `$defs` inside the property node instead of hoisting to schema root
### What happened?

When using `ArgTransform(type=SomePydanticModel)` to replace a parameter's type on an existing tool, the resulting schema has `$defs` nested inside the property node rather than at the schema root, while `$ref` values still use absolute paths (`#/$defs/...`) anchored at the root. This makes the schema invalid — the references dangle and any consumer that resolves `$ref` will raise a `KeyError`.

The root cause is in `_create_forwarding_transform`. `_apply_arg_transform` does:

```python
type_schema = get_cached_typeadapter(transform.type).json_schema()
new_schema.update(type_schema)  # dumps $defs straight into the property dict
```

`TypeAdapter.json_schema()` returns a self-contained schema with `$defs` at its top level. After `update`, those `$defs` are nested inside the property node.

The only place `$defs` get hoisted to root in `_create_forwarding_transform` is:

```python
if parent_defs:
    schema["$defs"] = parent_defs
    schema = compress_schema(schema)
```

But `parent_defs` comes from the *original* tool's schema root. If the original tool declared the parameter as `dict | None`, `parent_defs` is empty, the condition is false, and the `$defs` introduced by `ArgTransform(type=...)` are never hoisted.

**Fix:** strip `$defs` from `type_schema` in `_apply_arg_transform` before merging into the property schema, and collect them separately so `_create_forwarding_transform` can place them at the root.

### Example Code

```Python
import asyncio, json
from fastmcp import FastMCP
from fastmcp.tools.tool_transform import ArgTransform, TransformedTool

class Filter(BaseModel):
    value: str

mcp = FastMCP()

@mcp.tool()
def search(query: str, filters: dict | None = None) -> str:
    """Search tool"""
    return ""

tool = asyncio.run(mcp.get_tool("search"))
transformed = TransformedTool.from_tool(
    tool,
    transform_args={"filters": ArgTransform(type=list[Filter])}
)

schema = transformed.parameters
print(json.dumps(schema, indent=2))
print("$defs at root:", "$defs" in schema)                                          # False
print("$defs inside filters:", "$defs" in schema["properties"]["filters"])          # True
```

**Actual output** — `$defs` stranded inside `properties.filters`, `$ref` dangling:

```
{
  "type": "object",
  "properties": {
    "query": { "type": "string" },
    "filters": {
      "anyOf": [{ "additionalProperties": true, "type": "object" }, { "type": "null" }],
      "default": null,
      "$defs": {
        "Filter": {
          "properties": { "value": { "title": "Value", "type": "string" } },
          "required": ["value"],
          "title": "Filter",
          "type": "object"
        }
      },
      "items": { "$ref": "#/$defs/Filter" },
      "type": "array"
    }
  },
  "required": ["query"],
  "additionalProperties": false
}
```

$defs at root: False
$defs inside filters: True

**Expected output** — `$defs` at schema root, `$ref` resolvable:

```
{
  "type": "object",
  "$defs": {
    "Filter": {
      "properties": { "value": { "title": "Value", "type": "string" } },
      "required": ["value"],
      "title": "Filter",
      "type": "object"
    }
  },
  "properties": {
    "query": { "type": "string" },
    "filters": {
      "items": { "$ref": "#/$defs/Filter" },
      "type": "array"
    }
  },
  "required": ["query"],
  "additionalProperties": false
}
```

### Version Information

```Text
FastMCP version: 3.2.0
MCP version:     1.26.0
Python version:  3.13.11
Platform:        macOS-26.4.1-arm64-arm-64bit-Mach-O
```

SarthakB11:
`TransformedTool._apply_single_transform` calls `get_cached_typeadapter(transform.type).json_schema()` and merges the result into the per-property `new_schema` with `new_schema.update(type_schema)`. `TypeAdapter.json_schema()` returns a self-contained schema with `$defs` at its top level, so the `update` plants those defs inside the property node while the `$ref` values still point at `#/$defs/...` (the schema root). Any consumer that resolves refs raises `KeyError`.

`_create_forwarding_transform` already hoists the parent tool's pre-existing `$defs`, but when the parent declares the parameter as a plain type (e.g. `dict | None`), `parent_defs` is empty and the hoist branch never fires. `_apply_single_transform` now `pop`s `$defs` off the `TypeAdapter` schema before merging into the property schema and returns them alongside; the caller folds them into `parent_defs` so they land at the schema root via the same path that already handles parent defs (including `compress_schema`'s unused-def pruning).

```python
class Filter(BaseModel):
    field: str
    op: str

@base.tool
def search(filters: dict | None = None) -> list[str]: ...

base.add_tool_transform(
    "search",
    arg_transforms={"filters": ArgTransform(type=list[Filter])},
)
# Before: $defs lands inside the property node; refs to #/$defs/Filter resolve to nothing.
# After:  $defs hoists to the schema root; refs resolve cleanly.
```

Closes #4093.

### DIFF
--- a/src/fastmcp/tools/tool_transform.py
+++ b/src/fastmcp/tools/tool_transform.py
@@ -680,11 +680,15 @@ class TransformedTool(Tool):
             )
 
             if transform_result:
-                new_name, new_schema, is_required = transform_result
+                new_name, new_schema, is_required, extracted_defs = transform_result
                 new_props[new_name] = new_schema
                 new_to_old[new_name] = old_name
                 if is_required:
                     new_required.add(new_name)
+                # Hoist any $defs introduced by an ArgTransform(type=...) so that
+                # $ref values like "#/$defs/..." resolve at the schema root.
+                if extracted_defs:
+                    parent_defs.update(extracted_defs)
 
         schema = {
             "type": "object",
@@ -741,7 +745,7 @@ class TransformedTool(Tool):
         old_schema: dict[str, Any],
         transform: ArgTransform,
         is_required: bool,
-    ) -> tuple[str, dict[str, Any], bool] | None:
+    ) -> tuple[str, dict[str, Any], bool, dict[str, Any]] | None:
         """Apply transformation to a single parameter.
 
         This method handles the transformation of a single argument according to
@@ -754,8 +758,10 @@ class TransformedTool(Tool):
             is_required: Whether the original parameter was required.
 
         Returns:
-            Tuple of (new_name, new_schema, new_is_required) if parameter should be kept,
-            None if parameter should be dropped.
+            Tuple of (new_name, new_schema, new_is_required, extracted_defs) if
+            parameter should be kept, None if parameter should be dropped.
+            extracted_defs contains any $defs pulled off a type transform so the
+            caller can hoist them to the schema root.
         """
         if transform.hide:
             return None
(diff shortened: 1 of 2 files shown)


SarthakB11:
Good catch, pushed eaf3a15c. The hoist now raises `ValueError` on key collision when the colliding schemas differ; identical re-introductions are a no-op. Added a regression test (`test_arg_transform_type_raises_on_defs_name_collision`) that plants a different schema under `$defs.Foo` on the parent and confirms the transform raises with the expected message. Full `tests/tools/tool_transform/` (114 tests) passes.

jlowin (approved):
Thanks!

PR #4103, opened 2026-05-06
feat: support custom auth headers in from_openapi() via auth_header_name

### LINKED ISSUE #1574: FastMCP.from_openapi only works if the client sends Authorization header. Does not work with custom headers
### Description

FastMCP.from_openapi only works if the client sends Authorization header. Does not work with custom headers.

I start my server using , where auth_verifier is my custom class that validates the token. It seems that this only works when the client sends Authorization header and does not work if it does not exist. Is there a way I can send other than Authorization header from the MCP Client to the MCP server ? 

    server = FastMCP.from_openapi(
        openapi_spec=openapi_spec,
        client=http_client,
        name=config.get("server", {}).get("name", "fastmcp-server"),
        route_maps=route_maps,
        auth=auth_verifier,
    )

### Example Code

```Python
server = FastMCP.from_openapi(
        openapi_spec=openapi_spec,
        client=http_client,
        name=config.get("server", {}).get("name", "fastmcp-server"),
        route_maps=route_maps,
        auth=auth_verifier,
    )
```

The bearerauthbackend  code uses this in the fastmcp 

```python
auth_header = next(
    (conn.headers.get(key) for key in conn.headers if key.lower() == "authorization"),
    None,
)
if not auth_header or not auth_header.lower().startswith("bearer "):
    return None

token = auth_header[7:]  # Remove "Bearer " prefix
```

### Version Information

```Text
2.11.3
```

tomer-liran:
Adds `CustomHeaderAuthBackend` that reads tokens from a configurable HTTP header (e.g. `X-AUTH-ID`, `X-API-Key`) instead of the hardcoded `Authorization` header.

### Problem

`FastMCP.from_openapi()` only works when the MCP client sends an `Authorization` header. Upstream APIs that use custom credential headers (e.g. `X-AUTH-ID`, `X-ACCOUNT-ID`) cannot be used because `BearerAuthBackend` hardcodes the `authorization` header lookup.

Fixes #1574

### Changes

- Added `CustomHeaderAuthBackend` in `fastmcp/server/auth/auth.py` — an `AuthenticationBackend` that reads tokens from a configurable HTTP header
- Added `auth_header_name` parameter to `AuthProvider.__init__()` (defaults to `"authorization"` — backward-compatible)
- Modified `AuthProvider.get_middleware()` to select the right backend based on `auth_header_name`
- Added `auth_header_name` parameter to `FastMCP.__init__()` and `FastMCP.from_openapi()`

### Usage

```python
server = FastMCP.from_openapi(
    openapi_spec=spec,
    client=client,
    auth=auth_verifier,
    auth_header_name="x-auth-id",
)
```

### Testing

All 2,703 existing tests pass. No test regressions.

### DIFF
--- a/src/fastmcp/server/auth/auth.py
+++ b/src/fastmcp/server/auth/auth.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import json
+import time
 from typing import TYPE_CHECKING, Any
 from urllib.parse import urlparse
 
@@ -8,7 +9,10 @@ from mcp.server.auth.handlers.token import TokenErrorResponse
 from mcp.server.auth.handlers.token import TokenHandler as _SDKTokenHandler
 from mcp.server.auth.json_response import PydanticJSONResponse
 from mcp.server.auth.middleware.auth_context import AuthContextMiddleware
-from mcp.server.auth.middleware.bearer_auth import BearerAuthBackend
+from mcp.server.auth.middleware.bearer_auth import (
+    AuthenticatedUser,
+    BearerAuthBackend,
+)
 from mcp.server.auth.middleware.client_auth import (
     AuthenticationError,
     ClientAuthenticator,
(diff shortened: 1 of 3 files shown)


jlowin:
Thanks for the effort here, but I'm going to close this.

MCP authorization is `Authorization: Bearer <token>` per spec. Adding `auth_header_name` plumbing across `FastMCP`, `from_openapi`, `OAuthProvider`, and `RemoteAuthProvider` lets servers opt into non-conformant behavior, and it's permanent maintenance surface for something the spec doesn't endorse. There's also a design issue with `FastMCP.__init__` mutating `self.auth.auth_header_name` after the provider is constructed, but that's secondary to the spec question.

If the underlying need is forwarding a custom credential header to an upstream OpenAPI API, that's user-code territory: set default headers on the `httpx.AsyncClient` passed to `from_openapi`, or use a Starlette middleware to inject the header into the upstream request. Both are a few lines and don't require framework changes.

PR #4109, opened 2026-05-07
Fix None backend_* span attributes on un-renamed proxy components

### LINKED ISSUE #4108: Proxy*.get_span_attributes() returns None for backend_* keys, causing OTEL log spam
While working on a FastMCP based component recently, I hit an issue where it was logging

```
Invalid type NoneType for attribute 'fastmcp.proxy.backend_name' value. Expected one of ['bool', 'str', 'bytes', 'int', 'float'] or a sequence of those types
```

This was harmless, but noisy in the logs so I used an LLM tool to look into the issue.

The tool identified what looks to me like a genuine, if minor, problem, so I've raised an issue with the details per below.

I have used a tool to [create a **draft PR**](https://github.com/PrefectHQ/fastmcp/pull/4109) that I will review before un-marking draft and requesting review on.

**LLM-generated content follows**

---

When a proxied component is not renamed via `model_copy` (the common case for plain pass-through proxies, and also for transforms that override only `description`/`hidden`), its private `_backend_*` field stays `None`. `get_span_attributes()` returns that `None` for the `fastmcp.proxy.backend_*` keys, and OpenTelemetry rejects each on every proxied call with:

```
Invalid type NoneType for attribute 'fastmcp.proxy.backend_name' value. Expected one of ['bool', 'str', 'bytes', 'int', 'float'] or a sequence of those types
```

### MRE

```python
import asyncio
from fastmcp import FastMCP
from fastmcp.server.providers.proxy import ProxyProvider

backend = FastMCP("backend")

@backend.tool
def hello() -> str:
    return "hi"

async def main():
    proxy = create_proxy(backend)
    provider = next(p for p in proxy.providers if isinstance(p, ProxyProvider))
    tool = (await provider._list_tools())[0]
    print(tool.get_span_attributes())
    # {'fastmcp.provider.type': 'ProxyProvider',
    #  'fastmcp.proxy.backend_name': None,
    #  ...}

asyncio.run(main())
```

### Expected vs actual

- **Expected:** a string — the live `name` / `uri` / `uri_template`.
- **Actual:** `None`, and an OTEL warning logged for every proxied request.

### Notes

Same pattern in all four proxy component classes — `ProxyTool` (line 181), `ProxyResource` (288), `ProxyTemplate` (403), `ProxyPrompt` (486). Their `run()`/`read()` methods at lines 120 / 245 / 348 / 460 already use `self._backend_X or self.X` as a fallback, so the bug is only in `get_span_attributes()`.

🤖 Filed by Claude Code (Opus 4.7) on behalf of @ringerc.

ringerc:
To address https://github.com/PrefectHQ/fastmcp/issues/4108 I've used a tool to identify the cause of spurious

```
Invalid type NoneType for attribute 'fastmcp.proxy.backend_name' value. Expected one of ['bool', 'str', 'bytes', 'int', 'float'] or a sequence of those types
```

log messages and prepare a fix. Reading the proposed fix, it is pretty trivial; the only significant code here is the added test code, and I'm not convinced it adds enough value to be worth merging along with the code fix.

I've monkey-patched this fix into my use of FastMCP and confirmed it solves the problem.

**LLM-generated content follows**

---

`ProxyTool` / `ProxyResource` / `ProxyTemplate` / `ProxyPrompt` only stamp their private `_backend_*` field when `model_copy` runs with a `name` / `uri` / `uri_template` change, so un-renamed proxies leave it `None`. `get_span_attributes()` then emits `None` for the `fastmcp.proxy.backend_*` keys, and OpenTelemetry drops the attribute with a warning on every proxied call (`Invalid type NoneType for attribute 'fastmcp.proxy.backend_name' value...`).

The four `run()` / `read()` methods already fall back to the live identifier (`self._backend_name or self.name`, etc.) when forwarding to the backend. This PR applies the same fallback in `get_span_attributes()`, so the span attribute mirrors what was actually called instead of going silent.

```python
def get_span_attributes(self) -> dict[str, Any]:
    return super().get_span_attributes() | {
        "fastmcp.provider.type": "ProxyProvider",
        "fastmcp.proxy.backend_name": self._backend_name or self.name,
    }
```

Closes #4108.

🤖 Generated with Claude Code (Opus 4.7) on behalf of @ringerc. Marked draft pending @ringerc review.

### DIFF
--- a/src/fastmcp/server/providers/proxy.py
+++ b/src/fastmcp/server/providers/proxy.py
@@ -178,7 +178,7 @@ class ProxyTool(Tool):
     def get_span_attributes(self) -> dict[str, Any]:
         return super().get_span_attributes() | {
             "fastmcp.provider.type": "ProxyProvider",
-            "fastmcp.proxy.backend_name": self._backend_name,
+            "fastmcp.proxy.backend_name": self._backend_name or self.name,
         }
 
 
@@ -285,7 +285,7 @@ class ProxyResource(Resource):
     def get_span_attributes(self) -> dict[str, Any]:
         return super().get_span_attributes() | {
             "fastmcp.provider.type": "ProxyProvider",
-            "fastmcp.proxy.backend_uri": self._backend_uri,
+            "fastmcp.proxy.backend_uri": self._backend_uri or str(self.uri),
         }
 
 
@@ -400,7 +400,9 @@ class ProxyTemplate(ResourceTemplate):
     def get_span_attributes(self) -> dict[str, Any]:
         return super().get_span_attributes() | {
             "fastmcp.provider.type": "ProxyProvider",
-            "fastmcp.proxy.backend_uri_template": self._backend_uri_template,
+            "fastmcp.proxy.backend_uri_template": (
+                self._backend_uri_template or self.uri_template
+            ),
         }
 
 
@@ -483,7 +485,7 @@ class ProxyPrompt(Prompt):
     def get_span_attributes(self) -> dict[str, Any]:
         return super().get_span_attributes() | {
             "fastmcp.provider.type": "ProxyProvider",
-            "fastmcp.proxy.backend_name": self._backend_name,
+            "fastmcp.proxy.backend_name": self._backend_name or self.name,
         }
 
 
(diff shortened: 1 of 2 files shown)


jlowin (approved):

jlowin (approved):
Thanks @ringerc.

jlowin (approved):
Thanks @ringerc.

PR #4110, opened 2026-05-08
Add CORS middleware to /mcp StreamableHTTP route

troyharvey:
## Bug Report

Discovered while testing with the Claude Excel plugin.

image

## Summary

Fixes browser-based CORS preflight requests to the /mcp StreamableHTTP endpoint by:
- Importing cors_middleware from mcp.server.auth.routes
- Adding OPTIONS to http_methods for both auth and no-auth routes
- Wrapping the StreamableHTTP endpoint with cors_middleware

This aligns /mcp CORS handling with OAuth endpoints (/token, /register, etc.) which already use cors_middleware.

## Problem

Browser-based clients (e.g., Claude for Excel) make CORS preflight OPTIONS requests to /mcp and currently receive 405 Method Not Allowed, blocking the connection.

## Solution

The mcp.server.auth.routes module already exports a cors_middleware wrapper used by OAuth endpoints. This PR extends that same middleware to the /mcp route.

## Testing

The fix ensures:
- OPTIONS requests to /mcp return 200 with CORS headers
- Both auth and non-auth routes support CORS preflight
- Backwards compatible (GET/POST/DELETE still work as before)

### DIFF
--- a/src/fastmcp/server/http.py
+++ b/src/fastmcp/server/http.py
@@ -5,7 +5,7 @@ from contextlib import asynccontextmanager, contextmanager
 from contextvars import ContextVar
 from typing import TYPE_CHECKING
 
-from mcp.server.auth.routes import build_resource_metadata_url
+from mcp.server.auth.routes import build_resource_metadata_url, cors_middleware
 from mcp.server.lowlevel.server import LifespanResultT
 from mcp.server.sse import SseServerTransport
 from mcp.server.streamable_http import (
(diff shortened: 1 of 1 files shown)


jlowin:
Thanks for the report. FastMCP already supports this pattern through the existing `middleware=` kwarg on `mcp.run()` / `mcp.http_app()`, documented at https://gofastmcp.com/deployment/http#cors-for-browser-based-clients with the recipe for `Authorization`, `mcp-session-id`, and `expose_headers`.

Adding always-on CORS to the `/mcp` route diverges from upstream MCP SDK (which intentionally doesn't), and `allow_origins=["*"]` on an authenticated data plane is a security smell. Real deployments need to specify which origins they trust, which the existing kwarg already supports.

Closing in favor of the documented pattern. For the Excel plugin specifically, allowing the Office add-in origin via `allow_origins=[...]` should resolve it.

PR #4136, opened 2026-05-14
Cache docstring parsing and type hints

apocalypse9949:
What
Wrap `parse_docstring` and `get_type_hints` with `lru_cache` and replace all calls to use the cached versions during tool parsing. Also wraps them to gracefully fallback when encountering unhashable arguments.

Why
`parse_docstring` parsing via Griffe takes ~100-150 micro-seconds per call, and is called twice per function tool. `get_type_hints` takes ~2 micro-seconds and is called at least 4 times per tool registration path. In large server applications or when registering many schemas, this amounts to a measurable overhead that serves no purpose at runtime since functions and annotations do not change. 
FastMCP intentionally accepts callable instances and calls `parse_docstring` before unwrapping to `fn.__call__` (e.g. in `ParsedFunction.from_function`). Instances with `__hash__ = None` (common for `@dataclass(eq=True)`) raise `TypeError: unhashable type` during registration if the top-level method is wrapped with `lru_cache`. Handling unhashable parameters first before returning the cache prevents these failures.

Impact
Eliminates overhead during startup/registration by memoizing introspection paths.
`parse_docstring` goes from ~112 micro-seconds to ~0.08 micro-seconds.
`get_type_hints` goes from ~1.8 micro-seconds to ~0.09 micro-seconds.

Measurement
Can be measured using cProfile over `ParsedFunction.from_function` loop.

### DIFF
--- a/fastmcp_slim/fastmcp/server/dependencies.py
+++ b/fastmcp_slim/fastmcp/server/dependencies.py
@@ -17,7 +17,7 @@ from contextvars import ContextVar
 from datetime import datetime, timezone
 from functools import lru_cache
 from types import TracebackType
-from typing import TYPE_CHECKING, Any, Protocol, cast, get_type_hints, runtime_checkable
+from typing import TYPE_CHECKING, Any, Protocol, cast, runtime_checkable
 
 from mcp.server.auth.middleware.auth_context import (
     get_access_token as _sdk_get_access_token,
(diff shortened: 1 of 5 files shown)


jlowin:
@apocalypse9949 thanks for looking into registration performance. I’m going to close this because the caching boundary changes supported behavior.

`@lru_cache` hashes `fn` before `parse_docstring()` runs, so unhashable callable instances now fail during registration. FastMCP supports callable instances, and dataclass callables commonly have `__hash__ = None`.

A safe version would need to cache only hashable callables, keep the uncached path for everything else, and avoid the unrelated `loq.toml` churn. Given that this is a startup optimization rather than a correctness fix, I don’t think this PR is the right path forward.

PR #4140, opened 2026-05-15
docs: add Codex skills provider example

MukundaKatta:


strawgate:
From the contributing guide:

> Reference an issue. Every PR should address a tracked issue. If there isn't one, open an issue first. This isn't a permission step — you don't need to wait for a response. But the issue gives us context on the problem, and if a maintainer is already working on it, we can let you know before you invest time in code.

PR #4141, opened 2026-05-15
Support output_schema=False for transformed tools

### LINKED ISSUE #3834: TransformedTool: output_schema=False documented but not implemented
The `TransformedTool.from_tool()` docstring previously claimed `output_schema=False` would "disable output schema and structured outputs," but this was never implemented. Passing `False` would either raise a Pydantic validation error (on the `Tool` model) or get converted to `None` which is indistinguishable from "inherit from parent."

A proper implementation needs a way to distinguish "no schema was specified" (`NotSet` → inherit) from "explicitly disable structured output" (`False` → strip `structured_content` from results). This requires either a sentinel value that survives Pydantic model storage or a separate boolean field.

The false documentation was removed in #3823 to avoid misleading users.

MukundaKatta:
## Summary
- allow `Tool.from_tool(..., output_schema=False)` for transformed tools
- keep `None` semantics as schema-less but still auto-detecting structured content
- add an internal flag that strips structured content when `False` is explicitly requested
- cover dict returns, `ToolResult` returns, `None` vs `False`, and MCP tool schema output

Fixes #3834

## Testing
- `uv run ruff check fastmcp_slim/fastmcp/tools/base.py fastmcp_slim/fastmcp/tools/tool_transform.py tests/tools/tool_transform/test_schemas.py`
- `git diff --check`

Could not run `uv run pytest tests/tools/tool_transform/test_schemas.py` locally because collection fails before tests execute with `ImportError: FastMCP server support is not installed. Install fastmcp or fastmcp-slim[server]`, caused by a circular import while loading `fastmcp.server.providers.local_provider.decorators.tools`. I also tried installing `fastmcp_slim[server]` locally, but collection still failed at the same import path.

### DIFF
--- a/fastmcp_slim/fastmcp/tools/base.py
+++ b/fastmcp_slim/fastmcp/tools/base.py
@@ -7,6 +7,7 @@ from typing import (
     Annotated,
     Any,
     ClassVar,
+    Literal,
     TypeAlias,
     overload,
 )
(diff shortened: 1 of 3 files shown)


strawgate:
From the contributing guide:

PR #4144, opened 2026-05-15
fix(server): ignore missing event during eviction

### LINKED ISSUE #4143: event_store.store_event eviction can crash on concurrent delete (FileNotFoundError)
## Summary

`fastmcp.server.event_store.store_event` evicts an old `event_id` to enforce the ring-buffer cap. When two concurrent calls to `store_event` decide to evict the same `event_id`, the second `delete()` fails — `FileTreeStore.delete_entry` (in `py-key-value-aio`) does an `exists()` check followed by an unguarded `unlink()`, so the second call hits `FileNotFoundError` after the first one wins the race. The exception propagates as an unhandled error out of the SSE writer and message router tasks, producing ERROR-level tracebacks on every collision.

## Code

`fastmcp/server/event_store.py:122-125`:

```python
if len(event_ids) > self._max_events_per_stream:
    for old_id in event_ids[: -self._max_events_per_stream]:
        await self._event_store.delete(key=old_id)   # ← raises on concurrent delete
    event_ids = event_ids[-self._max_events_per_stream :]
```

The race is doubly observable because `streamable_http.py` runs both `sse_writer` (`:592`) and `message_router` (`:1021`) on the same session, both of which call `store_event` and trigger the same eviction.

## Sample traceback

```
ERROR mcp.server.streamable_http: Error in SSE writer
  File ".../mcp/server/streamable_http.py", line 592, in sse_writer
    await self._maybe_send_priming_event(...)
  File ".../mcp/server/streamable_http.py", line 285, in _maybe_send_priming_event
    priming_event_id = await self._event_store.store_event(...)
  File ".../fastmcp/server/event_store.py", line 124, in store_event
    await self._event_store.delete(key=old_id)
  ...
FileNotFoundError: [Errno 2] No such file or directory:
  '/data/state/events/fastmcp_events/4c54f4eb-89b3-4f7a-86be-5b1947645f1f.json'
```

Same shape repeats from `message_router` at `streamable_http.py:1021`.

## Impact

- **Functional**: events that *do* get stored are fine; the failed eviction leaves the ring buffer momentarily over-sized but the next successful eviction resolves it. No session corruption observed.
- **Operational**: every collision produces an ERROR + traceback in session logs. On a busy server this is constant noise that obscures real issues.

## Suggested fix

Two options, not mutually exclusive:

**(a) Belt-and-braces in `store_event`** — wrap the eviction loop's `delete()` call in try/except. The eviction is idempotent by intent; a missing key is success-by-another-name, not an error condition. Drop to DEBUG.

```python
if len(event_ids) > self._max_events_per_stream:
    for old_id in event_ids[: -self._max_events_per_stream]:
        try:
            await self._event_store.delete(key=old_id)
        except FileNotFoundError:
            # Race with another concurrent eviction or external removal —
            # the entry is already gone, which is what we wanted anyway.
            logger.debug("event_store: %s already evicted", old_id)
    event_ids = event_ids[-self._max_events_per_stream :]
```

This is robust against any underlying store that surfaces the same race.

**(b) Upstream fix in `py-key-value-aio`** — `FileTreeStore.delete_entry` should be idempotent (`unlink(missing_ok=True)` semantics). Filed sibling issue at [strawgate/py-key-value](https://github.com/strawgate/py-key-value). Doing (b) alone closes this particular store's race, but (a) defends against any other store that could surface a similar exception (e.g. a future Redis backend in a partition-tolerant deployment).

Recommendation: do both. (a) defends fastmcp; (b) fixes the specific backend.

## Provenance

Encountered while operating `pvliesdonk/markdown-vault-mcp` against a persistent file-tree event store. Tracked downstream as [markdown-vault-mcp#491](https://github.com/pvliesdonk/markdown-vault-mcp/issues/491). Sibling issue at [strawgate/py-key-value](https://github.com/strawgate/py-key-value) for the underlying `FileTreeStore.delete_entry` race.

Versions: fastmcp 3.2.4, py-key-value-aio 0.4.4.

MukundaKatta:
## Summary
- treat missing event entries during stream eviction as already-evicted success
- keep the stream event list trimmed after the race
- add regression coverage for FileNotFoundError during eviction

Fixes #4143.

## Tests
- uv run pytest tests/server/test_event_store.py -q
- uv run ruff check fastmcp_slim/fastmcp/server/event_store.py tests/server/test_event_store.py
- git diff --check

### DIFF
--- a/fastmcp_slim/fastmcp/server/event_store.py
+++ b/fastmcp_slim/fastmcp/server/event_store.py
@@ -121,7 +121,14 @@ class EventStore(SDKEventStore):
         # Trim to max events (delete old events)
         if len(event_ids) > self._max_events_per_stream:
             for old_id in event_ids[: -self._max_events_per_stream]:
-                await self._event_store.delete(key=old_id)
+                try:
+                    await self._event_store.delete(key=old_id)
+                except FileNotFoundError:
+                    logger.debug(
+                        "Event %s was already evicted from stream %s",
+                        old_id,
+                        stream_id,
+                    )
             event_ids = event_ids[-self._max_events_per_stream :]
 
         await self._stream_store.put(
--- a/tests/server/test_event_store.py
+++ b/tests/server/test_event_store.py
@@ -140,6 +140,23 @@ class TestEventStore:
         result = await event_store.replay_events_after(event_ids[3], callback)
         assert result == "stream-1"
 
+    async def test_missing_evicted_event_is_ignored(self, sample_message):
+        """Concurrent eviction can delete the same old event first."""
+        event_store = EventStore(max_events_per_stream=1, ttl=3600)
+        first_event_id = await event_store.store_event("stream-1", sample_message)
+
+        async def delete_raises_for_first_event(key: str):
+            if key == first_event_id:
+                raise FileNotFoundError("already deleted")
+
+        event_store._event_store.delete = delete_raises_for_first_event
+
+        second_event_id = await event_store.store_event("stream-1", sample_message)
+
+        stream_data = await event_store._stream_store.get(key="stream-1")
+        assert stream_data is not None
+        assert stream_data.event_ids == [second_event_id]
+
     async def test_multiple_streams_are_isolated(self, event_store):
         """Events from different streams should not interfere with each other."""
         msg1 = JSONRPCMessage(


strawgate:
From the contributing guide:

PR #4145, opened 2026-05-15
fix(client): forward cancelled tool calls upstream

### LINKED ISSUE #4124: `ProxyProvider` does not forward request cancellation to upstream
### What happened?

When a client sends a cancellation request to a proxy server, I would expect the upstream server to receive that cancellation. However, I've noticed that the upstream tool does not receive the cancellation and just keeps on running.

### Example Code

```Python
import asyncio

from fastmcp import FastMCP
from fastmcp.server.providers.proxy import ProxyClient


async def main():
    cancellation_received = asyncio.Event()

    # Upstream server with a slow tool
    upstream = FastMCP("Upstream")

    @upstream.tool
    async def slow_task() -> str:
        try:
            for i in range(30):
                print(f"  [upstream] working... {i + 1}/30")
                await asyncio.sleep(0.2)
            return "completed"
            print("  [upstream] got CancelledError!")
            cancellation_received.set()
            raise

    # Proxy server in front of the upstream
    proxy = create_proxy(ProxyClient(upstream))

    async with Client(proxy) as client:
        task = asyncio.create_task(client.call_tool("slow_task", {}))
        await asyncio.sleep(1.0)

        print("[client] cancelling call_tool task...")
        task.cancel()
        try:
            await task
            print("[client] task cancelled on client side")

        # Client is still connected — check if the upstream stops promptly
        print("[client] waiting 3s to see if upstream stops...")
        await asyncio.sleep(3.0)

        if cancellation_received.is_set():
            print("\n OK: Upstream received cancellation promptly")
        else:
            print("\n BUG: Upstream is still running 3s after client cancelled the call")
            print("      (cancellation was not forwarded through the proxy)")

    print("\n[done] Client session closed.")


asyncio.run(main())
```

### Version Information

```Text
FastMCP version:                                                          3.2.4
MCP version:                                                             1.27.1
Python version:                                                          3.14.3
Platform:                                   macOS-15.7.4-arm64-arm-64bit-Mach-O
FastMCP root path: /Users/amenck/repos/repro/.venv/lib/python3.14/site-packages
```

MukundaKatta:
## Summary
- send MCP cancellation notifications when a client-side tool call await is cancelled
- preserve cancellation delivery under a shield so proxy cleanup does not swallow the upstream notification
- add a stateful proxy regression test that verifies the backend tool receives cancellation

Fixes #4124.

## Tests
- uv run pytest tests/server/providers/proxy/test_stateful_proxy_client.py -k cancellation -q
- uv run pytest tests/server/providers/proxy/test_stateful_proxy_client.py -q
- uv run ruff check fastmcp_slim/fastmcp/client/mixins/tools.py tests/server/providers/proxy/test_stateful_proxy_client.py
- git diff --check

### DIFF
--- a/fastmcp_slim/fastmcp/client/mixins/tools.py
+++ b/fastmcp_slim/fastmcp/client/mixins/tools.py
@@ -2,10 +2,13 @@
 
 from __future__ import annotations
 
+import asyncio
 import uuid
 import weakref
+from contextlib import suppress
 from typing import TYPE_CHECKING, Any, Literal, overload
 
+import anyio
 import mcp.types
 from opentelemetry.trace import Status, StatusCode
 from pydantic import RootModel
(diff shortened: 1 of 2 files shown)


strawgate:
From the contributing guide:

PR #4146, opened 2026-05-15
fix(client): propagate tracing for list requests

### LINKED ISSUE #3998: OpenTelemetry: FastMCP client does not propagate trace context on initialize and list methods

MukundaKatta:
## Summary
- inject OpenTelemetry trace context into MCP _meta for tools/list, resources/list, resources/templates/list, and prompts/list
- keep the existing SDK list helpers when no trace context is active
- add regression assertions that server list spans are children of the matching client spans

Fixes #3998.

Note: this intentionally does not reimplement initialize tracing; the issue discussion calls that out as broader SDK/API work.

## Tests
- uv run pytest tests/client/telemetry/test_client_list_tracing.py -q
- uv run pytest tests/client/telemetry/test_client_tracing.py tests/client/telemetry/test_client_list_tracing.py -q
- uv run ruff check fastmcp_slim/fastmcp/client/mixins/tools.py fastmcp_slim/fastmcp/client/mixins/resources.py fastmcp_slim/fastmcp/client/mixins/prompts.py tests/client/telemetry/test_client_list_tracing.py
- git diff --check

### DIFF
--- a/fastmcp_slim/fastmcp/client/mixins/prompts.py
+++ b/fastmcp_slim/fastmcp/client/mixins/prompts.py
@@ -57,9 +57,24 @@ class ClientPromptsMixin:
         ):
             logger.debug(f"[{self.name}] called list_prompts")
 
-            result = await self._await_with_session_monitoring(
-                self.session.list_prompts(cursor=cursor)
-            )
+            propagated_meta = inject_trace_context()
+            if propagated_meta:
+                request = mcp.types.ListPromptsRequest(
+                    params=mcp.types.PaginatedRequestParams(
+                        cursor=cursor,
+                        _meta=propagated_meta,  # type: ignore[unknown-argument]  # pydantic alias  # ty:ignore[unknown-argument]
+                    )
+                )
+                result = await self._await_with_session_monitoring(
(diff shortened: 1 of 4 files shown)


strawgate:
From the contributing guide:

PR #4149, opened 2026-05-15
fix(server): avoid eager imports from server package

### LINKED ISSUE #4147: Regression: circular import breaks all fresh 3.3.0 installs
### What happened?

A fresh `pip install fastmcp==3.3.0` in a clean venv raises ImportError on
any import that touches `fastmcp.tools`. Same venv pinned to 3.2.4: no error.
Reproduces on Python 3.11 and 3.13, Linux.

### Example Code

```Python
python3.13 -m venv /tmp/v && /tmp/v/bin/pip install fastmcp==3.3.0
/tmp/v/bin/python -c "from fastmcp.tools import tool"



ImportError: cannot import name 'FunctionTool' from partially initialized
module 'fastmcp.tools.function_tool' (most likely due to a circular import)

The above exception was the direct cause of the following exception:

ImportError: FastMCP server support is not installed.
             Install `fastmcp` or `fastmcp-slim[server]`.
```

### Version Information

```Text
FastMCP version:                                             3.3.0
MCP version:                                                1.27.1
Python version:                                            3.13.12
Platform:                       Linux-6.18.5-x86_64-with-glibc2.39
FastMCP root path: /tmp/fastmcp_repro/lib/python3.13/site-packages
```

MukundaKatta:
## Summary
- make `fastmcp.server` exports lazy so importing nested server modules does not eagerly load the full server stack
- fix the circular import triggered by `from fastmcp.tools import tool`
- add a regression test for importing `fastmcp.tools`

Fixes #4147

## Tests
- uv run python -c "from fastmcp.tools import tool, FunctionTool; print(tool.__name__, FunctionTool.__name__)"
- uv run python -c "from fastmcp.server import FastMCP, Context, create_proxy; print(FastMCP.__name__, Context.__name__, callable(create_proxy))"
- uv run pytest tests/tools/test_imports.py tests/deprecated/test_import_server.py tests/tools/test_standalone_decorator.py -q
- uv run pytest tests/tools tests/server/test_app_state.py tests/server/http/test_http_middleware.py tests/server/auth/test_static_token_verifier.py -q
- uv run ruff check fastmcp_slim/fastmcp/server/__init__.py tests/tools/test_imports.py
- git diff --check

### DIFF
--- a/fastmcp_slim/fastmcp/server/__init__.py
+++ b/fastmcp_slim/fastmcp/server/__init__.py
@@ -2,14 +2,20 @@ import importlib
 
 from fastmcp import _install_hints
 
-try:
-    from .context import Context
-    from .server import FastMCP, create_proxy
-except ImportError as exc:
-    raise ImportError(_install_hints.SERVER_SUPPORT) from exc
-
 
 def __getattr__(name: str) -> object:
+    if name == "Context":
+        try:
+            from .context import Context
+        except ImportError as exc:
+            raise ImportError(_install_hints.SERVER_SUPPORT) from exc
+        return Context
+    if name in {"FastMCP", "create_proxy"}:
+        try:
+            from .server import FastMCP, create_proxy
+        except ImportError as exc:
+            raise ImportError(_install_hints.SERVER_SUPPORT) from exc
+        return {"FastMCP": FastMCP, "create_proxy": create_proxy}[name]
     if name == "dependencies":
         return importlib.import_module("fastmcp.server.dependencies")
     raise AttributeError(f"module {__name__!r} has no attribute {name!r}")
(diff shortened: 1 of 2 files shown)


jlowin:
Thanks for moving quickly on this. I’m closing this in favor of #4150.

The immediate lazy-export change avoids the failing import order, but it treats the symptom rather than the cause. The underlying problem is that component modules were importing through the server package for auth/task primitives, and the app/provider import graph could then loop back into tools. A fix at the `fastmcp.server.__init__` layer leaves that coupling in place and makes the framework harder to reason about.

Please also keep the contributing guidance in mind: an open issue is not an invitation to submit a PR. For anything that touches framework structure or multiple subsystems, propose the approach in the issue first. PRs that patch the observed failure without addressing the layer that caused it create review noise and often add more maintenance burden than they remove.

PR #4151, opened 2026-05-15
fix(server): preserve middleware protocol errors

### LINKED ISSUE #4126: ErrorHandlingMiddleware does not transform unhandled exceptions
### What happened?

The [MCP spec](https://modelcontextprotocol.io/specification/2025-11-25/server/tools#error-handling) says:

> Tools use two error reporting mechanisms:
> 1. **Protocol Errors**: Standard JSON-RPC errors for issues like:
> - Unknown tools
> - Malformed requests (requests that fail to satisfy [CallToolRequest schema](https://modelcontextprotocol.io/specification/2025-11-25/schema#calltoolrequest))
> - Server errors
> 2. **Tool Execution Errors**: Reported in tool results with `isError: true`:
> - API failures
> - Input validation errors (e.g., date in wrong format, value out of range)
> - Business logic errors

so I would expect FastMCP to treat unhandled errors in tools as generic "server errors" and raise them as **Protocol Errors**. Instead, unhandled errors get raised a ToolErrors which I believe are **Tool Execution Errors**.

Based on the code [here](https://github.com/PrefectHQ/fastmcp/blob/8209093871af25bc3ceb50bfbcec317632218afd/src/fastmcp/server/middleware/error_handling.py#L94-L118) it seems like it's intended to have
```
mcp.add_middleware(ErrorHandlingMiddleware(transform_errors=True))
```
in order to raise unhandled errors as `McpError(ErrorData(code=-32603, message=f"Internal error: {error!s}"))` which I believe correspond to **Protocol Errors**. 

I therefore expect the test in the Example Code to pass but instead it fails because a `ToolError` is raised. I assume because the error has already been converted to a `ToolError` before the `ErrorHandlingMiddleware` is run.

### Example Code

```Python
import pytest
from fastmcp import FastMCP
from mcp.shared.exceptions import McpError


@pytest.fixture
async def mcp_with_failing_tool():
    """Create a test MCP server with a tool that raises an unhandled exception."""
    mcp = FastMCP("test-exception-handling")
    mcp.add_middleware(ErrorHandlingMiddleware(transform_errors=True))

    @mcp.tool()
    def tool_that_raises_runtime_error() -> str:
        """A tool that raises an unhandled RuntimeError."""
        raise RuntimeError("This is an unhandled exception")

    yield mcp


@pytest.mark.asyncio
async def test_unhandled_exception_as_jsonrpc_error_with_middleware(
    mcp_with_failing_tool,
):
    """Test middleware route: unhandled exceptions become JSON-RPC errors."""
    async with Client(transport=mcp_with_failing_tool) as client:
        with pytest.raises(McpError) as exc_info:
            await client.call_tool("tool_that_raises_runtime_error", {})

        error = exc_info.value.error
        assert error.code == -32603
        assert "Internal error" in error.message
```

### Version Information

```Text
FastMCP version:                                                                              3.2.4
MCP version:                                                                                 1.27.0
Python version:                                                                              3.12.9
Platform:                                                              macOS-26.4.1-arm64-arm-64bit
```

MukundaKatta:
## Summary
- register FastMCP's call-tool handler directly so middleware-raised McpError values remain JSON-RPC protocol errors
- preserve the SDK call-tool normalization behavior for input validation, task results, tool results, structured output, and output schema validation
- add an integration regression test for ErrorHandlingMiddleware(transform_errors=True)

Fixes #4126

## Tests
- uv run ruff check src/fastmcp/server/mixins/mcp_operations.py tests/server/middleware/test_error_handling.py
- uv run pytest tests/server/middleware/test_error_handling.py::TestErrorHandlingMiddlewareIntegration::test_error_handling_middleware_transform_errors_to_protocol_error -q
- uv run pytest tests/server/middleware/test_error_handling.py tests/server/middleware/test_middleware.py::TestApplyMiddlewareParameter tests/server/test_input_validation.py -q
- git diff --check

### DIFF
--- a/fastmcp_slim/fastmcp/server/mixins/mcp_operations.py
+++ b/fastmcp_slim/fastmcp/server/mixins/mcp_operations.py
@@ -2,11 +2,13 @@
 
 from __future__ import annotations
 
+import json
 from collections.abc import Awaitable, Callable, Sequence
 from typing import TYPE_CHECKING, Any, TypeVar, cast
 
+import jsonschema
 import mcp.types
-from mcp.shared.exceptions import McpError
+from mcp.shared.exceptions import McpError, UrlElicitationRequiredError
 from mcp.types import ContentBlock
 from pydantic import AnyUrl
 
@@ -60,7 +62,8 @@ class MCPOperationsMixin:
         Exception: list_resource_templates SDK decorator doesn't pass the request,
         so we register that handler directly.
 
-        The call_tool decorator is from the SDK (supports CreateTaskResult + validate_input).
+        The call_tool handler is registered directly so middleware can surface
+        McpError as protocol errors instead of tool error results.
         The read_resource and get_prompt decorators are from LowLevelServer to add
         CreateTaskResult support until the SDK provides it natively.
         """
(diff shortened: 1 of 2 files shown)


strawgate:
From the contributing guide:

PR #4153, opened 2026-05-15
fix(server): make fastmcp.server top-level imports lazy (#4147)

### LINKED ISSUE #4147: Regression: circular import breaks all fresh 3.3.0 installs

manpreetsingh78:
Fixes #4147

## Problem

A fresh `pip install fastmcp==3.3.0` raises `ImportError` on any import that touches `fastmcp.tools`:

```
$ python -c "from fastmcp.tools import tool"
ImportError: cannot import name 'FunctionTool' from partially initialized module
'fastmcp.tools.function_tool' (most likely due to a circular import)
ImportError: FastMCP server support is not installed.
Install `fastmcp` or `fastmcp-slim[server]`.
```

## Root cause

`fastmcp/server/__init__.py` eagerly imports `Context` and `FastMCP` at package init time. When unrelated code does `from fastmcp.server.auth.authorization import AuthCheck` (or any other `fastmcp.server.*` submodule), Python first runs the `__init__.py`, which transitively pulls in:

```
fastmcp.server.__init__
  -> fastmcp.server.context
    -> fastmcp.server.low_level
      -> fastmcp.apps.config
        -> fastmcp.apps.app
          -> fastmcp.server.providers.filesystem
            -> fastmcp.server.providers.local_provider
              -> ...decorators.tools
                -> fastmcp.tools.function_tool   # ← still being initialized
```

The re-entry into the partially-initialized `fastmcp.tools.function_tool` raises `ImportError`, which the eager `try/except` in `server/__init__.py` then re-wraps as a misleading "server support not installed" message.

## Fix

Mirror the lazy `__getattr__` pattern that `fastmcp/__init__.py` itself already uses (see #3292). `Context`, `FastMCP`, and `create_proxy` are now resolved on attribute access instead of at package init time, so submodule imports like `fastmcp.server.tasks.config` no longer trigger the heavy import chain.

`__all__`, the `dependencies` shortcut, and the `ImportError(_install_hints.SERVER_SUPPORT)` wrapping are all preserved. The `TYPE_CHECKING` block keeps the names visible to static analyzers.

## Tests

Added `tests/tools/test_import_no_circular.py` with three subprocess-based regression tests (subprocess is required because `sys.modules` caching hides the regression after first successful import in the same interpreter):

- `test_import_fastmcp_tools_does_not_circular` — `from fastmcp.tools import tool`
- `test_import_function_tool_does_not_circular` — `from fastmcp.tools.function_tool import FunctionTool`
- `test_import_server_submodule_does_not_circular` — `from fastmcp.server.tasks.config import TaskConfig`

All three fail on `main` and pass with this fix.

## Verification

- Reproduced the bug in a fresh venv with `fastmcp==3.3.0` ✅
- Verified the fix resolves `from fastmcp.tools import tool` ✅
- All existing `tests/tools/` tests pass (314 passed, 1 skipped) ✅
- All public exports (`Context`, `FastMCP`, `create_proxy`) work via direct attribute access, `from` imports, and one-statement imports ✅
- All submodule imports (`tasks.config`, `auth.authorization`, `dependencies`, `context`, `server`) work ✅
- End-to-end functional test (FastMCP server + `@server.tool` + tool invocation) passes ✅
- `ty check` passes on the modified file ✅

### DIFF
--- a/fastmcp_slim/fastmcp/server/__init__.py
+++ b/fastmcp_slim/fastmcp/server/__init__.py
@@ -1,15 +1,39 @@
 import importlib
+from typing import TYPE_CHECKING
 
 from fastmcp import _install_hints
 
-try:
-    from .context import Context
-    from .server import FastMCP, create_proxy
-except ImportError as exc:
-    raise ImportError(_install_hints.SERVER_SUPPORT) from exc
+if TYPE_CHECKING:
+    from .context import Context as Context
+    from .server import FastMCP as FastMCP
+    from .server import create_proxy as create_proxy
 
 
+# --- Lazy imports to avoid circular imports (see #4147) ---
+# Eagerly importing Context/FastMCP at package init time pulls in the full
+# server import chain, which can re-enter partially-initialized modules
+# (e.g. fastmcp.tools.function_tool) and raise a misleading ImportError.
+# Defer these to __getattr__ so submodule imports like
+# `fastmcp.server.tasks.config` don't trigger the heavy import chain.
 def __getattr__(name: str) -> object:
+    if name == "Context":
+        try:
+            from .context import Context
+        except ImportError as exc:
+            raise ImportError(_install_hints.SERVER_SUPPORT) from exc
+        return Context
+    if name == "FastMCP":
+        try:
+            from .server import FastMCP
+        except ImportError as exc:
+            raise ImportError(_install_hints.SERVER_SUPPORT) from exc
+        return FastMCP
+    if name == "create_proxy":
+        try:
+            from .server import create_proxy
+        except ImportError as exc:
+            raise ImportError(_install_hints.SERVER_SUPPORT) from exc
+        return create_proxy
     if name == "dependencies":
         return importlib.import_module("fastmcp.server.dependencies")
     raise AttributeError(f"module {__name__!r} has no attribute {name!r}")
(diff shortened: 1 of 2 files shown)


jlowin:
Closing this in favor of #4150.

This takes the same lazy-export approach as #4149. It avoids the immediate failing import order, but it fixes the symptom rather than the cause. The underlying issue is that component modules were importing auth/task primitives through the server package, and the app/provider graph could then loop back into tools. The maintainer fix moves those primitives out of the server import path and keeps the existing server-facing imports as compatibility shims.

Please keep the contributing guidance in mind: an open issue is not an invitation to submit a PR. For framework-level changes, especially anything involving import structure or multiple subsystems, propose the approach in the issue first. PRs that patch the observed traceback without addressing the layer that caused it create review noise and maintenance burden.

PR #4156, opened 2026-05-15
fix: namespace resource links in tool results

### LINKED ISSUE #4154: Namespace transform does not rewrite ResourceLink URIs returned from proxied tools
## summary

With a `Namespace` transform on a proxy, listed resources are namespaced
correctly.
However, `ResourceLink.uri` values returned from a proxied
`tools/call` result keep the upstream URI instead of the proxied URI.

Tested on FastMCP `3.3.0`.

## minimal example

```python
from fastmcp.server.transforms import Namespace

proxy = create_proxy(upstream_client, name="Proxy-everything")
proxy.add_transform(Namespace("everything"))
```

An upstream resource looking like this:

```text
demo://resource/dynamic/text/2
```

will be exposed correctly as:

```text
demo://everything/resource/dynamic/text/2
```

But if an upstream tool returns:

```python
mcp.types.ResourceLink(
    type="resource_link",
    name="dynamic-text",
    uri="demo://resource/dynamic/text/2",
)
```

the proxied tool result still contains:

```text
demo://resource/dynamic/text/2
```

## Observed Behavior

```text
resources/list:
  upstream: demo://resource/dynamic/text/2
  proxied:  demo://everything/resource/dynamic/text/2

resources/read:
  demo://everything/resource/dynamic/text/2
  -> works

tools/call result content:
  ResourceLink.uri == demo://resource/dynamic/text/2
  -> raw upstream URI leaks through unchanged
```

## expected behavior

`ResourceLink` content returned by proxied tools should expose the same proxied
URI as `resources/list` and `resources/read`.

## scope of this issue

This is specifically about `mcp.types.ResourceLink` content inside `tools/call`
results. Normal `resources/list`, `resource_templates/list`, and
`resources/read` already behave correctly through the proxy.

Session-scoped resources returned by some tools may have additional lifecycle
constraints beyond URI rewriting. This issue is about ordinary ResourceLink URI
projection only.

### Example Code

```Python

```

### Version Information

```Text
FastMCP version:                                                                              3.3.0
MCP version:                                                                                 1.27.0
Python version:                                                                              3.14.3
Platform: Windows-11-10.0.26200-SP0
FastMCP root path: its a uv venv trust
```

he-yufeng:
## Summary
- apply Namespace URI projection to ResourceLink content returned from namespaced tools
- keep existing resource list/read behavior unchanged
- add a regression test covering a namespaced provider returning a ResourceLink from tools/call

Fixes #4154.

## To verify
- PYTHONUTF8=1 uv run pytest tests/server/providers/test_transforming_provider.py tests/server/transforms/test_resources_as_tools.py tests/server/transforms/test_prompts_as_tools.py -q --basetemp .tmp\pytest
- PYTHONUTF8=1 uv run ruff check fastmcp_slim/fastmcp/server/transforms/namespace.py tests/server/providers/test_transforming_provider.py
- PYTHONUTF8=1 uv run ty check fastmcp_slim/fastmcp/server/transforms/namespace.py tests/server/providers/test_transforming_provider.py
- PYTHONUTF8=1 uv run python -m py_compile fastmcp_slim/fastmcp/server/transforms/namespace.py tests/server/providers/test_transforming_provider.py

### DIFF
--- a/fastmcp_slim/fastmcp/server/transforms/namespace.py
+++ b/fastmcp_slim/fastmcp/server/transforms/namespace.py
@@ -4,7 +4,9 @@ from __future__ import annotations
 
 import re
 from collections.abc import Sequence
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Any, overload
+
+import mcp.types
 
 from fastmcp.server.transforms import (
     GetPromptNext,
@@ -13,16 +15,99 @@ from fastmcp.server.transforms import (
     GetToolNext,
     Transform,
 )
+from fastmcp.tools.base import Tool, ToolResult
 from fastmcp.utilities.versions import VersionSpec
 
 if TYPE_CHECKING:
     from fastmcp.prompts.base import Prompt
     from fastmcp.resources.base import Resource
     from fastmcp.resources.template import ResourceTemplate
-    from fastmcp.tools.base import Tool
+    from fastmcp.server.tasks.config import TaskMeta
 
 # Pattern for matching URIs: protocol://path
 _URI_PATTERN = re.compile(r"^([^:]+://)(.*?)$")
+
+
+class _NamespacedResultTool(Tool):
+    """Tool wrapper that keeps the wrapped tool's execution path intact."""
+
+    _tool: Tool
+    _namespace: Any
+
+    def __init__(self, tool: Tool, namespace: Namespace, name: str) -> None:
+        super().__init__(
+            name=name,
+            version=tool.version,
+            title=tool.title,
+            description=tool.description,
+            icons=tool.icons,
+            tags=tool.tags,
+            meta=tool.meta,
+            task_config=tool.task_config,
+            parameters=tool.parameters,
+            output_schema=tool.output_schema,
+            annotations=tool.annotations,
+            execution=tool.execution,
+            serializer=tool.serializer,
+            auth=tool.auth,
+            timeout=tool.timeout,
+        )
+        self._tool = tool
+        self._namespace = namespace
+
+    @overload
+    async def _run(
+        self,
+        arguments: dict[str, Any],
+        task_meta: None = None,
+    ) -> ToolResult: ...
+
+    @overload
+    async def _run(
+        self,
+        arguments: dict[str, Any],
+        task_meta: TaskMeta,
+    ) -> mcp.types.CreateTaskResult: ...
+
+    async def _run(
+        self,
+        arguments: dict[str, Any],
+        task_meta: TaskMeta | None = None,
+    ) -> ToolResult | mcp.types.CreateTaskResult:
+        result = await self._tool._run(arguments, task_meta=task_meta)
+        return self._transform_result(result)
+
+    async def run(self, arguments: dict[str, Any]) -> ToolResult:
+        return self._namespace._transform_tool_result(await self._tool.run(arguments))
+
+    def register_with_docket(self, docket: Any) -> None:
+        if not self.task_config.supports_tasks():
+            return
+        self._register_with_docket_as(docket, self.key)
+
+    def _register_with_docket_as(self, docket: Any, key: str) -> None:
+        fn = getattr(self._tool, "fn", None)
+        if fn is not None:
+            docket.register(fn, names=[key])
+            return
+
+        if isinstance(self._tool, _NamespacedResultTool):
+            self._tool._register_with_docket_as(docket, key)
+            return
+
+        docket.register(self.run, names=[key])
+
+    def get_span_attributes(self) -> dict[str, Any]:
+        return self._tool.get_span_attributes() | {
+            "fastmcp.component.key": self.key,
+        }
+
+    def _transform_result(
+        self, result: ToolResult | mcp.types.CreateTaskResult
+    ) -> ToolResult | mcp.types.CreateTaskResult:
+        if isinstance(result, mcp.types.CreateTaskResult):
+            return result
+        return self._namespace._transform_tool_result(result)
 
 
 class Namespace(Transform):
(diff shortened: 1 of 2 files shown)


jlowin (changes requested):
Thanks for the focused regression test here. I think this needs one small correction before merge: the ResourceLink copy currently bypasses Pydantic validation and stores the rewritten URI as a plain string.

In `fastmcp_slim/fastmcp/server/transforms/namespace.py`, `_transform_content_block()` does:

```python
return block.model_copy(update={"uri": self._transform_uri(str(block.uri))})
```

`model_copy(update=...)` does not validate the updated value, so `ResourceLink.uri` becomes `str` instead of `AnyUrl`. Pydantic then emits serializer warnings when the block is dumped. Please wrap the transformed value with `mcp.types.AnyUrl(...)` or otherwise rebuild the `ResourceLink` through validation so the field keeps the expected type.

jlowin:
It looks like this introduced some significant test regressions as well. Please make sure to run tests and check all behavior before opening a PR

he-yufeng:
Updated the PR to address the review/CI regression.

What changed:
- Stopped using `TransformedTool + forward_raw` for namespaced tools, because that bypassed the mounted child server dispatch path.
- Added a small wrapper that preserves the wrapped tool's `_run` / `run` path and only rewrites `ResourceLink` result URIs after execution.
- Wrapped transformed `ResourceLink.uri` with `AnyUrl` so the field keeps the MCP/Pydantic URI type.
- Preserved nested task registration by registering the innermost function under the outer namespaced key.

Validation run locally on Windows:
- `uv run --frozen pytest tests/server/providers/test_transforming_provider.py tests/server/tasks/test_task_mount.py tests/server/telemetry/test_delegate_method.py tests/server/telemetry/test_provider_tracing.py -q --basetemp .tmp/pytest` -> 67 passed
- `uv run --frozen ruff check fastmcp_slim/fastmcp/server/transforms/namespace.py tests/server/providers/test_transforming_provider.py`
- `uv run --frozen ty check fastmcp_slim/fastmcp/server/transforms/namespace.py`
- `uv run --frozen python -m py_compile fastmcp_slim/fastmcp/server/transforms/namespace.py tests/server/providers/test_transforming_provider.py`
- `git diff --check`

Note: on this Windows environment, pytest needs `PYTHONUTF8=1` and a repo-local `--basetemp` because the default user temp pytest directory returns `WinError 5`.

he-yufeng:
Rebased onto current `main` and re-ran the focused regression set after the URI-type fix.

Local validation on Windows:
- `uv run --frozen pytest tests/server/providers/test_transforming_provider.py tests/server/tasks/test_task_mount.py tests/server/telemetry/test_delegate_method.py tests/server/telemetry/test_provider_tracing.py -q --basetemp .tmp/pytest-4156` -> 67 passed
- `uv run --frozen ruff check fastmcp_slim/fastmcp/server/transforms/namespace.py tests/server/providers/test_transforming_provider.py` -> passed
- `uv run --frozen ty check fastmcp_slim/fastmcp/server/transforms/namespace.py` -> passed
- `uv run --frozen python -m py_compile fastmcp_slim/fastmcp/server/transforms/namespace.py tests/server/providers/test_transforming_provider.py` -> passed
- `git diff --check` -> passed

he-yufeng:
﻿Rebased onto current `main` again and re-ran the focused regression set.

Local validation on Windows:
- `uv run --frozen pytest tests/server/providers/test_transforming_provider.py tests/server/tasks/test_task_mount.py tests/server/telemetry/test_delegate_method.py tests/server/telemetry/test_provider_tracing.py -q --basetemp .tmp/pytest-4156` -> 67 passed
- `uv run --frozen ruff check fastmcp_slim/fastmcp/server/transforms/namespace.py tests/server/providers/test_transforming_provider.py` -> passed
- `uv run --frozen ty check fastmcp_slim/fastmcp/server/transforms/namespace.py` -> passed
- `uv run --frozen python -m py_compile fastmcp_slim/fastmcp/server/transforms/namespace.py tests/server/providers/test_transforming_provider.py` -> passed
- `git diff --check` -> passed

The latest branch keeps the `ResourceLink.uri` value wrapped as `mcp.types.AnyUrl(...)` and preserves the mounted child server execution path.

he-yufeng:
﻿Rebased onto current `main` again and re-ran the regression set that previously caught the mounted-server dispatch regressions.

Local validation on Windows:
- `$env:PYTHONUTF8='1'; .\.venv\Scripts\python.exe -m pytest tests\server\providers\test_transforming_provider.py tests\server\tasks\test_task_mount.py tests\server\telemetry\test_delegate_method.py tests\server\telemetry\test_provider_tracing.py -q --basetemp .tmp\pytest-4156-refresh` -> 67 passed
- `.\.venv\Scripts\python.exe -m ruff check fastmcp_slim\fastmcp\server\transforms\namespace.py tests\server\providers\test_transforming_provider.py` -> passed
- `.\.venv\Scripts\python.exe -m ruff format --check fastmcp_slim\fastmcp\server\transforms\namespace.py tests\server\providers\test_transforming_provider.py` -> passed
- `.\.venv\Scripts\python.exe -m ty check fastmcp_slim\fastmcp\server\transforms\namespace.py` -> passed
- `.\.venv\Scripts\python.exe -m py_compile fastmcp_slim\fastmcp\server\transforms\namespace.py tests\server\providers\test_transforming_provider.py` -> passed
- `git diff --check origin/main...HEAD` -> passed

jlowin:
Thanks for chasing this and for iterating after the CI feedback.

The behavior is right, but the implementation is solving too much of the framework inside `Namespace`. A namespaced tool that rewrites `ResourceLink` result URIs is a reasonable model; a bespoke `_NamespacedResultTool` inside `namespace.py` that has to preserve `_run`, `run`, task dispatch, docket registration, span attributes, and nested wrapper behavior is more coupling than I want in this transform.

The amount of code here is out of proportion to the use case because it is filling an abstraction gap. This should either be a reusable result-transforming tool abstraction, or an extension of `TransformedTool`, so `Namespace` only provides the small URI projection function and the shared tool wrapper owns the execution invariants.

The regression test is useful and should carry forward. I am closing this PR so the fix can be shaped around that abstraction instead of continuing with a one-off wrapper.

PR #4172, opened 2026-05-17
fix(server): wrap pydantic ValidationError from tool arg validation as fastmcp ValidationError

### LINKED ISSUE #4128: pydantic.ValidationError from argument validation should be wrapped as fastmcp's `ValidationError` before reaching middleware
### Enhancement

When a caller passes invalid arguments (e.g. a value that fails a `Field(le=10)` constraint), `pydantic-core`'s call validator raises a raw `pydantic.ValidationError`. In `server.py`, this is caught by the `except (ValidationError, PydanticValidationError)` clause and re-raised without conversion.

This means a raw `pydantic.ValidationError` — not a `fastmcp.exceptions.ValidationError` — propagates through the middleware chain and beyond. Any Sentry `ignore_errors=[ValidationError]` configuration has no effect on it, since the type doesn't match (and I think the Sentry integration should ignore fastmcp's `ValidationError` as 400 `HTTPException`s are ignored for FastAPI).

`fastmcp` should distinguish bad-call and server-bug `pydantic.ValidationError` converting the former for easy ignoring.

SarthakB11:
## Description

When invalid arguments are passed to a tool, `pydantic-core`'s call validator raises a raw `pydantic.ValidationError` which `FastMCP.call_tool` re-raises unchanged. The error then propagates through the middleware chain as a non-fastmcp type, so consumers filtering on the framework's own error taxonomy (e.g. a Sentry `ignore_errors=[fastmcp.exceptions.ValidationError]` filter) can't match it. The `call_tool` docstring already advertises `ValidationError: If arguments fail validation`, so the implementation had drifted from the contract.

This wraps the caught `PydanticValidationError` in `fastmcp.exceptions.ValidationError` with `from e`, preserving the structured pydantic error as `__cause__` so consumers that still want `.errors()` can walk the chain. The warning log just above the re-raise is unchanged, and the client-side wire contract (a `ToolError` reaches the client) is preserved.

The 10 existing `pytest.raises(ValidationError)` blocks in `tests/server/providers/local_provider_tools/test_parameters.py` that target `mcp.call_tool(...)` are migrated from `pydantic.ValidationError` to `fastmcp.exceptions.ValidationError` in the same PR, since they assert against the new error type. No `TypeAdapter`-direct tests exist in that file, so no `pydantic.ValidationError` raises are left intact.

The reporter notes in the issue thread that tool _bodies_ can also raise `pydantic.ValidationError` from their own bugs, and those should arguably stay distinguishable from bad-call errors. Distinguishing the two requires splitting argument validation from invocation inside `FunctionTool.run` (today `type_adapter.validate_python(arguments)` does both), which is a deeper refactor than this bug warrants.

Closes #4128.

## Contribution type

## Checklist

### DIFF
--- a/fastmcp_slim/fastmcp/server/server.py
+++ b/fastmcp_slim/fastmcp/server/server.py
@@ -55,6 +55,7 @@ from fastmcp.exceptions import (
     PromptError,
     ResourceError,
     ToolError,
+    ValidationError,
 )
 from fastmcp.mcp_config import MCPConfig
 from fastmcp.prompts import Prompt
@@ -1284,7 +1285,7 @@ class FastMCP(
                         name,
                         e.errors(include_url=False),
                     )
-                    raise
+                    raise ValidationError(str(e)) from e
                 except Exception as e:
                     logger.exception(f"Error calling tool {name!r}")
                     # Handle actionable errors that should reach the LLM
--- a/tests/server/providers/local_provider_tools/test_parameters.py
+++ b/tests/server/providers/local_provider_tools/test_parameters.py
@@ -109,7 +109,7 @@ class TestToolParameters:
         assert result.content[0].data == base64.b64encode(b"fake png data").decode()
 
     async def test_tool_with_invalid_input(self):
-        from pydantic import ValidationError
+        from fastmcp.exceptions import ValidationError
 
         mcp = FastMCP()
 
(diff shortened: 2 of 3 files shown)


strawgate:
Please follow the contributing guide and do not raise PRs without discussion on the issue first.

PR #4174, opened 2026-05-17
fix(ci): scope minimize-reviews concurrency by event name

strawgate:
A single review action fires two events at the same instant — `pull_request_review` (submitted) and `pull_request_review_comment` (created). Because the `minimize-resolved-reviews` concurrency group was keyed only on PR number, both sibling runs landed in the same group with `cancel-in-progress: true` and cancelled each other. Usually one survives, but under a burst both siblings can end `cancelled`, leaving a commit with no successful run of this workflow at all (observed on #4060's `codex/keycloak-fix` branch — e.g. runs 2767 and 2768 for the same SHA both cancelled).

Appending `github.event_name` to the concurrency group keeps the two sibling event types in separate groups so they no longer cancel each other. Same-PR runs of the *same* event still supersede cleanly, and the last run of each type always completes.

```yaml
group: minimize-reviews-${{ github.event.pull_request.number || github.event.issue.number }}-${{ github.event_name }}
```


strawgate:
@claude review

jlowin (approved):

PR #4179, opened 2026-05-18
chore(examples): use namespace in mount example

yuyua9:
## Summary

- replace deprecated `prefix=` usage in `examples/mount_example.py` with `namespace=`
- keep the example behavior and namespaces unchanged

## Why

Other examples have already moved to `namespace=`, and `prefix=` now emits a deprecation warning. Updating this example keeps the user-facing sample aligned with the current mount API.

## Validation

- `uv run python -m py_compile examples/mount_example.py`

### DIFF
--- a/examples/mount_example.py
+++ b/examples/mount_example.py
@@ -63,9 +63,9 @@ def check_app_status() -> dict[str, str]:
 
 
 # Mount sub-applications
-app.mount(server=weather_app, prefix="weather")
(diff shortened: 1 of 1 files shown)


yuyua9:
Closing this myself after recalibrating our submission gate. The change is technically correct, but it is only a tiny example-only cleanup without issue/maintainer backing, so it should not have been auto-submitted. Sorry for the noise.

PR #4180, opened 2026-05-18
Add per-key TTL to Context.set_state()

troyharvey:
`set_state` already persists values with a fixed 1-day TTL. This adds an optional `ttl` keyword argument so callers can control expiration per key — useful for short-lived caches (e.g., auth tokens, rate-limit windows) alongside long-lived session data.

```python
# Cache for 5 minutes
await ctx.set_state("user_profile", profile, ttl=300)

# Default 24-hour TTL still works
await ctx.set_state("firm_context", firm_data)
```

The parameter passes through to the underlying `AsyncKeyValue.put()` call. When `ttl` is `None` (default), the existing `_STATE_TTL_SECONDS` (86400) is used — no behavioral change for existing code.

---

### DIFF
--- a/docs/servers/context.mdx
+++ b/docs/servers/context.mdx
@@ -238,12 +238,12 @@ async def get_counter(ctx: Context) -> int:
 Each client session has its own isolated state—two different clients calling `increment_counter` will each have their own counter.
 
 **Method signatures:**
-- **`await ctx.set_state(key, value, *, serializable=True)`**: Store a value in session state
+- **`await ctx.set_state(key, value, *, serializable=True, ttl=None)`**: Store a value in session state
 - **`await ctx.get_state(key)`**: Retrieve a value (returns None if not found)
 - **`await ctx.delete_state(key)`**: Remove a value from session state
 
 <Note>
-State methods are async and require `await`. State expires after 1 day to prevent unbounded memory growth.
+State methods are async and require `await`. State expires after 1 day by default to prevent unbounded memory growth.
 </Note>
 
 #### Non-Serializable Values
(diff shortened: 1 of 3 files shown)


jlowin (changes requested):
Hi @troyharvey, thanks for the contribution but there’s too much sugar in this PR. The state identity problem is worth solving separately and I do want to preserve a piece of the PR.

`set_state(key, None)` is surprising behavior because `None` is a valid value to store, and `Context` already has an explicit `delete_state(key)` method for deletion.

I also don’t want `session_state_store="redis://..."` to auto-create a Redis store. The current API boundary is intentional: users pass a store object, and backend-specific configuration belongs on that store. That keeps FastMCP from becoming a pass-through surface for Redis or `py-key-value-aio` kwargs.

The `ttl=` pass-through on `set_state` is the piece I’m most comfortable with here. The state identity question is also important because MCP sessions/state are changing, so FastMCP needs a deliberate way to define what state belongs to when there is no durable protocol session. I don’t want to merge `set_state_key_prefix()` buried in this PR as that answer, but I do want to think about that API.

I don’t think the store failure behavior here is actually graceful degradation. If the external store is unavailable during `set_state`, the value is written only to request-local memory. When the store comes back, that write is gone, with no replay or reconciliation. A later request will read from the real store and see nothing. That silently breaks the persistence contract the external store was chosen for.

Please narrow this PR to `ttl=` only and remove the Redis URL shorthand, `set_state(None)` deletion semantics, store failure fallback, and user-scoped key prefixing. The state identity problem deserves a focused design pass.

PR #4186, opened 2026-05-19
Allow sampling handler fallback on client errors

### LINKED ISSUE #2162: Allow `sampling_handler` fallback even when client rejects sampling?
### Enhancement

I noticed that when calling `ctx.sample()`, FastMCP will use the server’s `sampling_handler` as a fallback only if the client doesn’t support sampling capability.  
However, if the client _does_ support sampling and the user/client rejects the request, FastMCP just raises `mcp.shared.exceptions.McpError: The user has denied permission to call this method.` and don't falls back to the provided handler.

That seems a bit inconsistent — fallback works for low-capability clients, but not when a “better” client explicitly rejects the sampling request.

---

I think although a user rejection may mean the tool should fail, in practice the meaning of rejection can be ambiguous:

1.  For example, in VS Code Copilot (which supports sampling primitives), if the user accepts sampling but then selects **zero** models, `ctx.sample()` will raise:
    
    ```yml
    mcp.shared.exceptions.McpError: The user has denied permission to call this method.
    ```
    
    This does not necessarily mean “don’t sample”, it could mean “don’t use VS Code’s provided models for sampling”.
2.  FastMCP allows passing a `sampling_handler` at initialization as a fallback, used automatically when the client does not support sampling. But there’s no way to debug this handler in MCP Inspector — rejecting the sampling request triggers the same `McpError`, otherwise I have to manually fill in a sample result.

**Suggestion:**
Add a configurable behavior (e.g. `sampling_handler_behavior` with values like `"always"`, `"fallback"`, and perhaps a new one like `"lazy"`/`"best-effort"`) meaning: always try client first, and if rejected, fall back to the server handler.

* * *

Of course this is not strictly necessary — for point (1) developers can use a CLI flag to set `sampling_handler_behavior="always"` and ask users to do so, or simply wrap their own LLM request helpers instead of using `sampling_handler` (which is what I’m currently doing); for point (2) they can use the FastMCP Client to test the handler.

I’m just raising it as a possible niche use‑case. If it’s too narrow, feel free to close as unplanned.

luyua9:
## Summary

- Adds `sampling_handler_behavior="fallback_on_error"` as an opt-in mode for server sampling handlers.
- Keeps the existing default `"fallback"` behavior unchanged: client sampling errors still propagate.
- Documents the new mode in both sampling docs pages.

Fixes #2162.

## Tests

- `uv run ruff check fastmcp_slim/fastmcp/server/sampling/run.py fastmcp_slim/fastmcp/server/server.py tests/server/sampling/test_sampling_handler_behavior.py`
- `uv run pytest -q tests/server/sampling/test_sampling_handler_behavior.py tests/server/sampling/test_prepare_tools.py`

### DIFF
--- a/docs/servers/sampling.mdx
+++ b/docs/servers/sampling.mdx
@@ -136,6 +136,7 @@ server = FastMCP(
 The `sampling_handler_behavior` parameter controls when the handler is used:
 
 - **`"fallback"`** (default): Use the handler only when the client doesn't support sampling. This lets capable clients use their own LLM while ensuring your tools still work with clients that lack sampling support.
+- **`"fallback_on_error"`**: Try client sampling first, then use the handler if the client sampling request fails. Use this when clients advertise sampling but may reject individual requests, while you still want a server-side handler as an opt-in fallback.
 - **`"always"`**: Always use the handler, bypassing the client entirely. Use this when you need guaranteed control over which LLM processes requests—for cost control, compliance requirements, or when specific model characteristics are essential.
 
 ## Structured Output
--- a/docs/v2/servers/sampling.mdx
+++ b/docs/v2/servers/sampling.mdx
@@ -478,6 +478,7 @@ server = FastMCP(
 The `sampling_handler_behavior` parameter controls when the handler is used:
 
 - **`"fallback"`** (default): Use the handler only when the client doesn't support sampling. This lets capable clients use their own LLM while ensuring your tools still work with clients that lack sampling support.
+- **`"fallback_on_error"`**: Try client sampling first, then use the handler if the client sampling request fails. Use this when clients advertise sampling but may reject individual requests, while you still want a server-side handler as an opt-in fallback.
 - **`"always"`**: Always use the handler, bypassing the client entirely. Use this when you need guaranteed control over which LLM processes requests—for cost control, compliance requirements, or when specific model characteristics are essential.
 
 <Note>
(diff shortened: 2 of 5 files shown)


strawgate:
McpError is very broad here, do we have an idea of what else this would fire on?

jlowin:
Note that sampling is about to be deprecated in the MCP spec (😢) which means that we may deprecate and subseqeuntly remove it from FastMCP when the MCP update lands next month, so I'm hesitant to add new surface area to it as that will need to be maintained for some period of time with limited ability to truly evangelize or enhance it.

jlowin:
Thanks for this, and apologies for the outcome — the work here was sound, but it's been overtaken.

FastMCP 4 removes server-initiated sampling entirely (#4648). The 2026-07-28 protocol has no server→client request channel, and our client now negotiates that revision by default, so `ctx.sample()`'s default outcome had become a runtime error. Rather than keep a method whose happy path requires opting out of the default protocol, we removed it and documented the migration: call an LLM directly from your server, with your own credentials. That takes `sampling_handler` off `FastMCP`, and with it the `sampling_handler_behavior` mode this PR adds — so there's no longer a place for this to land.

Client-side sampling is unaffected: `Client(sampling_handler=...)` stays, so a FastMCP client can still answer another server's sampling requests on a handshake-era connection.

Closing for that reason rather than any problem with the change itself.

PR #4188, opened 2026-05-19
test: import Azure identity before OBO tests

### LINKED ISSUE #4175: Flaky: TestAzureOBOIntegration times out on Windows (azure-identity imported inside the 5s test window)
## Problem

`tests/server/auth/providers/test_azure_scopes.py::TestAzureOBOIntegration` intermittently fails on `Tests: Python 3.10 on windows-latest` with a `pytest-timeout` (5s default) — not an assertion. Example: https://github.com/PrefectHQ/fastmcp/actions/runs/26002021291/job/76426873314 (the timed-out frame is `test_get_obo_credential_returns_configured_credential`).

## Cause

The tests fully mock Azure (no network). But `providers/azure.py` keeps `azure-identity` optional — it's imported only under `TYPE_CHECKING` and lazily inside `get_obo_credential`/`_require_azure_identity`. So the heavy `azure.identity` SDK (pulls `msal`, `cryptography`, many submodules) is imported the *first* time an OBO test runs — **inside the 5s per-test timeout window**. Whichever of the 6+ tests in the class runs first under xdist pays that cost; on a cold Windows runner it can exceed 5s. Order/timing-dependent → flaky.

## Suggested fix

Hoist the import out of the timed window with a module-level `pytest.importorskip("azure.identity.aio")` in the test file: the SDK is imported once at collection time (untimed), and the in-test `patch(...)` then hits the module cache. This also makes the module skip cleanly if the optional `azure` extra is absent, instead of hard-failing as it does today.

Not Windows-specific in principle; Windows just exposes it because cold imports are slower there.

he-yufeng:
﻿## Summary
- import `azure.identity.aio` once at collection time for the Azure scope/OBO test module
- keep the Azure runtime provider lazy; this only changes the test setup path

Fixes #4175.

## To verify
- python -m py_compile tests\server\auth\providers\test_azure_scopes.py
- uv run ruff check tests\server\auth\providers\test_azure_scopes.py
- uv run ruff format --check tests\server\auth\providers\test_azure_scopes.py
- PYTHONUTF8=1 TMP=.tmp\pytest-temp TEMP=.tmp\pytest-temp uv run pytest tests\server\auth\providers\test_azure_scopes.py -q -o cache_dir=.tmp\pytest-cache

### DIFF
--- a/tests/server/auth/providers/test_azure_scopes.py
+++ b/tests/server/auth/providers/test_azure_scopes.py
@@ -3,6 +3,8 @@
 import pytest
 from key_value.aio.stores.memory import MemoryStore
 
+pytest.importorskip("azure.identity.aio")
+
 from fastmcp.server.auth.auth import MultiAuth
 from fastmcp.server.auth.providers.azure import (
     OIDC_SCOPES,


strawgate:
please review the repository rules for contributing before opening a PR.

PR #4198, opened 2026-05-20
fix: avoid duplicated OAuth resource path

### LINKED ISSUE #4197: OAuthProvider.get_well_known_routes asymmetry: PRM route doubled-suffix when base_url.path == mcp_path
## Problem

`OAuthProvider.get_well_known_routes(mcp_path=...)` constructs its two well-known routes via two different code paths, producing **asymmetric output**:

- **`/.well-known/oauth-authorization-server`** derives its path from `issuer_url.path` only — it ignores the `mcp_path` argument.
- **`/.well-known/oauth-protected-resource`** derives its path from `(resource_base_url or base_url).path` AND additionally appends `mcp_path`.

When `base_url.path` already encodes the mount-path AND `mcp_path` mirrors that path (the natural pattern recommended by the official docs — see below), the PRM route ends up with a **doubled suffix** while the AS route remains correct.

## Minimal Reproduction

Tested against `fastmcp==3.3.1` on Python 3.14 (PyPI latest as of 2026-05-20).

```python
from fastmcp.server.auth.providers.in_memory import InMemoryOAuthProvider

# Pattern per https://gofastmcp.com/servers/auth/oauth-proxy — see citation below
provider = InMemoryOAuthProvider(base_url="https://api.example.com/api/mcp/v1")

# mcp_path mirrors the mount path — this is what FastMCP.http_app internal callers
# (create_streamable_http_app, create_sse_app) pass automatically.
routes = provider.get_well_known_routes(mcp_path="/api/mcp/v1")
for r in routes:
    print(r.path)
```

**Actual output:**

```
/.well-known/oauth-authorization-server/api/mcp/v1
/.well-known/oauth-protected-resource/api/mcp/v1/api/mcp/v1
```

The PRM route has the suffix doubled.

## Expected

Per RFC 9728 §3.1 (OAuth 2.0 Protected Resource Metadata) and RFC 8414 §3 (OAuth 2.0 Authorization Server Metadata), both routes should be constructed by the same "insert the well-known URI between host and path" rule and yield symmetric output:

```
/.well-known/oauth-authorization-server/api/mcp/v1
/.well-known/oauth-protected-resource/api/mcp/v1
```

RFC 9728 §3.1 (verbatim):

> Protected resources supporting metadata MUST make a JSON document containing metadata as specified in Section 2 available at a URL formed by inserting a well-known URI string into the protected resource's resource identifier between the host component and the path and/or query components, if any.

RFC 8414 §3 (verbatim):

> Authorization servers supporting metadata MUST make a JSON document containing metadata as specified in Section 2 available at a path formed by inserting a well-known URI string into the authorization server's issuer identifier between the host component and the path component, if any.

Both RFCs describe the same construction. The asymmetry in `get_well_known_routes` violates this structural symmetry for one of the two routes.

## Why this happens

The PRM-route path is computed via `_get_resource_url` (`fastmcp_slim/fastmcp/server/auth/auth.py:343-363`), which additively appends `mcp_path` to `(resource_base_url or base_url)`. The AS-route path is computed in `OAuthProvider.get_well_known_routes` (`auth.py:802-870`) directly from `issuer_url.path`, ignoring `mcp_path`.

## Why this matters — official docs lead users into this pattern

The official OAuth-Proxy docs at `https://gofastmcp.com/servers/auth/oauth-proxy` explicitly tell users to embed the mount-path into `base_url`:

> Public URL where OAuth endpoints will be accessible, including any mount path (e.g., `https://your-server.com/api`). This URL is used to construct OAuth callback URLs and operational endpoints. When mounting under a path prefix, include that prefix in `base_url`.

Following that guidance, then mounting the FastMCP sub-app under a non-trivial prefix (with `mcp_path` mirroring it), produces the RFC-violating PRM path shown above.

## This affects other projects too

Two community discussions report the same class of bug, both unanswered:

- https://github.com/PrefectHQ/fastmcp/discussions/1862 — "Using the OAuthProxy mounted in a FastAPI Server" (OPTIONS 404 on PRM under sub-mount)
- https://github.com/PrefectHQ/fastmcp/discussions/2011 — "OAuth endpoints for servers mounted on FastAPI custom routes"

## Repo references

- `fastmcp_slim/fastmcp/server/auth/auth.py:802-870` — `OAuthProvider.get_well_known_routes` (AS route construction)
- `fastmcp_slim/fastmcp/server/auth/auth.py:343-363` — `OAuthProvider._get_resource_url` (PRM additive append)
- `fastmcp_slim/fastmcp/server/auth/auth.py:723-800` — `OAuthProvider.get_routes` (calls `set_mcp_path` → triggers PRM bug)
- `tests/server/auth/test_oauth_mounting.py:65-108` — `test_well_known_with_mounted_app` (current canonical pattern that avoids the bug by splitting `base_url=mount-prefix` and `mcp_path=internal-path`)

## Question

Should this be scoped as a bug fix (the asymmetric path construction violates the structural symmetry both RFCs share, even though the canonical-pattern caller avoids it) or as an enhancement requiring a design-proposal first? Happy to follow whichever direction you prefer.

he-yufeng:
﻿## Summary
- avoid appending `mcp_path` when the configured resource/base URL already ends at that same path
- keep the existing canonical mounted case working (`base_url=/api`, `mcp_path=/mcp`)
- add coverage for the doubled protected-resource metadata route reported in #4197

Fixes #4197

## Test Plan
- `PYTHONUTF8=1 uv run pytest tests/server/auth/test_oauth_mounting.py::TestOAuthMounting::test_well_known_does_not_duplicate_matching_mcp_path tests/server/auth/test_oauth_mounting.py::TestOAuthMounting::test_well_known_with_mounted_app tests/server/auth/test_remote_auth_provider.py::TestRemoteAuthProvider::test_get_resource_url_uses_resource_base_url_when_provided -q --basetemp .tmp/pytest-well-known-path-final`
- `uv run ruff check fastmcp_slim/fastmcp/server/auth/auth.py tests/server/auth/test_oauth_mounting.py`
- `uv run ruff format --check fastmcp_slim/fastmcp/server/auth/auth.py tests/server/auth/test_oauth_mounting.py`
- `python -m py_compile fastmcp_slim/fastmcp/server/auth/auth.py tests/server/auth/test_oauth_mounting.py`
- `git diff --check`

### DIFF
--- a/fastmcp_slim/fastmcp/server/auth/auth.py
+++ b/fastmcp_slim/fastmcp/server/auth/auth.py
@@ -357,6 +357,11 @@ class AuthProvider(TokenVerifierProtocol):
             return None
 
         if path:
+            resource_path = urlparse(str(resource_base_url)).path.rstrip("/")
+            path_suffix = "/" + path.strip("/")
+            if resource_path == path_suffix:
+                return resource_base_url
+
             prefix = str(resource_base_url).rstrip("/")
             suffix = path.lstrip("/")
             return AnyHttpUrl(f"{prefix}/{suffix}")
--- a/tests/server/auth/test_oauth_mounting.py
+++ b/tests/server/auth/test_oauth_mounting.py
@@ -16,6 +16,7 @@ from starlette.routing import Mount
 from fastmcp import FastMCP
 from fastmcp.server.auth import RemoteAuthProvider
 from fastmcp.server.auth.oauth_proxy import OAuthProxy
+from fastmcp.server.auth.providers.in_memory import InMemoryOAuthProvider
 from fastmcp.server.auth.providers.jwt import StaticTokenVerifier
 
 
@@ -111,6 +112,21 @@ class TestOAuthMounting:
 
             # There will also be an extra route at /api/.well-known/oauth-protected-resource/mcp
             # (from the mounted MCP app), but we don't care about that as long as the correct one exists
+
+    async def test_well_known_does_not_duplicate_matching_mcp_path(self):
+        """base_url may already include the full externally mounted MCP path."""
+        auth_provider = InMemoryOAuthProvider(
+            base_url="https://api.example.com/api/mcp/v1"
+        )
+
+        well_known_routes = auth_provider.get_well_known_routes(mcp_path="/api/mcp/v1")
+        paths = {route.path for route in well_known_routes}
+
+        assert "/.well-known/oauth-authorization-server/api/mcp/v1" in paths
+        assert "/.well-known/oauth-protected-resource/api/mcp/v1" in paths
+        assert (
+            "/.well-known/oauth-protected-resource/api/mcp/v1/api/mcp/v1" not in paths
+        )
 
     async def test_mcp_endpoint_with_mounted_app(self, test_tokens):
         """Test that MCP endpoint works correctly when mounted.


jlowin:
Thanks for the quick patch. I'm leaving this closed because the doubled path comes from passing the full externally mounted MCP URL path as both `base_url.path` and `mcp_path`.

Adding de-duplication in `_get_resource_url()` would hide that misconfiguration and could collapse legitimate repeated path segments. The intended contract is covered in #4197: `base_url + mcp_path = actual externally-accessible MCP URL`.

PR #4210, opened 2026-05-22
Fix pip upgrades for fastmcp metapackage

### LINKED ISSUE #4207: installing fastmcp-slim breaks fastmcp if it is also installed
### What happened?

I have systems with fastmcp installed, but now when someone installs fastmcp-slim, it will break the fastmcp install (due to `from fastmcp import FastMCP` no longer works).

Having both packages exist is troublesome. Is there a plan to EOL fastmcp and we should migrate everything to use the fastmcp-slim ?

### Example Code

```Python

```

### Version Information

```Text
fastmcp == 3.3.1
fastmcp-slim == 3.3.1
```

jlowin:
Upgrading from a code-owning `fastmcp` wheel to the metadata-only metapackage can let pip uninstall files that `fastmcp-slim` just installed, because both distributions claimed `site-packages/fastmcp`. This keeps the same metapackage architecture while moving the slim implementation under a slim-owned import root and using a `.pth` file to keep the public import path stable.

The metapackage owns the `fastmcp` console entry point, while `fastmcp-slim` owns the importable implementation. That means pip's final install step recreates the CLI after uninstalling the old code-owning `fastmcp` wheel without leaving both distributions responsible for the same script.

```python
from fastmcp import FastMCP

app = FastMCP("demo")
```

Fixes #4207.

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -145,7 +145,7 @@ jobs:
           import fastmcp
           import fastmcp.settings
 
-          assert any(ep.name == "fastmcp" for ep in entry_points(group="console_scripts"))
+          assert not any(ep.name == "fastmcp" for ep in entry_points(group="console_scripts"))
 
           try:
               from fastmcp.cli import app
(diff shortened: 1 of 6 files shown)


jlowin:
Closing in favor of #4215. The redirect approach here works, but it adds permanent packaging indirection to fix what is a one-time pip migration off the pre-split `fastmcp` wheel — and adopting it is effectively a one-way door. Since the breakage only affects direct pip upgrades from 3.2 or earlier (uv is unaffected, and it's a self-limiting population), #4215 documents the `--force-reinstall` recovery instead and keeps the `fastmcp-slim` layout as-is.

PR #4232, opened 2026-05-25
feat: add on_connect/on_disconnect middleware hooks (closes #1894)

### LINKED ISSUE #1894: Enhance session start/end hook

Emeralden:
Closes #1894

Adds on_connect and on_disconnect lifecycle hooks to the Middleware base class, giving users access to the MCP session_id when a client connects or disconnects.

Changes:

Middleware.on_connect(context) — fires once per session before any messages are processed
Middleware.on_disconnect(context) — fires once per session after all messages, guaranteed via try/finally (even on error/crash)
Both hooks receive a Context object with session_id for per-client tracking
All registered middleware are notified independently (no call_next chain)
Tests: 6 new tests, 223 total passing

### DIFF
--- a/fastmcp_slim/fastmcp/server/low_level.py
+++ b/fastmcp_slim/fastmcp/server/low_level.py
@@ -237,6 +237,8 @@ class LowLevelServer(_Server[LifespanResultT, RequestT]):
         """
         Overrides the run method to use the MiddlewareServerSession.
         """
+        import fastmcp.server.context
+
         async with AsyncExitStack() as stack:
             lifespan_context = await stack.enter_async_context(self.lifespan(self))
             session = await stack.enter_async_context(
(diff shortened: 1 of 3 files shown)


jlowin:
The MCP spec is about to become stateless which is going to require a reexamination of all hooks that implicitly track sessions, so we should leave this out for now.

PR #4267, opened 2026-06-02
Add subject field to AccessToken initialization

### LINKED ISSUE #4266: Populate the subject property in the AccesToken in get_access_token()
### Problem

`get_access_token()` returns an `AccessToken` object which does not contain a `subject` property.

https://github.com/PrefectHQ/fastmcp/blob/0f4f78c46491a001ba0b409ef643a6649428a951/fastmcp_slim/fastmcp/server/dependencies.py#L519-L527

This makes early logging for authentication incomplete (and in my case made my troubleshooting of auth issue quite tedious). 

### Reproduction

Just invoke `get_access_token()` and print the output. It will return an `AccessToken` with the property subject set to `None` despite having a "sub" in the actual access token returned by the authorization server. 

### Expected

As per [RFC9068 - RFC JSON Web Token (JWT) Profile for OAuth 2.0 Access Tokens](https://www.rfc-editor.org/rfc/rfc9068.html#section-2.2) the "sub" claim is **REQUIRED** and must be provided by the authorization server. I suggest `subject` should be set with `subject=access_token_as_dict["subject"],`.

piaudonn:
Fixes #4266

Add `subject` property to the `AccessToken`.

```python
        return AccessToken(
            token=access_token_as_dict["token"],
            client_id=access_token_as_dict["client_id"],
            scopes=access_token_as_dict["scopes"],
            subject=access_token_as_dict["subject"],
            # Optional fields
            expires_at=access_token_as_dict.get("expires_at"),
            resource=access_token_as_dict.get("resource"),
            claims=access_token_as_dict.get("claims") or {},
        )
```


jlowin:
@piaudonn it sounds like this might require a change in the MCP SDK itself?

piaudonn:
@jlowin It looks like it is in the SDK now:

Updated `BaseModel`: 🔗 https://github.com/modelcontextprotocol/python-sdk/blob/main/src/mcp/server/auth/provider.py#L39C1-L46C82
PR: 🔗 https://github.com/modelcontextprotocol/python-sdk/pull/2686

Would need to update the **TokenVerifier** in: https://github.com/PrefectHQ/fastmcp/blob/0f4f78c46491a001ba0b409ef643a6649428a951/fastmcp_slim/fastmcp/server/auth/providers/jwt.py#L528-L534   
https://github.com/PrefectHQ/fastmcp/blob/0f4f78c46491a001ba0b409ef643a6649428a951/fastmcp_slim/fastmcp/server/auth/providers/jwt.py#L616-L622

And in:
https://github.com/PrefectHQ/fastmcp/blob/0f4f78c46491a001ba0b409ef643a6649428a951/fastmcp_slim/fastmcp/server/auth/providers/introspection.py#L285-L291

And I suppose vendor-based providers would need to be updated to reflect that?
Are we fine with this approach?

jlowin:
Yes, we could flow that through to vendor-specific implementations as long as its universal. Note that the SDK main branch tracks SDK v2 which is targeted for release I think next month

jlowin:
Thanks for this, and for your patience — you were right back in June that the SDK needed to land `subject` first, and it has.

One thing worth recording for anyone reading this later: the original one-line change would have been a no-op in practice. The SDK's `BearerAuthBackend` wraps whatever `TokenVerifier.verify_token()` returns directly, and `get_access_token()` early-returns when that object is already a FastMCP `AccessToken` — which every built-in verifier returns. So the conversion block never runs on a real authenticated request, and `subject` would still have come back `None`.

The fix therefore moved to the verifiers themselves: `JWTVerifier` and `StaticTokenVerifier`, the introspection verifier, and the vendor providers, each sourcing `subject` from the identifier it already extracted. The conversion block keeps its `subject` too, using `.get()` rather than an index so third-party verifiers returning the raw SDK type can't raise. This is the "flow it through to vendor-specific implementations, as long as it's universal" outcome we discussed on the issue.

Merging.

PR #4274, opened 2026-06-03
Add real Monty sandbox e2e coverage for CodeMode call_tool

### LINKED ISSUE #4263: CodeMode execute_tool returns empty result — async call_tool not awaited in sandbox
### What happened?

CodeMode's execute_tool returns an empty result when calling a registered tool. The tool is never actually executed — the async `call_tool` function is passed to the Monty sandbox as an external function, but the sandbox calls it without awaiting the coroutine.

```
<coroutine external_future(0)>
execute_tool result: content=[] structured_content=None meta=None
sys:1: RuntimeWarning: coroutine 'Provider.get_tool' was never awaited
```

### Example Code

```Python
import asyncio
from fastmcp.experimental.transforms.code_mode import CodeMode

mcp = FastMCP("Test", transforms=[CodeMode()])

@mcp.tool()
async def greet(name: str) -> dict:
    """Greet someone."""
    return {"greeting": f"Hello, {name}!"}

async def demo():
        # List tools — should show 'execute_tool'
        print(f"Tools: {[t.name for t in tools]}")

        # Call execute_tool to invoke greet
        result = await client.call_tool("execute", {
            "code": 'print(call_tool("greet", {"name": "World"}))'
        })
        print(f"Result: {result}")  # Expected: {"greeting": "Hello, World!"}
        # Actual: empty result

if __name__ == "__main__":
    asyncio.run(demo())
```

### Version Information

```Text
FastMCP version:                                                                                       3.3.1
MCP version:                                                                                          1.27.1
Python version:                                                                                       3.12.4
Platform:                                                                       macOS-15.7.3-arm64-arm-64bit
FastMCP root path: /Users/baadzianton/TrueConf/trueconf-server-mcp-server/.venv/lib/python3.12/site-packages
```

AlexlaGuardia:
Every end-to-end `execute` test for CodeMode runs through the exec-based `_UnsafeTestSandboxProvider`, so the real `MontySandboxProvider` + `call_tool` path — the one #4263 reported as broken — has no coverage. That's exactly the kind of gap that lets a regression in tool resolution through the sandbox slip past the suite, since the report could only be diagnosed by reproducing against real Monty rather than the test double.

This adds e2e coverage that drives `call_tool` through the actual `pydantic-monty` sandbox. The happy path resolves and returns correctly for both sync and async backend tools, string results round-trip, and sequential `call_tool` calls chain — none of which reproduce the reported "empty result." One test pins the reported symptom itself as a usage error: calling `call_tool` bare (no `await`, no `return`) yields an empty result because the block returns `None`, and the `coroutine ... was never awaited` warning is a cascading effect of that unawaited coroutine, not a separate defect.

```python
@requires_monty
async def test_code_mode_monty_bare_call_returns_empty() -> None:
    mcp = FastMCP("CodeMode Monty Bare Call")

    @mcp.tool
    def greet(name: str) -> str:
        return f"Hello, {name}!"

    mcp.add_transform(CodeMode(sandbox_provider=MontySandboxProvider()))

    result = await mcp.call_tool(
        "execute", {"code": "print(call_tool('greet', {'name': 'World'}))"}
    )
    assert result.content == []
    assert result.structured_content is None
```

The tests skip cleanly when `pydantic-monty` isn't installed. Closes #4263.

### DIFF
--- a/tests/experimental/transforms/test_code_mode.py
+++ b/tests/experimental/transforms/test_code_mode.py
@@ -1,5 +1,6 @@
 import asyncio
 import importlib
+import importlib.util
 import json
 from typing import Any
 
@@ -19,6 +20,11 @@ from fastmcp.experimental.transforms.code_mode import (
 )
 from fastmcp.server.context import Context
 from fastmcp.tools.base import Tool, ToolResult
+
+requires_monty = pytest.mark.skipif(
+    importlib.util.find_spec("pydantic_monty") is None,
+    reason="pydantic-monty is required for the real Monty sandbox provider",
+)
 
 
 def _unwrap_result(result: ToolResult) -> Any:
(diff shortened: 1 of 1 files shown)


AlexlaGuardia:
Rebased onto current main, green across the board (44/44, new Monty e2e cases included). Pure additive coverage for the real MontySandboxProvider call_tool path the test double was hiding. If you'd rather #4263 stay open as a documented footgun than get closed by this, say the word and I'll drop the Closes. Ready when you have a minute.

jlowin (approved):

PR #4303, opened 2026-06-09
Wrap tool argument validation errors

### LINKED ISSUE #4128: pydantic.ValidationError from argument validation should be wrapped as fastmcp's `ValidationError` before reaching middleware

cyq1017:
Fixes #4128.

## Summary
- wrap Pydantic argument validation failures from tool entry-point parsing as FastMCP ValidationError
- keep tool-body Pydantic failures on the existing ToolError path
- cover client and middleware behavior around argument validation

## Verification
- .venv/bin/python -m pytest tests/server/test_input_validation.py -q
- uv run prek run --files fastmcp_slim/fastmcp/tools/function_tool.py fastmcp_slim/fastmcp/server/server.py tests/server/test_input_validation.py

### DIFF
--- a/fastmcp_slim/fastmcp/server/server.py
+++ b/fastmcp_slim/fastmcp/server/server.py
@@ -35,7 +35,6 @@ from mcp.types import (
     ToolAnnotations,
 )
 from pydantic import AnyUrl
-from pydantic import ValidationError as PydanticValidationError
 from starlette.routing import BaseRoute
 from typing_extensions import Self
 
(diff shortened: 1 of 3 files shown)


jlowin:
Closed in #4392

PR #4329, opened 2026-06-15
fix: RateLimitingMiddleware supports async get_client_id callables

### LINKED ISSUE #4320: RateLimitingMiddleware.get_client_id passing an async callable silently disables rate limiting
### What happened?

`get_client_id` in `RateLimitingMiddleware` and `SlidingWindowRateLimitingMiddleware` is typed sync-only (`Callable[[MiddlewareContext], str]`). So when client id isn't embedded in the token (e.g., the token carries an ID but the real tenant/user id must be resolved from a database or external cache) the lookup requires async I/O - that is the real use-case issue Im facing.

Passing an async callable fails silently - the rate limit never triggers, and the only signal is a `RuntimeWarning: coroutine ... was never awaited`. On the other hand passing sync I/O to fetch client id blocks main asyncio loop.

**Note:**
FastMCP already accepts sync-or-async callables elsewhere (`DebugTokenVerifier.validate`) via the `inspect.isawaitable` pattern, which could be reused in this case as well.

**Expected:** Passing async callable works with per-tenant rate limiting using the resolved id.
**Actual:** Passing async callable is not supported and results in every request gets its own bucket keyed by a coroutine object, limits never apply, requests are never rejected.

### Fix

Happy to be assigned, I have a tested fix working branch here: https://github.com/Chotom/fastmcp/tree/async-rate-limiting-middleware-callbacks

### Example Code

```Python
from fastmcp import FastMCP
from fastmcp.server.middleware.rate_limiting import RateLimitingMiddleware

async def get_client_id(context):
    # token subject is opaque; real tenant id lives in storage
    return await tenant_store.resolve(context.fastmcp_context.session_id)

mcp = FastMCP("Server")
mcp.add_middleware(
    RateLimitingMiddleware(max_requests_per_second=5, get_client_id=get_client_id)
)
```

### Version Information

```Text
FastMCP version:                                                                                         3.4.2
MCP version:                                                                                            1.26.0
Python version:                                                                                        3.12.11
Platform:                                                                         macOS-26.3.1-arm64-arm-64bit
FastMCP root path: .../.venv/lib/python3.12/site-packages
```

tcconnally:
Previously `get_client_id` was typed as sync-only `Callable[[MiddlewareContext], str]`. Passing an async callable silently failed — the coroutine was never awaited, resulting in every request getting its own bucket and rate limiting being effectively disabled.

Added `inspect.isawaitable()` pattern (already used elsewhere in FastMCP, e.g. `DebugTokenVerifier.validate`) to support both sync and async callables.

Closes #4320

### DIFF
--- a/fastmcp_slim/fastmcp/server/middleware/rate_limiting.py
+++ b/fastmcp_slim/fastmcp/server/middleware/rate_limiting.py
@@ -1,6 +1,7 @@
 """Rate limiting middleware for protecting FastMCP servers from abuse."""
 
 import time
+import inspect
 from collections import defaultdict, deque
 from collections.abc import Callable
 from typing import Any
@@ -127,7 +128,11 @@ class RateLimitingMiddleware(Middleware):
         """
         self.max_requests_per_second = max_requests_per_second
         self.burst_capacity = burst_capacity or int(max_requests_per_second * 2)
-        self.get_client_id = get_client_id
+        self.get_client_id_result = get_client_id
+        if inspect.isawaitable(self.get_client_id_result):
+            self.get_client_id = await self.get_client_id_result
+        else:
+            self.get_client_id = self.get_client_id_result
         self.global_limit = global_limit
 
         # Storage for rate limiters per client


tcconnally:
I've requested assignment on the linked issue. Once assigned, please reopen this PR (or I can open a new one). The fix is ready to go.

PR #4345, opened 2026-06-21
docs: add audit/event-record recipe for tool-call middleware

### LINKED ISSUE #4244: Proposal: optional tool-call event hooks for downstream audit and learning layers
## Context

FastMCP is increasingly used as a tool/server layer in agent systems. Downstream systems often need a stable way to observe tool calls without wrapping every tool manually.

Examples include audit logging, policy checks, telemetry export, offline eval generation, and post-run analysis of repeated tool-use failures.

## Proposal

Would FastMCP be open to a small optional event-hook surface around tool calls?

A minimal shape could expose events such as:

- tool call started
- tool call completed
- tool call failed

Each event could include stable metadata only, for example:

- server name
- tool name
- tool schema or schema hash if available
- call id / trace id
- timestamp
- success or error summary
- optional redacted input/output summary

## Why this may be useful

This would let downstream integrations build on FastMCP without monkey-patching or wrapping every tool:

- audit logs can reference exact tool calls;
- telemetry exporters can share a common event stream;
- policy layers can record what happened;
- external analysis tools can inspect repeated failure patterns after the run.

## Non-goals

- Not proposing a full observability framework.
- Not proposing to store raw payloads by default.
- Not proposing any dependency on a specific downstream project.
- Not proposing behavior changes for existing users.

## Open questions

- Is there already a recommended hook/middleware surface for this?
- Should this live as public API, docs recipe, or plugin example?
- Would a small PR with a minimal event protocol and tests be welcome?

Happy to keep this external if it is out of scope for FastMCP core.

AlexlaGuardia:
Follows the discussion in #4244 on a tool-call event/audit surface.

The thread converged on a single conclusion: no new FastMCP API is needed — `on_call_tool` already exposes call start, the resolved `ToolResult` (so empty/`isError` detection), timing, and short-circuit. The gap is discoverability: every downstream project (Kavach, MCP Doctor, ChainWeaver, the original proposal) re-derives the same audit-record pattern.

This adds a canonical recipe to `docs/servers/middleware.mdx`:

- **Audit and Event Records** — one middleware that emits a structured, redacted record per call: tool name, call id, input schema hash, redacted argument summary, result class (`completed`/`empty`/`error`/`failed`), and duration. Raw inputs/outputs stay out of the default path.
- A short **deny/policy** variant on the same hook.
- A pointer to [OpenTelemetry](/servers/telemetry) for the telemetry-export half, making the OTel-vs-middleware split explicit (export → OTel; result-aware/runtime decisions → `on_call_tool`).

No public API changes — docs only. Code blocks use the documented `ToolResult` fields (`content`, `structured_content`, `is_error`) and `MiddlewareContext` attributes (`message`, `timestamp`, `fastmcp_context`), and parse clean.

Happy to adjust the framing, field set, or placement.

Closes #4244

### DIFF
--- a/docs/servers/middleware.mdx
+++ b/docs/servers/middleware.mdx
@@ -864,11 +864,19 @@ def _redact(arguments: dict | None) -> dict:
     return {key: "<redacted>" for key in (arguments or {})}
 
 
+def _call_id(context: MiddlewareContext) -> str | None:
+    """Request id when an MCP session is active (see Session Availability above)."""
+    ctx = context.fastmcp_context
+    if ctx is not None and ctx.request_context:
+        return ctx.request_id
+    return None
+
+
 class AuditMiddleware(Middleware):
     async def on_call_tool(self, context: MiddlewareContext, call_next):
         record = {
             "tool": context.message.name,
-            "call_id": getattr(context.fastmcp_context, "request_id", None),
+            "call_id": _call_id(context),
             "schema_hash": _schema_hash(context.message.arguments),
             "arguments": _redact(context.message.arguments),
             "received_at": context.timestamp.isoformat(),
--- a/docs/servers/middleware.mdx
+++ b/docs/servers/middleware.mdx
@@ -838,6 +838,74 @@ class ErrorLogger(Middleware):
 
 Catching and not re-raising suppresses the error entirely. Usually you want to log and re-raise.
 
+### Audit and Event Records
+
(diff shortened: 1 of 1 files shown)


AlexlaGuardia:
closed by the require-issue-link bot, not a review.. body referenced #4244 but no `Closes` keyword + I'm not assigned (external contributor). dgenio confirmed in the issue thread it matches his intent but can't do assignment or reopen from his side.

if @jlowin can assign me to #4244 or reopen-with-bypass, happy to push the keyword fix. or just say what the gate wants.

the Codex request_id flag is already fixed.. e207f1e5 adds a `_call_id` helper that guards on `ctx.request_context` before reading request_id. matches the Session Availability pattern in the docs.

jlowin (approved):

PR #4380, opened 2026-06-26
Add setup and teardown lifecycle hooks for tools

### LINKED ISSUE #4340: feat: add tool lifecycle hooks (setup/teardown) for resource cleanup on cancellation
## Problem

When an MCP tool is cancelled (via `CancelledError`, timeout, or client disconnect), there is no mechanism for the tool implementation to run cleanup logic. This means:

- Database connections, file handles, and network sockets leak
- Temporary files are not cleaned up
- Locks acquired during tool execution are never released
- Partial results are silently discarded (#4336)

The current workaround is `try/finally` inside every tool function, but this doesn't cover all cancellation paths — asyncio cancellation can interrupt between yield points, and `finally` blocks may not execute if the event loop is under pressure.

## Proposed Solution

Add optional `setup` and `teardown` hooks to the tool registration API:

```python
@server.tool(
    setup=lambda: acquire_lock("my_resource"),
    teardown=lambda: release_lock("my_resource"),
)
async def my_tool():
    ...
```

The `teardown` hook should:
1. Always execute, even on `CancelledError` or timeout
2. Run in a shielded context so it can't be cancelled itself
3. Accept the tool result (or exception) as context for conditional cleanup

## Related

- #4336 (CancelledError silently discards partial result + leaks resources)
- #4305 (custom timeout error hooks)
- MCP spec: tool execution lifecycle

## Impact

Any production MCP server with stateful operations (database queries, file I/O, external API calls) is affected. Without cleanup hooks, resource leaks accumulate over time and can cause server instability.

Emmanuel-Arokiaraj:
## Description

Fixes #4340

Tools that acquire resources during execution currently have to rely on inline `try`/`finally` cleanup, which is easy to miss and does not give FastMCP a consistent lifecycle boundary for cancellation, timeout, task execution, or app-discovered tools. This PR adds `setup` and `teardown` hooks to tool registration so resource acquisition and cleanup can be declared alongside the tool itself.

`setup` runs before each invocation, and `teardown` runs in a shielded scope after setup/tool execution is attempted. The teardown hook can inspect the raw `result`, the raised `exception`, and the `setup_result`, which lets tools clean up conditionally without every implementation duplicating lifecycle boilerplate.

```python
@mcp.tool(
    setup=lambda: acquire_lock("records"),
    teardown=lambda exception=None, setup_result=None: release_lock("records"),
)
async def update_record(record_id: str, value: str) -> str:
    return await write_record(record_id, value)
```

## Contribution type

- [x] Documentation improvement
- [x] Enhancement (maintainers typically implement enhancements — see [CONTRIBUTING.md](../CONTRIBUTING.md))

## Checklist

### DIFF
--- a/docs/servers/tasks.mdx
+++ b/docs/servers/tasks.mdx
@@ -68,6 +68,8 @@ When a client requests background execution, the call returns immediately with a
 Background tasks require async functions. Attempting to use `task=True` with a sync function raises a `ValueError` at registration time.
 </Warning>
 
+Tool lifecycle hooks also run for background tasks. If a task-enabled tool defines `setup` or `teardown`, FastMCP runs those hooks in the Docket worker around the tool function. The `teardown` hook receives the raw task result or exception before the task result is stored for the client.
+
 ## Execution Modes
 
 For fine-grained control over task execution behavior, use `TaskConfig` instead of the boolean shorthand. The MCP task protocol defines three execution modes:
--- a/docs/servers/tools.mdx
+++ b/docs/servers/tools.mdx
@@ -885,7 +885,7 @@ async def update_record(record_id: str, value: str) -> str:
 
 `setup` runs before every invocation. If it returns a value, FastMCP passes that value to `teardown` as `setup_result`.
 
-`teardown` runs after `setup` has been attempted, including when setup or the tool succeeds, raises an exception, times out, or is cancelled. It runs in a shielded cancellation scope so cleanup has a chance to finish even if the tool call is being cancelled. If setup or the tool failed, FastMCP preserves the original exception even if teardown also fails.
+`teardown` runs after `setup` has been attempted, including when setup or the tool succeeds, raises an exception, times out, or is cancelled. It runs in a shielded cancellation scope so cleanup has a chance to finish even if the tool call is being cancelled. If teardown raises, FastMCP logs the teardown failure and preserves the original invocation outcome: successful tool results are still returned, and setup or tool exceptions are still raised.
 
 The `teardown` hook may accept any of these keyword arguments:
 
@@ -899,6 +899,10 @@ The `teardown` hook may accept any of these keyword arguments:
 
 <ParamField body="setup_result" type="Any | None">
   The value returned by `setup`, if any.
+</ParamField>
+
+<ParamField body="timed_out" type="bool">
+  Whether the foreground invocation exceeded the tool's FastMCP `timeout`.
 </ParamField>
 
 Hooks can be sync or async. For sync tools, sync hooks follow the tool's `run_in_thread` setting: by default they run in the worker thread pool, and with `run_in_thread=False` they run inline on the event loop thread.
@@ -922,7 +926,9 @@ async def fetch_data(url: str) -> dict:
     ...
 ```
 
-Timeouts are specified in seconds as a float. When a tool exceeds its timeout, FastMCP returns an MCP error with code `-32000` and a message indicating which tool timed out and how long it ran. Both sync and async tools support timeouts—sync functions run in thread pools, so the timeout applies to the entire operation regardless of execution model.
+Timeouts are specified in seconds as a float. The timeout budget covers the foreground invocation's `setup` hook, tool execution, and result materialization. The `teardown` hook runs afterward in a shielded cancellation scope, outside the timeout budget.
+
+When a tool exceeds its timeout, FastMCP returns an MCP error with code `-32000` and a message indicating which tool timed out and how long it ran. Both sync and async tools support timeouts—sync functions run in thread pools, so the timeout applies regardless of execution model.
 
 <Note>
 Tools must explicitly opt-in to timeouts. There is no server-level default timeout setting.
--- a/docs/servers/tools.mdx
+++ b/docs/servers/tools.mdx
@@ -121,6 +121,14 @@ def search_products_implementation(query: str, category: str | None = None) -> l
   Execution timeout in seconds. If the tool takes longer than this to complete, an MCP error is returned to the client. See [Timeouts](#timeouts) for details.
 </ParamField>
 
+<ParamField body="setup" type="Callable[..., Any] | None">
+  Optional hook that runs before each tool invocation. It can be sync or async. Any value it returns is passed to `teardown` as `setup_result`.
+</ParamField>
+
+<ParamField body="teardown" type="Callable[..., Any] | None">
+  Optional hook that runs after each tool invocation, including failures, timeouts, and cancellation. It can accept `result`, `exception`, and `setup_result` keyword arguments for conditional cleanup.
+</ParamField>
+
 <ParamField body="version" type="str | int | None">
   <VersionBadge version="3.0.0" />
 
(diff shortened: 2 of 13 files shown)


tcconnally (commented):
## Code Review: Tool Lifecycle Hooks (#4380)

**Verdict: Approve with suggestions** (0 critical, 2 medium, 3 minor)

Great work on this feature. The design is clean: hooks propagate through all registration paths (standalone `@tool`, `@mcp.tool`, `@app.tool`, `@app.ui`, filesystem discovery, docket tasks), and the shielding with `anyio.CancelScope(shield=True)` in `_run_teardown` is exactly right. Test coverage is strong across success, failure, timeout, cancellation, sync/async combos, setup failure, and task background execution. The docs are clear with a good runnable example.

Below are findings from a 4-phase review (architecture → core logic → tests → security).

---

### 🔴 Medium

**1. Teardown failure during a successful tool call discards the tool result** — `function_tool.py` L433-435 in `_run_teardown`:

When the tool succeeds (`exception is None`) but the teardown hook raises, the teardown exception is re-raised. The client sees a teardown error and the successful tool result is lost. While "fail loudly" is defensible, the more common production expectation is: log the teardown failure but return the tool result. Consider making this behavior configurable, or at minimum add a prominent docs warning that teardown exceptions on success are fatal.

**2. `_run_raw_with_lifecycle` (docket path) bypasses Pydantic type coercion** — L391-404 vs `run()` L328-333:

The docket/task path calls `self.fn(**arguments)` directly, whereas `run()` routes through `type_adapter.validate_python(arguments)`. This means background tasks skip parameter validation and coercion that synchronous calls get. If this is intentional (docket handles validation separately), it should be documented in the method docstring.

---

### 🟡 Minor

**3. Hook execution mode is coupled to `self.run_in_thread`** — L448-451 in `_call_lifecycle_hook`:

Sync hooks follow the tool's `run_in_thread` setting. A sync tool with `run_in_thread=False` gets inline hooks, which could be correct for Windows COM scenarios but surprising for heavyweight cleanup. Conversely, there's no way to run an async teardown hook when the tool is sync with `run_in_thread=False`. Consider adding separate `setup_run_in_thread`/`teardown_run_in_thread` options long-term.

**4. Timeout budget is shared between setup and tool execution** — L323:

`anyio.fail_after(self.timeout)` wraps both setup and the tool function. If setup is slow, the tool body gets less time. The docs should clarify that `timeout` is total call time including hook execution.

**5. Teardown can't distinguish timeout from other errors** — L348:

When the tool times out, teardown receives a generic `McpError(code=-32000)` via the exception parameter. The original `TimeoutError` is suppressed by `from None`. For conditional cleanup (e.g., "only roll back if this was a timeout"), the hook has to pattern-match the error message string. Passing `timed_out: bool` or the original exception type would be more robust.

---

### ✅ What's solid

- **Shielding**: `anyio.CancelScope(shield=True)` ensures teardown can't be cancelled mid-cleanup
- **Setup failure → teardown**: Teardown still runs when setup fails, and correctly receives the exception + `setup_result=None`
- **Teardown + tool both fail → original preserved**: L433-439 correctly logs the teardown failure and re-raises the original tool exception
- **Context-aware kwargs**: `_lifecycle_hook_kwargs` uses `inspect.signature` to only pass kwargs the hook accepts — clean
- **Filesystem provider**: `extract_components` forwards `setup`/`teardown` from `ToolMeta` at L351-352 — caught this during the sweep
- **Test matrix**: 12 lifecycle hook tests + 3 setup/teardown tests in decorator tests + 4 task integration tests + 1 filesystem provider test = thorough

---

*Reviewed by Hermes Agent — [full methodology](https://github.com/PrefectHQ/fastmcp/pull/4380)*

Emmanuel-Arokiaraj:
Hi @tcconnally 
Thanks for the careful review. I addressed the lifecycle issues you called out:

- teardown failures after a successful tool call are now logged without replacing the successful tool result
- teardown hooks can now accept `timed_out: bool` for timeout-specific cleanup
- task/Docket execution now validates and coerces user arguments while preserving Docket dependency resolution for `Progress`, `CurrentDocket`, etc.
- docs now clarify teardown failure behavior and that the timeout budget covers setup + tool execution/result materialization, with teardown running afterward in a shielded scope

I also added regression coverage for teardown failure after success, timeout context passed to teardown, task argument coercion, and reran the relevant lifecycle/task/dependency tests.

tcconnally (commented):
## Second-Pass Review (after `bcd2e4a` fix)

The two follow-up commits address 3 of the 5 findings from the first pass cleanly:

✅ **Teardown exception on success now logs + preserves result** — `_run_teardown` L420-427 now logs both "completed" and "failed" cases instead of re-raising. The new test `test_teardown_failure_after_success_preserves_tool_result` validates this.

✅ **`timed_out: bool` passed to teardown** — L332 sets `timed_out = True` in the timeout handler, passed through at L358. Test `test_teardown_runs_on_timeout` verifies `received_timed_out is True`. Docs updated with `ParamField`.

✅ **`_run_raw_with_lifecycle` now has Pydantic coercion** — New `_coerce_task_arguments` method at L447-485 builds a per-call Pydantic model to coerce types. New test `test_tool_task_uses_pydantic_argument_coercion` confirms `"41"` → `42` for `int` params. Also now materializes generators on the task path (L387).

✅ **Timeout budget docs clarified** — Now states: "timeout budget covers setup hook, tool execution, and result materialization. Teardown runs outside the timeout budget."

✅ **Clean refactor** — `_execute_with_type_adapter` extracts the duplicated execution logic cleanly.

---

One finding on the new code:

### 🔴 Medium: `_coerce_task_arguments` creates a new Pydantic model per call

**Location:** `function_tool.py` L479-483

```python
model = create_model(
    "TaskArguments",
    __config__=ConfigDict(arbitrary_types_allowed=True),
    **fields,
)
```

`create_model` is called on **every background task invocation**. For high-throughput task systems, this is measurable overhead — dynamic model creation, field resolution, and GC pressure add up. The model depends only on `self.fn` (static for the tool's lifetime).

**Suggestion:** Cache the model on `FunctionTool` (e.g., `self._task_model` computed once in `from_function` or lazily on first use). The dependency chain is deterministic: `self.fn` → `without_injected_parameters(self.fn, ...)` → `inspect.signature(...)` → `create_model(...)`. All inputs are immutable after tool registration.

Similarly, `without_injected_parameters(self.fn, ...)` at L450 creates a new wrapper function per call — caching that too would save cycles.

---

Everything else from the first pass is resolved. No regression in the refactor. The new tests pass (confirmed by reading the test assertions — `"41"` → `42`, timed_out=True received).

Emmanuel-Arokiaraj:
Good catch @tcconnally . I updated the task argument coercion path to cache both pieces that were being rebuilt per invocation:

- the dependency-stripped task user function from `without_injected_parameters(...)`
- the generated Pydantic argument model used for task argument coercion

The cache lives on `FunctionTool` private attrs, so it does not affect schemas or serialized tool metadata. I also added a regression test that patches `create_model` and verifies two task calls only build the model once while still coercing arguments correctly.

tcconnally (approved):
## Third-Pass Review (after `48f49ec` caching fix)

All findings from both previous passes are now resolved. The caching implementation is solid:

### What changed
- **`_task_argument_model`** — PrivateAttr cached model, computed once on first use
- **`_task_user_fn`** — PrivateAttr cached `without_injected_parameters` wrapper, computed once
- **`_task_argument_model_ready`** — Boolean flag separates "no model needed" from "not computed yet"

### Why it's correct
- **Idempotent**: Both `create_model` and `without_injected_parameters` produce deterministic output from `self.fn` + `self.run_in_thread`, both immutable after `from_function`. Reentrant calls are safe — concurrent first-access computes twice but produces identical results.
- **Atomic writes**: `_task_argument_model` is assigned before `_ready`. No window where `_ready=True` but `_model=None`. CPython attribute assignment is atomic for simple references.
- **PrivateAttr isolation**: These fields are excluded from serialization (`model_dump`, `model_json_schema`), so they don't leak to MCP clients via `to_mcp_tool()`.
- **Test validates caching**: `test_tool_task_argument_model_is_cached` monkeypatches `create_model` and verifies exactly 1 call across 2 task invocations, plus validates both results (coercion + increment).

### Full finding resolution

| Finding | Pass | Status |
|---|---|---|
| Teardown failure discards tool result | 1 | ✅ Fixed in `bcd2e4a` |
| Missing Pydantic coercion in docket path | 1 | ✅ Fixed in `bcd2e4a` |
| No `timed_out` context for teardown | 1 | ✅ Fixed in `bcd2e4a` |
| Timeout budget docs unclear | 1 | ✅ Fixed in `bcd2e4a` |
| `_call_lifecycle_hook` coupled to `run_in_thread` | 1 | Design tradeoff — documented |
| `create_model` per-call overhead | 2 | ✅ Fixed in `48f49ec` |

### Edge cases verified
- **Empty model** (no coercible params): `_ready=True`, model stays `None`, `_coerce_task_arguments` returns `arguments` unchanged
- **`**kwargs`/`*args` params**: Skipped from model, pass through unvalidated (correct — can't coerce varargs)
- **Missing required params**: Not caught by Pydantic (field absent from `validation_input`), but function call fails with TypeError → caught by `except BaseException` → teardown runs → task result is error. Parity with sync path.
- **Extra keys in arguments** (DI params from Docket): Excluded from `validation_input`, pass through via `{**arguments, **validated}` merge

No regressions, no new issues. This PR is merge-ready. The test matrix covers all execution paths (foreground, background/task, filesystem-discovered, app-registered, all sync/async hook × tool combos, cancellation, timeout, setup failure, teardown failure, type coercion).

jlowin:
I don't see sufficient motivation in #4340 for this API expansion

PR #4403, opened 2026-06-27
fix(json-schema): keep string constraints when a format is set

### LINKED ISSUE #4404: json_schema_to_type drops minLength/maxLength/pattern when a string format is set
`json_schema_to_type` silently drops `minLength`, `maxLength`, and `pattern` for any string schema that also sets a `format`. `_create_string_type` returns early the moment a format is present, before the constraints block runs.

MRE:

```python
from pydantic import TypeAdapter
from fastmcp.utilities.json_schema_type import json_schema_to_type

t = json_schema_to_type({"type": "string", "format": "phone", "maxLength": 3})
TypeAdapter(t).validate_python("abcd")  # expected: ValidationError; actual: passes
```

`format: "phone"` falls back to `str`, but `maxLength` is never applied. Same for `uri-reference` and any other custom/unknown format. The module docstring lists both `format` and `minLength`/`maxLength`/`pattern` as supported, and `tests/utilities/json_schema_type/test_formats.py` only exercises formats on their own, so the combination is uncovered.

Expected: the length/pattern constraints are applied when the format resolves to a str-based type. Non-str formats (`date-time` -> `datetime`, `json` -> `Json`) carry no length/pattern and should stay as they are.

I have a small fix plus tests ready — happy to take this if a maintainer can assign it to me.

ikatyal2110:
_create_string_type returned early whenever a format was set, so minLength/maxLength/pattern were silently dropped for any string schema that also had a format — e.g. a custom format that falls back to str, or uri-reference. Now the constraints are applied when the format resolves to a str-based type (date-time, json etc. stay untouched). Added tests covering custom-format and uri-reference length constraints.

Fixes #4404

### DIFF
--- a/fastmcp_slim/fastmcp/utilities/json_schema_type.py
+++ b/fastmcp_slim/fastmcp/utilities/json_schema_type.py
@@ -269,12 +269,14 @@ def _create_string_type(schema: Mapping[str, Any]) -> type | Annotated[Any, ...]
     if "const" in schema:
         return Literal[schema["const"]]  # type: ignore
 
+    base: Any = str
     if fmt := schema.get("format"):
         if fmt == "uri":
             return AnyUrl
         elif fmt == "uri-reference":
-            return str
-        return FORMAT_TYPES.get(fmt, str)
+            base = str
+        else:
+            base = FORMAT_TYPES.get(fmt, str)
 
     constraints = {
         k: v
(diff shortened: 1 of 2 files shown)


jlowin (changes requested):
Thanks. This fixes the unknown-format and `uri-reference` cases, but it still drops string constraints for `format: email`. `EmailStr` is not a `str` subclass even though it validates to a string, so the `issubclass(base, str)` gate returns it unchanged. Compose `EmailStr` with `StringConstraints` and add regression coverage for `email` with `maxLength` and `pattern`. Keep `date-time` and `json` unchanged. Happy to re-review once this is addressed.

anujbolewar (commented):
Correct — the unconditional early return was silently dropping minLength/maxLength/pattern for any string format, and routing non-str formats past StringConstraints is the right boundary. One behavior nuance to call out: a custom format that previously fell back to bare str will now round-trip as Annotated str with constraints, which changes the inferred type for existing users of such custom formats. Worth listing as a changelog-behavior note in addition to the fix.

zzstoatzz:
Thanks @ikatyal2110 for tracking this down and for the clear issue and tests. The fix for custom formats and uri-reference is right. The remaining gap from review is `format: email`: EmailStr isn't a str subclass, so its length constraints were still dropped. Rather than leave this waiting, we're finishing it in a follow-up PR that keeps your change and credits you as co-author. Closing this in favor of that one.

PR #4406, opened 2026-06-27
Harden OAuth proxy redirect defaults

jlowin:
OAuth proxy clients should not turn a missing redirect allowlist into a browser redirect to arbitrary infrastructure. The default now accepts loopback callbacks only, validates dynamic client registrations against the active redirect policy, and rechecks stored consent-denial callbacks before sending the browser away.

Known external clients still work by opting into explicit callback patterns:

```python
from fastmcp.server.auth import OAuthProxy

auth = OAuthProxy(
    upstream_authorization_endpoint="https://provider.example.com/authorize",
    upstream_token_endpoint="https://provider.example.com/token",
    upstream_client_id="client-id",
    upstream_client_secret="client-secret",
    allowed_client_redirect_uris=[
        "http://localhost:*",
        "http://127.0.0.1:*",
        "https://app.example.com/oauth/callback",
    ],
)
```


jlowin:
Closing, this is not compatible with MCP design goals e.g. arbitrary redirects by default must be supported

PR #4411, opened 2026-06-28
Fix-issue-4284 : Add Auth0MCPProvider for Auth0 Auth for MCP

### LINKED ISSUE #4284: Auth0 now supports CIMD registration so an implementation is needed based on RemoteAuthProvider
### Enhancement

Here is the announcement: https://auth0.com/changelog#6SeJmK4wyY7sizFRsLJvvq

Currently, the `server/auth/providers/auth0.py` contains an implementation based on `OIDCProxy`.  We now need an implementation based on `RemoteAuthProvider`

vijaydeepsinha:
## Description

FastMCP already supports Auth0 via `Auth0Provider`, which proxies OAuth with fixed app credentials. Auth0 now ships **Auth for MCP** with tenant-level DCR/CIMD, where Auth0 is the authorization server and the MCP server is just a resource server that validates tokens.
This PR adds `Auth0MCPProvider` on top of `RemoteAuthProvider`, following the same pattern as `AuthKitProvider`: OIDC discovery, audience binding to the resource URL, AS metadata forwarding, and an `Auth0JWTVerifier` that treats Auth0's `permissions` claim as scopes.

Issue : https://github.com/PrefectHQ/fastmcp/issues/4284

```python
from fastmcp import FastMCP
from fastmcp.server.auth.providers.auth0 import Auth0MCPProvider
auth = Auth0MCPProvider(
    config_url="https://YOUR_TENANT.auth0.com/.well-known/openid-configuration",
    base_url="http://127.0.0.1:8000",
)

```

Closes #4284


jlowin (changes requested):
Thanks for putting this together. The `RemoteAuthProvider`-based Auth0 provider is the right shape for the issue, but I need the docs changes narrowed before this merges.

Please keep the Auth0-specific setup and decision guidance in `docs/integrations/auth0.mdx`. The new callouts in `docs/clients/auth/cimd.mdx`, `docs/servers/auth/oidc-proxy.mdx`, `docs/servers/auth/remote-oauth.mdx`, and the added Auth0 examples in `docs/servers/auth/authentication.mdx` make the generic auth docs carry integration-specific branching. Those pages should explain the framework concepts and provider categories, then link to provider integration guides when someone needs Auth0-specific setup details.

Concretely, remove the new Auth0-specific cross-page callouts and keep the Auth0 Auth for MCP walkthrough on the Auth0 integration page. It is fine for generic docs to mention provider categories or point readers to integration guides, but provider-specific tenant setup, CIMD behavior, and "use this instead of that" guidance belongs on the integration page.

Happy to re-review once the docs are tightened up.

vijaydeepsinha:
@jlowin Thanks for the feedback. I've removed the Auth0-specific callouts from the generic auth docs and kept the Auth for MCP walkthrough on docs/integrations/auth0.mdx only. Please have a look and suggest if any more changes are needed.

jlowin (approved):

jlowin:
Thanks for this, and for tightening the docs scope earlier — that landed exactly as asked, with the Auth0-specific setup kept on the integration page.

Reviewed the provider implementation before merging. Token validation is inherited unmodified from `JWTVerifier`, and the `permissions` handling is additive over the standard `scope`/`scp` claims rather than replacing them, so the normal fallback still applies when Auth0 doesn't send `permissions`. Audience binding to the resource URL only happens when FastMCP owns the verifier, so a caller-supplied `token_verifier` is left alone.

One fix went in on top of your branch: the module imported `httpx`, but FastMCP installs `httpx2` exclusively, so on a clean install `import fastmcp.server.auth.providers.auth0` raised `ModuleNotFoundError` — which would have broken the existing `Auth0Provider` too. That's now `httpx2`, along with the test file's patch targets, verified by installing a built wheel into an environment with no `httpx` present.

Merging.

vijaydeepsinha:
@jlowin  Thanks, for your thorough review and for catching the httpx/httpx2 packaging issue. I really appreciate the detailed feedback and the explanation behind the implementation review. This was my first open-source contribution, and am glad to see it through to completion. Going forward, I will pay closer attention to these integration and packaging nuances as well. Looking forward to contributing more in the future. Thanks again for the review and guidance.

PR #4444, opened 2026-07-07
Add URL-mode elicitation (ctx.elicit_url)

jlowin:
MCP SDK v2 introduced a second elicitation modality that FastMCP had no surface for: URL-mode elicitation, where instead of answering a schema form in-conversation, the user is directed to visit a URL out-of-band — OAuth consent screens, API-key entry, payment flows — so secrets never pass through the LLM context. This adds `ctx.elicit_url()` alongside the existing `ctx.elicit()`:

```python
from fastmcp import FastMCP, Context

mcp = FastMCP("Connections")

@mcp.tool
async def connect_account(ctx: Context) -> str:
    result = await ctx.elicit_url(
        "Authorize access to your account in your browser.",
        url="https://example.com/oauth/authorize?...",
    )
    if result.action == "accept":
        return "Account connected."
    return "Authorization declined."
```

The result mirrors `ctx.elicit`'s union (`AcceptedUrlElicitation | DeclinedElicitation | CancelledElicitation`), the elicitation id is auto-generated, and client-side handlers distinguish URL requests via `isinstance(params, ElicitRequestURLParams)` with no signature change. It rides the same elicitation channel as `ctx.elicit`, with identical protocol-era behavior; calling it from a background task raises a clear error (the task relay is form-mode only today).

Server and client docs gained URL-elicitation sections. Label: features.


jlowin:
Closing after design review rather than merging. As implemented, `accept` only signals that the user acknowledged the URL — not that the out-of-band flow completed. Making URL elicitation genuinely useful requires completion correlation (callback handling, cross-request state, `elicit_complete` signaling), which is substantial stateful infrastructure — and a blocking tool body is the wrong shape for arbitrarily-long external flows anyway. This modality fits the multi-round-trip model (return input-required, resume when the flow completes) far better, so it will be designed as part of that work instead of introduced in a stateful form first.

The branch is kept: the wire-type plumbing and client-side handler support should be reusable under the eventual design.

PR #4465, opened 2026-07-08
Add import-resolution CI gate for examples

jlowin:
Stale SDK-v1 idioms in `examples/` have now slipped past review three separate times during the v2 migration (`mcp.types` imports in #4452, camelCase attribute reads, list-indexed `call_tool` results) — because nothing in CI ever imports the examples. This adds `tests/test_examples_importable.py`: a ~0.6s gate that AST-parses every example file and verifies its top-level `fastmcp` / `mcp` / `mcp_types` imports actually resolve, name by name. Subtrees with their own `pyproject.toml` (like `examples/testing_demo`, which correctly pins v1) are treated as standalone and skipped. The gate fails on exactly the recurring bug class — an injected `from mcp.types import X` trips it immediately — without executing anything or requiring the examples' optional dependencies.

The sweep that validated the gate also caught the last four stragglers, fixed in the first commit: a camelCase `uriTemplate` read in `examples/skills/client.py`, a dead private-API call (`server._call_tool_mcp`) in the serializer example rewritten to use the in-memory `Client`, and two imports from the runtime-only `fastmcp.tools.tool` shim path that `ty` can't resolve, moved to their canonical modules.

Label: chores

### DIFF
--- a/examples/apps/qr_server/qr_server.py
+++ b/examples/apps/qr_server/qr_server.py
@@ -23,7 +23,7 @@ import base64
 import io
 
 import qrcode  # type: ignore[import-untyped]
-from mcp import types
+from mcp_types import ImageContent
 
 from fastmcp import FastMCP
 from fastmcp.apps import AppConfig, ResourceCSP
@@ -153,7 +153,7 @@ def generate_qr(
     img.save(buffer, format="PNG")
     b64 = base64.b64encode(buffer.getvalue()).decode()
     return ToolResult(
-        content=[types.ImageContent(type="image", data=b64, mimeType="image/png")]
+        content=[ImageContent(type="image", data=b64, mime_type="image/png")]
     )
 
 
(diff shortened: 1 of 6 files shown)


jlowin:
Good catch — confirmed real and fixed in d060e93e. The exclusion keyed off `pyproject.toml`, but only `examples/testing_demo` needs excluding (it has its own `uv.lock` — an independently-resolved graph targeting fastmcp v1 on purpose). The others just declare extras (qrcode/phue2/atproto) and share this tree's install. Narrowed the skip to `uv.lock` presence; that brought 20 more files into scope and caught exactly the regression this gate is for: `qr_server.py`'s `from mcp import types` + `mimeType=` (the field is now `mime_type`), which would raise on the current SDK. Gate now checks 120 files, green.

jlowin:
Superseded by #4466 — instead of a standalone import-resolution test, that PR folds examples/ into the existing ty static-analysis gate, which checks imports AND types on every CI run. Example fixes from here are carried forward there.

PR #4466, opened 2026-07-08
Add examples/ to the ty static-analysis gate

jlowin:
The `examples/` directory sits outside every CI gate — not type-checked, not import-checked, not run. During the SDK v2 migration this bit us four separate times: stale idioms (`from mcp import types`, `mimeType=`, `fastmcp.tools.tool`, `uriTemplate`, `server._call_tool_mcp`) shipped in examples and were only caught by hand, one commit at a time. The prior attempt (#4465) added a bespoke import-resolution test that walked `examples/` and asserted the `fastmcp`/`mcp` imports resolved — but that only ever checked imports, and it was a whole test file to maintain for a narrow slice of what actually goes stale.

This folds `examples/` into the existing `ty` static-analysis gate instead, so every CI run type-checks the examples the same way it checks the library — imports *and* types, on every SDK bump, with no new machinery. The examples were made compliant: the stale SDK idioms are fixed, and genuine type errors are addressed at the source (auth examples narrow `get_access_token()`'s `AccessToken | None` before reading `.claims`, matching the existing guard in `propelauth_oauth`; the sales-dashboard and quiz apps get `TypedDict`s so their heterogeneous data rows type cleanly; `text_me` drops a Liskov-violating `model_config` annotation). Subtrees that can't be checked against this tree's environment are excluded with a one-line reason each: `examples/testing_demo` pins fastmcp v1 on purpose, and `atproto_mcp` / `smart_home` / `apps/qr_server` / `providers/sqlite` / `memory.py` / `get_file.py` require third-party deps (`atproto`, `phue`, `qrcode`, `aiosqlite`, `asyncpg`/`numpy`/`pydantic_ai`/`pgvector`, `aiohttp`) that aren't installed here. The optional screenshot deps (`pyautogui`, `PIL`) are mapped to `Any` alongside the existing `prefab_ui.**` entry.

Before, a renamed SDK symbol in an example reached `main` and was found by a human reading the diff. After, `uv run ty check` fails in CI the moment an example drifts:

```toml
[tool.ty.src]
include = ["fastmcp_slim", "fastmcp_remote", "tests", "examples"]
```

The appropriate label is `chores` (or `enhancements`). ty is the static gate; runtime coverage is partial by nature — the in-memory examples (proxy, mount, search, versioning, the client/server pairs, serializer/tool-result demos) run green locally, while OAuth, sampling, and live-server examples can't run in CI without keys or external services.

### DIFF
--- a/examples/apps/qr_server/qr_server.py
+++ b/examples/apps/qr_server/qr_server.py
@@ -23,7 +23,7 @@ import base64
 import io
 
 import qrcode  # type: ignore[import-untyped]
-from mcp import types
+from mcp_types import ImageContent
 
 from fastmcp import FastMCP
 from fastmcp.apps import AppConfig, ResourceCSP
@@ -153,7 +153,7 @@ def generate_qr(
     img.save(buffer, format="PNG")
     b64 = base64.b64encode(buffer.getvalue()).decode()
     return ToolResult(
-        content=[types.ImageContent(type="image", data=b64, mimeType="image/png")]
+        content=[ImageContent(type="image", data=b64, mime_type="image/png")]
     )
 
 
--- a/examples/apps/quiz/quiz_server.py
+++ b/examples/apps/quiz/quiz_server.py
@@ -12,8 +12,6 @@ Usage:
 
 from __future__ import annotations
 
-from typing import TypedDict
-
 from prefab_ui.actions import SetState, ShowToast
 from prefab_ui.actions.mcp import CallTool, SendMessage
 from prefab_ui.app import PrefabApp
@@ -30,6 +28,7 @@ from prefab_ui.components import (
     Text,
 )
 from prefab_ui.rx import ERROR, RESULT, Rx
+from typing_extensions import TypedDict
 
 from fastmcp import FastMCP, FastMCPApp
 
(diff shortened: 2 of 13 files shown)


jlowin:
Confirmed and fixed in ff03cf01. You're right — `take_quiz` (the `@app.ui()` entry point) takes `questions: list[Question] | None`, so FastMCP builds a Pydantic schema over `Question`, and `typing.TypedDict` raises `PydanticUserError` on 3.10/3.11. The dev env is 3.12 (accepts it) and the ty gate is static, so it slipped through — exactly the compat gap this catches. Switched to `typing_extensions.TypedDict`.

For completeness: `MonthlyRevenue` in the sales-dashboard example was checked and left as `typing.TypedDict` — it's internal data (`list[MonthlyRevenue]`), never a tool parameter, so Pydantic never introspects it.

jlowin:
Confirmed and fixed in ff03cf01. You're right — `take_quiz` (the `@app.ui()` entry point) takes `questions: list[Question] | None`, so FastMCP builds a Pydantic schema over `Question`, and `typing.TypedDict` raises `PydanticUserError` on 3.10/3.11. The dev env is 3.12 (accepts it) and the ty gate is static, so it slipped through — exactly the compat gap this catches. Switched to `typing_extensions.TypedDict`.

For completeness: `MonthlyRevenue` in the sales-dashboard example was checked and left as `typing.TypedDict` — it's internal data (`list[MonthlyRevenue]`), never a tool parameter, so Pydantic never introspects it.

PR #4472, opened 2026-07-08
Restore HTTP host guard compatibility

jlowin:
FastMCP 3.4.3 made the Streamable HTTP Host/Origin guard default-on, which is the right direction for a major release but a patch-release compatibility break for existing ASGI, serverless, and reverse-proxy deployments. FastMCP 3.x should keep the guard implementation and explicit opt-ins without rejecting traffic that previously worked.

This restores the 3.x default to disabled while leaving strict and `auto` validation available for deployments that opt in intentionally. The DNS rebinding conformance scenario returns to the expected-failures baseline until v4 can redesign default-on behavior with a complete trusted-proxy story.

```python
from fastmcp import FastMCP

mcp = FastMCP("My Server")

app = mcp.http_app(
    host_origin_protection=True,
    allowed_hosts=["mcp.example.com"],
    allowed_origins=["https://app.example.com"],
)
```

### DIFF
--- a/docs/deployment/http.mdx
+++ b/docs/deployment/http.mdx
@@ -103,11 +103,11 @@ If you're mounting an authenticated server under a path prefix, see [Mounting Au
 
 ### Host and Origin Protection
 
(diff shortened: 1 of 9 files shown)


yinyunpan:
3.4.3 版本，好坑，healthcheck检查不过。原来这个问题。

PR #4484, opened 2026-07-09
Surface client extensions= and flip mode default to auto

jlowin:
Two approved client-surface changes. The first is simple: `fastmcp.Client` now accepts `extensions=` and `result_claims=` (the SDK's SEP-2133 client extension types), threaded through every transport. User extensions *compose* with FastMCP's internal task-status binding rather than clobbering it — the internal binding always leads, and a user extension that tries to bind the same method gets a clear duplicate-method error at connect. Covered by tests that fire both bindings on one live connection.

```python
from fastmcp import Client

client = Client("https://example.com/mcp", extensions=[MyVendorExtension()])
```

The second is the approved `mode` default flip from `"legacy"` to `"auto"` — and implementing it surfaced a sequencing consequence that deserves an explicit decision at merge time. With `auto`, a FastMCP client connecting to a FastMCP server negotiates the **modern 2026 era by default**, where the handshake-era features — `ctx.elicit()`, `ctx.sample()`, roots, `ping`, and `task=True` dispatch — are unavailable until the multi-round-trip work lands. About 250 tests tree-wide failed on the flip for exactly the reason user code would: they exercised those features over default connections. All are now pinned `mode="legacy"` where they test handshake-era behavior, and the full non-integration tree passes (6,146 tests). Three library seams were deliberately pinned legacy as part of this: `ProxyClient` backends (proxies forward server-initiated features), `MCPConfigTransport` (always fronts a proxy composite), and the `inspect` utility.

The practical read: **merging this before MRTR ships means elicitation/sampling/tasks are off by default for FastMCP↔FastMCP connections** (each fails with a clear era error, and `Client(..., mode="legacy")` is the one-line revert). If that's premature, the extensions surface in this PR stands alone and the flip commit can be deferred to land alongside MRTR. Both orderings are reasonable; flagging so the timing is a choice rather than a surprise. Change-register entry is marked breaking-behavior with the revert documented.

Label: enhancements, breaking change

### DIFF
--- a/docs/clients/client.mdx
+++ b/docs/clients/client.mdx
@@ -166,19 +166,19 @@ async with client:
 
 MCP has two protocol eras: the original *legacy* era, which begins every connection with an `initialize` handshake, and the *modern* era (protocol version `2026-07-28` and later), which a client discovers by probing the server's `server/discover` endpoint. The `mode` parameter controls which era the client negotiates when it connects.
 
-By default, `mode="legacy"`. This runs the initialize handshake and behaves identically to earlier FastMCP versions, so existing code connecting to any server keeps working unchanged.
+By default, `mode="auto"`. The client probes `server/discover` and adopts the modern protocol when the server responds; for any server that is not positive evidence of modern support, it falls back to the legacy handshake. This makes the default safe against a mixed fleet of legacy and modern servers.
 
 ```python
 from fastmcp import Client
 
-# Legacy handshake (the default)
-client = Client("https://example.com/mcp", mode="legacy")
+# Negotiate the newest era the server supports (the default)
+client = Client("https://example.com/mcp", mode="auto")
 ```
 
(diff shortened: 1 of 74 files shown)


jlowin:
Closing this — the two changes shouldn't ship together. The `mode="auto"` default flip turns elicitation / sampling / tasks off by default for FastMCP↔FastMCP connections until the multi-round-trip work lands (the ~250 test legacy-pins here are the tell), so it belongs with MRTR at release, not now (decision D10, "flip at release"). The `extensions=`/`result_claims=` surface is clean but was entangled with that churn, and `result_claims` needs to be wired through the FastMCP tool path to actually be usable.

We'll revisit the whole client surface — `extensions=`, `result_claims=` fully wired, and the `mode="auto"` default — as one coherent change alongside MRTR. Branch `feature/client-extensions-auto` is kept as salvage.

PR #4522, opened 2026-07-17
Fix docs banner contrast

### LINKED ISSUE #4499: Banner rendering on docs is unreadable in light mode
### What happened?

Image

### Example Code

```Python

```

### Version Information

```Text
n/a
```

jlowin:
The docs banner rendered its announcement at ~1.1:1 contrast in light mode: the custom CSS forced a near-white background while Mintlify hard-codes banner text to white (and wins the specificity fight), so the text was effectively invisible. An earlier pass swapped in solid dark `banner.color` values, which made the text readable but flattened the banner into a plain dark bar — and left a light/dark asymmetry, since a faint animated overlay read over the medium-slate dark base but vanished over the darker navy light base.

This restores an animated, branded background that stays readable in both themes. Because banner text is always white, the `::before` overlay now owns the background with a fully-opaque, deep brand-rainbow gradient (blue → violet → teal → orange → pink, looping) where every stop clears 7:1 against white — the palette echoes the FastMCP watercolor logo while guaranteeing legibility at every animation frame. The overlay is layered behind the text (`z-index`) so the announcement never gets muted, and a single gradient drives both themes, eliminating the asymmetry. A dark `banner.color` stays in `docs.json` as a graceful no-CSS fallback.

```css
#banner::before {
  background: linear-gradient(
    90deg,
    #1e40af 0%, #5b21b6 22%, #115e59 44%,
    #9a3412 66%, #9d174d 88%, #1e40af 100%
  );
  background-size: 250% 100%;
  animation: colorWave 18s ease-in-out infinite alternate;
}
```

Fixes #4499

### DIFF
--- a/docs/css/banner.css
+++ b/docs/css/banner.css
@@ -1,7 +1,5 @@
-/* Banner styling -- improve readability with better contrast */
+/* Banner animation and layout; theme colors are configured in docs.json. */
 #banner {
-  background: #f1f5f9 !important;
-  color: #1e293b !important;
   font-size: 0.95rem !important;
   font-weight: 600 !important;
   padding-top: 12px !important;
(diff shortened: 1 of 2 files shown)


jlowin:
note: requires local check

PR #4564, opened 2026-07-20
Make transformed tool `required` order deterministic

### LINKED ISSUE #4565: Transformed tool `required` order is nondeterministic across processes
### Description

The `required` array in a transformed tool's input schema is built from a `set`, so its order depends on Python's per-process hash randomization (`PYTHONHASHSEED`). The same server produces differently-ordered `tools/list` JSON on every fresh interpreter, which breaks snapshot testing.

### Example Code

```python
import asyncio
from fastmcp.tools import Tool
from fastmcp.tools.tool_transform import ArgTransform

mcp = FastMCP("snap")

@mcp.tool
def base(alpha: int, beta: str, gamma: float, delta: bool, epsilon: int) -> str:
    """Base tool."""
    return "x"

mcp.add_tool(Tool.from_tool(base, name="renamed", transform_args={"alpha": ArgTransform(name="a")}))

async def main():
    async with Client(mcp) as c:
        tools = await c.list_tools()
        print(tools[1].inputSchema["required"])

asyncio.run(main())
```

Run it a few times:

```
$ python repro.py
['delta', 'a', 'epsilon', 'gamma', 'beta']
$ python repro.py
['beta', 'gamma', 'a', 'delta', 'epsilon']
```

Expected: the same schema on every run.

### Version Information

```
FastMCP version: main (also reproduces on latest release)
Python version: 3.12.7
```

🤖 Generated with [GitHub Copilot CLI](https://github.com/features/copilot)

Kludex:
Fixes https://github.com/PrefectHQ/fastmcp/issues/4565 

Snapshot-testing `tools/list` output was impossible for servers using transformed tools: the JSON changed between test runs. The cause is that `TransformedTool` built its `required` array with `list(set)`, so the order depended on Python's per-process hash randomization (`PYTHONHASHSEED`) — every fresh interpreter could produce a different ordering of the same schema.

This PR makes `required` follow the schema's property order in both `_create_forwarding_transform` and `_merge_schema_with_precedence`, so the wire format is stable across processes (and easier to read). Verified by hashing `tools/list` JSON under several `PYTHONHASHSEED` values: previously every seed produced a different result; now they're identical.

```python
base = Tool.from_function(fn)  # fn(alpha, beta, gamma, delta, epsilon)
transformed = Tool.from_tool(base, transform_args={"alpha": ArgTransform(name="a")})

transformed.parameters["required"]
# before: ['delta', 'a', 'epsilon', 'gamma', 'beta']  (varies per process)
# after:  ['a', 'beta', 'gamma', 'delta', 'epsilon']  (always property order)
```

Label: bugs

🤖 Generated with [GitHub Copilot CLI](https://github.com/features/copilot)

### DIFF
--- a/fastmcp_slim/fastmcp/tools/tool_transform.py
+++ b/fastmcp_slim/fastmcp/tools/tool_transform.py
@@ -707,7 +707,8 @@ class TransformedTool(Tool):
         schema = {
             "type": "object",
             "properties": new_props,
-            "required": list(new_required),
+            # Iterate props (not the set) for deterministic ordering
+            "required": [p for p in new_props if p in new_required],
             "additionalProperties": False,
         }
 
@@ -899,7 +900,11 @@ class TransformedTool(Tool):
         result = {
             "type": "object",
             "properties": merged_props,
-            "required": list(final_required),
+            # Iterate props (not the set) for deterministic ordering; keep any
+            # required names not present in properties (sorted) rather than
+            # silently dropping them.
+            "required": [p for p in merged_props if p in final_required]
+            + sorted(final_required - set(merged_props)),
             "additionalProperties": False,
         }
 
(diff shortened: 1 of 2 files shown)


strawgate (approved):
Lgtm

PR #4567, opened 2026-07-20
fix(tools): order transformed tool required list deterministically

avzamelek:
`Tool.from_tool(...)` builds a transformed tool's `required` array from a `set`, so `list(new_required)` serializes in whatever order the interpreter's hash seed produces. The same server emits a differently-ordered `required` in its `tools/list` schema on every fresh process, which breaks snapshot tests and any consumer that expects a stable schema.

The fix orders `required` by property position — the same insertion order already used for `properties` — so it is deterministic and matches how JSON Schema and Pydantic conventionally present required fields. `new_required` stays a set for its membership checks; only the emitted list is ordered.

```python
mcp.add_tool(Tool.from_tool(base, name="renamed", transform_args={"alpha": ArgTransform(name="a")}))

# before: order varied per process, e.g. ['delta', 'a', 'epsilon', 'gamma', 'beta']
# after:  follows property order on every run: ['a', 'beta', 'gamma', 'delta', 'epsilon']

### DIFF
--- a/fastmcp_slim/fastmcp/tools/tool_transform.py
+++ b/fastmcp_slim/fastmcp/tools/tool_transform.py
@@ -707,7 +707,9 @@ class TransformedTool(Tool):
         schema = {
             "type": "object",
             "properties": new_props,
-            "required": list(new_required),
+            # Order `required` by property position so the schema is deterministic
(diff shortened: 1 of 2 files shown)


jlowin:
This was already implemented in #4564

PR #4572, opened 2026-07-20
Negotiate the best mutual protocol era by default

jlowin:
`fastmcp.Client` has always pinned itself to the legacy initialize handshake, even when talking to a server that speaks the modern stateless protocol. That default made sense while the modern era was new, but it inverts the normal relationship between a client and a server: the client refuses to ask what the server can do, and both ends settle for the older of the two regardless of what they share. Meanwhile every modern-only capability — multi-round-trip tools most importantly — is unreachable from a default client, so the feature FastMCP just shipped is invisible unless the caller knows to opt in.

This makes `mode="auto"` the default, so an ordinary `Client(...)` negotiates the best era both ends support and falls back to the handshake when the server is handshake-only. It also surfaces `extensions=` and `result_claims=` as thin passthroughs to the SDK session. Handshake-only capabilities — `ctx.elicit`, `ctx.sample`, `ctx.list_roots`, `ping`, session ids — now raise their existing clear era errors against a default client, and `mode="legacy"` is the one-line revert. SSE and multi-server-config connections stay on the handshake regardless, since neither supports the modern envelope.

```python
async with Client("http://localhost:8000/mcp") as client:
    # Was: 2025-11-25. Now negotiates the best mutual era.
    assert client.protocol_version == "2026-07-28"

async with Client("http://localhost:8000/mcp", mode="legacy") as client:
    assert client.protocol_version == "2025-11-25"
```

`ProxyClient` deliberately does not inherit this. A proxy's backend leg has to match the era its frontend negotiated — the two eras are mutually exclusive on one session — so it pins the handshake explicitly and a regression test holds that pin. Era-mirroring for proxies is a separate change.

Label: enhancements


zzstoatzz:
@jlowin is `mode="legacy"` required for most documented uses of `Client` now? reading the PR body suggests that the answer is no, but then i see a lot of snippets which gained the `mode` kwarg, passing `"legacy"`

jlowin:
@zzstoatzz good call this got overly complex.

PR #4578, opened 2026-07-20
Fix sampling tools capability issue #4577

### LINKED ISSUE #4577: examples/sampling/structured_output.py fails with ToolError: Client does not support sampling with tools
### What happened?

The examples/sampling/structured_output.py script fails during execution because the Client does not explicitly advertise the sampling.tools capability. While the SDK documentation suggests this should be handled automatically, the example requires manual configuration of sampling_capabilities to function correctly.
  
The following simple change seems to fix the example:
```
from mcp.types import SamplingCapability, SamplingToolsCapability

async with Client(
    mcp,
    sampling_handler=handler,
    sampling_capabilities=SamplingCapability(tools=SamplingToolsCapability()),
) as client:
    # ...
```

The doc at docs/clients/sampling.mdx#sampling-capabilities should be changed accordingly.

Thanks

### Example Code

```Python
Run uv run examples/sampling/structured_output.py
Observe the following error:
.exceptions.ToolError: Error calling tool 'analyze_sentiment': Client does not support sampling with  tools. The client must advertise the sampling.tools capability.
```

### Version Information

```Text
FastMCP version: 3.4.4
MCP version: 1.28.1
Python version: 3.12.13
Platform: Linux-6.18.24-x86_64-AMD_Ryzen_7_PRO_6850U_with_Radeon_Graphics-with-glibc2.42
FastMCP root path: /home/user/python-venvs/fastmcp3/lib/python3.12/site-packages
```

VectorCipher:
## Description

The "Sampling Capabilities" section claimed that providing a `sampling_handler` "automatically advertises full sampling capabilities to the server, including tool support." That stopped being true in `7a92b820`, which flipped the default to a bare `SamplingCapability()` but touched only `client.py` — the docs and the sampling examples were left describing the old behavior.

The visible symptom is a confusing failure: a server feature that calls tools under the hood (for example `result_type` for structured output) raises `Client does not support sampling with tools` even though the client passed a handler and the docs said tool support was automatic. This rewrites the section to state that tool execution is opt-in, shows how to advertise it, and names the error you get when it's missing.

```python
from mcp_types import SamplingCapability, SamplingToolsCapability

tools_capability = SamplingCapability(tools=SamplingToolsCapability())

client = Client(
    "my_mcp_server.py",
    mode="legacy",
    sampling_handler=my_handler,
    sampling_capabilities=tools_capability,
)
```

Scoped to documentation. This PR originally also updated `examples/sampling/structured_output.py` and `examples/sampling/tool_use.py`; those changes have been reverted here because both examples are being removed in the modern-protocol work, where server-initiated sampling no longer exists.

Closes #4577

## Contribution type

- [x] Documentation improvement

### DIFF
--- a/docs/clients/sampling.mdx
+++ b/docs/clients/sampling.mdx
@@ -181,7 +181,7 @@ Install the Google Gemini handler with `pip install 'fastmcp[gemini]'`.
 By default, providing a `sampling_handler` advertises a basic sampling capability (without tool execution) to the server. If your handler supports executing tools during sampling, you must explicitly advertise the tool capability:
 
 ```python
-from fastmcp.types import SamplingCapability, SamplingToolsCapability
+from mcp_types import SamplingCapability, SamplingToolsCapability
 
 client = Client(
     "my_mcp_server.py",
--- a/docs/clients/sampling.mdx
+++ b/docs/clients/sampling.mdx
@@ -178,7 +178,19 @@ Install the Google Gemini handler with `pip install 'fastmcp[gemini]'`.
 
 ## Sampling Capabilities
 
-When you provide a `sampling_handler`, FastMCP automatically advertises full sampling capabilities to the server, including tool support. To disable tool support for simpler handlers:
+By default, providing a `sampling_handler` advertises a basic sampling capability (without tool execution) to the server. If your handler supports executing tools during sampling, you must explicitly advertise the tool capability:
+
+```python
+from fastmcp.types import SamplingCapability, SamplingToolsCapability
+
+client = Client(
+    "my_mcp_server.py",
+    sampling_handler=my_handler,
+    sampling_capabilities=SamplingCapability(tools=SamplingToolsCapability()),
+)
+```
+
+To advertise sampling without tool support:
 
 ```python
 from mcp_types import SamplingCapability
--- a/docs/clients/sampling.mdx
+++ b/docs/clients/sampling.mdx
@@ -178,19 +178,24 @@ Install the Google Gemini handler with `pip install 'fastmcp[gemini]'`.
 
 ## Sampling Capabilities
 
-By default, providing a `sampling_handler` advertises a basic sampling capability (without tool execution) to the server. If your handler supports executing tools during sampling, you must explicitly advertise the tool capability:
+Providing a `sampling_handler` advertises basic sampling support. Tool execution during sampling is opt-in: advertise it explicitly when your handler can run tools.
 
 ```python
 from mcp_types import SamplingCapability, SamplingToolsCapability
 
+tools_capability = SamplingCapability(tools=SamplingToolsCapability())
+
 client = Client(
     "my_mcp_server.py",
+    mode="legacy",
     sampling_handler=my_handler,
-    sampling_capabilities=SamplingCapability(tools=SamplingToolsCapability()),
+    sampling_capabilities=tools_capability,
 )
 ```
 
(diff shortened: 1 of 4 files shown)


jlowin:
Thanks for tracking this down to `determine_handler_mode` and fixing both the examples and the docs.

One thing needs updating before this can merge: right after you opened this, `main` merged #4584, which trimmed `fastmcp.types` down to just `Textarea` (no more MCP SDK re-exports). Your `fastmcp_slim/fastmcp/types.py` change to re-export `SamplingToolsCapability` there no longer fits, and the `from fastmcp.types import SamplingCapability, SamplingToolsCapability` imports in both examples and the doc snippet won't work anymore either.

Rebase on `main` and import both from `mcp_types` directly instead, e.g.:

```python
from mcp_types import SamplingCapability, SamplingToolsCapability
```

That already matches how the rest of the codebase does it (`examples/task_elicitation.py`, and the doc's own pre-existing "disable tools" snippet), so you can drop the `fastmcp/types.py` change entirely.

jlowin (changes requested):
(please see above)

VectorCipher:
Ok I will look into it as soon as possible

VectorCipher:
Rebased on main and made the requested changes:

Dropped the fastmcp_slim/fastmcp/types.py change entirely — confirmed it now has zero diff against main.
Updated both the example and docs snippet to import from mcp_types directly instead of fastmcp.types, matching the pattern in examples/task_elicitation.py.

One thing worth flagging separately: I couldn't get examples/sampling/structured_output.py to run end-to-end on current main, because it now pins mcp==2.0.0b2, and that beta already enforces the SEP-2577 protocol changes — ctx.sample()/ctx.sample_step() raise ToolError: server-initiated sampling is not available on MCP 2026-07-28 connections there. This is unrelated to this PR's scope (capability advertisement + import paths), so I verified the actual changes here by inspection instead of a full run. Happy to open a separate issue for the ctx.sample() deprecation if that's useful — let me know.

jlowin:
Oh sorry this runs into a larger problem that I wasn't thinking through - FastMCP 4 will not support sampling at all as it has been dropped in the latest protocol update (as you noted, even the backwards compatible mechanisms in the SDK have changed). We could target the `release/3.x` branch with this PR

jlowin:
Thank you for the work here — you traced this to `determine_handler_mode`, rebased when `fastmcp.types` moved under you, and then found the deeper problem yourself and flagged it in this thread. That last comment is what prompted the change below.

FastMCP 4 removes server-initiated sampling entirely (#4648). The 2026-07-28 protocol has no server→client request channel, and our client now negotiates that revision by default — which is the real reason the example failed for you. It wasn't the tools capability: the client had negotiated a protocol where `sampling/createMessage` does not exist at all, so no capability configuration could have fixed it. Rather than keep an API whose default outcome is a runtime error, we removed it and documented the migration: call an LLM directly from your server.

That leaves nothing here to land. `examples/sampling/` is deleted, so the two example fixes have no target; the `fastmcp.types` re-export was already superseded by #4584; and the doc note describes a server-side feature that no longer exists.

One piece of your finding does survive, if it's what you actually needed: client-side sampling is unaffected. `Client(sampling_handler=..., mode="legacy")` still lets a FastMCP client answer another server's sampling requests, and `sampling_capabilities=` is still how you advertise tool support for it — see [Client Sampling](https://gofastmcp.com/clients/sampling).

Closing because the ground moved, not because of anything wrong with the change.

PR #4580, opened 2026-07-20
Let a server declare which protocol versions it serves

jlowin:
A FastMCP server serves every MCP protocol version at once, and a client negotiates whichever it and the server share. That is the right default, but it leaves an author who genuinely needs one version with no way to say so. A server built entirely on multi-round-trip guard tools needs the modern protocol; a server built on the server-initiated back-channel (`ctx.elicit`, `ctx.sample`, `ctx.list_roots`) needs the handshake protocol. Today either one accepts a client it cannot actually serve, then fails partway through a call with a confusing error instead of refusing the connection cleanly.

This adds an opt-in declaration: `FastMCP(protocol_versions=...)`. It is an allowlist expressed in the SDK's own constants, so authors pass the SDK's era tuples rather than any FastMCP-invented vocabulary, and those tuples grow on their own as the SDK adds revisions.

```python
from mcp_types.version import MODERN_PROTOCOL_VERSIONS

# This server's tools return InputRequiredResult, which only the modern protocol carries.
mcp = FastMCP("guarded", protocol_versions=MODERN_PROTOCOL_VERSIONS)
```

Enforcement is set membership, not a minimum — deliberately, because the versions are not a ladder. The modern protocol is numerically newer but *removed* the back-channel, so "newer" is not "a superset," and a minimum-version model could not express handshake-only at all. This mirrors the SDK's own stated position that protocol versions are an enumerated set rather than an ordered scalar. Both connection paths are covered: the handshake is refused before it commits, and modern connections are checked on the request itself. Refusals use the spec-standard `-32022` error carrying the server's supported list, so a `mode="auto"` client refused at discovery by a handshake-only server reads the supported versions out of the refusal and completes over the handshake on its own — the restriction steers negotiation rather than breaking it.

The default is unrestricted: every version the SDK supports is served. Flipping a server default would disconnect existing clients rather than degrade gracefully, so this is strictly opt-in. A startup check warns — never errors, and only when a set was declared — if the declared versions cannot carry a capability the server actually uses, through a capability map that a future protocol extension extends with a single entry.

Label: enhancements


jlowin:
Parking this rather than merging. A deep dive into the spec, both official SDKs, and real clients settled the design question underneath it:

- The Python SDK has no server-side version restriction at all — its handshake path counter-offers unconditionally, and the spec's handshake model has no server veto. Within-handshake pinning (the part this PR had to degrade to era membership) is unenforceable by design.
- The modern per-request `-32022` mechanics here are spec-native, but every real client today is handshake-era, so the knob would change nothing for actual traffic yet.
- The maintainer's call: mirror the SDK. Since the SDK we compose on has nothing here to mirror, we ship nothing new for now. Per-call era errors already cover mixed servers automatically.

The natural path to this capability is upstream parity — the TypeScript SDK already has `supportedProtocolVersions` on its server; the Python SDK lacks the equivalent. That ask is drafted in the v4 design notes. If it lands, FastMCP thin-wraps it, and the enforcement mechanics and research on this branch are the starting point. Branch is retained.

PR #4587, opened 2026-07-21
Preserve raw CallToolResult tool returns

### LINKED ISSUE #4549: Tool returning CallToolResult(isError=True) is serialized as text content; isError is dropped and the wire response reports success
### What happened?

I have a tool that returns `mcp.types.CallToolResult(..., isError=True)` directly instead of raising `ToolError`. The client receives `isError: False`, and the returned CallToolResult object is JSON-serialized into a text content block inside a fresh success envelope — so the error flag never reaches the client. The native OTel span for the call also reports UNSET rather than ERROR (raising an exception is traced as ERROR correctly).

I expected either the CallToolResult to be passed through as the result envelope (preserving isError), or a loud error telling me this return type is unsupported. If returning CallToolResult is simply not supported, feel free to treat this as a request to fail loudly instead of silently converting it.

### Example Code

```Python
import asyncio
from mcp.types import CallToolResult, TextContent

mcp = FastMCP("t")

@mcp.tool()
def failing_tool() -> CallToolResult:
    return CallToolResult(
        content=[TextContent(type="text", text="DB write failed")],
        isError=True,
    )

async def main():
    async with Client(mcp) as c:
        raw = await c.call_tool_mcp("failing_tool", {})
        print("wire isError:", raw.isError)          # False — expected True
        print("wire content:", raw.content[0].text)  # the whole CallToolResult as a JSON string

asyncio.run(main())
```

### Version Information

```Text
FastMCP version:                                                             3.4.4
MCP version:                                                                1.28.1
Python version:                                                             3.12.3
Platform:            Linux-6.18.33.2-microsoft-standard-WSL2-x86_64-with-glibc2.39
FastMCP root path: /home/thiru/fastmcp-bug-repro/venv/lib/python3.12/site-packages
```

LarryHu0217:
Returning an MCP `CallToolResult` from a FastMCP tool currently falls through generic result conversion. The protocol object is serialized into text inside a successful response, so its top-level `isError` flag and other protocol fields do not reach the client. A bare `CallToolResult` annotation also produces an envelope-shaped output schema even though the returned object controls the protocol response directly.

This preserves `CallToolResult` as a protocol-level escape hatch while retaining FastMCP's `ToolResult` as the friendlier API, and suppresses output schema generation for a bare protocol return annotation:

```python
return CallToolResult(
    content=[TextContent(type="text", text="upstream boom")],
    isError=True,
)
```

Fixes #4549.

Generated with OpenAI Codex.


zzstoatzz:
hi @LarryHu0217 - thanks for looking at this, but i'm not sure that rejecting `CallToolResult` is the way to go here. the upstream python SDK explicitly accepts a returned `CallToolResult`, preserves it unchanged, and suppresses the output schema for a bare `CallToolResult` annotation, and other official SDKs also treat it as a valid tool return type

this PR appears to fix the top-level `isError` symptom by replacing the result with a framework error, but in doing so discards the caller’s `content`, `structuredContent`, `_meta`, etc. it also continues advertising a generated output schema for a return type that is then rejected at runtime

could you rework this to preserve a returned `CallToolResult` as an escape hatch, while retaining `ToolResult` as FastMCP’s friendlier API? bare `CallToolResult` annotation should also avoid generating the current envelope-shaped output schema

otherwise, happy to take this over

zzstoatzz:
@LarryHu0217 looks like your rebase got tangled and pulled unrelated changes into the PR. can you cleanly rebase onto current `main` so only your intended `CallToolResult` changes are included?

zzstoatzz (approved):
lgtm!

PR #4643, opened 2026-07-26
fix: accept callable roots handlers (bound methods, partials)

### LINKED ISSUE #4638: Client roots handlers reject bound methods and functools.partial
`Client(..., roots=...)` accepts a `RootsHandler` typed as `Callable`, but valid callable objects are rejected unless they are plain functions. Bound methods and `functools.partial` objects both fail during client construction.

Minimal repro:

```python
import functools

server = FastMCP()


class RootsProvider:
    async def get_roots(self, _context):
        return ["file:///bound-method"]


async def get_roots(prefix, _context):
    return [f"file:///{prefix}"]


handlers = {
    "bound method": RootsProvider().get_roots,
    "partial": functools.partial(get_roots, "partial"),
}

for name, handler in handlers.items():
    try:
        Client(server, mode="legacy", roots=handler)
    except ValueError as exc:
        print(f"{name}: {exc}")
```

Both cases raise `ValueError: Invalid roots handler`.

Expected: both handlers are accepted because they satisfy the existing callable handler contract.

Actual: callback creation rejects them before the handler can run.

Version information:

```text
FastMCP: 0.0.1.dev127+37fb0ad
MCP: 2.0.0b2
Python: 3.12.12
```

Ranga-Prasath-22:
## Description

create_roots_callback() rejected bound methods, unctools.partial, lambdas, and callable class instances because it validated handlers with inspect.isfunction(), which only matches bare 	ypes.FunctionType. This is inconsistent with the RootsHandler type alias (Callable[...]) and with how sampling and elicitation handlers work — those perform no type check at all.

Replace inspect.isfunction(handler) with callable(handler) in astmcp_slim/fastmcp/client/roots.py. One-line change. Also removed a stale docstring in proxy.py that referenced the old inspect.isfunction() check.

Closes #4638

## Contribution type
- [x] Bug fix

## Checklist
- [x] This PR addresses an existing issue
- [x] I have read CONTRIBUTING.md
- [x] I have run validation and all checks pass

### DIFF
--- a/fastmcp_slim/fastmcp/client/roots.py
+++ b/fastmcp_slim/fastmcp/client/roots.py
@@ -37,7 +37,7 @@ def create_roots_callback(
     if isinstance(handler, list):
         # TODO(ty): remove when ty supports isinstance union narrowing
         return _create_roots_callback_from_roots(handler)  # type: ignore[arg-type]  # ty:ignore[invalid-argument-type]
-    elif inspect.isfunction(handler):
+    elif callable(handler):
         return _create_roots_callback_from_fn(handler)
     else:
         raise ValueError(f"Invalid roots handler: {handler}")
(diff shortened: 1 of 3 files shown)


jlowin:
Closing for #4639

PR #4675, opened 2026-07-27
Update v4 install instructions for the stable SDK pins

jlowin:
**Draft until #4655 merges.** Merging this before the SDK is stable would publish install instructions that don't work yet; merging #4655 without this would publish instructions that actively fail. They need to land together.

The published install snippet in the upgrade guide constrains `mcp==2.0.0b2` and `mcp-types==2.0.0b2`. Once #4655 moves FastMCP's own pins to `mcp>=2.0.0,<3.0.0`, those user-facing constraints become unsatisfiable — under PEP 440 a prerelease does not satisfy a `>=` bound on the same release, so `2.0.0b2` is excluded by `>=2.0.0` and `uv lock` fails on the exact command the section exists to make work.

With a stable SDK the constraint list also gets shorter, because `mcp` and `mcp-types` are no longer prereleases at all. The only prerelease left in the graph is `fastmcp-slim`, which arrives transitively from the `fastmcp` wrapper:

```toml
[project]
dependencies = ["fastmcp==4.0.0b1"]

[tool.uv]
constraint-dependencies = ["fastmcp-slim==4.0.0b1"]
```

The section now also states the pip path (`pip install "fastmcp==4.0.0b1"`, no `--pre` needed) and explains *why* uv needs the extra line, rather than leaving the reader to infer it from a block of pins.

Also corrects the internal release checklist in `dev-docs/v4-notes/`, which still said to cut `4.0.0` directly at SDK GA and referenced the older `2.0.0b1` pins. It now records that a beta ships on the stable SDK first, and flags that the published snippet has to change in the same commit as the pins.

Suggested label: `documentation`.

### DIFF
--- a/dev-docs/v4-notes/index.md
+++ b/dev-docs/v4-notes/index.md
@@ -22,7 +22,7 @@ FastMCP v4.0 is an engine swap. Three forces drive the major version:
 
 The migration merges to `main` and development continues there with subsequent PRs. Releases follow the SDK's own beta timeline:
 
-- **`main` carries the beta pins.** While the SDK is on `mcp==2.0.0b1` / `mcp-types==2.0.0b1`, `main` cuts **pre-releases** (`4.0.0b1`, `4.0.0b2`, …). No stable PyPI release goes out until `mcp 2.0.0` reaches GA — at which point the pins swap to the stable SDK and `4.0.0` ships. The pin-swap is a tracked checklist item on the [Known Gaps](known-gaps.md) page.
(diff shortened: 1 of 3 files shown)


jlowin:
Superseded by #4655, which now carries this fix. The install snippet has to change in the same commit as the dependency floor — a `mcp==2.0.0b2` constraint does not satisfy `mcp>=2.0.0`, so shipping them separately would leave the documented `uv lock` unsatisfiable in between.

PR #4691, opened 2026-07-28
Declarative elicitation: ask for tool parameters by annotating them

jlowin:
Asking a user for input means picking a mechanism and writing the exchange yourself. On handshake connections that's `ctx.elicit()`. On the modern protocol, where the back-channel is gone, it's returning an `InputRequiredResult` and hand-rolling the state machine — building `ElicitRequest` params, stashing values into `request_state`, parsing them back out next round. The two are mutually exclusive, so serving both eras means writing both paths and branching on the protocol version.

This adds a third option: declare which parameters come from the user and let FastMCP do the asking. Those parameters leave the input schema, get filled before the body runs, and — because the ask is no longer baked into the body's control flow — the framework picks the transport, so one function serves both eras.

An `Elicit` takes the same `message` and `response_type` as `ctx.elicit()`, because it describes the same request — you `await` the imperative one and `return` this one. A fixed question can go straight in the annotation; anything more is a **resolver**, a function returning `T | Elicit[T]` where a plain value means the answer was already known and nobody is asked.

```python
def which_airport(destination: str, traveller: str) -> str | Elicit[str]:
    profile = load_profile(traveller)
    if profile.home_airport:
        return profile.home_airport
    return Elicit(f"Which airport in {destination}?", response_type=str)


def confirm(destination: str, airport: str) -> Elicit[bool]:
    return Elicit(f"Book {destination} via {airport}?", response_type=bool)


@mcp.tool
async def book_flight(
    traveller: str,
    destination: Annotated[str, Elicit("Where would you like to fly?")],
    date: Annotated[str, Elicit("When would you like to fly?")],
    airport: Annotated[str, Elicit(which_airport)],
    cabin: Annotated[Cabin | None, Elicit("Which cabin?")] = None,
    approved: Annotated[bool, Elicit(confirm)] = False,
) -> str:
    ...
```

The model sees one parameter, `traveller`. The rest are the user's to answer, and the round structure falls out of what each question refers to rather than being written:

```
traveller = "regular"                    traveller = "newcomer"
  round 1  destination, date, cabin        round 1  destination, date, cabin
  round 2  approved                        round 2  airport
                                           round 3  approved
```

`destination`, `date` and `cabin` quote nothing, so they go out together instead of costing a round trip each. `airport` is only asked of the newcomer — the resolver already knew the returning traveller's home airport. `confirm` quotes two earlier answers, so it lands last; a confirmation that names what it confirms is ordered correctly by saying so, with no ordering knob involved. Declining `cabin` leaves its default and the booking proceeds; drop the default and declining fails the call instead.

Signature mistakes surface at import, not on the first production call: a resolver naming a value the tool doesn't have, two questions waiting on each other, or a resolver whose declared `Elicit[T]` can't fit its parameter. Answers are pinned to a digest of the exact text the client was shown, so redeploying reworded copy mid-conversation re-asks rather than crediting someone with an answer to a question they never saw.

The guard pattern stays, and the docs draw the line. Declared parameters resolve on every round, so a question built by a live search runs that search again and can offer flights that no longer exist by the time the answer arrives. When the work is expensive or non-deterministic, drive the rounds from the body and carry the result in `request_state`. Combining the two in one tool is rejected at registration — one call, one input channel.

The engine lives in a private module and is expected to move into [uncalled-for](https://github.com/chrisguidry/uncalled-for) once its `Annotated` path can inject a value rather than only run a dependency for its side effects. `Elicit` is exported from `fastmcp.elicitation` so that move stays invisible.


jlowin:
Holding this for https://github.com/chrisguidry/uncalled-for/pull/12

PR #4693, opened 2026-07-28
Search transform proxy tools (call_read_tool, call_write_tool, etc.) missing 'title' field

### LINKED ISSUE #4414: Search transform proxy tools (call_read_tool, call_write_tool, etc.) missing 'title' field
## Summary

When tool search mode is enabled, FastMCP generates proxy/discovery tools (`search_tools`, `call_read_tool`, `call_write_tool`, `call_delete_tool`). These generated tools are missing the standard `title` field that all other tools have.

Some MCP clients (notably ChatGPT's web MCP client) require the `title` field to register a tool. Without it, the tool is silently dropped and not available to the user.

## Environment

- FastMCP: 3.4.2
- ha-mcp: 7.9.0
- Client: ChatGPT (web) via Streamable HTTP

## Evidence

When `tools/list` is called with search mode ON, the response includes both regular tools and generated proxy tools:

**Regular tools (have `title`):**
```json
{
  "name": "ha_get_overview",
  "title": "Get System Overview",
  "description": "...",
  "inputSchema": {...},
  "outputSchema": {...},
  "annotations": {"readOnlyHint": true, "idempotentHint": true},
  "_meta": {"fastmcp": {"tags": ["Search & Discovery"]}}
}
```

**Generated proxy tools (missing `title`):**
```json
{
  "name": "ha_call_read_tool",
  "description": "...",
  "inputSchema": {...},
  "annotations": {"readOnlyHint": true},
  "_meta": {"fastmcp": {"tags": []}}
}
```

Notice: no `title` field, empty tags array, and `ha_call_read_tool`/`ha_call_write_tool`/`ha_call_delete_tool` also lack `outputSchema`.

## Impact

ChatGPT's MCP client silently drops all 4 proxy tools, making the entire search-based discovery mechanism non-functional for ChatGPT users. The server reports "Use ha_search_tools to find tools" in `ha_get_overview`, but the client has no `ha_search_tools` to call.

This creates a deadlock:
- With search mode ON: proxy tools are invisible → search doesn't work
- With search mode OFF: 84 tools / 274KB response → some tools still get dropped due to response size

## Suggested fix

Add a `title` field to all generated proxy tools:

| Tool | Suggested title |
|------|----------------|
| `search_tools` | "Search Tools" |
| `call_read_tool` | "Call Read Tool" |
| `call_write_tool` | "Call Write Tool" |
| `call_delete_tool` | "Call Delete Tool" |

Also consider adding `outputSchema` to the call_* tools for consistency with other tools.

Related issue on ha-mcp: https://github.com/homeassistant-ai/ha-mcp/issues/1711

loonister1:
The search transforms now set the MCP `title` on their synthetic `search_tools` and `call_tool` proxies, derived from each tool's configured name by a single private helper (`_synthetic_tool_title`) in `search/base.py`. No public constructor parameters were added, so the fix is confined to giving the generated tools a title that clients like ChatGPT require, and renamed tools (`call_read_tool` → "Call Read Tool") get correct titles automatically.

Fixes #4414

### Testing
[targeted] $ /Users/m2/Documents/PW/github-bot/work/PrefectHQ__fastmcp/4414/.contrib-venv/bin/python -m pytest -x -q --no-header tests/server/transforms/test_search.py

### Notes for reviewers
The source edits land under `fastmcp_slim/fastmcp/server/transforms/search/` while the tests import plain `fastmcp` and live under `tests/` — from the diff alone I cannot tell whether `fastmcp_slim/` is the real package root or a vendored/generated copy, which means I can't confirm the passing tests are exercising the changed code. Also verify `Tool.from_function` actually accepts and propagates `title` through to `to_mcp_tool()`. The two `.mdx` files are byte-identical edits; confirm `docs/v3/` isn't generated from `docs/`. Finally, both new tests assert exact dict equality over `await mcp.list_tools()`, which will break the moment the transform emits any additional tool — assert on the synthetic tools' titles specifically instead.

### DIFF
--- a/docs/servers/transforms/tool-search.mdx
+++ b/docs/servers/transforms/tool-search.mdx
@@ -139,6 +139,8 @@ mcp.add_transform(RegexSearchTransform(
 ))
 ```
 
+Both tools also carry an MCP `title` derived from their name — `find_tools` becomes "Find Tools" — since some clients drop tools that have no title.
+
 ## The `call_tool` Proxy
 
 The `call_tool` proxy forwards calls to the real tool. When a client calls `call_tool(name="search_database", arguments={...})`, the proxy resolves `search_database` through the server's normal tool pipeline — including transforms and middleware — and executes it.
--- a/docs/v3/servers/transforms/tool-search.mdx
+++ b/docs/v3/servers/transforms/tool-search.mdx
@@ -139,6 +139,8 @@ mcp.add_transform(RegexSearchTransform(
 ))
 ```
 
+Both tools also carry an MCP `title` derived from their name — `find_tools` becomes "Find Tools" — since some clients drop tools that have no title.
+
 ## The `call_tool` Proxy
 
 The `call_tool` proxy forwards calls to the real tool. When a client calls `call_tool(name="search_database", arguments={...})`, the proxy resolves `search_database` through the server's normal tool pipeline — including transforms and middleware — and executes it.
--- a/fastmcp_slim/fastmcp/server/transforms/search/base.py
+++ b/fastmcp_slim/fastmcp/server/transforms/search/base.py
@@ -112,6 +112,17 @@ def _schema_type(schema: Any) -> str:
     return "object" if "properties" in schema else "any"
 
 
+def _synthetic_tool_title(name: str) -> str:
+    """Derive a display title for a synthetic tool from its name.
+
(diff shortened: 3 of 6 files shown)


jlowin:
Superseded by #4694, which fixes this at the root (Tool.to_mcp_tool()) rather than per synthetic-tool-factory, so it also covers code mode, sessions, and prompts/resources-as-tools.

PR #4701, opened 2026-07-29
Parse skills frontmatter with yaml.safe_load, falling back to the line parser

### LINKED ISSUE #4416: SkillsDirectoryProvider: multiline description in frontmatter (using |) not parsed correctly
## Bug Description

The SkillsDirectoryProvider in fastmcp requires skill files to have a description in the frontmatter as a single line. Using YAML multiline syntax with `|` does not work.

## Example

This does NOT work:
```yaml
---
description: |
  This is a multiline
  description for the skill
---
```

But this works:
```yaml
---
description: This is a single line description
---
```

## Root Cause

The `parse_frontmatter` function in `fastmcp/server/providers/skills/_common.py` uses simple line-by-line key:value parsing without handling YAML multiline literals:

```python
for line in frontmatter_text.strip().split("\n"):
    if ":" in line:
        key, _, value = line.partition(":")
        # ... simple parsing
```

## Expected Behavior

The parser should either:
1. Support YAML multiline syntax (`|`, `>`, etc.)
2. Or clearly document that only single-line values are supported

## Environment

- fastmcp version: 3.4.2
- Python version: 3.11

## Related

This issue affects anyone creating skills with longer descriptions that naturally span multiple lines.

vishnujayvel:
`parse_frontmatter` splits skill frontmatter line by line on the first `:`, which means it only ever understood single-line `key: value` pairs. A `description: |` block — the idiomatic way to write a multi-line skill description — stored the literal string `|` and silently dropped every continuation line, so a skill with a paragraph-long description surfaced to the model with no description at all.

This parses the block with `yaml.load(..., Loader=yaml.BaseLoader)`, so block scalars and nested values work the way the frontmatter looks like it should work.

`BaseLoader` rather than `safe_load` is the load-bearing choice, and it's the one thing worth reviewing carefully. `SkillInfo.frontmatter` is a public mapping, and `safe_load` applies YAML 1.1 implicit scalar typing to all of it — `version: 1.10` becomes the float `1.1`, `enabled: yes` becomes `True`, a date-like value becomes a `datetime.date`. For a mapping callers can read, that's a backwards-incompatible change rather than an internal typing detail. `BaseLoader` supports block scalars without any implicit coercion, and registers no Python-object constructors, so it stays safe for untrusted frontmatter while returning every scalar as written. (This replaces an earlier revision of this PR that used `safe_load` plus a `description`-only recovery branch; see the review discussion below.)

One consequence worth stating explicitly: `description: null` yields the string `"null"`, not `None` — likewise `~` → `"~"` and a bare `key:` → `""`. That matches the pre-existing line parser, which did `value.strip()` on the raw text and produced exactly those values, so this preserves the original behavior rather than introducing a third one.

The malformed-YAML fallback is unchanged: real `SKILL.md` files in the wild contain unquoted values with `: ` in them, which PyYAML rejects outright. Falling straight through to `{}` would turn a cosmetic bug into a regression that loses the description entirely, so a `YAMLError` (or `RecursionError`, on deeply nested input) falls back to the original line parser and keeps whatever plain keys it can recover.

```python
# SKILL.md
# ---
# name: my-skill
# description: |
#   First line.
#   Second line.
# ---

frontmatter, body = parse_frontmatter(content)
frontmatter["description"]  # "First line.\nSecond line.\n"  (was: "|")
```

Tests assert scalar text is preserved across `version: 1.10` → `"1.10"`, `enabled: yes` → `"yes"`, a date-like value → the original string, and `description: null` → `"null"`.

Closes #4416

🤖 Generated with Claude Code


vishnujayvel:
Found a regression in my own patch before you spent time on it, so flagging and fixing it in
the same breath.

Moving to `yaml.safe_load` means we inherit YAML 1.1 typing, and PyYAML resolves the bare words
`yes`/`no`/`on`/`off` to booleans. Combined with the `str()` I added to keep
`SkillInfo.description` a string, that turned

```yaml
description: yes
```

into the description `"True"`. I checked it against this branch's parent rather than assuming:
pre-patch the naive line parser preserved `"yes"`, so this is something my change introduced,
not something it exposed. The same mechanism also hits numeric-looking scalars, where `1.10`
becomes `"1.1"`. Dates get coerced to `datetime.date` too, but there `str()` happens to render
back to the original text, so nothing visible changes — I'm mentioning it only so the table below
isn't read as claiming a fix that wasn't needed.

`0310ba6c` fixes the class rather than the boolean case: when `yaml.safe_load` returns a non-`str`
for `description`, recover the author's original text via the line-based parser that's already
in the module, and only fall back to `str()` if that recovery doesn't produce a string. String
descriptions short-circuit and are untouched, so the multiline block-scalar handling this PR
exists for is unaffected — I verified `|` and `>` both still round-trip.

One judgement call I'd like you to sanity-check. I originally added
`test_frontmatter_description_bool_coerced_to_str`, asserting `description: true` → `"True"`.
That's now inconsistent with the principle above, so I changed it to preserve `"true"` — the
literal source text — and renamed it accordingly. Applying "keep what the author wrote" to
`yes`/`no` but not to `true`/`false` would be an arbitrary carve-out. If you'd rather
`true`/`false` keep normalising to Python's `"True"`/`"False"` while only the word-form bools are
preserved, say so and I'll split it; I just didn't want to leave two different rules in the same
function without flagging it.

Verified after the change:

| input | before | after |
|---|---|---|
| `description: yes` | `"True"` | `"yes"` |
| `description: no` | `"False"` | `"no"` |
| `description: true` | `"True"` | `"true"` |
| `description: 1.10` | `"1.1"` | `"1.10"` |
| `description: 2026-07-29` | `"2026-07-29"` | `"2026-07-29"` |
| `description: hello world` | `"hello world"` | `"hello world"` |
| `description: \|` (block) | preserved | preserved |

`uv run pytest tests/server/providers/test_skills_provider.py` is green apart from two failures
that reproduce identically on the pre-patch code — `test_main_file_name_is_skill_md` and
`test_supporting_files_parameter` both time out on my machine because the suite walks my real
`~/.claude/skills` directory, which has a few hundred entries. I don't believe they're related to
this change, but I can't rule it out from here, so I'd rather name them than quietly report green.
I also confirmed each of the four new/changed assertions actually fails without the source fix,
rather than trusting that they would.

Since this adds a commit, the workflow runs will need your approval again — the earlier ones are
still sitting at `action_required` on the fork.

jlowin (changes requested):
`yaml.safe_load` changes the behavior of the entire public frontmatter mapping, while the recovery added here only repairs `description`. The previous parser preserved scalar text, but this branch still turns `version: 1.10` into the float `1.1`, `enabled: yes` into `True`, and date-like keys into `datetime.date` objects. `description: null` also remains `None` because the recovery skips null values. Since callers can access `SkillInfo.frontmatter`, these are backwards-incompatible changes rather than internal typing details.

Please use a loader that supports block scalars without implicit scalar coercion, such as `yaml.BaseLoader`, and retain the malformed-YAML fallback if desired. `BaseLoader` uses the basic constructor with no Python-object constructors, so it remains safe for untrusted frontmatter while preserving scalar text. The regression tests should assert preservation outside `description` as well; the current typed-scalar test instead locks in the behavior that needs to be avoided.

vishnujayvel:
Confirmed all five points against CPython 3.14.5 / PyYAML 6.0.3, and against our actual diff.

- `version: 1.10` — `yaml.safe_load` gives `1.1` (float); `yaml.load(doc, Loader=yaml.BaseLoader)` gives `"1.10"` (str).
- `enabled: yes` — `safe_load` gives `True`; `BaseLoader` gives `"yes"`.
- A date-like scalar (`2024-01-15`) — `safe_load` gives `datetime.date(2024, 1, 15)`; `BaseLoader` gives `"2024-01-15"`.
- `description: null` — confirmed against our code, not just the YAML behavior. The recovery block is gated on `parsed["description"] is not None`, so a null value never enters the non-str branch and stays `None` after `parse_frontmatter` returns.
- `BaseLoader` preserves block scalars: `description: |` round-trips to `"line one\nline two\n"` and `description: >` folds to `"folded one folded two\n"`, same as under `safe_load`. That was the reason we reached for `safe_load` in the first place, so this removes the blocker to switching.
- `BaseLoader` safety: it has no `construct_python_object` (or any other Python-object constructor) — confirmed via `hasattr`. Feeding it a `!!python/object/apply:os.system` tag doesn't execute anything; it just falls through to a plain untagged sequence. So it's safe for untrusted frontmatter.

You're also right that our added test locks in the behavior you're objecting to: `test_frontmatter_numeric_scalars_are_typed` asserts `frontmatter["version"] == 1.0` and `isinstance(frontmatter["version"], float)` — i.e. it pins the exact type coercion this review is asking us to remove. And the backwards-compat framing is correct: `SkillInfo.frontmatter` is public, so changing scalar types under `version`, `enabled`, dates, etc. isn't an internal detail, it's a break for any caller reading those keys.

Plan:
- Switch `parse_frontmatter` from `yaml.safe_load` to `yaml.load(frontmatter_text, Loader=yaml.BaseLoader)`.
- Keep the malformed-YAML → line-parser fallback as is.
- Drop the `description`-only bool/type recovery branch, since `BaseLoader` returns all scalars (including `description`) as strings already — the recovery logic becomes unnecessary once nothing is coerced.
- Replace `test_frontmatter_numeric_scalars_are_typed` with assertions that scalar text is preserved as-is across the cases you listed: `version: 1.10` → `"1.10"`, `enabled: yes` → `"yes"`, a date-like value → the original string, and `description: null` → `"null"`.

On that last one — worth stating explicitly since it's the one case where `BaseLoader` differs from both the current branch *and* from `safe_load`: `BaseLoader` returns the string `"null"`, not `None`. I checked that against the pre-PR parser rather than assuming, and it's the same value: the old line parser did `value.strip()` on the raw text, so `description: null` produced `"null"` there too (likewise `~` → `"~"` and a bare `key:` → `""`). So switching to `BaseLoader` restores the original behavior on nulls rather than introducing a third one. Flagging it because "null becomes the string 'null'" is surprising enough that I'd rather it be a stated decision than something you find in the diff.

Will push these changes to the branch.

zzstoatzz (approved):
lgtm! thank you @vishnujayvel

PR #4705, opened 2026-07-29
Don't cache error results in ResponseCachingMiddleware

### LINKED ISSUE #4395: 🐛 Bug Report: ResponseCachingMiddleware caches tool errors unconditionally
### What happened?

In `fastmcp/server/middleware/caching.py` (specifically within the `on_call_tool` hook), the result of a tool execution is cached unconditionally. There is no check to see if `tool_result.is_error` is True.

Because of this, if a tool fails transiently (e.g., due to a network blip, a timeout, or temporary bad input), that error gets cached for 1 hour by default. Every subsequent call to that tool with the same arguments will return the cached error until the TTL expires, rather than retrying the tool.

The fix is a single-line guard: `if not tool_result.is_error:` before putting the result into the cache.

### Example Code

```Python
import asyncio

mcp = FastMCP()
error_count = 0

@mcp.tool()
def flakey_tool() -> str:
    global error_count
    error_count += 1
    raise ValueError("Transient error")

async def demo():
    # Requires attaching ResponseCachingMiddleware to the server
        # First call fails and returns the error (error_count = 1)
        await client.call_tool("flakey_tool", {})
        
        # Second call returns the cached error (error_count is STILL 1, tool did not execute)
        await client.call_tool("flakey_tool", {})

if __name__ == "__main__":
    asyncio.run(demo())
```

### Version Information

```Text
FastMCP version:                                                                                    3.4.2               
MCP version:                                                                                       1.28.1               
Python version:                                                                                    3.14.6               
Platform:                                   Linux-6.18.33.2-microsoft-standard-WSL2-x86_64-with-glibc2.31
```

SaiMouli3:
`ResponseCachingMiddleware` stored every `ToolResult` it saw, including ones reporting failure. A tool that signals a transient problem by returning `is_error=True` — an upstream 503, a timed-out fetch, a rate-limited API — had that failure pinned in the cache for the full hour-long TTL. Every later call with the same arguments replayed the stale error and the tool never got a chance to run again, which is the opposite of what a caching layer should do with a result that is explicitly about this attempt rather than a stable answer.

This adds a guard next to the existing `InputRequiredToolResult` and non-`ToolResult` passthroughs, which already encode the same idea: some results describe a moment, not an answer, so they go back to the caller without being stored. Note that a tool that *raises* was never affected — the exception propagates past the middleware and is converted to an error result further up, so nothing reaches the cache. The bug is specific to error results that a tool returns.

```python
mcp.add_middleware(ResponseCachingMiddleware())

@mcp.tool
def fetch() -> ToolResult:
    response = call_upstream()
    if response.status == 503:
        return ToolResult("upstream unavailable", is_error=True)
    return ToolResult(response.body)

# before: the 503 was cached for an hour and fetch() never ran again
# after:  the next call retries; only the successful result is cached
```

Fixes #4395

🤖 Generated with Claude Code


jlowin (approved):

SaiMouli3 (commented):
Errors bypass the cache unconditionally here rather than becoming a configurable
negative-caching mode. `is_error=True` describes this attempt, not a stable answer,
so the safe default is to let the next call retry. If someone has a real case for
caching failures — an expensive upstream that fails deterministically — that's an
opt-in setting with its own TTL, and worth its own issue rather than being decided
by this guard.

PR #4706, opened 2026-07-29
Support RFC 6570 explode modifier for list-typed query parameters

### LINKED ISSUE #4378: Resource template query params fail for list-typed optional parameters
### What happened?

A resource template can be registered with an optional `list[str]` query parameter, and reading it without the query parameter works. But any query value for that parameter fails at read time with a Pydantic list validation error.

This makes list-style resource filters, such as tags or symbols, hard to use from a URI template. I would expect the template either to support a usable multi-value query form or reject this shape at registration time with a clearer error.

### Example Code

```Python
import asyncio

from pydantic import Field

mcp = FastMCP("demo")

@mcp.resource("items://{category}{?tags}")
def search(category: str, tags: list[str] = Field(default_factory=list)) -> dict:
    return {"category": category, "tags": tags}

async def main():
        for uri in [
            "items://books",
            "items://books?tags=alpha",
            "items://books?tags=alpha&tags=beta",
            "items://books?tags=alpha,beta",
        ]:
            try:
                result = await client.read_resource(uri)
                print(uri, "=>", result[0].text)
            except Exception as exc:
                print(uri, "=>", type(exc).__name__, str(exc).splitlines()[0])

asyncio.run(main())
```

### Version Information

```Text
fastmcp-slim 3.4.2
mcp 1.28.0
pydantic 2.13.4
Python 3.12.13
Windows 11
```

SaiMouli3:
A resource template could declare a `list[str]` query parameter and read cleanly without it, but any query value at all failed with a Pydantic list validation error. The cause is in `match_uri_template`: `parse_qs` returns every value for a key, and the code took `[0]` unconditionally. A list-typed parameter could therefore never receive more than one value, and the single string it did receive was rejected against `list[str]` at read time. There was no working multi-value form — repeated keys and comma-joined values both failed the same way.

This implements the RFC 6570 explode modifier, which is the syntax the spec already defines for exactly this and which the path side of the template grammar (`{var*}`) already supported. A query parameter declared `{?tags*}` collects every repetition into a list; a plain `{?tags}` stays a scalar and keeps its first-value-wins behavior. Because a list-typed parameter without the modifier can only ever fail, that shape is now rejected when the template is created rather than at read time, with an error naming the fix — the issue asked for a usable multi-value form or a clearer registration error, and the two fit together. `expand_uri_template` emits repeated keys for list values so templates still round-trip through match.

```python
@mcp.resource("items://{category}{?tags*}")
def search(category: str, tags: list[str] | None = None) -> dict:
    return {"category": category, "tags": tags or []}

# items://books                       -> tags=[]
# items://books?tags=alpha            -> tags=["alpha"]
# items://books?tags=alpha&tags=beta  -> tags=["alpha", "beta"]
```

A single value still arrives as a one-element list, so the parameter's type does not shift with the number of values supplied.

Fixes #4378

🤖 Generated with Claude Code


jlowin:
Thanks, making repeatability explicit with RFC 6570 explode syntax is the right contract. It keeps scalar behavior stable and fixes this at template matching rather than adding invocation-specific coercion.

Please address the existing postponed-annotation finding before this is ready. Resolve annotations with `get_type_hints(..., include_extras=True)` before `_is_collection_annotation`, including `Annotated` and unions, and add a regression test using `from __future__ import annotations` that verifies plain `{?tags}` is rejected while `{?tags*}` accepts one and repeated values.

Also keep the supported collection surface internally consistent. `_is_collection_annotation` currently accepts `set`, `tuple`, and `frozenset`, while `expand_uri_template` only expands `list` and `tuple` as repeated keys. For this issue, I would scope the feature to list parameters unless the other collection types are implemented and round-trip tested.

jlowin (dismissed):
Requesting changes for the two blockers in my comment above. Resolve postponed and `Annotated` annotations before enforcing the explode requirement, with regression coverage. Also make the supported collection annotations match expansion behavior; for this PR, scope the feature to list parameters unless the other collection types are implemented and round-trip tested. Happy to re-review once these are addressed.

SaiMouli3:
Both addressed.

Annotations are now resolved with `get_type_hints(..., include_extras=True)` (reusing `_resolve_param_hints`, which already tolerates partials) before the collection check, and `Annotated` is unwrapped alongside unions. Added a regression test that builds the function under `from __future__ import annotations`, so its raw signature is the string `"list[str] | None"`: it registers `{?tags}` without the fix and now raises, while `{?tags*}` accepts both one and repeated values. A separate test covers `Annotated[list[str] | None, Field(...)]`.

The check is also scoped to `list` only, matching what `expand_uri_template` emits as repeated keys — `set`, `tuple`, and `frozenset` are no longer flagged, since they don't round-trip today.

fastmcp_slim/fastmcp/resources/template.py
```diff
@@ -153,11 +197,18 @@ def _expand_query_block(match: re.Match[str]) -> str:
...
             underscored = name.replace("-", "_")
             if name in params:
-                parts.append(f"{quote(name)}={quote(str(params[name]))}")
+                value = params[name]
             elif underscored in params:
-                parts.append(f"{quote(name)}={quote(str(params[underscored]))}")
+                value = params[underscored]
+            else:
+                continue
+            values = value if isinstance(value, (list, tuple)) else [value]
```
jlowin: The expander repeats any runtime `list`/`tuple` even when this variable did not have `*`. For example, `expand_uri_template("test://x{?tags}", {"tags": ["a", "b"]})` now emits `test://x?tags=a&tags=b`, but matching that same non-exploded template returns only `{"tags": "a"}`. That contradicts the new documented scalar contract and breaks expand→match identity. Preserve whether this name was exploded before removing the suffix, and only emit repeated keys in that case; add a regression for a plain query variable with a sequence value.

jlowin (dismissed):
The postponed/`Annotated` annotation handling and list-only registration check now address the earlier blockers. One expansion bug remains: repeatable serialization must depend on the template's `*`, not only on the runtime value type.

SaiMouli3:
Good catch — fixed. `_expand_query_block` now records whether the name carried `*` before stripping the suffix, and only emits repeated keys in that case; a plain `{?tags}` serializes its value as a single key regardless of runtime type, which is the pre-existing behavior and matches what `match_uri_template` reads back.

Added a regression asserting `expand_uri_template("test://x{?tags}", {"tags": ["a", "b"]})` emits one `tags=` and round-trips through match unchanged.

SaiMouli3:
This should be ready for another look when you have time.

All three points are addressed: annotations are resolved before the check (so postponed and `Annotated` forms behave the same), the registration check is scoped to `list` only, and expansion now keys off the template's `*` rather than the runtime value's type — a plain `{?tags}` stays a single value whatever it's handed. Each has a regression test, and I checked they fail against the previous code before trusting them.

One small thing I changed on my own: the error text said "is a collection type", which stopped being accurate once the check narrowed to lists. It now says "is a list type".

Rebase-clean against current main. The one red job is `test_code_mode_max_tool_calls_enforced` on 3.13, which is unrelated to this PR — it fails on main too.

SaiMouli3:
Small correction to my last note: CI is fully green now. The 3.13 failure I mentioned was a flake and cleared on the latest run — all eight jobs pass.

zzstoatzz (approved):
The explode modifier now drives repeated query expansion, postponed and `Annotated` list annotations are resolved correctly, and non-exploded sequence behavior remains unchanged. The focused resource-template suite passes 154 tests; a conflict-free synthetic merge with current `main` passes the full suite (7,773 passed, 6 skipped, 16 xfailed) and all enforced prek checks.

zzstoatzz (changes requested):
Withdrawing my approval after comparing the implementation against current-main compatibility and the RFC 6570 reference behavior.

Two blockers remain:

1. `_is_list_annotation()` rejects any union containing a list. A currently valid scalar template such as `tags: str | list[str] | None` with `{?tags}` works on `main`—single and repeated query values arrive through the scalar branch—but this PR rejects it at registration. Please require explode only when the annotation cannot accept a scalar value, while retaining the intended rejection for `list[str] | None`.

2. Query expansion uses `quote()` with its default `safe="/"`, so `{?tags*}` with `["a/b"]` emits `?tags=a/b`. RFC 6570 §3.2.8 allows only unreserved characters here; the reference expansion and the upstream `uritemplate` implementation emit `?tags=a%2Fb`. Please use query-value encoding with no reserved safe characters and add a reserved-character regression.

The core repeated-key behavior and the original issue reproduction both work in end-to-end smoke tests; these findings are compatibility and RFC-conformance gaps around that core.

SaiMouli3:
Both addressed.

**1. Explode is now required only when the annotation can't accept a scalar.**

The old check returned true for *any* union containing a list, which rejected `str | list[str] | None` on a plain `{?tags}` — valid on `main`. It now requires explode only when every non-`None` member is a list:

```python
def _requires_explode(annotation: Any) -> bool:
    if annotation is list:
        return True
    origin = get_origin(annotation)
    if origin is list:
        return True
    if origin is Annotated:
        return _requires_explode(get_args(annotation)[0])
    if origin in (Union, UnionType):
        members = [a for a in get_args(annotation) if a is not type(None)]
        return bool(members) and all(_requires_explode(a) for a in members)
    return False
```

So `list[str] | None` is still rejected, while `str | list[str] | None` registers and reads back through the scalar branch unchanged.

**2. Query values are percent-encoded as unreserved-only.**

`quote()` was running with its default `safe="/"`, so `{?tags*}` with `["a/b"]` emitted `?tags=a/b`. Query expansion now passes `safe=""` for both the name and the value, per RFC 6570 §3.2.8:

```python
if exploded and isinstance(value, (list, tuple)):
    parts.extend(
        f"{quote(name, safe='')}={quote(str(v), safe='')}" for v in value
    )
else:
    parts.append(f"{quote(name, safe='')}={quote(str(value), safe='')}")
```

Path expansion is untouched — a wildcard `{path*}` still keeps `/`, since it may legitimately span segments.

Round-trip on the reserved set:

```python
>>> u = expand_uri_template("test://x{?tags*}", {"tags": ["a/b", "c&d=e", "f g", "h+i", "ü"]})
'test://x?tags=a%2Fb&tags=c%26d%3De&tags=f%20g&tags=h%2Bi&tags=%C3%BC'
>>> match_uri_template(u, "test://x{?tags*}")
{'tags': ['a/b', 'c&d=e', 'f g', 'h+i', 'ü']}
```

Four new assertions cover this — the scalar-union registration and read, plus reserved characters in exploded and non-exploded query values. I checked each one fails against the previous code before trusting it.

One note on the contract boundary: the check tests "every member is a list", not "no member is scalar-compatible", so an exotic all-collection union like `list[str] | dict[str, str]` isn't flagged. Distinguishing those needs a scalar/collection taxonomy that felt like more machinery than the case warrants, and the behavior there matches `main` rather than regressing it. Happy to tighten it if you'd rather the rule read literally.

Also updated `docs/servers/resources.mdx`, which claimed any list-typed parameter raises without `*` — no longer true for the scalar-union case.

Full suite passes on the current rebase (7,780 passed, 6 skipped, 16 xfailed); the two red skills-provider tests are pre-existing parallel flakes and pass serially. prek is clean.

zzstoatzz:
Thanks so much @SaiMouli3. The explode-modifier design here was the right contract, and your commits are carried into #5206 with authorship preserved. Since #5206 now contains this work plus the final compatibility fixes, I'm closing this one as superseded in favor of #5206. Really appreciate you working through several rounds of review on this.

PR #4725, opened 2026-08-02
fix(cli): preserve valid MCP servers during discovery

cyq1017:
When CLI discovery parsed the full mcpServers mapping at once, one malformed entry caused every valid server to be dropped. This change parses each entry independently, preserving valid servers while logging and skipping malformed entries. The regression test covers a config with one valid command server and one malformed entry.

Validation: `compileall`, Ruff lint, Ruff format, and `git diff --check` pass. The focused pytest command could not collect in this checkout because `mcp_types` is not installed.

🤖 Generated with Claude Code


jlowin:
Already implemented in #4714

PR #4726, opened 2026-08-02
fix(auth): forward CIMD setting to OIDC providers

cyq1017:
Auth0Provider, AWSCognitoProvider, and OCIProvider now expose the existing OIDCProxy `enable_cimd` option and forward it to the base class. The default remains enabled, while callers can pass `enable_cimd=False` to disable CIMD for providers that do not want client metadata document support. Provider-level regression tests cover both the opt-out and the unchanged default.

Validation: `compileall`, Ruff lint, Ruff format, and `git diff --check` pass. The focused pytest command could not collect in this checkout because `mcp_types` is not installed.

🤖 Generated with Claude Code


jlowin:
Already implemented in #4719

PR #4751, opened 2026-08-04
Serialize_by_alias=False ignored for nested Pydantic models

### LINKED ISSUE #4744: serialize_by_alias=False ignored for nested Pydantic models
### What happened?

**Description:**

When a tool returns a dict containing a Pydantic model, serialize_by_alias=False on the nested model is ignored aliases are always emitted. Returning the model directly as the tool's return value works correctly; the issue only affects models nested inside a dict.

**Expected:**
`{"status": "success", "item": {"real_name": "hello"}}`

**Actual:**
`{"status": "success", "item": {"alias_name": "hello"}}`

Note:
This is related to but different from issue #1197 which only addressed returning a model directly.

### Example Code

```Python
from pydantic import BaseModel, ConfigDict, Field
import fastmcp

mcp = fastmcp.FastMCP()

class MyModel(BaseModel):
    model_config = ConfigDict(serialize_by_alias=False)
    real_name: str = Field(..., serialization_alias="alias_name")

@mcp.tool()
def my_tool() -> dict:
    return {"status": "success", "item": MyModel(real_name="hello")}

```

### Version Information

```Text
fastmcp 3.4.4, pydantic 2.13.4
```

loonister1:
`by_alias` applies to a whole payload, so tool result serialization resolved it once from the outermost value. A model returned on its own honored its `serialize_by_alias` config, but the moment it was nested in a dict or list the container's default won and the model was serialized with aliases anyway. Serialization now recurses through plain containers and resolves the setting per model, so each one is dumped the way it asked to be.

The output schema is derived from the return annotation, so it has to agree with that: schema resolution now unwraps container types to find the model whose config it should honor. Both sides recognize the same containers — dicts, lists, tuples, sets, and frozensets — so a `list[Model]` or `set[Model]` return still validates against the schema clients check it with.

```python
class Biofile(BaseModel):
    model_config = ConfigDict(serialize_by_alias=False)
    id: str = Field(alias="_id")


@mcp.tool
def get_biofiles() -> dict[str, Biofile]:
    return {"a": Biofile(_id="1")}


# before: {"a": {"_id": "1"}}
# after:  {"a": {"id": "1"}}
```

Fixes #4744

### DIFF
--- a/fastmcp_slim/fastmcp/tools/base.py
+++ b/fastmcp_slim/fastmcp/tools/base.py
@@ -83,6 +83,19 @@ def resolve_serialize_by_alias(value: Any) -> bool:
     return True if configured is None else configured
 
 
+def _dump_nested_models(value: Any, fallback: Callable[[Any], Any] | None) -> Any:
+    """Serialize models found in *value*, each with its own alias setting."""
+    if isinstance(value, dict):
+        return {key: _dump_nested_models(item, fallback) for key, item in value.items()}
+    if isinstance(value, list | tuple):
+        return [_dump_nested_models(item, fallback) for item in value]
+    if isinstance(value, BaseModel):
+        return pydantic_core.to_jsonable_python(
+            value, by_alias=resolve_serialize_by_alias(value), fallback=fallback
+        )
+    return value
+
+
 def to_jsonable_by_alias(
     value: Any, *, fallback: Callable[[Any], Any] | None = None
 ) -> Any:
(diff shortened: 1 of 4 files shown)


jlowin (changes requested):
Thanks. The schema and runtime changes need to recognize the same container types. `_unwrap_model()` now descends through `set[Model]` and `frozenset[Model]`, but `_dump_nested_models()` only recurses through dictionaries, lists, and tuples. On this head, a set containing a model with `serialize_by_alias=False` advertises field names in the output schema but emits aliases at runtime, so `Client.call_tool()` rejects the structured result.

Please either serialize sets and frozensets recursively or constrain schema unwrapping to exactly the containers handled at runtime, then add an end-to-end client regression for a set return. Please also condense the PR description to the repository's PR format and remove the testing summary. Happy to re-review once those are addressed.

loonister1:
Good catch — fixed by serializing sets and frozensets recursively, so the runtime now covers exactly the containers the schema unwraps (dict, list, tuple, set, frozenset). Added an end-to-end client regression for a `set[Model]` return; it fails on the previous head with the validation error you described.

One related tightening in `_unwrap_model` while I was there: it now skips `Annotated` metadata instead of searching every arg, so a metadata object that happens to be a model class can't flip the schema's alias mode for an unrelated return type.

Description condensed and the testing section removed.

jlowin:
Thanks for working through this and for following up on the set and frozenset mismatch. Your changes helped expose that the complexity here was a signal about the abstraction, not just another container case. To make the recursive approach correct, FastMCP has to reproduce Pydantic's traversal rules at runtime and independently traverse type annotations for schemas, then keep both lists of supported containers synchronized.

That complexity points to the fix belonging at a different layer. #4771 uses Pydantic's public `TypeAdapter` serializer so each nested model's configuration is handled natively, and only customizes schema generation enough to apply that same configuration per model. That removes the parallel serialization system entirely.

The work you put into making the original approach correct directly clarified why the fix belongs at the Pydantic API boundary instead.

PR #4770, opened 2026-08-05
Sanitize forwarded request metadata where the proxy copies it

### LINKED ISSUE #4769: Prevent proxy connection metadata from crossing protocol eras
A proxy can negotiate a modern frontend connection and a legacy backend connection independently, then forward the frontend request `_meta` wholesale. This puts modern connection-owned fields such as `io.modelcontextprotocol/protocolVersion`, `clientInfo`, and `clientCapabilities` onto the legacy backend request, where they disagree with the backend session.

Proxy protocol policy should be explicit (`independent` or `mirror`), client factories should remain request-independent, and the outbound backend connection should rebuild or omit its own reserved metadata while preserving progress, tracing, task, and application metadata.

jakekaplan:
The proxy copied each inbound request's `_meta` onto its backend request verbatim, including the reserved connection envelope: `io.modelcontextprotocol/protocolVersion`, `clientInfo`, and `clientCapabilities`. Those keys describe the frontend connection, and the spec makes them per-request, sender-owned fields — on the backend hop the sender is the proxy. A modern backend session overwrites them with its own stamp, but a handshake-era backend receives them verbatim and rejects the call as a protocol violation (or, as with the upstream that surfaced this, hangs). The practical failure: a modern client could not call tools through a proxy whose upstream only speaks the handshake era.

`_forwardable_request_meta` now drops the three connection-owned keys at the proxy's single read of frontend request metadata, so the guarantee holds for every allowed backend client — `ProxyClient`, a plain `Client` from a custom factory, or a reused connected client. Progress tokens, tracing, task state, and vendor metadata pass through unchanged, and — because the spec's `_meta` model is uniform across request types — they now pass through for proxied resources, templates, and prompts too, which previously forwarded no request metadata at all. A modern backend stamps its own envelope; a handshake backend never sees one.

```python
# modern client → proxy → handshake-era backend
proxy = FastMCPProxy(client_factory=lambda: ProxyClient(backend, mode="legacy"))

async with Client(proxy, mode="auto") as client:
    # Previously rejected; the backend now sees only hop-safe metadata.
    await client.call_tool("list_projects", meta={"progressToken": "p1"})
```

Fixes #4769


zzstoatzz (approved):
small, weakly held nit but lgtm!

edit: whoops! new commit right as i reviewed, re-reviewing

zzstoatzz (approved):
lgtm after new commit!

PR #4772, opened 2026-08-05
Encrypt task context snapshots at rest

### LINKED ISSUE #4747: Allow task context snapshots to redact or encrypt credentials and HTTP headers
### Description

FastMCP Tasks 4.0.0b1 captures authenticated request context for background execution by serializing both:

- the complete access token via `AccessToken.model_dump_json()`
- all inbound HTTP headers via `get_http_headers(include_all=True)`

`TaskContextSnapshot.to_json()` then persists those fields as a JSON value in the configured Docket backend. With a distributed Redis/Valkey backend, bearer credentials and the `Authorization` header are therefore stored as plaintext backend values. A `rediss://` connection protects transport but does not change the stored representation.

This is particularly consequential for authentication schemes where the inbound bearer value is a long-lived refresh token, but it also expands the storage boundary for ordinary access tokens and arbitrary sensitive headers.

### Requested capability

Please provide a supported extension point that lets a server do one or more of the following before a task snapshot is persisted:

1. redact or allowlist captured HTTP headers;
2. encrypt/decrypt the serialized task-context payload with an application-supplied codec or key provider;
3. replace credentials with an opaque, short-lived server-side handle that the worker resolves after authorization.

The restore path should fail closed if protected context cannot be recovered. It would also help if the Tasks documentation explicitly described the credential-at-rest implications of distributed Docket backends.

### Current application mitigation

Downstream, we are failing startup for authenticated HTTP plus non-`memory://` Docket unless an operator explicitly opts into the risk. That is intentionally conservative, but it prevents safe distributed background workers until FastMCP exposes a protected snapshot mechanism.

### Version

- FastMCP / fastmcp-tasks: 4.0.0b1
- Docket (pydocket): 0.20.1

Related downstream work: https://github.com/KuudoAI/amazon_ads/issues/10

chrisguidry:
A background task runs long after the request that submitted it has ended, but it still needs to know who asked for the work. FastMCP captures that identity at submission time in a task context snapshot: the caller's full `AccessToken` and every inbound HTTP header, including `Authorization`. That snapshot goes to the Docket backend for the task's TTL. With Redis or Valkey it is stored as plaintext JSON, so anyone who can read the backend can read the tokens. A `rediss://` URL encrypts the connection, not the stored value.

Set `FASTMCP_TASKS_ENCRYPTION_KEY` and the snapshot becomes a Fernet token instead:

```bash
export FASTMCP_TASKS_ENCRYPTION_KEY=$(python -c "import secrets; print(secrets.token_urlsafe(32))")
```

Every server and worker on the queue must set the same key, because the process that restores a snapshot is rarely the one that captured it.

Restore fails closed. A worker that cannot decrypt a snapshot fails the task rather than run the tool with no caller: running an authorization-sensitive tool as an anonymous user is worse than not running it. So tasks submitted before the key was set fail once a worker with the key picks them up, and rotating a key does the same to tasks in flight. Drain the queue first.

Three decisions a reviewer would otherwise have to reconstruct. The key is scoped to the tasks package, on a new `TasksSettings` class, because it protects task snapshots and nothing else; other task-owned state can later derive its own key from it under a different salt. The codec sits at the serialization boundary, so Docket stays the only Redis client and `memory://` and cluster URLs work unchanged. Header capture is untouched, because an allowlist would silently break a tool that reads an unlisted header.

This covers the snapshot only. Tool arguments and mid-task input answers are still plaintext.

Closes #4747

🤖 Generated with Claude Code

<details>
<summary>Session context</summary>

Started from issue #4747, reported by @tspicer, who is blocking distributed background workers downstream until FastMCP offers a protected snapshot mechanism.

The issue asked for three possible extension points: header redaction or allowlisting, an encryption codec, and opaque short-lived credential handles. @chrisguidry's comments on the issue pointed at reusing `FernetEncryptionWrapper` and layering encryption into a `py-key-value` read/write pipeline, and named the open question as whether to add a tasks-scoped key setting or one global one.

Four design questions were settled before implementation:

- **Key setting**: initially one global `FASTMCP_ENCRYPTION_KEY` on `fastmcp.settings`; maintainer review rescoped it to `FASTMCP_TASKS_ENCRYPTION_KEY` in `fastmcp-tasks`, since the key protects task snapshots, not all FastMCP data at rest.
- **Scope**: the snapshot only. Encrypting `:args`, `:input_responses`, `:request_state`, and the outstanding `:input:N:requests` was considered and deferred.
- **Mechanism**: a Fernet codec at the serialization boundary, over routing the writes through a `py-key-value` store wrapped in `FernetEncryptionWrapper`. The store route needs its own Redis URL and connection, duplicating what Docket already owns, and `memory://` would need a separate store choice.
- **Headers**: unchanged. An allowlist and a redaction hook were both considered and dropped.

`cryptography` is declared by `fastmcp-tasks`, which owns the codec. `fastmcp-slim` still receives it transitively through authlib; whether slim should declare it for its own auth imports is a separate packaging question.

Tests were written first and confirmed red before any implementation.

Two things reviewers should know:

1. **No real Redis was exercised.** The whole task suite runs on `memory://`, which is `burner-redis` speaking the Redis protocol in-process. The codec sits above the storage layer, so the payload is a string either way, but the distributed path is not covered by test.
2. **The `uv.lock` diff was hand-applied.** A plain `uv sync` rewrote `exclude-newer` to a null value and churned about 30 unrelated marker lines. The two lines here are the `cryptography` entry alone, and `uv lock --check` passes.

Verification: full suite 7050 passed, 5 skipped, 16 xfailed. `prek run --all-files` clean. The 26 loq violations it reports are pre-existing and unenforced; no file in this change appears among them.

Separately, a pre-existing bug was found and left alone: the `:snapshot` key's TTL is never refreshed and never deleted. `refresh_current_leg_ttl` in `input_store.py` slides only the leg-pointer keys, so a long-parked multi-leg task can outlive its snapshot, after which restore returns silently and the leg runs unauthenticated.

</details>

### DIFF
--- a/fastmcp_tasks/fastmcp_tasks/context.py
+++ b/fastmcp_tasks/fastmcp_tasks/context.py
@@ -297,10 +297,11 @@ async def restore_task_snapshot(key: str = TaskKey()) -> None:
     still runs, and sync helpers return ``None`` as they would have before
     the snapshot was captured.
 
-    A snapshot that will not decrypt is the one exception. When the deployment
-    configured an encryption key, an unreadable snapshot means the caller cannot
-    be recovered, so the task fails rather than running under no identity at all
-    (#4747).
+    Configuring an encryption key changes that contract. The operator asked for
+    fail-closed protection, so any failure to retrieve, decrypt, parse, or apply
+    the snapshot, including a snapshot that is simply missing, escapes this
+    dependency and fails the task, rather than running the tool without the
+    submitting caller's identity (#4747).
     """
     try:
         parts = parse_task_key(key)
@@ -311,6 +312,11 @@ async def restore_task_snapshot(key: str = TaskKey()) -> None:
     from fastmcp.server.dependencies import get_server
     from fastmcp_tasks.dependencies import _current_docket
 
+    # Resolved before anything can fail: a misconfigured key (e.g. an empty
+    # string) raises here and fails the task, and the except blocks below read
+    # it to pick between the fail-open and fail-closed contracts.
+    codec = snapshot_codec()
+
     try:
         docket = get_server()._docket
     except RuntimeError:
@@ -318,6 +324,11 @@ async def restore_task_snapshot(key: str = TaskKey()) -> None:
     if docket is None:
         docket = _current_docket.get()
     if docket is None:
+        if codec is not None:
+            raise RuntimeError(
+                "No Docket backend is available to retrieve the protected "
+                "task snapshot, so the submitting caller cannot be recovered."
+            )
         return
 
     task_scope = parts["task_scope"]
@@ -328,8 +339,12 @@ async def restore_task_snapshot(key: str = TaskKey()) -> None:
                 docket.key(f"{task_redis_prefix(task_scope)}:{task_id}:snapshot")
             )
         if raw is None:
-            return
-        codec = snapshot_codec()
+            if codec is None:
+                return
+            raise RuntimeError(
+                "The task's context snapshot is missing (its TTL may have "
+                "expired), so the submitting caller cannot be recovered."
+            )
         if codec is not None:
             raw = codec.decode(raw)
         snapshot = TaskContextSnapshot.from_json(raw)
@@ -350,6 +365,15 @@ async def restore_task_snapshot(key: str = TaskKey()) -> None:
         )
         raise
     except Exception:
+        if codec is not None:
+            _logger.error(
+                "Failed to restore the protected task snapshot for %s. The task "
+                "will fail rather than run without the submitting caller's "
+                "identity.",
+                key,
+                exc_info=True,
+            )
+            raise
         _logger.warning("Failed to restore task snapshot for %s", key, exc_info=True)
 
 
--- a/fastmcp_tasks/fastmcp_tasks/context.py
+++ b/fastmcp_tasks/fastmcp_tasks/context.py
@@ -251,7 +251,7 @@ class TaskContextSnapshot:
     ) -> None:
         """Store this snapshot as a single Redis key.
 
-        The stored value is encrypted when a ``FASTMCP_ENCRYPTION_KEY`` is
+        The stored value is encrypted when a ``FASTMCP_TASKS_ENCRYPTION_KEY`` is
         configured: this payload carries the caller's bearer token and headers,
         and a distributed backend keeps it where the backend's operators can
         read it (#4747).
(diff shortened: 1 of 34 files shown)


fastmcp_tasks/fastmcp_tasks/context.py
```diff
@@ -313,12 +329,26 @@ async def restore_task_snapshot(key: str = TaskKey()) -> None:
             )
         if raw is None:
```
jlowin, line 358: With `FASTMCP_ENCRYPTION_KEY` set, this still fails open when the snapshot is missing. `raw is None` returns, and the broad `except Exception` below also swallows JSON, schema, and context-restore failures. In both cases Docket proceeds to run the tool without the submitting caller.

I reproduced this end to end by skipping the snapshot write and separately by making `from_json()` fail after successful decryption; both tasks completed and returned `no-token`. Missing snapshots are already reachable through the TTL issue noted in the PR description.

Once encryption is configured, any failure to retrieve, decrypt, parse, or apply the protected snapshot needs to escape the dependency and fail the task. Limiting that behavior to `SnapshotDecryptionError` does not provide the stated fail-closed contract.

fastmcp_tasks/fastmcp_tasks/encryption.py
```diff
@@ -0,0 +1,109 @@
...
+    HKDF; a short passphrase goes through PBKDF2 instead, which is slower but
+    survives the weaker input.
+    """
+
+    def __init__(self, material: str) -> None:
+        from cryptography.fernet import Fernet
+
+        from fastmcp.server.auth.jwt_issuer import derive_jwt_key
+
+        if len(material) >= _MINIMUM_HIGH_ENTROPY_LENGTH:
```
jlowin: Length is not evidence that this is high-entropy key material. A value like `correct horse battery staple` takes the HKDF branch and can be tested cheaply offline by exactly the backend reader this feature protects against.

The OAuth code referenced here does not use its 12-character threshold this way: it only warns below 12, then PBKDF2-derives every user-supplied string. Treat this environment-provided material as low entropy by default, or require and validate an explicitly encoded random key. Empty values also need to be rejected; `FASTMCP_ENCRYPTION_KEY=''` currently creates a working codec with a universally reproducible key.

jlowin (approved):

jlowin:
actually checking one thing one sec

jlowin (changes requested):
Thanks, the snapshot encryption behavior looks right, but the configuration is scoped too broadly.

`FASTMCP_ENCRYPTION_KEY` establishes a framework-wide contract even though this PR only protects task context snapshots. It does not protect task arguments or input responses, and other FastMCP storage has its own key material. Please move this setting into `fastmcp-tasks` and expose it as `FASTMCP_TASKS_ENCRYPTION_KEY`. This leaves room to use domain-separated derived keys for other task-owned state later without implying that one setting encrypts all FastMCP data at rest.

`fastmcp-tasks` should also declare `cryptography` directly because it imports and owns this codec. `fastmcp-slim[server]` currently receives `cryptography` through Authlib and has existing auth imports of its own, so whether slim should also declare it directly is a separate packaging question. This task feature should not be the reason the dependency is owned there.

The codec can stay at the snapshot serialization boundary. Happy to re-review once the setting and dependency ownership are scoped to the tasks package.

chrisguidry:
Got it!  Good catch, I kinda forgot that it's a separate package now :D

jlowin (approved):

PR #4776, opened 2026-08-05
Forward proxy server metadata across protocol eras

### LINKED ISSUE #4775: Support forwarding proxy server metadata
FastMCP proxies currently forward upstream instructions through separate legacy `initialize` and modern `server/discover` paths, while other useful upstream server metadata is dropped. Using `ProxyProvider` directly also provides no equivalent behavior, which prevents controlled gateways from preserving backend identity and namespaced metadata.

Add a composable proxy metadata middleware shared by `FastMCPProxy` and direct `FastMCP` + `ProxyProvider` gateways. It should normalize metadata from either protocol era and forward optional instructions, server identity, and namespaced `_meta`, while leaving protocol versions, capabilities, cache policy, result type, and unknown top-level claims owned by the frontend gateway. Discovery needs a typed middleware result hook so this does not depend on raw dictionary patching.

jakekaplan:
### Before

Proxy gateways forwarded upstream instructions through two unrelated paths: `ProxyInitializeMiddleware` for legacy clients and a replacement `server/discover` handler for modern clients. They behaved differently when the backend was unavailable, applied only to `FastMCPProxy`, and discarded backend identity and namespaced `_meta`.

### Now

A typed `on_discover` hook lets one composable `ProxyMetadataMiddleware` serve both protocol eras, whether the gateway uses `FastMCPProxy` or `ProxyProvider` directly. The middleware uses the backend client's existing lifecycle and negotiation mode rather than cloning or reconfiguring it. The gateway builds its normal response first, then merges optional backend instructions and `_meta`; gateway instructions and `_meta` values win, operational claims are never copied, and `identity="upstream"` separately opts into the backend identity when available. If the backend is unavailable, the client can still connect to the gateway; subsequent component failures follow the gateway's existing `provider_error_strategy`.

```python
from fastmcp import FastMCP
from fastmcp.server.providers.proxy import ProxyMetadataMiddleware, ProxyProvider

provider = ProxyProvider(create_backend_client)
gateway = FastMCP(
    "Gateway",
    providers=[provider],
    middleware=[ProxyMetadataMiddleware(provider, identity="upstream")],
)
```

Fixes #4775

🤖 Generated with OpenAI Codex


jakekaplan:
Sorry still need to cleanup some regressions here

jlowin:
Nice cleanup. One lifecycle edge I ran into while reviewing: when `ProxyMetadataMiddleware` probes a pinned-modern `StatefulProxyClient`, `client.new()` preserves the stateful subclass. That clone's `__aexit__` is intentionally a no-op, but the clone is not registered for the frontend connection's stateful cleanup, so its backend session remains connected after the frontend disconnects.

I reproduced this with the registered client closed and the metadata probe still connected after teardown. The disposable probe should be force-disconnected in a `finally` block, while preserving the current fail-open behavior if cleanup repeats the original transport error. I have a small regression test and fix locally and can push it once the branch settles.

jlowin:
"I" did all those things, yes

jakekaplan:
I ended up removing the disposable probe entirely instead of adding cleanup around it. The clone only existed to force an `auto` negotiation so we could read real metadata from a pinned client, but we can send a non-adopting `server/discover` request on the factory-provided client instead.

That feels cleaner because the middleware no longer creates a second session, changes the configured mode, or takes responsibility for a client lifecycle it doesn’t own. Stateful clients now naturally remain attached to their registered connection cleanup, while connected and ordinary clients follow the same path.

jlowin (approved):

PR #4793, opened 2026-08-09
fix(openapi): extract parameter-level example and examples

### LINKED ISSUE #4792: OpenAPI parameter-level `example` is dropped from generated tool input schemas
### What happened?

When an OpenAPI document declares an example on the parameter object
(`parameters[].example`), that value never reaches the generated MCP tool input
schema. The same parameter's `default` does reach it, so a spec that carries
both ends up exposing the default and silently losing the example.

There is a second, worse case. If a parameter carries `parameters[].example`
*and* its schema carries an example, the schema-level value is the one that
shows up in the tool schema — that is, the value the OpenAPI specification says
should be overridden wins, and the overriding one is discarded.

Expected: `startYearMonth` exposes `201601`, and `override` exposes
`PARAM_LEVEL`.
Actual: `startYearMonth` has no example at all, and `override` exposes
`SCHEMA_LEVEL`.

This matters for hand-curated specs: an example value is often what tells a
model that a date parameter wants `201601` rather than `2016-01-01`, and
curating it at the parameter level — the place the spec defines for it —
currently has no effect.

Reproduced on 3.4.6 and 2.14.7 with the script below, and on 4.0.0b2 with the
same script using `httpx2` instead of `httpx`.

Related: #1876, where parameter descriptions were dropped in the same
conversion path.

### Example Code

```python
# ASCII-only repro: parameter-level `example` is dropped from the MCP tool input schema.
import asyncio
import json

import httpx
from fastmcp import FastMCP

SPEC = {
    "openapi": "3.1.0",
    "info": {"title": "Example Drop Demo", "version": "1.0.0"},
    "paths": {
        "/trade": {
            "get": {
                "operationId": "get_trade",
                "summary": "Monthly trade totals",
                "parameters": [
                    {
                        "name": "startYearMonth",
                        "in": "query",
                        "required": True,
                        "description": "Start month, YYYYMM.",
                        "example": "201601",
                        "schema": {"type": "string"},
                    },
                    {
                        "name": "pageNo",
                        "in": "query",
                        "required": False,
                        "description": "Page number.",
                        "example": 3,
                        "schema": {"type": "integer", "default": 1},
                    },
                    {
                        "name": "override",
                        "in": "query",
                        "required": False,
                        "description": "Parameter example must override the schema example.",
                        "example": "PARAM_LEVEL",
                        "schema": {"type": "string", "examples": ["SCHEMA_LEVEL"]},
                    },
                ],
                "responses": {
                    "200": {
                        "description": "OK",
                        "content": {"application/json": {"schema": {"type": "object"}}},
                    }
                },
            }
        }
    },
}


async def main() -> None:
    async with httpx.AsyncClient(base_url="https://api.example.com") as api_client:
        server = FastMCP.from_openapi(openapi_spec=SPEC, client=api_client, name="demo")
        async with Client(server) as mcp_client:
            tools = await mcp_client.list_tools()
            tool = next(t for t in tools if t.name == "get_trade")
            props = tool.inputSchema["properties"]
            print(json.dumps(props, indent=2, sort_keys=True))
            print("---")
            print("startYearMonth has example :", "example" in props["startYearMonth"] or "examples" in props["startYearMonth"])
            print("pageNo has example        :", "example" in props["pageNo"] or "examples" in props["pageNo"])
            print("pageNo default preserved  :", props["pageNo"].get("default"))
            print("override surfaced value   :", props["override"].get("examples") or props["override"].get("example"))


asyncio.run(main())
```

Output (identical on 3.4.6, 2.14.7, and 4.0.0b2):

```
{
  "override": {
    "description": "Parameter example must override the schema example.",
    "examples": [
      "SCHEMA_LEVEL"
    ],
    "type": "string"
  },
  "pageNo": {
    "default": 1,
    "description": "Page number.",
    "type": "integer"
  },
  "startYearMonth": {
    "description": "Start month, YYYYMM.",
    "type": "string"
  }
}
---
startYearMonth has example : False
pageNo has example        : False
pageNo default preserved  : 1
override surfaced value   : ['SCHEMA_LEVEL']
```

### Version Information

```
FastMCP version:                                                          3.4.6
MCP version:                                                             1.29.0
Python version:                                                          3.11.9
Platform:                                             Windows-10-10.0.19045-SP0
```

Also reproduced on 2.14.7 (httpx 0.28.1) and on 4.0.0b2 (httpx2 variant of the
same script).

doneman536:
Preserve parameter-level example and examples in OpenAPI specs, resolve references, and dump Pydantic models to JSON-serializable dictionaries.

issue #4792

## Description

When an OpenAPI document declares `example` or `examples` at the parameter level (`parameters[].example`), those values were dropped during conversion to the generated MCP tool input schema. If a parameter carried both a parameter-level example and a schema-level example or default, the parameter-level example was silently lost.
This change extracts parameter-level `example` and `examples` during OpenAPI parsing, ensuring parameter-level declarations take precedence over schema-level examples while preserving defaults. References within `examples` are resolved and Pydantic models are serialized to JSON-compatible dictionaries.

Closes #4792

## Contribution type

- [*] Bug fix (simple, well-scoped fix for a clearly broken behavior)

## Checklist

- [*] This PR addresses an existing issue (or fixes a self-evident bug)
- [*] I have read [CONTRIBUTING.md](../CONTRIBUTING.md)
- [*] I have added tests that cover my changes
- [*] I have run `uv run prek run --all-files` and all checks pass
- [*] I have self-reviewed my changes
- [*] If I used an LLM, it followed the repo's contributing conventions (not generic output)


doneman536:
Can u verify once again

doneman536:
Reverted the uv.lock and Convert parameter examples to JSON Schema values is already done . Please check it once

jlowin (changes requested):
Thanks. This needs to be reduced to the contract reported in #4792. The issue is specifically about singular `Parameter.example`, which is already a raw value and does not require reference resolution. In `_extract_parameters`, copy that value into the generated property schema, remove the schema-level `example` or `examples` it overrides, and preserve `default`.

Supporting plural `examples` introduces a separate contract involving named Example Objects, references, and `externalValue`. The current implementation also invents precedence between `example` and `examples`, even though OpenAPI defines them as mutually exclusive, and treats an Example Object without `value` as though the wrapper object itself were the example value.

Please remove the plural and reference handling and replace the 213-line fixture with focused tests for singular propagation, default preservation, and a singular parameter example overriding schema-level examples. The current tests no longer cover that last case from the issue. This should be roughly 5 to 10 production lines, not 52.

doneman536:
@jlowin

Thanks for the clarification , I've refactored the PR to focus strictly on the singular `Parameter.example` contract reported in #4792 : 

Simplified implementation: Reduced `_extract_parameters` down to 5 lines for singular `Parameter.example`, preserving `default` and clearing schema-level overrides.

Removed extra complexity: Stripped out plural `examples` processing and `$ref` resolution.

Focused tests: Replaced test fixtures with targeted tests for singular propagation, default preservation, and overriding schema-level examples.

jlowin (approved):

PR #4802, opened 2026-08-10
Support CallArgument and Depends bindings from uncalled-for 0.4.0

chrisguidry:
uncalled-for 0.4.0 adds explicit argument references: `CallArgument()` lets a dependency factory read an argument of the function it serves, and `Depends(factory, **bindings)` supplies factory arguments at the declaration site (https://github.com/chrisguidry/uncalled-for/pull/12).

FastMCP's resolver now opens a `frame_scope()` around dependency resolution, with the sanitized user arguments as the frame's provided values. A `CallArgument` can reference a tool call's public parameters, but a caller-supplied value for a dependency parameter name is still stripped before resolution. `CallArgument` and `CycleError` are re-exported from `fastmcp.dependencies`, and the dependency-injection docs cover both features.

Changes:

- Raise the `uncalled-for` floor to 0.4.0 and exempt it from the one-week dependency cooldown (0.4.0 shipped today).
- Resolve each dependency parameter through the call's resolution frame. With no `CallArgument` in play, the frame resolves exactly what the old loop resolved.
- Let `CycleError` propagate unwrapped, since it already names the cyclic reference path.
- Add tests for the new surface through `call_tool`, `render_prompt`, and `resolve_dependencies`, including the argument-strip security check.
- Extend `docs/servers/dependency-injection.mdx` with Call Arguments and Bindings sections.


jlowin:
one thing i noticed with the Windows split: `burner-redis<0.1.7` forces pydocket 0.20.1, but 0.20.1's task resolver doesn't establish an uncalled-for frame. That means `CallArgument` works for foreground tools on Windows but fails for `task=True` tools.

I reproduced this on the exact PR head. The same background task completed with pydocket 0.24.1, then failed after switching to pydocket 0.20.1 with `RuntimeError: A resolution frame is required here`.

I think this either needs pydocket 0.24.1 to be unblocked on Windows before landing, or the Windows background-task limitation should be explicit and covered by a test so the platform difference is intentional.

chrisguidry:
:doh: @jlowin thanks for catching that, I didn't notice we had that Windows-specific pin.  I've got an agent on the case to fix `burner-redis`

fastmcp_tasks/pyproject.toml
```diff
@@ -56,16 +56,13 @@ dependencies = [
...
-    # no Python traceback) running the memory:// backend under pytest-xdist —
-    # reproduced on GitHub Actions windows-latest, confirmed absent on
-    # macOS/Linux with the same versions (full suite green there under the
-    # identical upgraded dependencies). pydocket only floors it at >=0.1.6, so
-    # capping pydocket alone is not enough: a resolver is free to pick the
-    # newest burner-redis satisfying that floor regardless. Pin burner-redis
-    # directly on Windows only (which in turn caps pydocket to <0.20.2 there,
-    # the last release that doesn't itself require burner-redis>=0.1.7) until
-    # upstream ships a fix — other platforms are unaffected and stay unpinned.
-    "burner-redis<0.1.7; sys_platform == 'win32'",
```
chrisguidry, line 70: @jlowin looks like it was just an artifact of history for the Rust portion of burner-redis and we're good to unpin it!

zzstoatzz (approved):
lgtm!

PR #4818, opened 2026-08-11
Keep earlier consent CSRF tokens valid within a transaction

### LINKED ISSUE #4817: Second render of a consent page invalidates the first, breaking the OAuth flow
Loading the consent page twice for one transaction breaks the flow. The second render replaces the transaction's `csrf_token`, so the form from the first render is already dead when the user clicks Approve. They get "Invalid or expired consent token" on a transaction that hasn't expired, and retrying doesn't help because the retry does the same thing again.

We hit this in production behind an `OAuthProxy` (Google upstream, Claude.ai as the client). One user could never connect: 19 attempts over about six and a half hours, every one failing. Load balancer logs made the pattern obvious once we looked:

| GETs before the POST | POST result | Transactions |
| --- | --- | --- |
| 1 | 302 | 96 |
| 2 | 400 | 19 |

No exceptions in either direction over 30 days of logs. Every duplicate GET failed, every single GET succeeded. The duplicate came from that one user's browser (same IP, two GETs about 400ms apart, no referer on either) — a preload, an extension re-fetching the URL, something along those lines. Everyone else was fine, which is why it read as "one user's machine is broken" rather than a server bug.

### Reproduction

```python
txn_id, _ = await _start_flow(proxy, "client", "http://localhost:9090/callback")
app = Starlette(routes=proxy.get_routes())

with TestClient(app, base_url="https://myserver.com") as client:
    first = _extract_csrf(client.get(f"/consent?txn_id={txn_id}").text)
    client.get(f"/consent?txn_id={txn_id}")  # reload, preload, re-fetch

    # The user submits the page they had open.
    r = client.post(
        "/consent",
        data={"action": "approve", "txn_id": txn_id, "csrf_token": first},
        follow_redirects=False,
    )

assert r.status_code == 302  # actual: 400, "Invalid or expired consent token"
```

Expected: the form the user is looking at can be submitted.
Actual: 400, and the message says the token expired when it didn't.

### One thing worth flagging

`MCP_CONSENT_STATE` already handles this on the client side — `_set_consent_cookie` merges each new token into a list with the comment "supports concurrent flows". The transaction only stores one token, so the two halves of the double-submit check disagree about whether more than one render is allowed. The cookie says yes, the transaction says no.

I have a fix for this and would be glad to open a PR if you want it. It keeps a bounded list of issued tokens on the transaction so the cookie and the server agree, while still minting a fresh token per render so a token stays unique to the browser that received it and the double-submit check keeps its meaning. Happy to leave it to you instead if you'd rather own the approach — the report is the part I wanted to get to you either way.

Version: reproduced on 3.4.4 (what we run) and present unchanged on `main` through 4.0.0b2.

trevhud:
Fixes #4817.

Each render of the consent page overwrote the transaction's `csrf_token`. Load the page twice (a reload, a browser preload, an extension re-fetching the URL) and the form from the first render is already dead. The user clicks Approve, gets "Invalid or expired consent token" on a transaction that hasn't expired, and retrying just repeats it.

### What changed since the review

The first version kept a bounded list of tokens on the transaction, which has the lost-update race you described: two renders read the same transaction, each append, and the second write drops the first. The cookie had the same shape of problem, since overlapping responses each build their `Set-Cookie` from the same inbound `Cookie` header, so one entry always loses.

Both sides are independently addressable now:

- Each token is stored under its own key — `ConsentCSRFToken`, keyed by `_hash_token`, in a `mcp-consent-csrf-tokens` collection. A render writes only its own key, so nothing it writes is being written by another render. `_submit_consent` looks the token up instead of scanning a list.
- Each token gets its own cookie, `MCP_CONSENT_STATE_<digest>`, rather than one list shared by all of them. Separate names never collide, so the order overlapping responses arrive in stops mattering.

Every render still mints its own token, so the property `TestCSRFDoubleSubmit` defends is intact: the only token matching a stored one lives in the browser that received it.

### Scoping and bounds

Approve and deny expire only the cookies carrying the completed transaction's tokens, so a second flow the browser still has open survives. Previously the entire list went. The submitted token is deleted server-side too, so it can't be replayed.

The cookie count is bounded at `_MAX_CSRF_TOKENS`, oldest first, using an issued-at carried inside the cookie so eviction needs no shared server-side bookkeeping.

Server-side I did not add a count bound. Reading the entries back to count them is the read-modify-write this change exists to remove, and the 15-minute TTL already bounds them. Say the word if you'd rather have a hard cap and I'll find a way to do it without reintroducing the race.

### Tests

- `test_overlapping_renders_both_submittable` — two GETs held at the same point inside the handler by a gate, released together, both responses applied to one cookie jar, then each form submitted in turn (parametrized over which one). Fails on the previous version.
- `test_completing_one_flow_leaves_another_submittable` and `test_completing_a_flow_removes_only_its_own_state` — two pending flows, completing one leaves the other alone. Both fail on the previous version.
- `test_consent_state_cookies_are_bounded` — covers the limit. Worth saying plainly: this one passes on the previous version too. It locks in the new representation rather than catching the old bug.

`asyncio.Barrier` would have been the obvious gate, but it's 3.11+ and the floor here is 3.10, hence the hand-rolled one.

Full suite green (`uv run pytest -n auto`, 7181 passed), as is `uv run prek run --all-files`. No JWT flakes this time, for whatever that's worth.

### Back-compat

A consent page rendered by the previous version is still submittable after an upgrade. With no record under the token's key, `_submit_consent` falls back to `csrf_token` on the transaction and to the old flat cookie. Both are read, never written.

### Context

Found this in production behind an `OAuthProxy` in front of Google, with Claude.ai as the client. One user could never connect: every attempt preceded by a duplicate GET, every one failing. The numbers are in the issue.

### DIFF
--- a/fastmcp_slim/fastmcp/server/auth/oauth_proxy/consent.py
+++ b/fastmcp_slim/fastmcp/server/auth/oauth_proxy/consent.py
@@ -20,7 +20,7 @@ from pydantic import AnyUrl
 from starlette.requests import Request
 from starlette.responses import HTMLResponse, RedirectResponse
 
-from fastmcp.server.auth.oauth_proxy.models import ProxyDCRClient
+from fastmcp.server.auth.oauth_proxy.models import OAuthTransaction, ProxyDCRClient
 from fastmcp.server.auth.oauth_proxy.ui import create_consent_html
 from fastmcp.server.auth.redirect_validation import (
     build_client_redirect,
@@ -36,7 +36,25 @@ if TYPE_CHECKING:
 # Keeps the Cookie header bounded to avoid hitting reverse proxy header limits.
 _MAX_REMEMBERED_CLIENTS = 25
 
+# Maximum number of concurrently valid CSRF tokens per transaction. A handful
+# of renders is normal (reload, preload, a re-fetch); anything beyond this is
+# not a flow a user is going to complete, and the bound keeps both the stored
+# transaction and the MCP_CONSENT_STATE cookie from growing without limit.
+_MAX_CSRF_TOKENS = 10
+
 logger = get_logger(__name__)
+
+
+def _valid_csrf_tokens(txn_model: OAuthTransaction) -> list[str]:
+    """CSRF tokens currently accepted for a transaction, oldest first.
+
+    Falls back to the single `csrf_token` for transactions created before
+    `csrf_tokens` existed, so a consent flow already in progress survives an
+    upgrade instead of failing at the point the user clicks Approve.
+    """
+    if txn_model.csrf_tokens:
+        return txn_model.csrf_tokens
+    return [txn_model.csrf_token] if txn_model.csrf_token else []
 
 
 class ConsentMixin:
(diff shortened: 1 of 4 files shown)


jlowin (changes requested):
Thanks. Keeping a fresh token for each render is the right security model here; reusing one transaction-wide token would weaken the double-submit protection.

The transaction update still has a lost-update race. Two overlapping GETs can both read the same transaction, append different tokens, and then overwrite each other in `_transaction_store`, leaving only the last writer's token valid. `AsyncKeyValue` does not expose compare-and-swap, and the new test performs the GETs sequentially, so it does not exercise this path. The browser-side list has the same problem when overlapping responses are built from the same incoming cookie. Please use independently addressable token state, or another mechanism that works across processes sharing a backend, and add a synchronized overlapping-GET regression that proves forms from both responses remain submittable after both responses are applied.

The bound also currently applies only to `OAuthTransaction.csrf_tokens`. `_show_consent_page` still appends every token to one flat `MCP_CONSENT_STATE` cookie without applying `_MAX_CSRF_TOKENS`, and approve or deny clears that entire list, including tokens belonging to other pending transactions. Please make the browser-side state bounded and transaction-scoped, remove only the completed transaction's state, and cover both the limit and two simultaneously pending flows.

Happy to re-review once these are addressed.

zzstoatzz (approved):
this lgtm! will cc @jlowin as the original reviewer though

jlowin (approved):

PR #4835, opened 2026-08-14
fix(auth): cache OIDC discovery configuration

### LINKED ISSUE #4837: OIDC discovery is fetched repeatedly for identical configuration
### What happened?

Every call to `OIDCConfiguration.get_oidc_configuration()` hits the discovery URL again, even when the URL and options are unchanged. This has caused Auth0 throttling for a user in #4054.

I expected the second call to reuse the first successful result. It currently makes another request.

### Example Code

```python
from unittest.mock import MagicMock, patch

from httpx2 import Response
from pydantic import AnyHttpUrl

from fastmcp.server.auth.oidc_proxy import OIDCConfiguration

url = AnyHttpUrl("https://example.com/.well-known/openid-configuration")
document = {
    "issuer": "https://example.com",
    "authorization_endpoint": "https://example.com/authorize",
    "token_endpoint": "https://example.com/token",
    "jwks_uri": "https://example.com/jwks",
    "response_types_supported": ["code"],
    "subject_types_supported": ["public"],
    "id_token_signing_alg_values_supported": ["RS256"],
}

with patch("httpx2.get") as get:
    response = MagicMock(spec=Response)
    response.json.return_value = document
    get.return_value = response

    for _ in range(2):
        OIDCConfiguration.get_oidc_configuration(
            config_url=url,
            strict=True,
            timeout_seconds=10,
        )

    print(get.call_count)  # 2
```

### Version Information

FastMCP `main` at `ba283ddb4c46789493e6fed5e2adc59c1ea76255`; Python 3.12.10; Windows.

sui-ni2:
## Summary

- Cache successfully validated OIDC discovery metadata in-process to avoid repeated requests for identical provider configuration.
- Bound the cache to 128 entries with a 5-minute TTL so discovery metadata does not remain stale indefinitely.
- Cache serialized configuration snapshots and revalidate a copied snapshot on hits, so callers never share the same mutable `OIDCConfiguration` instance.
- Do not cache discovery or validation failures.
- Clear the process-local discovery cache between auth tests to prevent cross-test state leakage.

## Regression coverage

Coverage verifies that:

- identical discovery calls issue one upstream request;
- expired entries are fetched again;
- mutating one returned configuration does not affect later callers;
- failed discovery is retried rather than cached.

Fixes #4837.
Refs #4054 (Bug 2).

### DIFF
--- a/fastmcp_slim/fastmcp/server/auth/oidc_proxy.py
+++ b/fastmcp_slim/fastmcp/server/auth/oidc_proxy.py
@@ -10,6 +10,7 @@ This implementation is based on:
 """
 
 from collections.abc import Sequence
+from functools import lru_cache
 from typing import Any, Literal
 
 import httpx2
@@ -148,6 +149,7 @@ class OIDCConfiguration(BaseModel):
         return self
 
     @classmethod
+    @lru_cache(maxsize=128)
     def get_oidc_configuration(
         cls, config_url: AnyHttpUrl, *, strict: bool | None, timeout_seconds: int | None
     ) -> Self:
--- a/tests/server/auth/test_oidc_proxy.py
+++ b/tests/server/auth/test_oidc_proxy.py
@@ -384,6 +384,36 @@ def validate_get_oidc_configuration(oidc_configuration, strict, timeout_seconds)
 class TestGetOIDCConfiguration:
     """Tests for getting OIDC configuration."""
 
+    @pytest.fixture(autouse=True)
+    def clear_configuration_cache(self):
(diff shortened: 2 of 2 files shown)


sui-ni2:
@codex address that feedback

Please keep this strictly scoped to #4837. Address both review findings: (1) the discovery cache must have a bounded lifetime rather than indefinite process lifetime, and (2) repeated callers must not share the same mutable `OIDCConfiguration` instance. Prefer a small stdlib-only implementation with regression tests for cache reuse, expiry/refetch, and caller isolation. Do not broaden into provider refactors or unrelated auth changes.

PR #4863, opened 2026-08-19
fix(auth): cache OIDC discovery configuration

### LINKED ISSUE #4837: OIDC discovery is fetched repeatedly for identical configuration

sui-ni2:
## Summary

- cache successful OIDC discovery documents in-process to avoid repeated identical upstream fetches;
- bound the cache to 128 entries with a 5-minute TTL;
- cache raw discovery documents and deep-copy/revalidate them per caller so mutable `OIDCConfiguration` instances are never shared;
- keep `strict` as a per-call validation option rather than part of the cache key;
- do not cache failed discovery or failed first-call validation.

## Regression coverage

Adds focused tests for:

- repeated-call cache reuse;
- expiry and refetch;
- caller mutation isolation;
- failed discovery retry;
- per-call `strict` validation without an unnecessary refetch.

Fixes #4837.

### DIFF
--- a/fastmcp_slim/fastmcp/server/auth/oidc_proxy.py
+++ b/fastmcp_slim/fastmcp/server/auth/oidc_proxy.py
@@ -9,8 +9,11 @@ This implementation is based on:
     OAuth 2.0 Authorization Server Metadata - https://datatracker.ietf.org/doc/html/rfc8414
 """
 
+from collections import OrderedDict
 from collections.abc import Sequence
-from typing import Any, Literal
+from copy import deepcopy
+from time import monotonic
+from typing import Any, ClassVar, Literal
 
 import httpx2
 from key_value.aio.protocols import AsyncKeyValue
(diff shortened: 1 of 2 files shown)


sui-ni2:
Closing this as a duplicate of #4861. #4861 already contains the reviewed #4837 fix, including the follow-up cache-key isolation change for `OIDCConfiguration` subclasses. Keeping a single PR avoids splitting review and CI state.

PR #4879, opened 2026-08-22
Add pluggable identity assertion replay storage

### LINKED ISSUE #4846: IdentityAssertion (SEP-990): jti replay cache is per-process — support a pluggable store, like KeyValueResponseCacheStore
### Description

As of `v4.0.0b3`, `IdentityAssertionValidator` keeps its jti replay cache in process memory:

```python
# fastmcp_slim/fastmcp/server/auth/identity_assertion.py
self._jti_cache: dict[str, float] = {}
self._jti_cache_max_size = 10000
```

The docstring is candid that this "is per-process, so replay protection is not shared across horizontally-scaled workers." In practice that means SEP-990's replay protection is defeated by exactly the deployments identity assertion is aimed at: any server running more than one replica (or any autoscaling platform — Cloud Run, ECS, Kubernetes HPA) will accept the same assertion once per instance. We verified this cross-instance replay in a two-process trial. The known-gaps page currently classifies multi-replica concerns as "deployment configuration rather than protocol gaps", but this one is a security property of the validator itself, not a deployment knob — there is no configuration that fixes it today.

### Proposed solution

Accept a pluggable KV store on `IdentityAssertionValidator`, following the precedent FastMCP already set with `KeyValueResponseCacheStore` (memory / Redis / filetree backends behind one interface). Something like:

```python
IdentityAssertionValidator(..., jti_store=SomeKeyValueStore(...))
```

with the current in-process dict as the default. The important part is that it's the *store interface* that's pluggable, not a hard dependency on any particular backend — deployments that already run Postgres (or any KV-capable store) can back replay state with what they have, without adding Redis to their stack. A TTL-honoring `setnx`-style primitive (set-if-absent with expiry) is all the validator needs for an atomic jti burn.

Raising this before 4.0 GA in case it has API-shape implications. Happy to provide the cross-instance repro details if useful.

jlowin:
Identity assertion replay state currently lives in each process, so the same ID-JAG can be accepted once per replica. This routes consumed JTI state through py-key-value's `AsyncKeyValue` contract, with a bounded `MemoryStore` default and any shared backend accepted through `IdentityAssertion(jti_store=...)`.

py-key-value 0.4.x does not yet expose a conditional insert. FastMCP serializes exchanges within one validator, but replicas sharing a store perform a non-atomic `get` followed by `put`; simultaneous exchanges on different replicas can race and both succeed. Sequential replay is rejected, store failures fail closed, and an atomic follow-up is proposed in [py-key-value#369](https://github.com/strawgate/py-key-value/pull/369).

```python
from key_value.aio.stores.redis import RedisStore

from fastmcp.server.auth import IdentityAssertion

identity_assertion = IdentityAssertion(
    trusted_issuers=["https://login.example.com"],
    jti_store=RedisStore(url="redis://localhost:6379"),
)
```

Closes #4846

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -157,6 +157,7 @@ Because the docs land *before* the tag exists, derive the entry from the maintai
 
 - Python ≥ 3.10 with full type annotations
 - Follow existing patterns and maintain consistency
+- **Use `py-key-value` for stateful framework features.** New caches, replay tracking, token or session persistence, and other pluggable state should accept the existing `AsyncKeyValue` abstraction instead of introducing a feature-specific storage protocol. Add a specialized interface only when the required semantics cannot be represented by `AsyncKeyValue`, and document that constraint explicitly.
 - **Prioritize readable, understandable code** - clarity over cleverness
 - Avoid obfuscated or confusing patterns even if they're shorter
 - Each feature needs corresponding tests
--- a/docs/servers/auth/oauth-proxy.mdx
+++ b/docs/servers/auth/oauth-proxy.mdx
@@ -758,8 +758,56 @@ identity_assertion=IdentityAssertion(
 
 The subject asserted in the ID-JAG flows into the normal FastMCP auth context. Tools read it through `get_access_token()` exactly as they would for any other token, because the proxy issues the access token through its own token factory.
 
+By default, replay protection uses a bounded `InMemoryJTIReplayStore`, so each
+server process tracks its own consumed assertions. A horizontally-scaled
+deployment should pass a shared implementation of `JTIReplayStore` as
+`IdentityAssertion(jti_store=...)`. Its `check_and_store(jti, expires_at)` method
+must atomically reserve a previously unseen JTI until the supplied Unix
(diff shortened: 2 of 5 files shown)


zzstoatzz (changes requested):
The distributed replay reservation must be atomic before this can ship. The current `get()` followed by `put()` allows simultaneous exchanges of the same ID-JAG through different replicas—or merely different validators sharing one store—to both mint tokens. That is the security property #4846 specifically requires; documenting the race does not make the advertised shared replay protection safe.

The proposed upstream primitive in `strawgate/py-key-value#369` now exists on its branch, but that PR remains open with unresolved correctness findings around Redis expiry and wrapper semantics. Once a sound version is released, this change should require an atomic conditional-write capability for shared stores and use it for the JTI reservation. Please include a deterministic concurrent test using two validators sharing one store; the current cross-instance test is sequential and cannot catch this race.

The existing process-local default may remain bounded, but a user-supplied store must not silently downgrade the single-use guarantee it is intended to provide.

PR #4880, opened 2026-08-22
fix(versions): deterministic meta.fastmcp.versions ordering

### LINKED ISSUE #4867: Equivalent version metadata order depends on registration order
## Problem

`dedupe_with_versions()` injects the available component versions into `meta.fastmcp.versions`. When two version strings are PEP 440-equivalent but spelled differently, the injected list depends on registration order.

## Minimal reproduction

```python
from fastmcp.tools import Tool
from fastmcp.utilities.versions import dedupe_with_versions


def make_tool(version: str) -> Tool:
    return Tool.from_function(lambda: None, name="demo", version=version)


forward = dedupe_with_versions(
    [make_tool("1"), make_tool("1.0")], key_fn=lambda component: component.name
)
reverse = dedupe_with_versions(
    [make_tool("1.0"), make_tool("1")], key_fn=lambda component: component.name
)

print(forward[0].meta["fastmcp"]["versions"])
print(reverse[0].meta["fastmcp"]["versions"])
```

On current `main`, this produces the two input orders respectively:

```text
['1', '1.0']
['1.0', '1']
```

## Expected behavior

Equivalent inputs should produce the same metadata ordering regardless of registration order. This would match the deterministic equivalent-version selection added in #4058.

Reported with Codex assistance and reproduced against `main` at `609f79b8a118cd6c4bb58a5341bafd76e42e5a2b`.

feiiiiii5:
## Problem

`dedupe_with_versions()` injects a `meta.fastmcp.versions` list onto deduplicated components, ordered by version descending. For PEP 440-equivalent spellings (`"1"` and `"1.0"`), `parse_version_key()` returns equal keys — and since `sorted()` is stable, the list order then depends on **registration order**:

```python
from fastmcp import FastMCP
m = FastMCP()

@m.tool.version("1")\nasync def demo(): ...
@m.tool.version("1.0")\nasync def demo_1_0(): ...
# meta.fastmcp.versions == ['1', '1.0']

# ...but register "1.0" first and you get ['1.0', '1']
```

Same component tree, different published metadata depending on import/registration order.

## Root cause

`fastmcp/utilities/versions.py:330-334` sorts with `key=parse_version_key(v)` alone. The repo already solved this exact determinism class for component selection in #4058: `version_sort_key` adds a raw-string tie-breaker so equivalent spellings always order lexicographically. This sort simply predates that convention.

## Fix

One line + comment:

```python
unique_versions = sorted(versions, key=lambda v: (parse_version_key(v), v), reverse=True)
```

Equivalent spellings now always order lexicographically regardless of registration order; primary (non-equivalent) version ordering is unchanged.

## Testing

- New `TestDedupeVersionMetadataDeterminism` in `tests/server/versioning/test_versioning.py`: equivalent spellings are order-independent both ways, primary descending order preserved (`['2.0', '1.10', '1']`), unversioned components excluded, malformed versions tie-break lexicographically (`['latest', 'LATEST']`).
- Repro script on main @ 49ff7f5 produced `['1','1.0']` vs `['1.0','1']` before; both orders → `['1.0', '1']` after.
- `tests/server/versioning/` (versioning + calls) and `tests/server/transforms/test_catalog.py`: all pass.

Fixes #4867


jlowin (approved):

jlowin:
/marvin fix the CI failure for static analysis

feiiiiii5:
Friendly nudge: this is approved and mergeable on my side. Happy to address anything else if needed.

zzstoatzz (approved):
lgtm! thank you @feiiiiii5

PR #4893, opened 2026-08-24
fix: support modern protocol in multi-server clients

### LINKED ISSUE #4892: Multi-server clients cannot negotiate the modern protocol
### What happened?

In FastMCP 4, a client created from a configuration containing multiple servers always negotiates the legacy handshake era, even when every configured backend supports the modern `2026-07-28` protocol. This prevents modern-only behavior such as multi-round-trip guard tools from working through the multi-server client.

```python
client = Client(
    {
        "mcpServers": {
            "weather": {"url": "https://weather.example.com/mcp"},
            "calendar": {"url": "https://calendar.example.com/mcp"},
        }
    }
)

async with client:
    assert client.protocol_version == "2026-07-28"
```

When both servers support the modern protocol, the assertion currently receives `2025-11-25`. The aggregate should negotiate the best protocol era shared by its connected backends and use that era consistently across every connection leg.

### Version Information

FastMCP 4.0.0b4

zzstoatzz:
A multi-server `MCPConfigTransport` currently declares itself legacy-only before connecting and builds backend proxy clients under the handshake era, so even an entirely modern configuration is silently downgraded. That makes the multi-server client an exception to FastMCP 4's normal protocol negotiation and prevents modern-only interactions such as multi-round-trip guard tools.

The connecting client's requested mode now reaches every backend client. Under `auto`, the transport discovers the connected backend eras before negotiating with its in-process router: an all-modern configuration stays modern end to end, while a mixed configuration reconnects every leg under the shared legacy era so the proxy chain never straddles incompatible interaction models. All-modern configurations follow normal modern semantics; applications that rely on handshake-era server-initiated sampling or elicitation can pin `mode="legacy"`.

```python
client = Client(
    {
        "mcpServers": {
            "weather": {"url": "https://weather.example.com/mcp"},
            "calendar": {"url": "https://calendar.example.com/mcp"},
        }
    }
)

async with client:
    assert client.protocol_version == "2026-07-28"
```

<details>
<summary>Why does this change existing multi-server behavior?</summary>

Multi-server aggregation is implemented by mounting proxy clients behind an in-process router. Those proxy clients historically defaulted to the handshake era because that era supports server-initiated sampling, elicitation, and roots. When FastMCP 4 changed the general `Client` default to `mode="auto"`, the aggregate retained that internal proxy default and was marked legacy-only to prevent different eras from meeting within the proxy chain.

That safety measure made the transport's implementation detail override the public client's requested behavior: even when every configured backend supported July, the aggregate silently stayed on the handshake protocol. Resolving one shared era across the whole chain removes the mismatch without retaining the silent downgrade. The compatibility tradeoff is therefore visible now: all-modern configurations use modern result-based interactions by default, while applications that require handshake-era server-initiated interactions can select them explicitly with `mode="legacy"`.

</details>

Fixes #4892

🤖 Generated with Codex


chrisguidry (approved):
I love how surgical the changes and new tests were, it's easy to trust this.

parkedwards (approved):

PR #4904, opened 2026-08-25
add independent client groups

### LINKED ISSUE #4903: Add independent multi-server client coordination
### What is the problem?

`Client(config)` combines multiple configured servers behind one in-process proxy and therefore exposes one logical MCP connection with one shared protocol era. That is useful when callers want a single aggregate endpoint, but it does not cover applications that need to coordinate several independent server connections—for example, one handshake-era server and one modern server—without translating either connection through a proxy.

Today those applications must rebuild the same configuration parsing, connection lifecycle, tool namespacing, collision detection, and call routing around several ordinary `Client` instances.

### What should FastMCP provide?

Add an experimental coordinator for named, independent `Client` instances. Each client should retain its own transport, negotiated protocol version, capabilities, handlers, and lifecycle. The coordinator should aggregate namespaced tool discovery and route calls to the client that advertised each tool, without constructing an MCP server or proxy.

Connection ownership must remain optional: callers can manage every client context themselves, or use the coordinator as a convenience context manager. Entering already-connected clients should compose with FastMCP's existing context refcounting rather than taking ownership away from the caller.

This should remain explicitly distinct from `Client(config)`: one coordinates multiple real connections, while the other presents one proxied MCP endpoint.

zzstoatzz:
`ClientGroup` coordinates one managed FastMCP client per server: each connection negotiates its own protocol era, and the group owns lifecycle, collision-checked `{server}_{tool}` namespacing, and call routing. Unlike `Client(config)`, no proxy stands in the middle and no backend drags the fleet onto its era; a one-client group is a supported way to get namespacing alone.

```mermaid
flowchart LR
    app[application] --> group[ClientGroup]
    group --> legacy["Client<br/>mode = legacy"]
    group --> modern["Client<br/>mode = auto"]
    legacy -->|2025-11-25| old[legacy server]
    modern --> |2026-07-28| new[modern server]
```

```python
from fastmcp import Client
from fastmcp.client.group import ClientGroup

group = ClientGroup(
    {
        "legacy": Client("legacy_server.py", mode="legacy"),
        "modern": Client("https://modern.example.com/mcp", mode="auto"),
    }
)

async with group:
    tools = await group.list_tools()
    result = await group.call_tool("modern_get_weather", {"city": "Chicago"})
```

Fixes #4903

<details>
<summary>the gap this fills</summary>

Downstream tool adapters that build agent tools over FastMCP clients face a forced choice today. `Client(config)` composes several servers behind one aggregate endpoint, so the proxy chain shares one interaction model: one handshake-era backend reconnects every modern backend under the handshake era. Keeping one client per server preserves each negotiated era, but then nothing owns the composition — callers concatenate per-client tool lists and inherit the collisions, the lifecycle coordination, and the routing themselves.

`ClientGroup` is the object between those two: independent negotiation per server, with collision-checked namespacing, shared lifecycle, and call routing owned in one place.

The upstream SDK group is the closest prior art, but its two connection paths split the behavior needed here:

| path | modern negotiation | group-owned lifecycle | calls through FastMCP |
| --- | --- | --- | --- |
| SDK `connect_to_server()` | no | yes | no |
| SDK `connect_with_session()` | externally | no | no |
| FastMCP `ClientGroup` | yes, per server | yes | yes |

`connect_with_session()` can aggregate a modern session after a separate FastMCP client connects it, but the caller must keep both lifecycle layers coordinated. `ClientGroup` combines independent negotiation and ownership while preserving FastMCP authentication, tracing, caching, result parsing, progress handling, and multi-round tool behavior.

</details>

<details>
<summary>lifecycle and routing</summary>

Callers can let the group own every connection with `async with group`, or connect the clients themselves and use the group only for discovery and routing. FastMCP's existing reference counting makes nested ownership safe, and a failed group entry unwinds clients that were already connected. The group adds no session handling of its own: each client owns its transport and session exactly as it does standalone, so legacy stateful sessions need nothing special.

Tool discovery runs concurrently and records each public name's server, client, and upstream name. The first routed call loads a shared catalog snapshot; explicit `list_tools()` calls refresh it. Unknown names fail locally after discovery, and collisions are rejected.

`resolve_tool()` exposes route provenance — the server name, the owning `Client`, and the upstream tool name — because adapters need more than routed calls. Per-tool behavior that must run on the real connection (session-driven input loops, handler context, interceptors) binds to `route.client`; the group aggregates names without ever standing between an adapter and the client that owns the tool.

`ClientGroup.from_config` accepts the standard `mcpServers` shape with an optional per-server `mode` field, so configuration-driven callers get per-server protocol pinning without constructing clients by hand.

</details>

<details>
<summary>scope</summary>

The initial surface is intentionally tool-only and uses FastMCP's existing `{server}_{tool}` convention: the group's server names are the namespace. A per-client `tool_name_prefix` kwarg on `Client` was explored as an alternative home for naming (#4932) and closed in favor of keeping collision policy here, with the composition object that detects collisions. This PR does not add protocol translation, an agent framework, resource URI policy, prompt aggregation, or a new transport. Naming configurability, dynamic membership, and expansion beyond tools should follow concrete usage before the surface widens.

</details>

### DIFF
--- a/dev-docs/v4-notes/client-groups.md
+++ b/dev-docs/v4-notes/client-groups.md
@@ -1,0 +1,33 @@
+---
+title: Client Groups
+---
+
+**Status: In review (#4904).** `ClientGroup` coordinates independent FastMCP clients — one managed connection per server — with collision-checked tool namespacing and call routing. User-facing usage is documented at [Client Groups](https://gofastmcp.com/clients/client-groups).
+
+## TL;DR
+
+Applications that hand tools from several servers to one agent previously chose between two bad shapes. `Client(config)` composes servers behind an in-process proxy, so every backend shares one negotiated protocol era — one handshake-era backend reconnects every modern backend under the handshake era — and concurrent calls funnel through one frontend session. One `Client` per server keeps each negotiated era, but nothing owns the composition: callers hand-roll namespacing, collision detection, lifecycle, and routing.
+
+`ClientGroup` is the object between those two: a mapping of named clients where each connection negotiates independently, tools are published as `{server}_{tool}`, and calls route back through the client that advertised the tool. It lives at `fastmcp.client.group` — in the main namespace, since the surface is additive and small.
+
+## Motivation
+
+The immediate driver was per-server protocol pinning: agent integrations need `mode="legacy"` on one server and `mode="auto"` on another, which a single aggregate connection cannot express. Namespacing requests (a prefix for a single server's tools) kept arriving alongside it, because the same integrations feed multiple servers' tools to one model.
+
+## Decisions
+
+**Group over kwarg.** A `tool_name_prefix` kwarg on `Client` was fully implemented as a comparison (#4932, closed). It was rejected because a prefix is collision policy, and collision policy belongs with the composition object that detects collisions. It also would have made the stable single-server client present names the server never advertised. The group's dict keys are the namespace; a one-client group is the supported way to get namespacing alone.
+
+**No synthetic frontend.** There is no aggregate session and no protocol translation. `resolve_tool()` returns the server name, the owning `Client`, and the upstream tool name, because tool adapters need to bind per-tool behavior (session-driven input loops, handlers, interceptors) to the real connection. The group never stands between an adapter and the client.
+
+**Invariants over flexibility.** Membership is immutable after construction (routes hold the advertising client and would go stale under mutation). Group entry is race-guarded and connects clients concurrently, unwinding successes on partial failure. A known route only requires its own client to be connected, so one dead server does not fail calls routed to healthy servers; full-fleet connectivity is required only for catalog discovery, which queries everyone.
+
+**Tool-only for now.** Resources have URI-based identity with different collision rules, and prompts a different invocation surface. Expanding beyond tools should follow concrete usage rather than declaring those policies by analogy.
+
+## Performance
+
+Measured against two real local streamable-HTTP servers (ratios are the finding, not the absolute numbers): routed calls through the group are indistinguishable from direct client calls (p50 1.89ms both), while the `Client(config)` proxy adds ~65% per call. Under concurrency the difference compounds — 200 concurrent calls across both servers completed ~6x faster through the group (~145ms vs ~875ms), because the group fans out over independent sessions while the proxy serializes through one. Group entry connects clients concurrently, so entry latency stays roughly one handshake deep instead of growing linearly with server count.
+
+## Relationship to the upstream SDK group
+
+The MCP Python SDK's `ClientSessionGroup` is the prior art. Its `connect_to_server()` owns lifecycle but only speaks the classic handshake; `connect_with_session()` can register an externally negotiated modern session but does not own it. `ClientGroup` combines independent modern negotiation with group-owned lifecycle while keeping calls on the configured FastMCP client surface (auth, tracing, caching, result parsing, progress, multi-round tools).
--- a/docs/clients/client-groups.mdx
+++ b/docs/clients/client-groups.mdx
@@ -4,9 +4,9 @@ sidebarTitle: Client Groups
 icon: layer-group
 ---
 
-<Warning>
-`ClientGroup` is experimental. Its API may change before it moves out of `fastmcp.experimental`.
-</Warning>
+import { VersionBadge } from '/snippets/version-badge.mdx'
+
+<VersionBadge version="4.0.0" />
 
 A `ClientGroup` coordinates several independent FastMCP clients without combining them behind a proxy server. Each client keeps its own connection, negotiated protocol version, capabilities, and handlers. The group adds namespaced tool discovery and routes each call back to the client that advertised the tool.
 
@@ -20,7 +20,7 @@ Construct the clients explicitly when each server needs its own handlers, authen
 import asyncio
 
 from fastmcp import Client
-from fastmcp.experimental.client_group import ClientGroup
+from fastmcp.client.group import ClientGroup
 
 legacy_client = Client("legacy_server.py", mode="legacy")
 modern_client = Client("https://modern.example.com/mcp", mode="auto")
@@ -50,6 +50,8 @@ asyncio.run(main())
 
 Tool names are prefixed with the configured client name by default. For example, a `get_weather` tool exposed by the `modern` client becomes `modern_get_weather`.
 
+A group with a single client is a supported way to get namespacing alone: the one server's tools are presented under its configured name, with no other behavior change.
+
 The first routed call loads the tool catalog lazily. After a successful load, unknown tool names fail locally rather than repeating discovery against every server. Call `list_tools()` explicitly to refresh the routes when servers add or remove tools dynamically.
 
 ## Bind tools to their owning client
@@ -72,7 +74,7 @@ The group aggregates names and detects collisions; it never stands between the a
 `ClientGroup.from_config` creates one client per server rather than passing the entire configuration through a proxy. A FastMCP-specific `mode` field can select the protocol behavior for each server:
 
 ```python
-from fastmcp.experimental.client_group import ClientGroup
+from fastmcp.client.group import ClientGroup
 
 config = {
     "mcpServers": {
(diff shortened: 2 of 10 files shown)


zzstoatzz:
I tested the direct upstream path, and it works across independently negotiated protocol eras. This should be the baseline before deciding whether FastMCP needs another group abstraction.

The important detail is to connect each FastMCP client first and register its active session with `ClientSessionGroup.connect_with_session()`. The SDK group's `connect_to_server()` path calls `initialize()` itself, which made every server in my test negotiate the handshake era.

```python
import asyncio
from contextlib import AsyncExitStack

from fastmcp import Client
from mcp import ClientSessionGroup
from mcp_types import Implementation


async def main() -> None:
    clients = {
        "docs": Client("https://docs.mcp.cloudflare.com/mcp", mode="auto"),
        "wiki": Client("https://mcp.deepwiki.com/mcp", mode="legacy"),
    }

    async with AsyncExitStack() as stack:
        for client in clients.values():
            await stack.enter_async_context(client)

        group = await stack.enter_async_context(
            ClientSessionGroup(
                component_name_hook=lambda component, server: (
                    f"{server.name}_{component}"
                )
            )
        )

        for name, client in clients.items():
            await group.connect_with_session(
                Implementation(name=name, version="example"),
                client.session,
            )

        print(
            "protocols:",
            {name: client.protocol_version for name, client in clients.items()},
        )
        print("tools:", sorted(group.tools))

        result = await group.call_tool(
            "wiki_read_wiki_structure",
            {"repoName": "jlowin/fastmcp"},
            read_timeout_seconds=30,
        )
        print("wiki call failed:", result.is_error)


asyncio.run(main())
```

This produced a modern `docs` connection (`2026-07-28`) beside a legacy `wiki` connection (`2025-11-25`), exposed five namespaced tools, and routed the real tool call successfully.

For consumers that want raw MCP tool results, this may already be the complete solution. The narrower question for this draft is whether keeping calls on the FastMCP `Client` surface—rather than routing through the registered raw sessions—removes enough additional work to justify a parallel abstraction.

zzstoatzz:
<details>
<summary>the full client-group investigation and why this PR has its current shape</summary>

## where this started

The original question was whether FastMCP could compose servers that negotiate different MCP protocol eras without translating between them or replacing the existing client stack.

The first important distinction was between two kinds of composition:

- `Client(config)` is **endpoint composition**. It mounts configured backends behind one in-process FastMCP proxy and exposes one outer MCP connection. That design is correct for a unified aggregate endpoint, but the proxy chain necessarily participates in one compatible interaction model; it cannot preserve several independently negotiated protocol eras through the one frontend session.
- `ClientGroup` is **application composition**. It keeps one real FastMCP `Client` per server, lets every client negotiate independently, and routes each call back through the client that owns that server connection.

That distinction ruled out changing `Client(config)`. The mixed-era use case needs parallel sessions, not a smarter protocol proxy.

## what the upstream SDK already provides

The upstream official MCP Python SDK's `ClientSessionGroup` was the obvious prior art, so both of its connection paths were tested rather than assumed to be insufficient.

With the SDK versions evaluated for this PR, `connect_to_server()` creates and initializes the session itself using the classic handshake path. It provides group-owned lifecycle, but it could not negotiate the modern `2026-07-28` interaction model.

`connect_with_session()` can route an already-initialized modern session, but session creation and lifecycle remain external. A caller using a configured FastMCP client must therefore keep the FastMCP client context, the underlying session, and the SDK group coordinated separately. Calls also no longer flow through the configured FastMCP client surface.

That leaves a concrete gap rather than a competing reimplementation:

| path | independent modern negotiation | group-owned lifecycle | calls through configured FastMCP clients |
| --- | --- | --- | --- |
| SDK `connect_to_server()` | no | yes | no |
| SDK `connect_with_session()` | externally supplied | no | no |
| this `ClientGroup` | yes, per server | yes | yes |

The upstream SDK group remains the better primitive for raw classic-session aggregation, especially when prompts, resources, and dynamic membership are required. This PR is narrower: tool composition over independently configured FastMCP clients.

## alternatives explored

A disposable convergence spike tested whether the FastMCP client could instead become a very thin wrapper over upstream session-bound operations. The experiment removed roughly 200 net lines and passed 597 client tests, but it was not merge-safe: it had to bypass the upstream client constructor and write private extension state. There is not yet a public upstream session-bound operations abstraction that preserves FastMCP's configured transport and lifecycle behavior without depending on private internals.

A consumer-side composition spike also tested the opposite direction: keep the FastMCP client intact and bind generated tool wrappers to the client that discovered them. It loaded 19 tools successfully and preserved route provenance through an interception layer. That validated independent-client composition, but it does not provide shared lifecycle ownership, catalog collision handling, or a reusable FastMCP-level routing primitive.

The resulting design therefore composes existing clients rather than duplicating or bypassing them.

## protocol experiments

The implementation was exercised against three simultaneous servers: one local filesystem server over stdio and two remote HTTP servers. The group held `2025-11-25` and `2026-07-28` connections at the same time and routed calls without a synthetic frontend session or protocol translation.

The stress passes covered:

- simultaneous mixed-era connection and discovery;
- concurrent discovery and concurrent tool calls;
- caller-owned clients that are already connected;
- group-owned clients entered through `async with group`;
- nested group contexts using FastMCP's existing reference-counted lifecycle;
- cleanup when a later client fails during group entry;
- disconnected-client errors;
- one server dying after discovery;
- repeated unknown-tool calls; and
- concurrent cold calls racing to initialize the catalog.

Those experiments produced several implementation changes rather than only validating the first draft.

## discovery and routing changes

Initial tool discovery was changed from sequential calls to concurrent `gather()` calls so one slow server does not serialize every other catalog request.

A lock now protects lazy cold-start discovery. Concurrent first calls share one catalog load instead of each querying every server.

The catalog is intentionally a snapshot:

- the first routed call lazily loads it;
- `list_tools()` explicitly refreshes it;
- unknown names fail locally once a snapshot exists; and
- an unknown name does not trigger repeated full-catalog network amplification.

Collisions are rejected rather than silently choosing a server. Public names currently follow the established `{server}_{tool}` convention.

The group records more than a callable mapping. `resolve_tool()` returns a public `ToolRoute` containing the server name, originating client, and upstream tool name. That gives adapters and observability layers provenance without reverse-engineering prefixes.

## lifecycle decisions

Both lifecycle styles are supported deliberately.

A caller can let the group own all connections:

```python
async with group:
    result = await group.call_tool("modern_get_weather", {"city": "Chicago"})
```

Or the caller can connect clients independently and use the group only for discovery and routing. Group entry reuses FastMCP's existing reentrant reference counting, so nested ownership does not close a client prematurely. If group entry fails partway through, clients already entered by that attempt are unwound.

Calls always return through the originating FastMCP client. That preserves its transport inference, authentication and TLS configuration, HTTP customization, pagination, tracing, result parsing, progress handling, handler behavior, and multi-round tool semantics.

## naming investigation

Earlier repository requests were reviewed because generated names have repeatedly surfaced as an integration concern. Prior discussions requested custom separators, no prefix, and complete generated-name overrides. The prototypes around those requests also showed why naming policy expands quickly once it becomes configurable.

For this experimental surface, `_` remains the conservative default because it has the broadest compatibility across tool-name consumers. Separator customization, prefix removal, callbacks, aliases, and collision policy were intentionally deferred until there is concrete usage that can define one coherent API rather than several overlapping knobs.

## why the surface is experimental and tool-only

The group lives under `fastmcp.experimental` because the architectural primitive is useful but the long-term breadth is not settled. This PR does not add:

- protocol translation;
- another transport;
- an agent abstraction;
- resource URI conflict policy;
- prompt aggregation semantics;
- dynamic group membership; or
- configurable naming policy.

Prompts and resources are not merely more entries in the same dictionary: resource URIs have different identity and collision rules, while prompts and tools have different invocation surfaces. Keeping the first version tool-only avoids declaring those policies accidentally.

## validation completed

The focused suite grew to 10 tests covering discovery, routing, provenance, collisions, lifecycle ownership, caller-managed lifecycle, reentrancy, failure cleanup, snapshot refresh, and concurrent cold discovery.

The implementation also passed the full repository suite during development, with more than 7,250 passing tests, plus Ruff, formatting, typing, and the repository pre-commit checks. The mixed-era and failure-mode stress programs were run separately from the unit suite because they exercised real stdio and HTTP connections.

## current conclusion

The PR's current shape is intentionally small: it fills the independent-negotiation plus owned-lifecycle gap while reusing FastMCP's configured single-server client as the unit of composition. It does not compete with the upstream SDK group where that group already fits, and it does not mutate the proxy-backed semantics of `Client(config)`.

The unresolved product decision is not whether the implementation can route mixed-era tools—it can—but whether this reusable group abstraction belongs in FastMCP now, should remain parked under `experimental`, or should wait for a future public upstream session-bound client seam. That is why the PR remains a draft.

</details>

jlowin (approved):

PR #4932, opened 2026-08-27
feat(client): add tool_name_prefix to Client

zzstoatzz:
Applications that hand tools from several servers to one agent need collision-free tool names, but the servers may need different per-connection configuration (protocol `mode`, auth, handlers) — which rules out `Client(config)`'s single aggregate endpoint. Today the only options are a proxy-backed multi-server client or hand-rolled renaming around each `Client`.

This adds `tool_name_prefix` to `Client`: a literal string prepended to tool names in `list_tools()` results and stripped from the name passed to `call_tool()`. The prefix is presentation-only and tools-only. The wire-level `list_tools_mcp()` / `call_tool_mcp()` methods, direct `client.session` access, and server-initiated messages (notifications, sampling) all carry the server's unprefixed names — the server never sees the prefix. Calling `call_tool()` with an unprefixed name raises `ValueError`, so the names an agent discovers are always the names it can call.

```python
from fastmcp import Client

weather = Client("https://weather-api.example.com/mcp", mode="auto", tool_name_prefix="weather_")
assistant = Client("https://assistant-api.example.com/mcp", mode="legacy", tool_name_prefix="assistant_")

async with weather:
    tools = await weather.list_tools()   # weather_get_forecast, ...
    result = await weather.call_tool("weather_get_forecast", {"city": "Chicago"})
```

The string is used verbatim (no separator is appended), so the entire API is one string — no separator, override, or callback knobs. Draft: this is one of two candidate homes for client-side tool naming; compare with the experimental `ClientGroup` in #4904, whose `{server}_{tool}` naming would delegate to this parameter if it lands.

https://claude.ai/code/session_01DrQJXPeMepQgJay5xF91yE

### DIFF
--- a/docs/clients/tools.mdx
+++ b/docs/clients/tools.mdx
@@ -29,6 +29,29 @@ async with client:
 
 Arguments are passed as a dictionary. For multi-server clients, tool names are automatically prefixed with the server name (e.g., `weather_get_forecast` for a tool named `get_forecast` on the `weather` server).
 
+## Tool Name Prefixes
+
+<VersionBadge version="4.0.0" />
+
+When an application composes tools from several independent clients — for example, providing tools from multiple servers to the same agent — tool names can collide. The `tool_name_prefix` parameter namespaces a single client's tools without involving a proxy or a multi-server configuration:
+
+```python
+from fastmcp import Client
+
+weather = Client("https://weather-api.example.com/mcp", tool_name_prefix="weather_")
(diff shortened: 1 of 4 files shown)


zzstoatzz:
Closing in favor of the group-first direction: tool naming is collision policy, and collision policy belongs with the composition object that detects collisions — the experimental ClientGroup in #4904 — rather than as a permanent kwarg on the stable Client. Leaving this branch up as the on-record comparison for the next time a per-client prefix is requested.

Issue #4927, opened 2026-08-27
Lifespan teardown does not run on SIGTERM, only SIGINT (leaks resources on container shutdown)


Aztec03hub:
Lifespan teardown does not run when the server is stopped with `SIGTERM`. It runs on `SIGINT` only. Docker, Kubernetes and Cloud Run all stop containers with `SIGTERM`, so any lifespan that closes an HTTP client, flushes a buffer, or releases a connection silently does not run on a normal container shutdown.

Not specific to 4.0 - reproduced identically on 3.4.7, so this appears longstanding.

### Environment

- `fastmcp==4.0.0b4` (`mcp` 2.1.1) **and** `fastmcp==3.4.7` (`mcp` 1.29.1), same result on both
- Python 3.11.15, Linux, HTTP transport
- `#4118` and `#3480` are both present in these versions and do not cover this path

### Reproduction

```python
import pathlib
from fastmcp import FastMCP
from fastmcp.server.lifespan import lifespan

EV = pathlib.Path("events.txt")
def say(m):
    with EV.open("a") as f:
        f.write(m + "\n")

@lifespan
async def a(server):
    say("A-startup")
    yield {"db": 1}
    say("A-shutdown")

mcp = FastMCP(name="repro", lifespan=a)

@mcp.tool
def ping() -> str:
    return "pong"

if __name__ == "__main__":
    mcp.run(transport="http", host="127.0.0.1", port=8944)
```

Start it, then `kill -TERM <pid>`. Compare against `kill -INT <pid>`.

Evidence is written to a file rather than stdout deliberately, so the result cannot be an artifact of output buffering at process death.

### Expected

`A-shutdown` is written for both signals.

### Actual

| Signal | `events.txt` contents |
|---|---|
| `SIGTERM` | `A-startup` only - 3 of 3 runs |
| `SIGINT` | `A-startup` then `A-shutdown` |

Verified with the venv interpreter invoked directly (no `uv run` wrapper) and the target PID confirmed through `/proc/<pid>/cmdline`, because an earlier version of this test signalled the wrong process and produced a false pass.

### Analysis

uvicorn's own shutdown sequence runs in **both** cases. Under `SIGTERM` the log still shows:

```
Shutting down
Waiting for application shutdown.
Application shutdown complete.
Finished server process
```

[...]

PR #4943, opened 2026-08-29
fix(server): defer HTTP signal replay until lifespan teardown

zzstoatzz:
Stopping an HTTP or SSE server with `SIGTERM` could kill the process before FastMCP's outer lifespan released user resources, even after Uvicorn reported application shutdown complete. Keep the existing lifecycle ownership from #4446, but defer Uvicorn's signal replay until asynchronous FastMCP teardown finishes. Direct task cancellation also completes ASGI shutdown before propagating.

Once Uvicorn stops serving, restore the caller's signal handlers during final lifespan cleanup. A second `SIGTERM` with the default handler can therefore terminate a stuck cleanup instead of being captured into shutdown flags that no longer have a consumer. Custom handlers keep their existing behavior. The initial `SIGTERM` still terminates the process after successful cleanup, so cleanup belongs in the lifespan rather than code following `mcp.run()`:

```python
from fastmcp import FastMCP
from fastmcp.server.lifespan import lifespan
import httpx2

@lifespan
async def app_lifespan(server):
    async with httpx2.AsyncClient() as client:
        yield {"http_client": client}

mcp = FastMCP("My server", lifespan=app_lifespan)
mcp.run(transport="http")
```

Refs #4927. Its stdio follow-up is a separate SDK cancellation issue; this change covers HTTP and SSE.

🤖 Generated with OpenAI Codex

### DIFF
--- a/fastmcp_slim/fastmcp/server/mixins/transport.py
+++ b/fastmcp_slim/fastmcp/server/mixins/transport.py
@@ -354,20 +354,30 @@ class TransportMixin:
             config_kwargs["log_level"] = default_log_level_to_use
 
         with temporary_log_level(log_level):
-            async with self._lifespan_manager():
-                config = uvicorn.Config(app, host=host, port=port, **config_kwargs)
-                server = uvicorn.Server(config)
-                path = getattr(app.state, "path", "").lstrip("/")
-                mode = " (stateless)" if stateless_http else ""
-                display_host = _format_host_for_url(host)
-                logger.info(
-                    f"Starting MCP server {self.name!r} with transport {transport!r}{mode} on http://{display_host}:{port}/{path}"
-                )
+            config = uvicorn.Config(app, host=host, port=port, **config_kwargs)
+            server = uvicorn.Server(config)
+            path = getattr(app.state, "path", "").lstrip("/")
+            mode = " (stateless)" if stateless_http else ""
+            display_host = _format_host_for_url(host)
+            logger.info(
+                f"Starting MCP server {self.name!r} with transport {transport!r}{mode} on http://{display_host}:{port}/{path}"
+            )
 
+            try:
                 if sockets is not None:
                     await server.serve(sockets=sockets)
                 else:
                     await server.serve()
+            except anyio.get_cancelled_exc_class():
+                # Uvicorn does not run shutdown when serve() itself is
+                # cancelled. Complete its graceful shutdown under a shield so
+                # the ASGI lifespan (and therefore FastMCP's lifespan) exits
+                # before propagating cancellation to the caller.
+                server.should_exit = True
+                if server.started:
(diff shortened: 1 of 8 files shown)


chrikrah:
Confirmed on `6ef0965` against uvicorn 0.35.0 and 0.51.0, with a lifespan whose teardown awaits
before it records:

| build | signal | exit | events written |
|---|---|---|---|
| main `0796584` | SIGTERM | 143 | `startup` |
| `6ef0965` | SIGTERM | 143 | `startup`, `shutdown` |
| both | SIGINT | 0 | `startup`, `shutdown`, and the line after `mcp.run()` |

One consequence worth a decision. Teardown now runs inside `capture_signals()`, and uvicorn's
`handle_exit` escalates to `force_exit` only on a repeated SIGINT:

```python
if self.should_exit and sig == signal.SIGINT:
    self.force_exit = True
```

So a slow teardown blocks shutdown and further SIGTERMs are swallowed. With a teardown that sleeps 30
seconds, three SIGTERMs one second apart left the process alive after 12 seconds and it took SIGKILL to
end it. On main the first SIGTERM ended it in 2 seconds. Kubernetes reaches the same place through
`terminationGracePeriodSeconds`, but `kill -TERM` twice no longer works, and that is what an operator
tries first.

Separately, `mcp.run()` still does not return under SIGTERM, because the replayed signal ends the
process with 143. Code after `run()` and any outer `finally` run on SIGINT only. Is that intended?

I did not test stdio or Windows.

chrikrah (commented):
Confirmed on `6ef0965` against uvicorn 0.35.0 and 0.51.0, with a lifespan whose teardown awaits
before it records:

| build | signal | exit | events written |
|---|---|---|---|
| main `0796584` | SIGTERM | 143 | `startup` |
| `6ef0965` | SIGTERM | 143 | `startup`, `shutdown` |
| both | SIGINT | 0 | `startup`, `shutdown`, and the line after `mcp.run()` |

One consequence worth a decision. Teardown now runs inside `capture_signals()`, and uvicorn's
`handle_exit` escalates to `force_exit` only on a repeated SIGINT:

```python
if self.should_exit and sig == signal.SIGINT:
    self.force_exit = True
```

So a slow teardown blocks shutdown and further SIGTERMs are swallowed. With a teardown that sleeps 30
seconds, three SIGTERMs one second apart left the process alive after 12 seconds and it took SIGKILL to
end it. On main the first SIGTERM ended it in 2 seconds. Kubernetes reaches the same place through
`terminationGracePeriodSeconds`, but `kill -TERM` twice no longer works, and that is what an operator
tries first.

Separately, `mcp.run()` still does not return under SIGTERM, because the replayed signal ends the
process with 143. Code after `run()` and any outer `finally` run on SIGINT only. Is that intended?

I did not test stdio or Windows.

Issue #4920, opened 2026-08-27
v4 beta: out-of-band `tools/list_changed` is silently dropped on the `2026-07-28` era (delivered on `2025-11-25`)


daniel-k:
### Summary

A server that changes its tool surface from a background task — not while
servicing a request — cannot notify connected clients once the connection
negotiates the sessionless `2026-07-28` era, which is what `mode="auto"` picks.
Forcing `mode="legacy"` on the same server, same code, same fastmcp build
delivers the notification. Nothing is raised or logged; the notification just
never arrives.

### Repro

```python
import asyncio, contextlib, os, fastmcp
from fastmcp import Client, FastMCP
from fastmcp.client.messages import MessageHandler
from fastmcp.server.middleware import Middleware

MODE = os.environ.get("MODE", "auto")
captured = []

class Capture(Middleware):
    async def on_message(self, context, call_next):
        fctx = getattr(context, "fastmcp_context", None)
        if fctx is not None:
            with contextlib.suppress(Exception):
                captured.append(fctx.session)
        return await call_next(context)

class Recorder(MessageHandler):
    def __init__(self):
        super().__init__(); self.changed = asyncio.Event()
    async def on_tool_list_changed(self, message):
        self.changed.set()

async def main():
    server = FastMCP("repro"); server.add_middleware(Capture())

    @server.tool
    def hello() -> str: return "hi"

    rec = Recorder()
    async with Client(server, message_handler=rec, mode=MODE) as c:
        await c.list_tools()
        print(f"fastmcp={fastmcp.__version__} mode={MODE} negotiated={c.protocol_version}")
        for s in captured:                      # out-of-band: no request in flight
            with contextlib.suppress(Exception):
                await s.send_tool_list_changed()
        try:
            await asyncio.wait_for(rec.changed.wait(), timeout=3)
            print("RESULT: RECEIVED")
        except TimeoutError:
            print("RESULT: NOT RECEIVED")

asyncio.run(main())
```

```
MODE=legacy  ->  negotiated=2025-11-25   RESULT: RECEIVED
MODE=auto    ->  negotiated=2026-07-28   RESULT: NOT RECEIVED
```

Sessions are captured in both cases (2 of them); only delivery differs. Same
result over real Streamable HTTP (`http_app(transport="streamable-http")` behind
uvicorn), so it isn't an in-memory-transport artifact. fastmcp 3.4.7 always
delivers, since it doesn't negotiate the modern era.

### Guess at the cause

[...]

PR #4954, opened 2026-08-31
server: warn when modern-era change notifications are silently dropped

zzstoatzz:
on `2026-07-28` connections the SDK silently discards connection-scoped change notifications — the spec only delivers them through a `subscriptions/listen` stream, which fastmcp doesn't publish to yet (#4920). the send reports success for a message that can never arrive, which cost the reporter a debugging session. `ctx.session` now returns a thin proxy that logs a warning (once per notification method) when one of these sends is discarded, pointing at the tracking issue.

a warning rather than an error: the same send may also serve legacy connections, where it still delivers — code that notifies best-effort across both eras keeps working.

```
notifications/tools/list_changed cannot be delivered on a 2026-07-28 connection: ...
Track support at https://github.com/PrefectHQ/fastmcp/issues/4920.
```

subscription support (`SubscriptionBus` publication + client listen streams) is the real fix and stays tracked in #4920.

bufo-heralds-an-incident

### DIFF
--- a/fastmcp_slim/fastmcp/server/context.py
+++ b/fastmcp_slim/fastmcp/server/context.py
@@ -7,11 +7,12 @@ from contextlib import contextmanager
 from contextvars import ContextVar, Token
 from dataclasses import dataclass
 from logging import Logger
-from typing import Any, Literal, cast, overload
+from typing import Any, ClassVar, Literal, cast, overload
 
 import mcp_types
 from mcp import LoggingLevel, ServerSession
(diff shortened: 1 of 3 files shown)


zzstoatzz:
parking this: the upstream SDK already ships both halves of subscriptions/listen (ListenHandler + SubscriptionBus server-side, a listen() client driver), so the real fix in #4920 is wiring rather than a large feature. rather than teach users a session-send surface we'd deprecate, we'll wire the bus properly as the fast follow.

PR #4970, opened 2026-09-01
codemode: raise legible errors in the sandbox; render enums and defaults in `get_schema`

### LINKED ISSUE #4969: CodeMode friction when driven by a model: silent in-sandbox call failures, pydantic-voiced errors, thin `get_schema` output
Notes from driving a CodeMode server as its intended consumer — an LLM given only `search` / `get_schema` / `execute`, doing a multi-step task against a server with several mounted backends. The catalog/search side worked well (specific queries return small, useful results; well-authored backend errors came through crisply). Three points of friction, in descending order of how much they hurt:

### 1. a failed `call_tool` inside `execute` returns the error as the call's *value*

```python
repos = await call_tool("tangled-mcp_search", {"query": "spaces", "kind": "repos"})
# `kind` is not a valid argument. no exception — `repos` is now this string:
# "1 validation error for call[search]\nkind\n  Unexpected keyword argument ..."
return {"repos": repos}
```

The sandbox code continued as if the call succeeded, and the error string flowed into the final result as data. In a longer chain the model carries that forward and reasons about it as if it were a result — nothing in the control flow signals that step 2 of 4 failed.

### 2. argument errors speak pydantic, not MCP

The error above, in full:

```
1 validation error for call[search]
kind
  Unexpected keyword argument [type=unexpected_keyword_argument, input_value='repos', input_type=str]
    For further information visit https://errors.pydantic.dev/2.13/v/unexpected_keyword_argument
```

Three legibility problems for a model reading it: it names the tool by its internal un-namespaced identity (`call[search]`, not the `tangled-mcp_search` the caller used); it links a human documentation site; and it never states which arguments the tool *does* accept, so the recovery step is another `get_schema` round-trip (or another guess). The same voice appears when miscalling the meta-tools themselves (`get_schema` with a wrong parameter name).

### 3. `get_schema` output is often too thin to prevent guess-and-check

```
**Parameters**
- `window` (string)
- `sort` (string)
- `limit` (integer)
```

No enums, defaults, or required markers. The only reason I knew `sort="trending"` was valid is that this particular tool's prose description said so — a tool with a terse description leaves the model guessing string values and learning by error, which combined with (1) means the wrong guesses can also travel silently.

Minor, for completeness: result envelopes vary per backend tool (some return `{"result": [...]}`, some a bare list), so robust `execute` code ends up defensive (`found.get("result", found)`); and the error in (1) suggests validation happens against the upstream tool identity rather than the name the sandbox exposed.

No strong opinions on mechanism — filing this as observed friction from real use. CodeMode is marked experimental and this is exactly the kind of thing the experiment should surface: when the whole surface is `search`/`get_schema`/`execute`, error and schema legibility *are* the UX.

zzstoatzz:
CodeMode previously passed backend error results into sandbox code as ordinary values. Failed `call_tool` calls now raise catchable exceptions naming the tool; validation failures also list valid parameters. Detailed schemas include enum values and defaults so agents can construct valid calls with less guesswork.

```text
call_tool('tangled-mcp_search') failed: … Unexpected keyword argument …
Valid parameters for tangled-mcp_search (* = required): query*, limit
```

Closes #4969.

bufo-heralds-an-incident

generated with Codex (GPT-6)


strawgate:
I agree with the three open Codex threads: error results should preserve their human-readable text, `const` values should appear in the detailed schema, and the docs should reflect the new detailed output.

One additional case: I tested the validation path through the public APIs with a mounted tool. For a tool exposed as `orders_search`, an invalid argument still produces `1 validation error for call[search]` and includes the Pydantic documentation URL. That seems contrary to this helper's stated goal of using the caller-facing identity, and it gives the model an internal name it cannot call.

The original Pydantic error is retained as the `ValidationError.__cause__`, so this could be rendered from `cause.errors(include_url=False)` as concise field/message pairs before appending the exposed tool's parameter list. A regression test with a mounted/namespaced tool should catch both the internal-name and URL leakage.

zzstoatzz:
@strawgate thanks! agree, updated accordingly

strawgate:
One part of the earlier error-result thread still appears unresolved. If an is_error result contains only non-text content, this falls back to _unwrap_tool_result, which stringifies the content. For an ImageContent, that includes the full base64 payload in the exception and can consume a large amount of model context; empty content produces only call_tool('name') failed: .
FastMCP’s client already treats empty and non-text error content as a generic Tool '<name>' returned an error. Could CodeMode use the same error-specific policy: prefer text, then structured content, then a stable generic fallback, without stringifying non-text blocks?

zzstoatzz:
addressed in c2a06121. `is_error` results now go through `_error_detail`: text content first, then `structured_content`, then a fixed `tool returned an error`. non-text blocks are never stringified, so an image error no longer puts its base64 payload in the exception, and empty content no longer yields a bare `call_tool('name') failed: `. test covers both the empty and image-only cases.

also pushed e27d257d since your last look: union enum values are only listed when every non-null branch is enumerated, so `Literal["a"] | int` no longer renders a misleading "one of".

PR #4971, opened 2026-09-01
perf: avoid duplicate startup for mixed-era backends

### LINKED ISSUE #4896: Reduce duplicate backend startup during mixed-era negotiation
Under `mode="auto"`, a mixed-era multi-server configuration first connects every backend to discover its era, then closes the entire set and reconnects every backend under the legacy handshake so all proxy legs share one interaction model.

That is correct but can be expensive for process-backed transports: each modern stdio backend is spawned twice before the aggregate is ready. Explore whether mixed-era negotiation can preserve the single-era invariant without reconnecting every healthy backend, or otherwise reduce the duplicate startup work for expensive transports.

ningmao-hlyz:
Closes #4896

When using a multi‑server MCPConfig with `mode="auto"`, mixed modern/legacy server fleets previously started every backend once for protocol discovery, shut them down, then restarted all backends under the legacy handshake.
Process‑backed transports incurred redundant startup cost even after the final protocol era was determined.

This fix prepares each transport only once and prioritizes connections to known legacy‑only transports:
- If a legacy backend connects successfully, remaining servers are opened directly in legacy mode
- If the legacy connection fails, healthy modern backends continue auto‑negotiation

Mount order is preserved from configuration, and the composite client still exposes a single shared protocol era.

Among the shipped transports, only SSE declares `legacy_only`, so this avoids duplicate startup for mixed fleets containing an SSE backend. Backends discovered to be legacy only after connection still use the existing fallback when a rebuild is needed.


zzstoatzz:
hi @ningmao-hlyz - thanks for the PR

verified on a mixed in-memory fleet: `main` starts every backend twice, this starts each once, same negotiated era either way.

two small things:

- `_prepare_transports` reports a transport *construction* failure as `"Failed to connect to MCP server %r, skipping"` — nothing has attempted a connection at that point. debug-log it and let the connection loop report where it actually happens.
- `legacy_only` is only set by SSE among the shipped transports, so this optimizes a mixed fleet containing an SSE backend rather than mixed fleets generally. say so in the description.

`ClientGroup` seems to avoid this failure bc we have one client per server, each negotiating its own era, nothing restarted, and the modern backend is not downgraded

leaving in draft while i decide whether we take this or fold it into the wider multi-server work

ningmao-hlyz:
> hi @ningmao-hlyz - thanks for the PR
> 
> verified on a mixed in-memory fleet: `main` starts every backend twice, this starts each once, same negotiated era either way.
> 
> two small things:
> 
> * `_prepare_transports` reports a transport _construction_ failure as `"Failed to connect to MCP server %r, skipping"` — nothing has attempted a connection at that point. debug-log it and let the connection loop report where it actually happens.
> * `legacy_only` is only set by SSE among the shipped transports, so this optimizes a mixed fleet containing an SSE backend rather than mixed fleets generally. say so in the description.
> 
> `ClientGroup` seems to avoid this failure bc we have one client per server, each negotiating its own era, nothing restarted, and the modern backend is not downgraded
> 
> leaving in draft while i decide whether we take this or fold it into the wider multi-server work

Thanks for the review and verification.

I’ve made the two requested adjustments:
- transport construction failures are logged at debug level during preparation, while the connection loop reports failures as connection failures;
- the description now clarifies that only SSE currently declares `legacy_only` among the shipped transports, so the practical target is a mixed SSE/modern fleet.

As for the design discussion around `ClientGroup`, I believe it should be addressed in a follow‑up conversation. It requires thoughtful design and thorough validation, and it is reasonable to keep the existing implementation as‑is for now.

Please let me know if there is anything else I can help with, I’m happy to assist.

zzstoatzz (approved):
lgtm!

PR #4999, opened 2026-09-03
fix: don't infer an output schema for unconstrained sequences

### LINKED ISSUE #4998: Bare `list` return annotation duplicates content blocks into structuredContent
### Problem

A tool annotated `-> list` that returns a content block alongside metadata sends the content block's payload twice: once as a real content block, and once again serialised into `structuredContent`. For an image that means the base64 is duplicated on the wire, and clients mishandle it — in our case claude.ai rendered nothing inline until we removed the duplication.

### Reproduction

```python
import asyncio, json
from mcp.types import ImageContent

PNG = ("iVBORw0KGgoAAAANSUhEUgAAAAEAAAABCAYAAAAfFcSJAAAADUlEQVR42mP8"
       "z8BQDwAEhQGAhKmMIQAAAABJRU5ErkJggg==")

mcp = FastMCP("repro")

@mcp.tool
def with_annotation() -> list:
    return [ImageContent(type="image", data=PNG, mimeType="image/png"), {"note": "metadata"}]

@mcp.tool
def without_annotation():
    return [ImageContent(type="image", data=PNG, mimeType="image/png"), {"note": "metadata"}]

async def main():
    async with Client(mcp) as c:
        for name in ("with_annotation", "without_annotation"):
            r = await c.call_tool(name)
            blob = json.dumps(r.structured_content) if r.structured_content else ""
            print(f"{name:20s} blocks={[b.type for b in r.content]} "
                  f"structured={len(blob)}B duplicated={PNG[:32] in blob}")

asyncio.run(main())
```

On `main` (a338cfbb):

```
with_annotation      blocks=['image', 'text'] structured=222B duplicated=True
without_annotation   blocks=['image', 'text'] structured=0B   duplicated=False
```

### Expected vs actual

Expected: the image is sent once, as an image content block, as it is without the annotation.

Actual: adding `-> list` duplicates it into `structuredContent`.

### The bit that isn't obvious

This looks like an oversight in a suppression that already exists rather than a missing feature. `ImageContent`, `ContentBlock`, `list[ContentBlock]` and `Any` all correctly infer no output schema today. That works via the `replace_type` swap in `function_parsing.py`, which substitutes `_UnserializableType` for MCP content types **inside** the annotation — so a bare `list` has no item type to match, slips past it, and infers `{"result": {"items": {}, "type": "array"}}`.

That schema constrains nothing — which is precisely why `Any` is excluded from inference already — but its existence is enough to force structured content, and the content blocks get serialised into it.

So `list[ContentBlock]` is safe and `list` is not, which seems like the wrong way round given `list` is the annotation people reach for when returning mixed content.

### Impact

Two client-side bug reports look like this same duplication from the other end: anthropics/claude-code#70280 (base64 injected into model context, blowing the tool-output token cap) and openai/codex#10334 (`content[]` dropped when `structuredContent` is present).

The per-tool workaround is `output_schema=None`, which also discards the schema for tools that wanted one, and requires knowing about this in the first place.

I have a fix and tests ready if you'd like a PR.

fsecada01:
Fixes #4998

### What

`list`, `tuple`, `list[Any]` and `tuple[Any, ...]` now infer no output schema, matching `Any`.

### Why this layer

The suppression for MCP content types already exists — it's the `replace_type` swap that puts `_UnserializableType` in place of `ImageContent`, `AudioContent` and friends. It works by matching types *inside* the annotation, so it covers `list[ContentBlock]` but has nothing to match in a bare `list`, which is why `list` infers a schema and `list[ContentBlock]` doesn't.

Fixing it at the conversion site instead wouldn't work: the schema is advertised at `list_tools` time, before the runtime value is known, so suppressing `structured_content` there would leave the tool advertising a schema it doesn't satisfy. Inference is the layer where this is decidable.

The rule is the same one already applied to `Any`: don't infer a schema that constrains nothing. `{"result": {"items": {}, "type": "array"}}` permits every value that no schema permits, and its only effect is to force structured content — which for a content-block result means duplicating the payload.

Parameterised annotations are untouched: `list[int]`, `tuple[int, str]`, `dict`, `str` all infer exactly as before.

### Tests

Two existing parametrized sweeps included `list` among many annotations as a wrapped-schema case; `list` is dropped from both. New tests cover the annotations that now infer nothing, the parameterised ones that still do, and a regression test that a returned image doesn't appear in `structuredContent`.

`uv run pytest -n auto` — 7304 passed. `uv run prek run --all-files` — clean.

### Confirmed against a real client

We hit this on a remote MCP server whose `get_result` returns `[ImageContent, record]` under a `-> list` annotation. claude.ai showed no inline image while the base64 was duplicated into `structuredContent` (3533 B, mostly base64); with the duplication removed (415 B, no base64) the image renders inline. Same server, same tool, same client — the duplication was the difference.

### DIFF
--- a/docs/servers/tools.mdx
+++ b/docs/servers/tools.mdx
@@ -546,7 +546,7 @@ def get_user_data(user_id: str) -> dict:
 
 #### Primitives and Collections
 
-When your tool returns a primitive type (int, str, bool) or a collection (list, set), FastMCP needs a return type annotation to generate structured content. The annotation tells FastMCP how to validate and serialize the result.
+When your tool returns a primitive type (int, str, bool) or a collection (list, set), FastMCP needs a return type annotation to generate structured content. The annotation tells FastMCP how to validate and serialize the result. The annotation has to say something about the items: a bare `list`, `tuple`, or `Sequence` (or `list[Any]`) describes nothing, so FastMCP infers no output schema for it, the same as for `Any`. Use `list[int]`, `list[dict]`, and so on when you want structured output for a collection.
 
 Without a type annotation, the tool only produces `content`:
 
--- a/fastmcp_slim/fastmcp/tools/function_parsing.py
+++ b/fastmcp_slim/fastmcp/tools/function_parsing.py
@@ -51,6 +51,23 @@ def _contains_prefab_type(tp: Any) -> bool:
     return False
 
 
+def _is_unconstrained_sequence(tp: Any) -> bool:
+    """A `list`/`tuple` that says nothing about its items.
+
+    Bare `list`, bare `tuple`, `list[Any]`, `tuple[Any, ...]`. The schema these
+    produce -- ``{"result": {"items": {}, "type": "array"}}`` -- constrains
+    nothing, which is why `Any` itself is already excluded from inference. These
+    are its sequence-shaped equivalent.
+    """
+    tp = _unwrap_type_alias(tp)
+    if tp in (list, tuple):
+        return True
+    if get_origin(tp) not in (list, tuple):
+        return False
+    args = [a for a in get_args(tp) if a is not Ellipsis]
+    return bool(args) and all(a is Any for a in args)
+
+
 def _unwrap_type_alias(tp: Any) -> Any:
     """Resolve a PEP 695 ``type X = ...`` alias to its underlying value.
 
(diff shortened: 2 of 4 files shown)


zzstoatzz (dismissed):

zzstoatzz (commented):
looks right to me overall and diagnosis makes sense.

nits:
- guard only catches `list`/`tuple` by identity, so `typing.List`, `list | None`, `Annotated[list, ...]`, and `Sequence` still duplicate
- docs under "Primitives and Collections" in `docs/servers/tools.mdx` should say a bare `list` no longer infers a schema

PR #5064, opened 2026-09-11
Honor OpenAPI explode defaults for pipe and space delimited query arrays

### LINKED ISSUE #5060: OpenAPI query arrays ignore the style-specific explode default
When an OpenAPI query array uses `style: pipeDelimited` or `style: spaceDelimited` and omits `explode`, the generated tool sends repeated query parameters. APIs expecting the declared delimiter therefore receive a different parameter representation. Explicitly setting `explode: false` works for pipeDelimited in the same example.

Expected: the decoded query parameter is `ids=a|b` for pipeDelimited, and `ids=a b` for spaceDelimited. Actual: both produce `ids=a&ids=b`.

The [OpenAPI 3.1 Parameter Object](https://spec.openapis.org/oas/v3.1.0.html#parameter-object) specifies a false default for explode when style is not form. The default form case correctly produces repeated keys.

## Reproduction

Run against FastMCP main at `0796584bb12606721011ac60a4522cafe4f5d497` with its locked dependencies (Python 3.12.13, MCP 2.0.0, httpx2 2.7.0, Windows). This uses the public Client API and captures the outgoing request without external network access.

```python
"""Public-API reproduction: inspect array serialization with omitted explode."""

import asyncio

import httpx2


async def inspect_query(style: str, explicit_false: bool) -> None:
    parameter = {
        "name": "ids",
        "in": "query",
        "required": True,
        "style": style,
        "schema": {"type": "array", "items": {"type": "string"}},
    }
    if explicit_false:
        parameter["explode"] = False
    spec = {
        "openapi": "3.1.0",
        "info": {"title": "Query reproduction", "version": "1"},
        "paths": {
            "/inspect": {
                "get": {
                    "operationId": "inspect",
                    "parameters": [parameter],
                    "responses": {"200": {"description": "OK"}},
                }
            }
        },
    }

    def capture(request: httpx2.Request) -> httpx2.Response:
        print(
            style,
            "false" if explicit_false else "omitted",
            request.url.params.multi_items(),
        )
        return httpx2.Response(200, json={"ok": True})

    async with httpx2.AsyncClient(
        base_url="http://local.test", transport=httpx2.MockTransport(capture)
    ) as http:
        server = FastMCP.from_openapi(openapi_spec=spec, client=http)
            tool = (await client.list_tools())[0]
            await client.call_tool(tool.name, {"ids": ["a", "b"]})


async def main() -> None:
    for style, explicit_false in [
        ("pipeDelimited", False),
        ("pipeDelimited", True),
        ("spaceDelimited", False),
        ("form", False),
    ]:
        await inspect_query(style, explicit_false)


if __name__ == "__main__":
    asyncio.run(main())
```

Output:

```text
pipeDelimited omitted [('ids', 'a'), ('ids', 'b')]
pipeDelimited false [('ids', 'a|b')]
spaceDelimited omitted [('ids', 'a'), ('ids', 'b')]
form omitted [('ids', 'a'), ('ids', 'b')]
```

Related: #1543 / #3595 addressed explicitly setting explode to false. This example leaves that field out and still reproduces on the commit above.

Investigation and reproduction prepared with AI assistance and executed locally.

gyanu2507:
Fixes #5060.

When a query array used `style: pipeDelimited` or `style: spaceDelimited` and omitted `explode`, the generated tool sent repeated keys (`ids=a&ids=b`). Setting `explode: false` already produced `ids=a|b` / `ids=a b`.

OpenAPI 3.1 defaults explode to true only for `form`. Other styles default to false. The director now follows that, so omitted explode on pipe/space delimited arrays joins with the declared delimiter. Form arrays are unchanged.

Tested with `TestQueryParameterSerialization.test_pipe_delimited_explode_omitted` and `test_space_delimited_explode_omitted`.

### DIFF
--- a/fastmcp_slim/fastmcp/utilities/openapi/director.py
+++ b/fastmcp_slim/fastmcp/utilities/openapi/director.py
@@ -290,8 +290,10 @@ class RequestDirector:
         Serialize query parameter values according to their OpenAPI style/explode settings.
 
         By default (style=form, explode=true), list values are passed through as-is
-        so httpx repeats the key (e.g. values=a&values=b). When explode=false,
-        list values are joined with the style-appropriate delimiter:
+        so httpx repeats the key (e.g. values=a&values=b). When explode is omitted,
+        OpenAPI 3.1 defaults explode to true only for ``form``; other styles
+        (pipeDelimited, spaceDelimited, deepObject, ...) default to false.
+        When explode=false, list values are joined with the style-appropriate delimiter:
           - form (default): comma  (values=a,b)
           - pipeDelimited:  pipe   (values=a|b)
           - spaceDelimited: space  (values=a%20b)
@@ -308,7 +310,12 @@ class RequestDirector:
         for key, value in query_params.items():
             param_info = param_lookup.get(key)
             if param_info is not None:
-                explode = param_info.explode if param_info.explode is not None else True
+                style = param_info.style or "form"
+                if param_info.explode is not None:
+                    explode = param_info.explode
+                else:
+                    # OpenAPI Parameter Object: explode defaults to true only for form.
+                    explode = style == "form"
                 if isinstance(value, dict):
                     if not value:
                         continue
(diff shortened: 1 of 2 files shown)


zzstoatzz:
Thank you for the contribution! #5060 was addressed in #5065, which preserves existing object-parameter behavior while fixing the array defaults. Closing this as superseded.

PR #5073, opened 2026-09-12
Preserve repeated fields for default multipart string arrays

### LINKED ISSUE #5072: Multipart string arrays are sent as Python list strings
A generated OpenAPI tool sends `tags=["a", "b"]` as one multipart field containing `['a', 'b']`. A backend reading repeated `tags` fields therefore receives one incorrect value instead of two. Expected: `{"tags": ["a", "b"]}`; actual: `{"tags": ["['a', 'b']"]}`.

Reproduced on main commit `58d903c84b0a3fe7596a15f9810ac138bda713dd`, Python 3.12.13 / Windows, with its frozen uv.lock dependencies. This example runs in-process and requires no external server.

```python
import asyncio

import httpx2
from fastapi import FastAPI, Request

app = FastAPI()


@app.post("/submit")
async def receive(request: Request):
    return {"tags": (await request.form()).getlist("tags")}


async def main():
    spec = {
        "openapi": "3.1.0",
        "info": {"title": "Multipart array", "version": "1"},
        "paths": {
            "/submit": {
                "post": {
                    "operationId": "submit",
                    "requestBody": {
                        "required": True,
                        "content": {
                            "multipart/form-data": {
                                "schema": {
                                    "type": "object",
                                    "properties": {
                                        "tags": {
                                            "type": "array",
                                            "items": {"type": "string"},
                                        }
                                    },
                                    "required": ["tags"],
                                }
                            },
                        },
                    },
                    "responses": {"200": {"description": "OK"}},
                }
            }
        },
    }
    async with httpx2.AsyncClient(
        base_url="http://test", transport=httpx2.ASGITransport(app=app)
    ) as http:
        async with Client(
            FastMCP.from_openapi(openapi_spec=spec, client=http)
        ) as client:
            result = await client.call_tool("submit", {"tags": ["a", "b"]})
            print(result.structured_content)


asyncio.run(main())
```

The [OpenAPI multipart rules](https://spec.openapis.org/oas/v3.1.0.html#special-considerations-for-multipart-content) describe a separate section for each array element. No explicit encoding is provided in this example.

Prepared with OpenAI Codex assistance; reproduction executed locally.

devispace:
Generated OpenAPI tools send non-empty string arrays in multipart bodies as a single Python list string, so a receiver sees `"['a', 'b']"` instead of two `tags` fields. Use ordered multipart entries to send each string as its own same-name part.

Retain explicit per-property encoding metadata so this change applies only to the default encoding path. Existing empty-array, non-string-array, object, and file-part behavior is preserved. This does not implement the previously unsupported explicit encoding rules.

Fixes #5072.

Prepared with OpenAI Codex assistance.
Validation is incomplete on this Windows environment: the full run before the final empty-array compatibility guard had 7,341 passes and 34 failures. The final OpenAPI subset had 390 passes with the network performance case run separately. The other 33 failing test IDs matched an unmodified checkout of the same commit and locked dependencies; the GitHub schema test also timed out on the baseline. Full prek passes except for four Windows fcntl attribute errors in an unchanged test file. The final empty-array guard was checked with the OpenAPI suite and full static checks, not another full-suite run. This PR is a draft pending maintainer review and CI validation.

### DIFF
--- a/fastmcp_slim/fastmcp/utilities/openapi/director.py
+++ b/fastmcp_slim/fastmcp/utilities/openapi/director.py
@@ -96,7 +96,7 @@ class RequestDirector:
         # Step 6: Handle request body — dispatch on declared content type.
         # httpx body kwargs (json, content, data, files) are mutually exclusive
         # but all accept None, so we set exactly one and pass all to Request().
-        files: dict[str, Any] | None = None
+        files: list[tuple[str, Any]] | None = None
         data: dict[str, Any] | None = None
         if body is not None:
             if declared_content_type == "multipart/form-data" and isinstance(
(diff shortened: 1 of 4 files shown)


zzstoatzz:
superseded by #5121, which builds on this implementation — thank you @devispace, you're credited in the commit

PR #5074, opened 2026-09-12
fix(auth): only map definitive upstream refresh rejections to invalid_grant

### LINKED ISSUE #5002: OAuthProxy maps temporarily_unavailable and unexpected refresh failures to invalid_grant and reflects exception details
### What happened?

When an `OAuthProxy` has valid refresh-token state but the upstream OAuth provider rejects a refresh with the standard `temporarily_unavailable` error, FastMCP returns a `TokenError` with `error="invalid_grant"`.

The client-facing `error_description` also contains the upstream provider's raw error description. The same behavior occurs for transport exceptions and unexpected local exceptions such as `RuntimeError`: they are all converted to `invalid_grant`, and their exception text is included in the resulting description.

Actual result from the example below:

```text
invalid_grant
Upstream refresh failed: temporarily_unavailable: private-upstream-detail
```

Expected behavior:

- Only a definitive refresh-credential rejection should be returned as `invalid_grant`.
- A standard `temporarily_unavailable` response should remain distinguishable from an invalid credential so clients do not discard otherwise valid refresh credentials.
- Unexpected local exceptions should remain internal server failures rather than being represented as OAuth credential failures.
- Client-facing responses should not contain raw upstream or internal exception details.

This reproduces on FastMCP 4.0.1, the latest published release. The same catch-all behavior remains on `main` at commit `75d50ffb683ba502e86bf833411c46234749270b`.

This is distinct from #2461: no token or JTI mapping is missing in this reproduction. All local refresh state is present and valid; only the upstream refresh attempt fails temporarily.

### Example Code

```Python
import asyncio
import time

from fastmcp.server.auth import AccessToken, OAuthProxy, TokenVerifier
from fastmcp.server.auth.oauth_proxy.models import (
    JTIMapping,
    UpstreamTokenSet,
)
from fastmcp.server.auth.oauth_proxy.upstream import OAuthError
from mcp.server.auth.provider import RefreshToken, TokenError
from mcp.shared.auth import OAuthClientInformationFull


class UnusedVerifier(TokenVerifier):
        return None


class FailingOAuthClient:
    async def refresh_token(
        self,
        *args: object,
        **kwargs: object,
    ) -> dict[str, object]:
        raise OAuthError(
            error="temporarily_unavailable",
            description="private-upstream-detail",
        )

    async def aclose(self) -> None:
        pass


class ReproProxy(OAuthProxy):
    def _create_upstream_oauth_client(self) -> FailingOAuthClient:
        return FailingOAuthClient()


async def main() -> None:
    proxy = ReproProxy(
        upstream_authorization_endpoint="https://idp.example/authorize",
        upstream_token_endpoint="https://idp.example/token",
        upstream_client_id="upstream-client",
        token_verifier=UnusedVerifier(),
        base_url="https://mcp.example",
        client_storage=MemoryStore(),
        jwt_signing_key="0123456789abcdef0123456789abcdef",
    )
    proxy.get_routes("/mcp")

    client = OAuthClientInformationFull(
        client_id="mcp-client",
        redirect_uris=["http://localhost/callback"],
        token_endpoint_auth_method="none",
    )
    refresh_jti = "refresh-jti"
    upstream_token_id = "upstream-token-id"

    refresh_value = proxy.jwt_issuer.issue_refresh_token(
        client_id=client.client_id,
        scopes=["read"],
        jti=refresh_jti,
        expires_in=3600,
    )

    # Seed a complete, valid mapping to isolate the upstream failure behavior.
    await proxy._jti_mapping_store.put(
        key=refresh_jti,
        value=JTIMapping(
            jti=refresh_jti,
            upstream_token_id=upstream_token_id,
            created_at=time.time(),
        ),
        ttl=3600,
    )
    await proxy._upstream_token_store.put(
        key=upstream_token_id,
        value=UpstreamTokenSet(
            upstream_token_id=upstream_token_id,
            access_token="old-access-token",
            refresh_token="old-refresh-token",
            refresh_token_expires_at=time.time() + 3600,
            expires_at=time.time() - 1,
            token_type="Bearer",
            scope="read",
            client_id=client.client_id,
            created_at=time.time(),
        ),
        ttl=3600,
    )

    try:
        await proxy.exchange_refresh_token(
            client,
            RefreshToken(
                token=refresh_value,
                client_id=client.client_id,
                scopes=["read"],
                expires_at=int(time.time()) + 3600,
            ),
            ["read"],
        )
    except TokenError as exc:
        print(exc.error)
        print(exc.error_description)


if __name__ == "__main__":
    asyncio.run(main())
```

### Version Information

```Text
FastMCP: 4.0.1
MCP: 2.1.1
httpx2: 2.12.0
Python: 3.14.3
Platform: macOS 26.6.1, arm64

FastMCP main checked at:
75d50ffb683ba502e86bf833411c46234749270b
```

fszcd:
## Root cause

`OAuthProxy.exchange_refresh_token` wrapped the upstream refresh call in a catch-all:

```python
except Exception as e:
    raise TokenError("invalid_grant", f"Upstream refresh failed: {e}") from e
```

Every failure mode — an upstream `temporarily_unavailable`, a transport error, an unexpected local exception — surfaced to the client as 400 `invalid_grant` with the raw exception text in `error_description`. `invalid_grant` tells the client its refresh credential is dead, so clients discard otherwise valid refresh tokens on a transient blip; the embedded text also reflects upstream and internal details into the client-facing response (#5002).

## Fix

Split the exception handling at the refresh call site:

- Upstream `OAuthError` with `error == "invalid_grant"` (definitive rejection of the stored refresh credential) → 400 `invalid_grant` with a generic description.
- Any other upstream OAuth error (e.g. `temporarily_unavailable`), transport failure, or unexpected local exception → generic `RuntimeError("Upstream token refresh failed")`, which the MCP SDK's `TokenHandler` does not map to a 400, so it surfaces as a 500 with no details echoed. Full details stay in the server logs.

This matches the conformance constraint noted on the issue: the SDK's `TokenErrorCode` has no `temporarily_unavailable` member, so "distinguishable from a credential failure" means 500 for everything that is not a definitive rejection.

The transparent upstream refresh path (`_try_transparent_refresh`) is unchanged — it already degrades gracefully through its caller.

## Tests

New `TestUpstreamRefreshErrorMapping` in `tests/server/auth/oauth_proxy/test_tokens.py` seeds a full refresh session and drives `exchange_refresh_token` with a mocked upstream client:

- upstream `invalid_grant` → `TokenError("invalid_grant")`, description does not echo the upstream detail
- upstream `temporarily_unavailable` → not a `TokenError`, no upstream detail in the message
- local `RuntimeError` → not a `TokenError`, no internal detail in the message

All three fail on unpatched `main` (each maps to `invalid_grant` with details echoed) and pass with the fix. Full `tests/server/auth/` suite: 1353 passed. `ruff check`/`ruff format --check` clean on both files. (`ty`'s five `fcntl` diagnostics and the `validate-pyproject` hook failure reproduce identically on a clean checkout of `main` on this Windows machine — pre-existing, unrelated.)

Fixes #5002

### DIFF
--- a/fastmcp_slim/fastmcp/server/auth/oauth_proxy/proxy.py
+++ b/fastmcp_slim/fastmcp/server/auth/oauth_proxy/proxy.py
@@ -109,7 +109,7 @@ from fastmcp.server.auth.oauth_proxy.models import (
     _hash_token,
 )
 from fastmcp.server.auth.oauth_proxy.ui import create_error_html
-from fastmcp.server.auth.oauth_proxy.upstream import AsyncOAuth2Client
+from fastmcp.server.auth.oauth_proxy.upstream import AsyncOAuth2Client, OAuthError
 from fastmcp.server.auth.redirect_validation import (
     build_client_redirect,
     is_redirect_uri_allowed_for_application_type,
@@ -1816,9 +1816,25 @@ class OAuthProxy(OAuthProvider, ConsentMixin):
                     **self._extra_token_params,
                 )
             logger.debug("Successfully refreshed upstream token")
-        except Exception as e:
+        except OAuthError as e:
+            if e.error == "invalid_grant":
+                # The upstream provider definitively rejected the stored
+                # refresh credential; only here should the client discard it.
+                logger.error("Upstream refresh token rejected: %s", e)
+                raise TokenError(
+                    "invalid_grant", "Upstream refresh token was rejected"
+                ) from e
+            # Any other upstream OAuth error (e.g. temporarily_unavailable) is
+            # not a credential failure; surface it as a server error so the
+            # client retries instead of discarding a valid refresh token.
             logger.error("Upstream token refresh failed: %s", e)
-            raise TokenError("invalid_grant", f"Upstream refresh failed: {e}") from e
+            raise RuntimeError("Upstream token refresh failed") from e
+        except Exception as e:
+            # Transport failures and unexpected local errors are server
+            # failures, not credential rejections. Details stay in the logs;
+            # the client-facing response must not echo them.
+            logger.error("Upstream token refresh failed: %s", e)
+            raise RuntimeError("Upstream token refresh failed") from e
 
         # Update stored upstream token
         # In refresh flow, we know there's a refresh token, so default to 1 hour
--- a/tests/server/auth/oauth_proxy/test_tokens.py
+++ b/tests/server/auth/oauth_proxy/test_tokens.py
@@ -9,7 +9,7 @@ import pytest
 from key_value.aio.stores.memory import MemoryStore
 from mcp.server.auth.handlers.token import TokenErrorResponse
 from mcp.server.auth.handlers.token import TokenHandler as SDKTokenHandler
-from mcp.server.auth.provider import AuthorizationCode
+from mcp.server.auth.provider import AuthorizationCode, TokenError
 from mcp.shared.auth import OAuthClientInformationFull
 from pydantic import AnyUrl
 
@@ -31,6 +31,7 @@ from fastmcp.server.auth.oauth_proxy.models import (
     UpstreamTokenSet,
     _hash_token,
 )
+from fastmcp.server.auth.oauth_proxy.upstream import OAuthError
 from fastmcp.server.auth.providers.jwt import JWTVerifier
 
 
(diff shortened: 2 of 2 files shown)


fszcd:
Closing this for now as the contribution workflow is being consolidated. If the change remains useful, it can be reopened or resubmitted later. Thank you for the review time.

PR #5101, opened 2026-09-14
Redact client values from invalid tool-argument logs

### LINKED ISSUE #5100: Invalid tool arguments are logged verbatim, including the arguments that were valid
### What happened?

When a `tools/call` fails argument validation, FastMCP logs a WARNING containing pydantic's full error list, so the values the client sent end up on stderr (or wherever the server's logs go):

- For a missing required argument, the logged `input` is the whole arguments dict, which includes the arguments that were valid.
- For a wrong-type value, `input` is that value.
- For an unknown key, `loc` is the key name the client sent.

Tool arguments often carry user content (prompts, documents, tokens), and a malformed call is exactly when a client is most likely to send it unexpectedly. I expected the log to say which tool failed and why (the error type and the field), without the values or client-chosen key names. Upstream `main` (58d903c) has the same log call.

Output of the example below:

```
WARNING  Invalid arguments for tool 'ask': [{'type': 'missing_argument', 'loc': ('question',), 'msg': 'Missing required argument', 'input': {'context': 'SECRET-1'}}]
WARNING  Invalid arguments for tool 'ask': [{'type': 'string_type', 'loc': ('question',), 'msg': 'Input should be a valid string', 'input': ['SECRET-2']}]
WARNING  Invalid arguments for tool 'ask': [{'type': 'unexpected_keyword_argument', 'loc': ('SECRET-3',), 'msg': 'Unexpected keyword argument', 'input': 1}]
```

### Example Code

```python
import asyncio

mcp = FastMCP()


@mcp.tool
def ask(question: str, context: str = "") -> str:
    return question


async def demo():
        # missing `question`; the `context` that was sent is valid
        await client.call_tool("ask", {"context": "SECRET-1"}, raise_on_error=False)
        # wrong type
        await client.call_tool("ask", {"question": ["SECRET-2"]}, raise_on_error=False)
        # unknown key
        await client.call_tool("ask", {"question": "q", "SECRET-3": 1}, raise_on_error=False)


if __name__ == "__main__":
    asyncio.run(demo())
```

### Version Information

```
FastMCP version:                                                           4.0.3
MCP version:                                                               2.2.0
Python version:                                                           3.14.3
Platform:                                    macOS-26.6.2-arm64-arm-64bit-Mach-O
```

gyanu2507:
Fixes #5100

A bad `tools/call` is logged with pydantic's full error list, including `input` (the values the client sent) and unexpected-key names. That is exactly when a caller is most likely to have stuffed a prompt, document, or token into the payload.

`_public_validation_errors()` keeps type, loc, and message. It drops `input`, and unexpected-key `loc` entries become `("<unexpected>",)`.

Test: `tests/server/test_input_validation.py::TestExpectedToolFailureLogging::test_invalid_arguments_log_omits_client_values`.


zzstoatzz:
Thanks for working on this. We confirmed the remaining leaks through validator messages and client-controlled validation locations. We’re taking this fix forward in #5106 so the logging policy covers those paths consistently; closing this proposal in favor of that replacement.

Issue #5098, opened 2026-09-13
from_openapi: cookie parameters with array/object values send Python repr strings


AWUQAQ:
### What happened?

With FastMCP.from_openapi(), cookie parameters are stringified with str() before the request is built. Scalars are fine (verbose → verbose=true), but non-scalar values leak Python reprs into the Cookie header: an array-typed cookie sends session_tags=['a', 'b'] and an object-typed cookie sends prefs={'theme': 'dark'}.

Expected per OpenAPI cookie rules (form style, the only style valid for cookies): arrays join with commas and objects alternate key/value pairs, e.g. session_tags=a,b; prefs=theme,dark. (The httpx cookies mapping cannot express repeated names, so explode=true necessarily collapses to the same single-string form.)

#4897 fixed this class of bug for header/path parameters and treated cookies as safe because they "are stringified" — stringifying is only correct for scalars, so non-scalar cookies are still affected.

### Example Code

```Python
import asyncio
import json

import httpx2
from fastmcp import FastMCP, Client

SPEC = {
    "openapi": "3.1.0",
    "info": {"title": "demo", "version": "1"},
    "servers": [{"url": "http://test"}],
    "paths": {
        "/things": {
            "get": {
                "operationId": "list_things",
                "parameters": [
                    {
                        "name": "session_tags",
                        "in": "cookie",
                        "schema": {"type": "array", "items": {"type": "string"}},
                    },
                    {
                        "name": "prefs",
                        "in": "cookie",
                        "schema": {
                            "type": "object",
                            "additionalProperties": {"type": "string"},
                        },
                    },
                ],
                "responses": {"200": {"description": "ok"}},
            }
        }
    },
}


async def main() -> None:
    seen: dict = {}

    async def handler(request: httpx2.Request) -> httpx2.Response:
        seen["cookie"] = request.headers.get("cookie")
        return httpx2.Response(200, json={"ok": True})

    client = httpx2.AsyncClient(transport=httpx2.MockTransport(handler))
    mcp = FastMCP.from_openapi(SPEC, client=client)

    async with Client(mcp) as c:
        await c.call_tool(
            "list_things",
            {"session_tags": ["a", "b"], "prefs": {"theme": "dark"}},
        )

    print("Cookie header sent:", seen["cookie"])
    # Expected: session_tags=a,b; prefs=theme,dark
    # Actual:   session_tags=['a', 'b']; prefs={'theme': 'dark'}


if __name__ == "__main__":
[...]

PR #5105, opened 2026-09-14
fix(openapi): serialize cookie parameters per the form style

### LINKED ISSUE #5104: Cookie parameters with array/object values are serialized as Python reprs
`RequestDirector.build` routes cookie parameters through `_query_scalar_to_str`, which stringifies non-scalars with `str()`. An array cookie goes on the wire as its Python repr instead of the OpenAPI form serialization.

```python
route = HTTPRoute(
    path="/things",
    method="GET",
    parameters=[
        ParameterInfo(name="id", location="cookie", schema={"type": "array", "items": {"type": "integer"}}),
        ParameterInfo(name="color", location="cookie", schema={"type": "object"}),
    ],
)
req = RequestDirector(spec=None).build(route, {"id": [3, 4, 5], "color": {"R": 100, "G": 200}})
print(req.headers["cookie"])
# actual:   id=[3, 4, 5]; color={'R': 100, 'G': 200}
# expected: id=3,4,5; R=100; G=200   (form style, explode=true is the cookie default)
```

Per the OpenAPI serialization rules, cookie parameters use the form style: arrays join with commas (`id=3,4,5`), objects explode into one cookie pair per property by default (`R=100; G=200`) and join as name,value pairs when `explode: false` (`color=R,100,G,200`). Scalar and boolean cookies are unaffected.

This is the same class as #4897 (path/header parameters), one location over: query parameters got full style/explode handling, path and header parameters got the simple style in #5091, and cookies were left with scalar-only handling. Reproduced twice with identical output. I have a fix ready with regression tests and can open a PR.

Harsh23Kashyap:
Closes #5104

Cookie parameters with array or object values were stringified with Python reprs (`id=[3, 4, 5]`, `color={'R': 100, 'G': 200}`). Cookie parameters always use the OpenAPI form style, so this routes them through a dedicated serializer:

- arrays join with commas (`id=3,4,5`), whether exploded or not;
- objects explode into one cookie pair per property by default (`R=100; G=200`), and join as name,value pairs when `explode: false` (`color=R,100,G,200`);
- empty arrays and objects are omitted, matching the query-parameter handling;
- scalars and booleans keep the existing `_query_scalar_to_str` behavior.

Regression tests in `tests/utilities/openapi/test_director.py` cover the array, exploded-object, and `explode: false` object cases: all three fail on main, pass with the patch, and the full director suite (65 tests) passes unchanged.


Harsh23Kashyap:
Closing as a duplicate: #5098 reported the same cookie-serialization bug about a day before my related issue, and #5099 already carries an equivalent fix (same form-style serialization in director.py, with tests).

PR #5108, opened 2026-09-14
fix(auth): send Azure account prompt upstream

### LINKED ISSUE #5107: AzureProvider sends prompt=select_account to the consent page instead of Entra
### What happened?

`AzureProvider` sends `prompt=select_account` to its own consent page instead of to Microsoft Entra.

With default settings, starting an OAuth flow redirects to `https://<server>/consent?txn_id=...&prompt=select_account`. The parameter ends up on FastMCP's local consent URL, and the redirect that follows to `login.microsoftonline.com` carries no `prompt` at all. Setting `require_authorization_consent=False` sends it correctly, so the behavior differs between consent modes.

Entra then falls back to silent SSO against whatever session the browser already holds. On a tenant where users have several Microsoft sessions, or one stale session, this shows up as sign-in attempts that hang after the user approves consent and then succeed on retry. There is no error and no log line, so it is hard to attribute.

Expected: `prompt=select_account` reaches Azure's authorization endpoint regardless of the consent setting.

`OAuthProxy.authorize()` returns the local `/consent` URL unless consent is disabled or external, so `AzureProvider.authorize()` appends the parameter to the wrong URL.

### Example Code

```Python
import asyncio

from mcp.server.auth.provider import AuthorizationParams
from mcp.shared.auth import OAuthClientInformationFull
from pydantic import AnyUrl

async def demo():
    provider = AzureProvider(
        client_id="test-client-id",
        client_secret="test-client-secret",
        tenant_id="test-tenant-id",
        base_url="https://myserver.com",
        required_scopes=["read"],
    )
    client = OAuthClientInformationFull(
        client_id="c1",
        redirect_uris=[AnyUrl("https://client.example.com/cb")],
    )
    params = AuthorizationParams(
        state="st",
        scopes=["read"],
        code_challenge="ch",
        redirect_uri=AnyUrl("https://client.example.com/cb"),
        redirect_uri_provided_explicitly=True,
    )

    # Expected: a login.microsoftonline.com URL carrying prompt=select_account
    print(await provider.authorize(client, params))


if __name__ == "__main__":
    asyncio.run(demo())
```

### Version Information

```Text
FastMCP version:                                      4.0.3
MCP version:                                          2.2.0
Python version:                                     3.13.15
Platform:                         Windows-11-10.0.26200-SP0
```

Harsh23Kashyap:
Fixes #5107.

`AzureProvider.authorize()` currently appends `prompt=select_account` to whichever URL the proxy returns. With the default consent screen, that is FastMCP's local `/consent` URL, so Microsoft never receives the parameter.

This moves the parameter into `_build_upstream_authorize_url()`, which is used by both the consent redirect and the direct path. The local consent URL now contains only the transaction ID.

The regression test covers both consent modes.

Tests:
- `pytest -q tests/server/auth/providers/test_azure.py` (54 passed)
- `ruff check fastmcp_slim/fastmcp/server/auth/providers/azure.py tests/server/auth/providers/test_azure.py`
- `ruff format --check fastmcp_slim/fastmcp/server/auth/providers/azure.py tests/server/auth/providers/test_azure.py`


zzstoatzz:
Thanks for jumping on this, @Harsh23Kashyap. Moving the prompt into _build_upstream_authorize_url is the right idea. The issue reporter's PR #5162 makes the same fix and also keeps select_account away from B2C user flows (the case Codex flagged here), so we're consolidating on that one and closing this as a duplicate. Appreciate the contribution!

PR #5118, opened 2026-09-15
Fix #5010: Investigate Windows CI worker crashes in credential tests

### LINKED ISSUE #5010: Investigate Windows CI worker crashes in credential tests
The Python 3.10 Windows unit-test job lost two xdist workers while running credential-storage tests in [this CI run](https://github.com/PrefectHQ/fastmcp/actions/runs/33938089591/attempts/1). The PR changes only the smart-home example; this is not yet established as a regression or a confirmed flake.

The reported tests were `test_windows_credential_state_allows_only_the_current_user` and `test_malformed_credential_state_has_a_safe_error` in `tests/cli/deploy/test_credentials.py`. The job finished with 7,248 tests passed and two worker-crash failures. The failed job has been rerun on the unchanged commit.

The workflow already documents Windows subprocess/timeout-related worker crashes and references #2715 and #2726. Investigate whether these failures share that cause; the test names alone do not establish the cause. No minimal reproducer has been isolated yet. Expected behavior is that the workers survive and failures, if any, are reported as ordinary test failures.

nandanadileep:
Fixes #5010

## What changed
Added `timeout=15` to the PowerShell subprocess in `_restrict_windows_access` so a hung ACL call raises `TimeoutExpired` (wrapped as `StateFileError`) instead of triggering pytest-timeout's worker-killing `os._exit` on Windows.

## Verification
`python -m pytest -q -x` passed

This is a draft so the final diff can be reviewed before it is marked ready.


zzstoatzz:
Thanks @nandanadileep for taking a look at the Windows worker crashes. I'm going to close this one: the suite's pytest-timeout is 5s, so a 15s subprocess timeout would still let pytest-timeout kill the worker first, and #5010 is still an open investigation without an identified root cause. If you find a reproducer or root cause, please add it to #5010 so we can decide on the fix there.

PR #5119, opened 2026-09-15
Fix #5115: OAuthProxy accepts self-contained ID-JAG tokens even when identity_ass

### LINKED ISSUE #5115: OAuthProxy accepts self-contained ID-JAG tokens even when identity_assertion is not configured (signing-key-only identity forgery)
### Summary

`OAuthProxy.load_access_token()` accepts a self-contained ID-JAG access token (a proxy-signed JWT carrying `fastmcp_grant: "id_jag"`) **even when `identity_assertion` is not configured**. In that state the proxy never issues such tokens, but it still honours them, and for that branch it returns the JWT payload as the identity **without the upstream token swap / re-validation** that every other proxy token goes through.

Net effect: anyone who holds the proxy's `jwt_signing_key` (or a derived key) can mint a token with arbitrary `sub`/`email`/`upstream_claims` and it will be accepted as an authenticated user, with no upstream (Google/GitHub/…) session behind it. A token signed with the same key but **without** the marker is correctly rejected because there is no JTI mapping — so the ID-JAG branch is currently the only path where the signing key alone equals identity.

### Version

- fastmcp 4.0.3 (latest on PyPI at the time of writing), Python 3.11
- `src/fastmcp/server/auth/oauth_proxy/proxy.py`, `load_access_token()` — the branch guarded by `payload.get("fastmcp_grant") == _ID_JAG_GRANT_MARKER` checks the revocation set but not `self._identity_assertion`.

### Reproduction

```python
import asyncio

prov = GoogleProvider(client_id="x", client_secret="y", base_url="https://example.com",
                      jwt_signing_key="k" * 64, client_storage=MemoryStore())
prov.set_mcp_path("/mcp")
issuer = prov.jwt_issuer
assert prov._identity_assertion is None  # feature NOT configured

# "Attacker" holds only the signing key. No login, no upstream token, no JTI mapping in storage.
forged = issuer.issue_access_token(client_id="forged", scopes=["openid"], jti="forged-1", subject="attacker",
                                   extra_claims={"fastmcp_grant": "id_jag", "email": "victim@example.com"})
res = asyncio.run(prov.load_access_token(forged))
print(res.subject, res.claims.get("email"))        # -> attacker victim@example.com   (accepted)

plain = issuer.issue_access_token(client_id="forged", scopes=["openid"], jti="forged-2", subject="attacker",
                                  extra_claims={"email": "victim@example.com"})
print(asyncio.run(prov.load_access_token(plain)))  # -> None  ("JTI mapping not found")  (rejected, as expected)
```

Output on 4.0.3:

```
attacker victim@example.com
None
```

### Expected

When `identity_assertion` is `None`, the proxy has never issued an ID-JAG token, so a token carrying the marker should be treated as invalid (return `None`, ideally with a log line), exactly like an unknown-JTI token. Something like:

```python
if payload.get("fastmcp_grant") == _ID_JAG_GRANT_MARKER:
    if self._identity_assertion is None:
        logger.warning("Rejected ID-JAG token: identity assertion is not configured (jti=%s...)", jti[:16])
        return None
    ...
```

Even when the feature *is* configured, it may be worth documenting that ID-JAG tokens are self-contained and therefore make the signing key equivalent to identity — so deployments that keep the key in a shared secret store understand the blast radius.

### Why it matters

For deployments that use the proxy's verified upstream claims (e.g. `claims["email"]`) to authorize or to attribute actions downstream (per-user cloud sessions, audit trails), the normal token-swap path gives a useful property: the signing key alone is not enough, you also need a live upstream session in storage. The ungated ID-JAG branch silently removes that property for everyone, including servers that never opted into SEP-990.

### Workaround we are using

In a `Middleware.on_call_tool`, reject any access token whose claims contain `fastmcp_grant`; normal token-swap claims never carry that key, so there is no functional impact.

Happy to open a PR with the gate + a test if that's welcome.

nandanadileep:
Fixes #5115

## What changed
Added a guard in `OAuthProxy.load_access_token()` that returns `None` with a warning when a token carrying the `fastmcp_grant: "id_jag"` marker is presented without `identity_assertion` configured, closing the identity-forgery path.

## Verification
`python -m pytest -q -x tests/server/mount/test_proxy.py` passed


zzstoatzz (approved):
verified the repro on main and the guard on this head; regression test to follow separately.

PR #5122, opened 2026-09-15
Fix #5120: OAuthProxy sends its own refresh token to the upstream revocation endp

### LINKED ISSUE #5120: OAuthProxy sends its own refresh token to the upstream revocation endpoint
### What happened?

When Claude Code triggers revocation through OAuthProxy, my upstream revocation endpoint receives a FastMCP-issued refresh JWT. My authorization server cannot find this token because it issued a different upstream refresh token.

The received JWT has my FastMCP server as its issuer and token_use: "refresh". It does not match any token in my authorization server’s database.

To reproduce:
- Configure OAuthProxy with upstream client credentials and upstream_revocation_endpoint.
- Authenticate through Claude Code.
- Trigger re-authentication and inspect the token received by the upstream revocation endpoint.

**Expected:** FastMCP sends the associated upstream refresh token so authorization server can revoke it.
**Actual:** revoke_token() deletes the local refresh-token metadata, then forwards token.token unchanged:
https://github.com/PrefectHQ/fastmcp/blob/258d5ef3d61fc6ad51fc6c3581f5508f39fc36eb/fastmcp_slim/fastmcp/server/auth/oauth_proxy/proxy.py#L2374 

**Possible resolution:** `exchange_refresh_token()` already resolves the FastMCP refresh JWT to the stored upstream token using its JTI. Could the refresh-token branch of `revoke_token()` use the same lookup?

https://github.com/PrefectHQ/fastmcp/blob/258d5ef3d61fc6ad51fc6c3581f5508f39fc36eb/fastmcp_slim/fastmcp/server/auth/oauth_proxy/proxy.py#L1794-L1806

### Version Information

```Text
FastMCP version:                                                                                                 4.0.3
MCP version:                                                                                                     2.2.0
Python version:                                                                                                 3.13.9
Platform:                                                                          macOS-26.6.2-arm64-arm-64bit-Mach-O
```

nandanadileep:
Fixes #5120

## What changed
In `revoke_token()`, when revoking a RefreshToken, the code now resolves the FastMCP-issued refresh JWT to the upstream refresh token via JTI mapping (the same lookup used by `exchange_refresh_token()`), so the upstream revocation endpoint receives the token it actually issued rather than the FastMCP wrapper JWT.

## Verification
`python -m pytest -q -x tests/server/mount/test_proxy.py` passed


zzstoatzz:
thanks, the approach is right: resolving the upstream token through the JTI mapping is the causal fix. two things before merge:

- a regression test. there are no tests exercising the upstream revocation POST today, so this would be the first; `tests/server/auth/oauth_proxy/test_tokens.py` has the fixtures for minting a refresh token and asserting on the outbound request. the PR body points at `tests/server/mount/test_proxy.py`, which is the mount proxy, so drop that reference.
- when the JTI lookup fails, the fallback still POSTs the fastmcp JWT upstream at `debug`. that is the bug from #5120 made quiet. log at `warning` and skip the upstream call, since the JWT can never match.

the refresh JTI mapping surviving local revocation is pre-existing and out of scope here; we'll follow up on that separately. i've assigned #5120 to you so the issue-link gate clears.

Ujemny:
Hey @zzstoatzz,
Is it something i could take over? I would like to apply your suggestions.

Thank you :)

zzstoatzz:
> Hey @zzstoatzz, Is it something i could take over? I would like to apply your suggestions.
> 
> Thank you :)

yes @Ujemny go for it, thanks!

PR #5133, opened 2026-09-16
fix(resources): strip URI fragment before matching query params

### LINKED ISSUE #5129: Resource-template query values include URI fragments
Resource-template query parameters include the URI fragment in their value. I expected fragments to stay separate from the query, as they do in standard URI parsing.

```python
from fastmcp.resources.template import match_uri_template

print(
    match_uri_template(
        "test://items?filter=active#section",
        "test://items{?filter}",
    )
)
```

This prints:

```python
{"filter": "active#section"}
```

Expected:

```python
{"filter": "active"}
```

Version: FastMCP 4.0.4, Python 3.12.13, Linux.

ege-arhan:
match_uri_template leaves the URI fragment stuck onto the last query value: `test://items?filter=active#section` against `test://items{?filter}` parses filter as `'active#section'`. Per RFC 3986 the fragment is client-side only and should never leak into query data.

Fix is 3 lines: partition on `#` before splitting path/query in `fastmcp_slim/fastmcp/resources/template.py`. Walked through old and new logic on the bug case, a fragment-free URI, and a path+fragment case; py_compile clean.

Fixes #5129


zzstoatzz:
Thanks for the fix, @ege-arhan. Partitioning on '#' before the query split is a clean way to do it. The issue reporter has an open PR for #5129 (#5132), and under our contributing guidelines the reporter gets first claim, so we're closing this as a duplicate. For future PRs: the new test_query_value_ignores_fragment fails locally, because get_data takes a required `id` that isn't in the template, so running the suite before pushing would catch that. Appreciate you taking the time!

PR #5155, opened 2026-09-18
fix: serialize OpenAPI collection cookies

### LINKED ISSUE #5098: from_openapi: cookie parameters with array/object values send Python repr strings

Michael-WhiteCapData:
OpenAPI cookie parameters currently pass through scalar stringification, so arrays and objects are sent as Python repr strings such as `['a', 'b']` and `{'theme': 'dark'}`. Those values do not follow OpenAPI's form-style cookie serialization and can be misread by the upstream API.

This adds a cookie-specific serializer that preserves the existing scalar behavior while joining array items and object key/value pairs with commas:

```text
tags=["alpha", "beta"]                  → tags=alpha,beta
preferences={"theme": "dark"}          → preferences=theme,dark
```

The regression test exercises both collection shapes in one request, including JSON-style boolean serialization inside an object.

Fixes #5098

🤖 Generated with OpenAI Codex


Michael-WhiteCapData:
Closing this in favor of #5099, which predates this PR and implements the same OpenAPI collection cookie serialization. Keeping one implementation avoids duplicate review effort. Thanks to the maintainers and the original contributor.

PR #5158, opened 2026-09-18
fix: refresh BM25 result tool metadata

### LINKED ISSUE #5152: BM25 search returns an old input schema after registering a newer tool version
### What happened?

After a BM25 search, registering version 2 of a tool with a different input constraint but unchanged names and descriptions leaves search advertising version 1's input schema. The current tool accepts `running`, but search still says only `queued` is valid. Following that schema causes a validation error. Expected: search returns the current tool's schema.

Reproduced on current main; CodeMode's default search is affected too. The same problem was noted in an earlier [review comment](https://github.com/PrefectHQ/fastmcp/pull/3154#discussion_r2862070827).

### Example Code

```python
import asyncio
import json
from typing import Literal

from fastmcp.server.transforms.search import BM25SearchTransform

server = FastMCP("example", transforms=[BM25SearchTransform()])


@server.tool(name="job_status", version="1")
def old(state: Literal["queued"]) -> str:
    """Check job status."""
    return state


def new(state: Literal["running"]) -> str:
    """Check job status."""
    return state


async def main() -> None:
        await client.call_tool("search_tools", {"query": "status"})
        server.tool(name="job_status", version="2")(new)
        result = await client.call_tool("search_tools", {"query": "status"})
        advertised = json.loads(result.content[0].text)[0]
        current = await server.get_tool("job_status")
        print("advertised:", advertised["inputSchema"]["properties"]["state"])
        print("current:", current.parameters["properties"]["state"])
        assert advertised["inputSchema"] == current.parameters


asyncio.run(main())
```

Output before the assertion fails:

```text
advertised: {'const': 'queued', 'type': 'string'}
current: {'const': 'running', 'type': 'string'}
```

### Version Information

FastMCP `4.0.6.dev2+9f983e9` (main `9f983e945b96ba19652b0b6757baf06a2d76a7fc`), MCP `2.0.0`, Python `3.12.10`, macOS `27.0` arm64.

Prepared with AI assistance.

Michael-WhiteCapData:
BM25 caches both its relevance index and the `Tool` objects returned for matching indices. Replacing a versioned tool can leave the searchable text unchanged, so the index stays valid but search continues serializing the old tool's schema and metadata.

This keeps the cached index when its ordered searchable documents are unchanged while refreshing the associated tool objects on every search. It also makes document order part of the cache key so each BM25 index always maps to the same catalog positions.

```python
server.tool(name="job_status", version="2")(new)
result = await client.call_tool("search_tools", {"query": "status"})
# result now advertises version 2's input schema
```

Fixes #5152

🤖 Generated with OpenAI Codex


zzstoatzz:
superseded by #5153 from the issue reporter, which resolves results against the current catalog and carries broader tests. thanks for the fix, the diagnosis matched.

PR #5159, opened 2026-09-18
fix: exclude URI fragments from template matches

### LINKED ISSUE #5129: Resource-template query values include URI fragments

Michael-WhiteCapData:
Resource-template matching currently splits the query directly from the original URI, which leaves a trailing fragment attached to the final query value. A URI such as `test://items?filter=active#section` therefore exposes `active#section` to the resource handler instead of `active`.

This removes the fragment before splitting and parsing the path and query. The same normalization also prevents a fragment from becoming part of a trailing path parameter.

```python
match_uri_template(
    "test://items?filter=active#section",
    "test://items{?filter}",
)
# {"filter": "active"}
```

Fixes #5129

🤖 Generated with OpenAI Codex


zzstoatzz:
duplicate of #5133, already assigned on #5129. thanks.

PR #5164, opened 2026-09-18
Expose FileSystemProvider discovery failures

### LINKED ISSUE #5135: FileSystemProvider should expose original discovery failures after initialization
### Enhancement

With `FileSystemProvider`, a configured module can fail to import while the provider continues loading the remaining components and the server still starts. `discover_and_import()` returns those failures in `DiscoveryResult.failed_files`, but after provider initialization there is no supported way for the application to inspect that original discovery result.

This matters for applications that need to distinguish an intentionally empty/conditional provider from a server that booted successfully but failed to load a configured capability. Re-running discovery is not a safe workaround because module imports can have top-level side effects and execute twice.

Concrete use case: a tool module raises during import because required configuration is missing; the server boots and `list_tools()` omits that tool, but the application cannot inspect the provider's original discovery failure through a public API.

**Expected:** applications can observe whether the provider's original discovery had failures, so they can choose their own tolerant or strict policy without re-importing modules.

**Actual:** the structured failure result is transient during provider loading; afterward only warnings/private implementation state remain.

spark-328:
## Summary

- retain the latest filesystem discovery failures on `FileSystemProvider`
- expose them through a copied, read-only-by-convention `discovery_failures` mapping
- refresh the mapping when `reload=True` performs another discovery pass

This lets applications apply their own strict startup policy after a configured module fails to import, without re-importing modules and risking top-level side effects.

Fixes #5135

## Test plan

- `uv run --frozen pytest tests/fs/test_provider.py -q` (26 passed)
- `uv run --frozen ruff check ...`
- `uv run --frozen ruff format --check ...`
- `uv run --frozen ty check fastmcp_slim/fastmcp/server/providers/filesystem.py tests/fs/test_provider.py`


zzstoatzz:
thanks, this is the right shape: a snapshot on the provider, refreshed per load, no re-import. two things to finish it:

- **registration failures too.** after discovery, `_load_components` loops `_register_component` and logs on exception. a component that imported fine but failed to register is just as missing from `list_tools()`, so it belongs in the same mapping. record it there with the exception type and message (`f"{type(exc).__name__}: {exc}"`), and add a test; monkeypatching `FileSystemProvider._register_component` to raise is the simplest way to exercise it.
- **name it `failed_files`**, matching `DiscoveryResult.failed_files`, so the vocabulary stays the same across the discovery result and the provider.

no constructor flag is needed; with the property, a strict startup is one line in the app. i've assigned #5135 to you so the issue-link gate clears.

spark-328:
Addressed both points in 7da47d0:

- registration failures now populate `failed_files`, formatted as `{ExceptionType}: {message}`
- renamed to `failed_files` to match `DiscoveryResult.failed_files`
- added a regression test using a monkeypatched `_register_component`

Thanks!

zzstoatzz (approved):
lgtm!

PR #5169, opened 2026-09-19
fix(server): fall back to the highest authorized version on direct lookup

### LINKED ISSUE #4950: Versioned component lookup does not fall back after authorization filtering
### What happened?

When multiple versions of a component share an identifier, list operations filter component-level authorization before exposing the highest available version, but direct lookups select the highest version before applying authorization.

For an unauthenticated caller with:

- `calc@1.0` public
- `calc@2.0` requiring the `admin` scope

`list_tools()` exposes `calc` from v1, but `get_tool("calc")` returns `None` and `call_tool("calc")` reports `Unknown tool`.

The same inconsistency affects resources, resource templates, and prompts.

### Minimal reproduction

```python
from fastmcp import FastMCP
from fastmcp.server.auth import require_scopes

mcp = FastMCP()

@mcp.tool(version="1.0")
def calc() -> int:
    return 1

@mcp.tool(version="2.0", auth=require_scopes("admin"))
def calc() -> int:
    return 2
```

With no `admin` scope:

```text
list_tools()       -> calc (v1)
get_tool("calc")   -> None
call_tool("calc")  -> Unknown tool
```

This reproduces both in-process and over a local Streamable HTTP server using Uvicorn.

### Expected behavior

Direct lookup and invocation should resolve the highest version authorized for the current request, consistent with the component list. An explicitly requested version should continue to be authorized independently.

### Version information

Reproduced against `upstream/main` at `977ba66c`, Python 3.12.

gshklovs:
## Description

`list_tools()` filters component auth before the client sees the highest
version, but `get_tool()` picked the highest version first and only then
ran the auth check. When that version was unauthorized `_get_tool` returned
`None`, and `get_tool` treated `None` as "not found" and returned early,
never reaching the existing fallback that already selects the highest
enabled and authorized version. So an unauthenticated caller saw `calc`
(v1) in `tools/list` but got `Unknown tool` from `tools/call`. Resources,
resource templates and prompts had the same shape.

This moves the early return so a `None` from the highest-version lookup
falls through to the same fallback the disabled-version case already uses
(#3421). Explicit version requests are unchanged: an unauthorized explicit
version still returns `None`, and the hashed-name dispatch path still runs
its own auth check.

Since a miss now runs the fallback scan, `call_tool` tries hashed-name
dispatch first when the name parses as `<hash>_<local_name>` and only
falls back to `get_tool` on a miss, so mcp-apps hashed calls never pay
the scan; only genuinely unknown names do.

Closes #4950

## Contribution type

## Checklist


zzstoatzz (changes requested):
The version/auth fallback addresses #4950 at the causal early return, but the bundled `call_tool()` routing reorder changes supported display-name precedence.

Tool names matching `<12 lowercase hex>_<name>` are currently valid; that pattern is parsed as a possible backend address but is not reserved or rejected at registration. On `main`, normal display-name resolution wins and hashed routing is only attempted on a miss. This PR tries hashed routing first, so a root tool whose valid display name equals an app tool's hashed address silently calls the app backend instead.

I reproduced this on exact head `ec8372dc`: a root tool named `hashed_backend_name("contacts", "save")` returns `"display tool"` on `main`, but the same call returns `"app backend"` on this PR. Please keep the existing display-name-first behavior (the safest scoped fix), or establish and enforce a reserved addressing namespace as a separate compatibility decision. The fallback/auth tests otherwise pass (`79 passed`).

PR #5180, opened 2026-09-20
Keep shared stdio sessions alive until the last client exits

### LINKED ISSUE #5179: Exiting a shared non-keep-alive stdio client closes other active clients
Two clients can connect to the same `PythonStdioTransport(keep_alive=False)`, but exiting one closes the subprocess while the other is still inside its connection context. The remaining client's next tool call raises `MCPError: Connection closed`.

I expect the shared connection to remain usable until its last client exits, then close as requested by `keep_alive=False`.

```python
import asyncio
from pathlib import Path
from tempfile import TemporaryDirectory

from fastmcp import Client
from fastmcp.client.transports import PythonStdioTransport


async def main():
    with TemporaryDirectory() as directory:
        server = Path(directory) / 'server.py'
        server.write_text(
            'import os\nfrom fastmcp import FastMCP\n'
            'mcp = FastMCP("test")\n'
            '@mcp.tool\ndef pid() -> int:\n    return os.getpid()\n'
            'mcp.run()\n'
        )
        transport = PythonStdioTransport(server, keep_alive=False)
        async with Client(transport) as first:
            async with Client(transport) as second:
                pid = (await second.call_tool("pid")).data
            assert (await first.call_tool("pid")).data == pid
            print('remaining client still connected')


asyncio.run(main())
```

Reproduced on current `main` at `9c35c017cd89e4d50a9f512c8eafde68c301ec70`, Python 3.12.14, MCP 2.0.0, macOS ARM64. The transport already tracks active sessions but disconnects unconditionally when a non-keep-alive context exits.

🤖 Investigated with OpenAI Codex.

Aurelucien:
Fixes #5179

Exiting one of two clients sharing a non-keep-alive stdio transport currently closes the subprocess underneath the remaining client. Automatic cleanup now waits for the last active session and holds the existing connection lock through teardown, so a new connection cannot reuse a session being closed. Explicit `disconnect()` is unchanged.

```python
transport = PythonStdioTransport("server.py", keep_alive=False)
async with Client(transport) as first:
    async with Client(transport) as second:
        await second.call_tool("pid")
    await first.call_tool("pid")  # Still uses the same live subprocess.
# The subprocess exits after the last client leaves.
```

🤖 Generated with OpenAI Codex.


zzstoatzz (approved):
lgtm!

zzstoatzz:
one follow-up uncertainty: sharing a single Client already supports nested/concurrent use; separate Clients sharing one transport also share the underlying session, so callback/settings ownership is less clear. before extending that pattern further, i'd like to check whether proxy internals rely on it and what contract we intend to support. not a requested change to this fix.

PR #5181, opened 2026-09-20
Preserve binary resource contents in response cache

### LINKED ISSUE #5178: ResponseCachingMiddleware breaks binary resource responses
# ResponseCachingMiddleware breaks binary resource responses

Adding `ResponseCachingMiddleware()` changes the behavior of resources returning bytes. UTF-8-compatible bytes become a text response on the second read; non-UTF-8 bytes fail on the first read with an internal server error.

```python
import asyncio

from fastmcp.server.middleware.caching import ResponseCachingMiddleware

server = FastMCP("binary-cache-repro")
server.add_middleware(ResponseCachingMiddleware())


@server.resource("data://sample", mime_type="application/octet-stream")
def sample() -> bytes:
    return b"hello"  # Also try b"\x89PNG\r\n\x1a\n".


async def main() -> None:
        for _ in range(2):
            print(await client.read_resource("data://sample"))


asyncio.run(main())
```

Expected: both reads return `BlobResourceContents` containing the original bytes, as they do when the middleware is removed.

Actual: `b"hello"` returns `BlobResourceContents` on the first read and `TextResourceContents` on the second. The PNG header raises `MCPError: Internal server error` on both reads. The server traceback ends in the cache adapter's `dump_python(value, mode="json")` with `UnicodeDecodeError: 'utf-8' codec can't decode byte 0x89 in position 0: invalid utf-8`.

Reproduced on main at `9c35c017cd89e4d50a9f512c8eafde68c301ec70`, Python 3.12.13, Pydantic 2.13.4, and py-key-value-aio 0.4.5, using the default in-memory cache. No external services are needed.

OpenAI Codex helped investigate and run this reproduction.

brennanbutler01:
Fixes #5178. `ResponseCachingMiddleware` currently turns UTF-8-compatible bytes into text on cache hits and fails to cache non-UTF-8 bytes. Store binary resource content as base64 with an explicit byte/text marker, then decode it when unwrapping the cached result. Existing text entries remain readable, and MIME types and metadata are preserved.

For example, a resource returning `b"\x89PNG\r\n\x1a\n"` now returns the same `BlobResourceContents` on both the first read and cache hits. OpenAI Codex helped prepare the fix and regression tests.


zzstoatzz (approved):

zzstoatzz (approved):
lgtm!

PR #5194, opened 2026-09-21
Detect client extensions from clientCapabilities, not clientInfo

### LINKED ISSUE #5192: `client_supports_extension` returns False when the client advertises the extension without `clientInfo`
### What happened?

A client advertises `io.modelcontextprotocol/ui` in the 2026-07-28 per-request envelope and sends no `clientInfo`, which is optional on that protocol version. A tool on the server then calls `ctx.client_supports_extension("io.modelcontextprotocol/ui")`.

Expected: `True`.

Actual: `False`, while `session.client_capabilities.extensions` on that same request holds `{"io.modelcontextprotocol/ui": {}}`. Adding the optional `clientInfo` key flips it to `True`. A server that gates behavior on the check drops the feature for a conformant client, with no error to point at.

Output (the MRE posts the envelope itself because FastMCP's own `Client` sends `clientInfo` and so only reaches the passing branch):

```
fastmcp 4.0.5, mcp 2.2.0, protocol 2026-07-28

raw tools/call, envelope capabilities advertise the UI extension:
  without clientInfo:
    {"client_supports_extension": false, "session.client_capabilities.extensions": {"io.modelcontextprotocol/ui": {}}, "session.client_params": null}
  with clientInfo:
    {"client_supports_extension": true, "session.client_capabilities.extensions": {"io.modelcontextprotocol/ui": {}}, "session.client_params": "present"}
```

`client_supports_extension` in `fastmcp_slim/fastmcp/server/low_level.py` returns early when `session.client_params is None`, but on 2026-07-28 the SDK sets `session.client_capabilities` either way and synthesizes `client_params` only when `clientInfo` arrives, which is the point of the "Prefer this over `client_params.capabilities`" note on that property in `mcp/server/session.py`.

Reproduced on fastmcp 4.0.5 / mcp 2.2.0 / Python 3.12.0, and checked `main` at 9c35c017cd89e4d50a9f512c8eafde68c301ec70, where `low_level.py` is unchanged. Happy to open a PR if that helps.

### Example Code

```python
"""client_supports_extension is False when the 2026-07-28 envelope advertises the
extension in clientCapabilities but carries no clientInfo."""

import importlib.metadata as md
import json
import socket
import threading

import httpx2
import uvicorn

import fastmcp
from fastmcp import FastMCP
from fastmcp.server.dependencies import get_context

UI = "io.modelcontextprotocol/ui"
PROTOCOL_VERSION = "2026-07-28"

mcp = FastMCP("repro")


@mcp.tool
def supports_ui() -> str:
    ctx = get_context()
    session = ctx.session
    caps = session.client_capabilities
    return json.dumps(
        {
            "client_supports_extension": ctx.client_supports_extension(UI),
            "session.client_params": None if session.client_params is None else "present",
            "session.client_capabilities.extensions": None if caps is None else caps.extensions,
        },
        sort_keys=True,
    )


def call(url: str, *, with_client_info: bool) -> str:
    meta = {
        "io.modelcontextprotocol/protocolVersion": PROTOCOL_VERSION,
        "io.modelcontextprotocol/clientCapabilities": {"extensions": {UI: {}}},
    }
    if with_client_info:
        meta["io.modelcontextprotocol/clientInfo"] = {"name": "repro-client", "version": "1.0"}
    body = {
        "jsonrpc": "2.0",
        "id": 1,
        "method": "tools/call",
        "params": {"name": "supports_ui", "arguments": {}, "_meta": meta},
    }
    headers = {
        "Accept": "application/json, text/event-stream",
        "MCP-Protocol-Version": PROTOCOL_VERSION,
        "Mcp-Method": "tools/call",
        "Mcp-Name": "supports_ui",
    }
    text = httpx2.post(url, json=body, headers=headers, timeout=15.0).text
    for line in text.splitlines():  # streamable HTTP answers with SSE
        if line.startswith("data: "):
            text = line[len("data: ") :]
            break
    return str(json.loads(text)["result"]["content"][0]["text"])


def main() -> None:
    print(f"fastmcp {fastmcp.__version__}, mcp {md.version('mcp')}, protocol {PROTOCOL_VERSION}")
    with socket.socket() as sock:
        sock.bind(("127.0.0.1", 0))
        port = int(sock.getsockname()[1])
    config = uvicorn.Config(mcp.http_app(stateless_http=True), host="127.0.0.1", port=port, log_level="error")
    server = uvicorn.Server(config)
    thread = threading.Thread(target=server.run, daemon=True)
    thread.start()
    while not server.started:
        threading.Event().wait(0.05)
    url = f"http://127.0.0.1:{port}/mcp"  # /mcp, not /mcp/ - the trailing slash 307-redirects
    try:
        print("\nraw tools/call, envelope capabilities advertise the UI extension:")
        print("  without clientInfo:")
        print("    " + call(url, with_client_info=False))
        print("  with clientInfo:")
        print("    " + call(url, with_client_info=True))
    finally:
        server.should_exit = True
        thread.join(timeout=10)


if __name__ == "__main__":
    main()
```

### Version Information

```text
FastMCP version:                                                          4.0.5
MCP version:                                                              2.2.0
Python version:                                                          3.12.0
Platform:                                             Windows-11-10.0.26200-SP0
```

yang0228:
`ctx.client_supports_extension()` read the client's capabilities through `session.client_params`, which on the 2026-07-28 per-request envelope is only synthesized when the client sends both `clientCapabilities` and the optional `clientInfo`. A client that advertises an extension in `_meta` without identifying itself — fully conformant, and exactly the case the SDK's "prefer `client_capabilities` over `client_params.capabilities`" note anticipates — was reported as unsupported, so a server gating a feature on the check silently dropped it with no error to point at. The capabilities fact is recorded on the connection either way, so the helper now reads `session.client_capabilities` first and falls back to `client_params` only for sessions that don't expose it.

```python
meta = {
    "io.modelcontextprotocol/protocolVersion": "2026-07-28",
    "io.modelcontextprotocol/clientCapabilities": {"extensions": {UI_EXTENSION_ID: {}}},
    # no `clientInfo`: optional on this protocol version
}
```

An envelope like that used to answer `False` for `ctx.client_supports_extension(UI_EXTENSION_ID)` while `ctx.session.client_capabilities.extensions` held the extension; it now answers `True`. The by-handshake clients reach the same answer through the unchanged params path.

Fixes #5192


zzstoatzz:
Thanks for digging into this, @yang0228. The diagnosis is right, and the raw-envelope end-to-end test is a nice touch. #5196 from the issue reporter makes the same change without the fallback path, so we're going ahead with that one and closing this as a duplicate. We may bring your wire-level test over when we land it. Thanks again!

PR #5195, opened 2026-09-21
Fix TokenCache handling for expired access tokens

### LINKED ISSUE #5193: TokenCache keeps tokens whose expires_at is zero
`TokenCache.set()` accepts an access token with `expires_at=0` and gives it the cache's full configured TTL. A later `get()` returns that token as a cache hit, even though its expiry timestamp is already in the past.

I expected an already expired token to miss the cache.

```python
from fastmcp.server.auth.auth import AccessToken
from fastmcp.utilities.token_cache import TokenCache

cache = TokenCache(ttl_seconds=300)
cache.set(
    "expired",
    AccessToken(
        token="expired",
        client_id="demo",
        scopes=[],
        expires_at=0,
    ),
)
print(cache.get("expired"))
```

Actual:

```text
(True, AccessToken(token='expired', client_id='demo', scopes=[], expires_at=0, resource=None, subject=None, claims={}))
```

Expected: `(False, None)`.

Version: FastMCP `0.0.1.dev1+3e20bba`, Python 3.12.13, Linux.

pradeep-ramola:
## Description

Fixes #5193.

`TokenCache.set()` now treats `expires_at=0` as an explicit token expiry instead of falling back to the cache TTL. If the effective expiry is already in the past, the cache removes any existing entry for that token and returns a miss on later reads.

## Contribution type

## Checklist

Validation run locally:

- `.venv/Scripts/python.exe -m pytest tests/utilities/test_token_cache.py` (`24 passed`)
- `.venv/Scripts/python.exe -m ruff check fastmcp_slim/fastmcp/utilities/token_cache.py tests/utilities/test_token_cache.py`
- `.venv/Scripts/python.exe -m ruff format --check fastmcp_slim/fastmcp/utilities/token_cache.py tests/utilities/test_token_cache.py`

I could not run `uv run prek run --all-files` because `uv` is not installed in this Windows environment.


zzstoatzz (changes requested):
The expired-token check needs to happen before capacity enforcement. When the cache is full and this method receives an already-expired token under a new key, `_enforce_size_limit()` currently evicts the oldest valid entry; the later `expires_at <= now` branch then drops the new token too. For example, with `max_size=1`, caching `"valid"` and then setting a distinct token with `expires_at=0` leaves both keys as misses.

Please calculate/reject the effective expiry before `_enforce_size_limit()`, while retaining the `pop()` behavior so an expired replacement still clears an existing entry under the same key. A regression case with a full cache and a distinct expired key would cover the ordering contract.

pradeep-ramola:
Updated, thanks for catching that ordering issue. The effective expiry is now calculated and rejected before `_enforce_size_limit()` runs, while still popping the existing cache key for expired replacements.

I also added a regression case for a full cache with a distinct expired token (`max_size=1`) to ensure the valid entry is not evicted.

Validated locally:
- `.venv\Scripts\python.exe -m pytest tests\utilities\test_token_cache.py` (25 passed)
- `.venv\Scripts\python.exe -m ruff check fastmcp_slim\fastmcp\utilities\token_cache.py tests\utilities\test_token_cache.py`
- `.venv\Scripts\python.exe -m ruff format --check fastmcp_slim\fastmcp\utilities\token_cache.py tests\utilities\test_token_cache.py`

PR #5206, opened 2026-09-22
fix(resources): support exploded list query parameters

### LINKED ISSUE #4378: Resource template query params fail for list-typed optional parameters

zzstoatzz:
List-typed resource query parameters currently register successfully but fail as soon as a value is supplied. This adds the RFC 6570 explode form, `{?tags*}`, so repeated query keys reach the resource function as a stable list while plain `{?tags}` remains scalar.

This draft carries forward @SaiMouli3's implementation from #4706 with its commit authorship preserved. It narrows the registration guard so an existing scalar-capable union such as `str | list[str] | None` remains valid without explode, and it encodes exploded values according to RFC 6570's unreserved-character rule.

```python
@mcp.resource("items://{category}{?tags*}")
def search(category: str, tags: list[str] | None = None) -> dict:
    return {"category": category, "tags": tags or []}

# items://books?tags=alpha&tags=beta
# → tags=["alpha", "beta"]
```

Closes #4378.

AI assistance: the original implementation was generated with Claude Code. The compatibility audit and follow-up revisions used Pi with OpenAI GPT-5.6; I reviewed the resulting diff and test evidence.


SaiMouli3:
Thanks for carrying this forward. I'd landed the same two fixes on #4706 just before this went up — the union guard came out character-for-character the same, which is reassuring. Two small gaps I hit on the way that this branch still has:

**Non-exploded query values aren't encoded.** `safe=''` is applied only to the exploded branch, so the plain form keeps `quote()`'s default `safe='/'`:

```
{?q}     with "a/b"  →  q=a/b
{?tags*} with "a/b"  →  tags=a%2Fb
```

RFC 6570 §3.2.8 is form-style expansion, where the `?` operator sets the encoding rule rather than the `*` modifier — so unreserved-only applies to the non-exploded form too, and the reference expansion emits `q=a%2Fb`. It round-trips either way, since `parse_qs` is happy with a bare `/`, so this is conformance rather than a live bug. Collapsing the three branches to one encoding call covers it:

```python
if exploded and isinstance(value, (list, tuple)):
    parts.extend(
        f"{quote(name, safe='')}={quote(str(v), safe='')}" for v in value
    )
else:
    parts.append(f"{quote(name, safe='')}={quote(str(value), safe='')}")
```

**The docs contradict the new test.** `docs/servers/resources.mdx` still reads:

> Declaring a list-typed parameter without the `*` raises an error when the template is created.

and rule 4 below it says **List-typed query parameters** must use the explode modifier. After narrowing the guard that's no longer true for `str | list[str] | None` — which is exactly what `test_union_with_scalar_query_param_does_not_require_explode` now asserts. I reworded both lines to say "can only be a list" and to note the scalar-accepting union stays valid.

Both are in `ecd25141` on `fix/4378-exploded-query-params` if it's easier to cherry-pick than to redo. Happy either way — glad this is getting to a merge.

One contract boundary worth naming either way: the guard tests "every member is a list" rather than "no member is scalar-compatible", so an all-collection union like `list[str] | dict[str, str]` isn't flagged. Distinguishing those needs a scalar/collection taxonomy that seemed heavier than the case warrants, and it matches `main` rather than regressing it.

SaiMouli3:
Codex's finding on `{?tag*}` with a scalar parameter is real, and it's a regression rather than a pre-existing gap. Worth spelling out because the base behavior isn't obvious:

On the PR base (`3820f87b`), that template is rejected at registration:

```
ValueError: Query parameters {'tag*'} must be optional function parameters
            with default values
```

`extract_query_params` didn't strip the `*` back then, so `tag*` never matched a function parameter and the template was refused up front. Once explode is honored, the name resolves, registration succeeds, and the failure moves to read time:

```
matched:     {'category': 'books', 'tag': ['alpha']}
read FAILED: ValidationError — tag: Input should be a valid string,
             input_value=['alpha'], input_type=list
```

So a clean startup error becomes a per-read validation error. That's the shape `AGENTS.md:197-199` is about.

I took the second of Codex's two options — reject at creation — since it's symmetric with the guard already there, and the alternative (re-narrowing a list back to a scalar inside `read()`) is the side-channel-compensating-for-a-symptom pattern that same rule warns against. The two guards now read as mirrors:

- annotation can *only* be a list, no `*` → error
- `*` present, annotation can't hold a list → error

```python
def _accepts_list(annotation: Any) -> bool:
    """Whether an annotation can hold a list."""
    if annotation in (inspect.Parameter.empty, Any, object):
        return True
    if isinstance(annotation, str):
        return True
    if annotation is list:
        return True
    origin = get_origin(annotation)
    if origin is list:
        return True
    if origin is Annotated:
        return _accepts_list(get_args(annotation)[0])
    if origin in (Union, UnionType):
        return any(_accepts_list(arg) for arg in get_args(annotation))
    return False
```

The permissive cases matter as much as the rejecting one — untyped parameters, `Any`, and unresolved string annotations under the `NameError` fallback all have to keep registering, or the guard starts refusing templates that work today. Covered by tests alongside the rejection: `list[str] | None`, `str | list[str] | None`, `Annotated[list[str] | None, Field()]`, and an unannotated parameter all still register.

It's `7fe5d310` on `fix/4378-exploded-query-params`, cherry-pickable onto this branch along with `ecd25141` from my earlier note. Full suite green (7,784 passed); the red skills-provider tests are pre-existing parallel flakes that pass serially.

zzstoatzz:
@SaiMouli3 can you dial back the agent here? i took your pr over because its clearly an agent blindly looping on pr events and missing details

zzstoatzz:
@codex review

